PR #12, opened 2016-08-07
Add windows 64 bit encryption dll

hhu94:
## Description
Add windows 64 bit encryption dll

## Motivation and Context
Windows 64 bit users couldn't use the old encrypt.dll.

## How Has This Been Tested?
Needs testing in both windows 32 and 64. I have not tested extensively.

## Types of changes

## Checklist:


pogom/utils.py
```diff
@@ -286,8 +287,12 @@ def send_to_webhook(message_type, message):
 
 def get_encryption_lib_path():
     lib_path = ""
+    # win32 doesn't mean necessarily 32 bits
     if sys.platform == "win32":
-        lib_path = os.path.join(os.path.dirname(__file__), "encrypt.dll")
+        if platform.machine().endswith('64'):
```
GamingMaster2000: I think this should be something like:
     platform.architecture()[0] == '64bit'

instead of .machine()

since AMD64 machine can run 32bit python instances. The .architecture() changes with the architecture of the python instance running rather than the machine.

ytanventuretech:
Tested under Windows 7 64bit for the new encrypt64bit.dll file, makes it working as compared to the existing "encrypt.dll" file.

pogom/utils.py
hhu94: I think the .dll needs to match the operating system's version, not the python instance's version. Did your test conclude that this PR does not work?

There is already a successful 64bit windows test in the comments.

GamingMaster2000: Yeah, the PR doesn't work correctly for me.
 
On 64bit Windows 10, running 64bit Python it works fine.
On 64bit Windows 10, running 32bit Python it fails "Exception while downloading map: [Error 193] %1 is not a valid Win32 application"

hhu94: Did applying your proposed change solve the issue?

GamingMaster2000: Yeah but would be good to test on a few more systems (particularly 32bit Windows if possible) to make sure. Unfortunately I don't have access to any 32bit systems at the moment.

hhu94: I'll commit the change and wait for more people to test it then. Thanks.

hhu94:
@ytanventuretech Could you retest with the commit I just pushed? Thank you :)

To everyone else, I'd appreciate more testers for this PR.

marauder37:
+1 from Barie

SimonTellsAll:
Getting error:


```
C:\Users\xxx\Desktop\PokemonGo-Map-testing>python runserver.py
  File "runserver.py", line 100, in <module>
    if not any(position):
TypeError: 'NoneType' object is not iterable
```

PR #14, opened 2016-08-07
Problem installing 

AlejandroBlackburn:
Hi, I have a little problem, I followed the steps that tells the wiki for installation in Windows but I get the following error: 

> Command "python setup.py egg_info" failed with error code 1 in c:\users\alejan~1\appdata\local\temp\pip-build-1clihd\pyproj\

 I've tried with easy-setup and git bash and in both cases I get the same error.
Additional when I try to use runserver.py gives me the following error: 

> 2016-08-07 01:01:18,009 [      MainThread][     runserver][CRITICAL] It seems `pgoapi` is not installed. You must run pip install -r requirements.txt again

Information in the wiki says that only I have to replace files and also serves to realize that.

Please help me.


Octylpy:
Not a pull request. Refer to the discord #help channel for assistance.

PR #20, opened 2016-08-07
Updated scalingo config to fit new repository

ahleksb:
As scalingo requires its own config file it needs to be updated to fit to the new repository.


mlindner:
+1 Please add.

PR #27, opened 2016-08-07
Add FreeBSD 10 encryption lib

hasosoft:
Compiled encryption library for FreeBSD 10

## Description
Added binary lib and added switch for freebsd-10 in pogom/utils.py

## Motivation and Context
The linux binary wouldn't work under freebsd 10 even with the linux compat libs

## How Has This Been Tested?
Compiled with only " '&' within '|' " warnings and has been running succesfully on my system for about 30 minutes now.

## Types of changes

## Checklist:


marauder37:
@hasosoft unfortunately the ARM change has conflicted this, would you mind rebasing?

hasosoft:
Fixed

marauder37:
I tested on OSX and it didn't asplode, so in it gets. Thanks!

hasosoft:
Well I doubt very many people will have need for this but if they do I'll keep updating this if they release a new version since I will probably have to fix it for my own system anyways.

marauder37:
we're fiddling with it to try and support more versions since I think the fbsd ABI has been the same since about 2.2

https://github.com/PokemonGoMap/PokemonGo-Map/pull/109

hasosoft:
The main problem I think was that, even with the linux compat libs it still depended on GLIBC_2.14, where as the BSD libc would do specs wise but the dependencies are hardcoded in.

I think if you compile it against FreeBSD libc it should work for Linux as well as BSD based systems but I'm not a 100% sure on that since I'm not really experienced in (cross) compiling stuff other than that I know how it should work and can type 'make'.

hasosoft:
Ok never mind that's not going to work. But larger FreeBSD support should be doable, I think pretty much everyone is on 9 or 10 though unless you're running on some archaic embedded system.

jbirch:
First off, thanks for this! I don't think the check for FreeBSD is correct though; on my system `sys.platform` returns 'freebsd10', not 'freebsd-10'

```
[jbirch@plumbus ~/PokemonGo-Map]$ ./venv/bin/python
Python 2.7.11 (default, Jun 21 2016, 01:19:47) 
[GCC 4.2.1 Compatible FreeBSD Clang 3.4.1 (tags/RELEASE_34/dot1-final 208032)] on freebsd10
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.platform
'freebsd10'
>>> 
[jbirch@plumbus ~/PokemonGo-Map]$ uname -a
FreeBSD plumbus 10.2-RELEASE-p7 FreeBSD 10.2-RELEASE-p7 #0: Mon Nov  2 14:19:39 UTC 2015     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
```

hasosoft:
True, I thought I fixed that but it didn't make it through, it's in #109 though, don't know why that one hasn't been pulled yet.

PR #31, opened 2016-08-07
Add support for basic authentication to protect web app

hannobo:
## Description
This adds support for basic authentication to limit access to the web app to people with the correct credentials. Useful when run on a public server for a single person only or a limited group of people.

Usage parameter:
`--basic-auth-user admin --basic-auth-pass mysecret`
or short:
`-ba admin -bp mysecret`

If parameter -ba is omitted (default), no authentication is requested, thus being compatible with existing installations.

## How Has This Been Tested?
On Heroku with own installation, with and without password

## Types of changes

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -6,6 +6,7 @@ import logging
 
 from flask import Flask, jsonify, render_template, request
 from flask.json import JSONEncoder
+from flask_basicauth import BasicAuth
 from flask_compress import Compress
 from datetime import datetime
 from s2sphere import *
@@ -24,6 +25,15 @@ class Pogom(Flask):
     def __init__(self, import_name, **kwargs):
         super(Pogom, self).__init__(import_name, **kwargs)
         compress.init_app(self)
+        
+        # optional basic authentication
+        args = get_args()
+        if args.basic_auth_user:
+            self.config['BASIC_AUTH_USERNAME'] = args.basic_auth_user
+            self.config['BASIC_AUTH_PASSWORD'] = args.basic_auth_pass
+            self.config['BASIC_AUTH_FORCE'] = True
+            basic_auth = BasicAuth(self)
+        
         self.json_encoder = CustomJSONEncoder
         self.route("/", methods=['GET'])(self.fullmap)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -67,6 +67,10 @@ def get_args():
                         default='127.0.0.1')
     parser.add_argument('-P', '--port', type=int,
                         help='Set web server listening port', default=5000)
+    parser.add_argument('-bu', '--basic-auth_user',
+                        help='Optional basic authentication user to protect web app')
+    parser.add_argument('-bp', '--basic-auth_pass',
+                        help='Optional basic authentication password to protect web app')
     parser.add_argument('-L', '--locale',
                         help='Locale for Pokemon names (default: {},\
                         check {} for more)'.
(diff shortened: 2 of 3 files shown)


BillyWM:
I recall there was a PR or suggestion via issues for this before. Auth should be outside the scope of the map IMO. The server admin can add basic auth through server config if they want it (basic auth is pretty poor anyway. Passwords are sent in the clear....)

danieliser:
@BillyWM - As a pro web developer I would agree. But seeing as my 10 year old set this up last night for himself I would say this is a good feature. Easy to document it among the exisitng setup where as doing it for the webserver itself not so much.

marauder37:
I guess I'm -0 on this, it adds two flags which isn't great but it's quite compact. Long-options/config keys only have hyphens in them, not underscores, so changing them to `--basic-auth-user` would get me to at least +0

hannobo:
Also, we are providing links for quick setup with Heroku etc. where there is no other option to limit the web access or configure a web server. It just prevents countless copies of open installations on the internet that people will abuse and spam Niantic.

hannobo:
@marauder37 oops, that was indeed not intended (as can be seen from the description). will fix it in a second.

marauder37:
Thanks for the update, I'm mildly convinced that this will be helpful for people doing 1-click installs, it will need some other opinions though

danieliser:
@marauder37 @hannobo - Would it be possible to move these into the config file as an optional method as well? IE -- rather than -?

Also a comma list would be awesome too (multiple user:pass combos), but I'm not picky ;)

marumari:
I think this is much better implemented upstream, via Apache or nginx or whatever frontend web server you use.

andrewda:
@marumari I agree. The only issues is as @danieliser pointed out above - many users don't have the technical know-how to get Apache or Nginx setup and just want a basic map. That being said, those users probably shouldn't be hosting it publicly. I guess overall I'm still a :-1: for this, but it *could* be useful under the right circumstances.

marumari:
I mean, it's not excessively complex to get nginx setup.  My own setup would look like this:

```nginx
location / {
    proxy_pass http://somehost.somedomain.tls:12345;
    auth_basic "Pokemon Go Map";
    auth_basic_user_file /path/to/htpasswd;
}
```

Honestly, I'd way rather see support for TLS (since my mapping server is far, far away from my web server, on the tiniest VPS I can find) than I would for basic auth.  :)

BillyWM:
Wouldn't hurt to put instructions in a wiki page for, say, setting up basic auth on Heroku. Even if it goes in the .ini here, people still have to write config to use it (user/password list). Making them go write server config is only a bit more complicated

DiscoTim:
I think most users would benefit from this. I don't want to have setup and run a separate server just to have some basic authentication. I'd like to run this on my desktop and be able to access it from my phone without leaving the app wide open to the public.

marumari:
Leaving aside that it's kind of a bad architectural decision to bolt on an entire web server, I think that this patch is probably not the best way to go about it.  It supports only a single user with a single password that is stored and transmitted in the clear.

If we do go about this, at the very least should probably just implement your standard htpasswd file.  Users can probably achieve roughly the same security by simply running their mapping service on a random high port.  Nobody scans these high ports, since it's way too slow.

invisiblek:
I'm +1 on this. I like the simplicity. Sure its not the most secure thing ever, but its simple and optional.

medghaim:
Like the last PR that suggested this, I'm not a fan. I think all auth is outside the scope of this project.

adizam:
I don't think authentication should be outside the scope of this project as a whole. Now that this lives inside its own org (pokemongomap).. perhaps an additional github repository is needed to show some nginx examples of multi-user authentication, upstream proxy routing, etc.

PR #34, opened 2016-08-07
Adds webhook field required by Pokelyzer

Brideau:
## Description
Pokelyzer has a webhook listener for receiving data from PokemonGo-Map, but with the new event for 'lured' Pokemon, this missing field for that data causes the integration to break. This addition fixes it.

## How Has This Been Tested?
This is a relatively small change so I simply tested it up making the specified change and then confirming that it worked with a local installation of Pokelyzer.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -339,7 +339,8 @@ def parse_map(map_dict, step_location):
                         'longitude': f['longitude'],
                         'disappear_time': calendar.timegm(lure_expiration.timetuple()),
                         'is_lured': True
-                    }
+                    },
+                    last_modified_time: f['last_modified_timestamp_ms'],
                     send_to_webhook('pokemon', webhook_data)
                 else:
                     lure_expiration, active_pokemon_id = None, None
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -339,7 +339,8 @@ def parse_map(map_dict, step_location):
                         'longitude': f['longitude'],
                         'disappear_time': calendar.timegm(lure_expiration.timetuple()),
                         'is_lured': True
-                    }
+                    },
+                    last_modified_time: f['last_modified_timestamp_ms'],
```
andrewda: This won't work...

Brideau: Why do you say that? I have this running now and no errors are being thrown.

Brideau: Shit, sorry. Nevermind. I added this in the wrong spot in this request. One moment.

andrewda: You're defining the `last_modified_time` property outside of the `webhook_data` dictionary. Assuming you're trying to add the `last_modified_time` property to the `webhook_data` dictionary, you should add the contents of line 343 to the line immediately after `'is_lured': True`.

Brideau: Yeah it was a dump error. I did it right locally and tried to save time by just editing the file in Github, but I stick it in the right spot. The most recent commit should fix this though, so I believe it's good now.

andrewda:
LGTM :+1:

PR #35, opened 2016-08-07
[Request] It is possible to add RSS functionality?

ccarderr:
Like in title, it is possible to add RSS functionality?

User could choose some pokemons and generate the RSS link that could be used on smartphones for fast notification when they appear nearby.
What do you think? :]


andrewda:
Pull requests are not a place for discussion.

Gol-D-Ace:
Trust me when I say that this will happen more often without the Issues page @andrewda 
We have from time to time the same issue with OBSProject that has far lower activity in PRs.

andrewda:
@Gol-D-Ace Unfortunately, yes, you're right. We're discussing re-opening the issues soon.

PR #36, opened 2016-08-07
Added tracker page

adamhwang:
## Description

Added /tracker page to play PoGo as it was meant to be played
- Follows user and refreshes every 5 seconds

Optional silhouette query string option:
- /tracker?shadow=notify - silhouettes mons on your notify list
- /tracker?shadow=not-notify  - silhouettes mons _not_ on your notify list
- /tracker - no silhouettes

## Motivation and Context

Made it because Niantic won't

## How Has This Been Tested?

Tested on Chrome desktop and mobile and FF mobile

image

## Types of changes


templates/tracker.html
```diff
@@ -0,0 +1,45 @@
...
+            </div>
+        </div>
+    </div>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/knockout/3.4.0/knockout-min.js"></script>
+    <script src="static/dist/js/tracker.min.js"></script>
+    <script defer src="https://maps.googleapis.com/maps/api/js?key={{ gmaps_key }}&amp;libraries=places,geometry"></script>
+</body>
+</html>
```
andrewda: Please indent with two spaces.

andrewda:
Other than my note above, looks good! :+1:

andrewda:
Also, just a small change - it might be better to silhouette the Pokémon *NOT* on the notification list, and color in the ones which are on your notification list. In my opinion, that would be much nicer and make much more sense.

adamhwang:
Fixed the indention.

I was hoping to have just a stand alone page without options, but I can see how some may or may not want the silhouette features.  Turned that into a query string param:

/tracker?shadow=notify - silhouettes mons on your notify list
/tracker?shadow=not-notify  - silhouettes mons _not_ on your notify list
/tracker - no silhouettes

andrewda:
@adamhwang Good idea. Will test it out with the new parameters in a bit.

JZ6:
That's a great idea, but would it slow then any features? imo if you create a separate project with it using our repo I wouldnt mind.

danieliser:
@JZ6 - It shouldn't based on the fact that it is query string based. The only time it should effect anything at all if when your looking at it. When on normal scanning it should have no effect.

prtsh716:
getting this error please help me
capture

adamhwang:
The only server side change is to serve up the new view. So @JZ6 it won't slow anything down and @prtsh716 it could not possibly have caused your issue.

JZ6:
alright understood

hhu94:
This is very cool! I'll test it in a bit.

hhu94:
I'm having issues on Chrome on Android. This is the scenario:

1. Run the scanner and visit /tracker on mobile in incognito, seemed to work fine.
2. Stop scanner and delete pogom.db.
3. Run the scanner again and visit /tracker on same mobile but different incognito tab. Old pokemon remain in the tracker, even though they weren't returned by raw_data.

I'm doing this in incognito mode, so I don't know how the old pokemon are being cached. Can you try and reproduce this?

adamhwang:
@hhu94 I can't seem to reproduce that behavior.  processPokemon() is the only method that adds pokemon, and that is only called from the raw_data request.  if you re-open the mobile page, assuming you didn't spoof a brand new location, you _should_ see the same pokemon when the scanner starts back up again, right?

are you sure they aren't returned in raw_data?  can you open up the map page on desktop or mobile and confirm the pokemon showing in the tracker aren't re-scanned on the map?

nevrending:
@hhu94 In my experience, incognito shares the same session or localStorage data across tabs in the same incognito window. Perhaps Chrome mobile also behaves this way. Can you try to repeat the scenario, but close Chrome completely first (e.g force quit it / remove from recent apps list) before visiting /tracker again at step 3?

hhu94:
@adamhwang Yes, they weren't returned by raw_data, I made sure to look at the json. I tested with a single account and scan delay 10 so there is no way it could reach the farther pokemon immediately.

@nevrending I'll try that later.

invisiblek:
bump

straitpick:
Any update for this? currently just a blank nearby screen no matter where my location is or the amount of pokemon around me, cheers

AndreasEnemyr:
When i enter `/tracker` its just empty? :(

Eskof:
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9 in position 35: invalid continuation byte

got this :/

chuyskywalker:
I think this ded.

cural0:
Works on newest dev, but only if server is ran without -fl or -nsc

tracker.js tries to change location of scan and when it can't, it doesn't continue

darkelement1987:
I downloaded the latest release (3.2.0) but the tracker wasn't included. Also can it be fixed to work with the location marker when using -fl

adamhwang:
Hi sorry all, I stopped following this after the PR was closed.  This will only work when the location is not fixed.  It essentially piggybacked off the track location feature, just in a separate view.  I'm happy to accept any PR or make anybody contributor for this PR branch if they're interested, but I've since stopped playing the game after they broke the APIs.

cural0:
`+      navigator.geolocation.getCurrentPosition(function (position) {
 +        currentLatLng(getLatLng(position));
 +        updateLatLng(currentLatLng()).then(startWatch);
 +      });`
startWatch depends on  updateLatLng being succesfull
When ran on server with -fl / -nsc this function will return a 403 error and startWatch will not start.

I've made a 'dirty' fix for it, two functions changed

updateLatLng changed to 

`      var updateLatLng = function (latLng) {
		  $.post("/next_loc?lat=" + latLng.lat() + "&lon=" + latLng.lng());
		  return lastLatLng = latLng;
      };`

and navigator.geolocation.getCurrentPosition(function (position) { changed to 

`      navigator.geolocation.getCurrentPosition(function (position) {
        currentLatLng(getLatLng(position));
        updateLatLng(currentLatLng());
        startWatch();
      });
`

When  you view /tracker in chrome using Inspect Element, the 403 error will still be there, but the /tracked will show pokemon as it's supposed to do.

Tested against fakepgoapi, works as expected.
If server is ran without -fl / -nsc it works too, server properly changes search location

I say it's a dirty fix because I don't really know how to code in JS, so I dont know if it's a proper way to code this. 
C

PR #37, opened 2016-08-07
Added possibility to change step limit via frontend

Mr-Groch:
## Description

Added to global config 'STEP_LIMIT'. Initial value readed from get_args(). Used in search threads instead of args.step_limit and possible to change via web nav panel. Changes are instant, because this changes next_location (sets actual scan location and reset scanner threads).

EDIT:
Also added -nst (--no-step-control) arg to disable step limit change from website

## Motivation and Context

Dynamic step limit is usefull when I travel from country side to big city. In big city big step_limit is useless, because i can't travel fast in straight direction, I just wanna quick search results.

## How Has This Been Tested?

Changing step limit without changing location, with lcoation change, after location change and before location change.

steplimit

## Types of changes

## Checklist:


marumari:
My only comment is that if `fixed-location` is set to `true`, then this should probably disable itself as well.  My reasoning is that users that have set that probably already have a "public" instance, and simply pulling down updates with git and running things with their current setup shouldn't grant access to this kind of powerful setting.

Alternatively, it could be disabled by default, and require you set `step-control: true`a in the config. That honestly might be the better way to go about things.

pogom/utils.py
```diff
@@ -87,6 +87,9 @@ def get_args():
     parser.add_argument('-nsc','--no-search-control',
                         help='Disables search control',
                         action='store_false', dest='search_control', default=True)
+    parser.add_argument('-nst','--no-step-control',
```
invisiblek: i don't see the nead for this

shouldn't we just wrap it in with fixed-location?

marumari: That also seems like a fine solution!

marumari:
Okay, I have some code review comments:

- You shouldn't bind onkeypress directly on the tag, you should use addEventListener (or $.on())
- The input type should be number, not text
- It probably shouldn't set styles on the input either, and instead use css

Mr-Groch:
OK, I will remove '-nst','--no-step-control' from config and use fixed-location for this. But need help with html, css things :)

Mr-Groch:
I have read something about html5 number input fiields and changed input field to number with min of 1. So onkeypress is not required any more to filter non numeric inputs. Thanks @marumari !

I have also decided to not remove -nst parameter - I can imagine situation when fixed location is not required but owner don't want others to change this i portant parameter.

I have made step control depend on fixed location - when fixed location is set, then step control will be also disabled

marumari:
Commit looks a lot cleaner now, nice work.

SebCorbin:
👍

sadimusi:
:+1:

Mr-Groch:
And what about this PR?

mrfantomex:
+1 to this request

Chlodochar:
Tested and seems to working fine. Code looks good. :+1: from me. Another collaborator must do the merge.

It would help if others test this as well and mention this in their comments.

sadimusi:
I've been running this for a while now and it's working exactly as expected 👍

Mr-Groch:
I found there is Feature Request for this:

http://feathub.com/PokemonGoMap/PokemonGo-Map/+1

Imoq:
I applied this PR along with #737 but I am not getting the box below the "Search" switch where I should be able to input the new step size. Did I do something wrong? I did run npm install / npm run build after pulling both PRs, and I do have the "scan only spawnpoints" switch (thank you) but not the one to modify steps.

Mr-Groch:
@Imoq If you wanna this with #737 I suggest to merge with my other branch https://github.com/Mr-Groch/PokemonGo-Map/tree/step_and_search_control. Those 2 independent have conflicts with each other.

git remote add Mr-Groch https://github.com/Mr-Groch/PokemonGo-Map.git
git fetch Mr-Groch step_and_search_control
git merge Mr-Groch/step_and_search_control

Imoq:
Cool, @Mr-Groch, thanks. Followed the instructions from latet devel and now I have the two new cool functions. I hope this gets merged into main soon, since everytime there is something new we will get out of sync no?

Thanks!

Mr-Groch:
And what about this PR? Can I get some feedback?

Mr-Groch:
With new commits to develop branch, Location & Search Settings needs ordering. I have put all search settings on top, and user location on bottom, now it looks like this:

loc_search_reorganize

primaxius:
Not everything needs to be in the frontend.

PR #49, opened 2016-08-07
FLASK env value not needed

chuyskywalker:
We shouldn't need to install flask on its own; and if we do `requirements.txt` should be considered broken.


chuyskywalker:
travis-ci was happy with it.

Chlodochar:
I don't see why it's still needed. Travis still runs good. :+1: for merging.

PR #52, opened 2016-08-07
Open marker settings menu by default, set height style to be the size…

invisiblek:


hhu94:
Tested and working +1

PR #70, opened 2016-08-08
Dev

mcc1:
add zh_tw to grunt minify
fix a typo in translation
add locales option to example config file
update readme


chuyskywalker:
This looks like it should be fine. :+1:

config/config.ini.example
```diff
@@ -25,6 +25,7 @@
 # Webserver settings
 #host:                  # address to listen on (default 127.0.0.1)
 #port:                  # port to listen on (default 5000)
+#locale:                # pokemon translation
```
JZ6: why was this not in the original?

PR #75, opened 2016-08-08
Optional altitude for configuration

huanga:
Because we don't all live underground at altitude 0, we shouldn't be submitting that. This PR enables us to submit altitude as a configuration, while retaining backwards compatibility for old underground people.

## Description
Modified regex to support a optional third set of values for position.
Modified parsed position to accept the optional third set of values.

## Motivation and Context
Makes it more difficult for Niantic to ban all that submits altitude 0.

## How Has This Been Tested?
Built docker image and running for my own production environment now.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -92,7 +92,7 @@ if __name__ == '__main__':
     res = prog.match(args.location)
     if res:
         log.debug('Using coords from CLI directly')
-        position = (float(res.group(1)), float(res.group(2)), float(res.group(3))
+        position = (float(res.group(1)), float(res.group(2)), float(res.group(3)))
     else:
         log.debug('Lookig up coords in API')
         position = util.get_pos_by_name(args.location)
--- a/runserver.py
+++ b/runserver.py
@@ -88,7 +88,7 @@ if __name__ == '__main__':
 
 
     # use lat/lng directly if matches such a pattern
-    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+),?(\d+\.?\d+?)?$")
+    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+),?\s?(\-?\d+\.?\d+?)?$")
     res = prog.match(args.location)
     if res:
         log.debug('Using coords from CLI directly')
--- a/runserver.py
+++ b/runserver.py
@@ -92,7 +92,7 @@ if __name__ == '__main__':
     res = prog.match(args.location)
     if res:
         log.debug('Using coords from CLI directly')
-        position = (float(res.group(1)), float(res.group(2)), float(res.group(3)))
+        position = (float(res.group(1)), float(res.group(2)), float(res.group(3) if (res.group(3) != None) else 0.0))
     else:
         log.debug('Lookig up coords in API')
         position = util.get_pos_by_name(args.location)
--- a/runserver.py
+++ b/runserver.py
@@ -88,11 +88,11 @@ if __name__ == '__main__':
 
 
     # use lat/lng directly if matches such a pattern
-    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+)$")
+    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+),?(\d+\.?\d+?)?$")
     res = prog.match(args.location)
     if res:
         log.debug('Using coords from CLI directly')
-        position = (float(res.group(1)), float(res.group(2)), 0)
+        position = (float(res.group(1)), float(res.group(2)), float(res.group(3))
     else:
         log.debug('Lookig up coords in API')
         position = util.get_pos_by_name(args.location)


JZ6:
wouldnt a random altitude from 10 - 300 be better? Less options and cleaner code

runserver.py
```diff
@@ -88,11 +88,11 @@
 
 
     # use lat/lng directly if matches such a pattern
-    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+)$")
+    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+),?(\d+\.?\d+?)?$")
```
chuyskywalker: `,?` should probably be `,?\s?` like the other set

huanga:
You're correct in terms of cleaner code. Ternary statements are kind of code smell IMO as well. However, my intent is to put in closer to real values so that it is harder for the accounts to be flagged when compared against other players. Perhaps I can make it randomize if it is not provided?

chuyskywalker:
> wouldnt a random altitude from 10 - 300 be better?

I don't think so. Given the option, I'd rather have a third option which can be accurate if we're trying to mask the scan. If altitude for a location is 20 and gets reported as 300, unless people have learned to hover, it's gonna get pretty easy to spot that as a bot.

chuyskywalker:
That said...

> Makes it more difficult for Niantic to ban all that submits altitude 0.

There are so, so many flags that an account is running from automation that it's absurd to think this is the straw that breaks the camel's back. It's just not really going to be useful to make this aspect less fake.

runserver.py
huanga: You're correct. I missed this because I am used to `12.3,12.3,12.3` instead of `12.3, 12.3, 12.3`. I will fix this now and update the PR.

huanga:
> There are so, so many flags that an account is running from automation that it's absurd to think this is the straw that breaks the camel's back. It's just not really going to be useful to make this aspect less fake.

Personally, my thought process is, the less we show up on radar, however simple as that may be, the better. I'm sure they have ways to flag it using other machine learning approaches, but the harder we make it the better it is for us :)

marumari:
Instead of being either random or static, perhaps it could be random within a range?  I could set mine to 820-850, and it would just jump around inside there?  That behaves a lot more like actual devices as well, which completely suck at determining your altitude consistently.

marumari:
Another thought!  Instead of requiring users to enter an altitude, it could simply take the startup address and use Google's Elevation API to get the elevation:

https://developers.google.com/maps/documentation/elevation/start

chuyskywalker:
I'm actually a :-1: on this. Adding extra features to "hide" the mapping is completely innefective since Niantic can straight up read this codebase (and this PR, hi guys!). Unless setting altitude _fixes_ the scanner, I don't want to add it. Adding it would just be another line of code we'd need to support, another param to pass around, and another value to track, validate, and use. Unless it's fixing something, I'd rather keep the code without it.

runserver.py
```diff
@@ -88,11 +88,11 @@
...
-    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+)$")
+    prog = re.compile("^(\-?\d+\.\d+),?\s?(\-?\d+\.\d+),?\s?(\-?\d+\.?\d+?)?$")
     res = prog.match(args.location)
     if res:
         log.debug('Using coords from CLI directly')
-        position = (float(res.group(1)), float(res.group(2)), 0)
+        position = (float(res.group(1)), float(res.group(2)), float(res.group(3) if (res.group(3) != None) else 0.0))
     else:
         log.debug('Lookig up coords in API')
         position = util.get_pos_by_name(args.location)
```
marumari: If you want to allow people to set their altitude in config.ini, you'll probably also want:

position = (position[0], position[1], float(args.altitude)

marumari:
@chuyskywalker: I think that's why I'm a fan of using the Elevation API.  We wouldn't need to have a preference or anything.  You don't even need an API key to query it:

curl 'https://maps.googleapis.com/maps/api/elevation/json?locations=43.743165,-90.21056'

runserver.py
huanga: Not sure if args are available from config.ini; I'll double check when I get a desktop environment (mobile right now).

chuyskywalker:
My point about it still being extra code to implement and maintain, to no effective benefit is the main idea though. 

Additionally, using the API is, honestly, even worse since there would now be a new endpoint which we have to keep working in order to not have people's app's bomb out.

marumari:
```python
>>> import requests
>>> altitude = requests.get('https://maps.googleapis.com/maps/api/elevation/json?locations=43.743165,-90.21056').json()[u'results'][0][u'elevation']
>>> altitude
349.6358337402344
```

We're already using this endpoint, or otherwise people wouldn't be able to see their maps.

huanga:
The application is already tightly coupled with Google Maps API. If that end point craps out because maps API is failing, the entire application would be moot anyway. It is nice to think we can make clean decoupled applications, but it is pointless to try to decouple something that is an integral part of he application.

chuyskywalker:
Fair enough, if I have to pick I like the API better.

marumari:
Want me to code something up quick and make a PR, or do you want to take this on @huanga?

huanga:
I'll investigate using that API tonight after work and evening plans. Thanks!

huanga:
@marumari it would be amazing if you could drive this! I'm at work and have some evening plans tonight, so it won't be for another while before I can get this to a merge able state.

marumari:
Okay, I'm on it!

chuyskywalker:
I'll close this PR then since @marumari will have to open another in any case.

marumari:
See PR #171.

PR #109, opened 2016-08-08
Cleanup in encryption libs, CentOS (libc 2.12) lib

cichacz:
## Description
Since CentOS in most cases ships with libc 2.12, included encryption libs are unusable. I've compiled lib from here: http://pgoapi.com/
Also moved libs to separate folder, and cleaned up a bit function

## Motivation and Context
Wanted to make it working on CentOS without changing other (working) libs

## How Has This Been Tested?
Checked if server returns data

## Types of changes

## Checklist:


cichacz:
Also, it's only for 64 bit, so if somebody can compile it for 32 bit version it would be great.

marauder37:
Since you have this open can you do something else? change `freebsd-10` to `freebsd` so we can fix it for the dude running freebsd9

cichacz:
Done :)

marauder37:
thanks, now this needs testers

wschoot:
Works for me on CentOS 6 with SCL python (2.7), thanks!

hasosoft:
libencrypt-freebsd-64.so works on ubuntu 16.04 as well, (Ubuntu GLIBC 2.23-0ubuntu3).

`ldd pogom/libencrypt/libencrypt-centos-x86-64.so`
	

> linux-vdso.so.1 =>  (0x00007ffd151f3000)
> 	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc16b7ff000)
> 	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc16b436000)
> 	/lib64/ld-linux-x86-64.so.2 (0x000055e2bdd05000)

`ldd pogom/libencrypt/libencrypt-linux-x86-64.so `
	

> linux-vdso.so.1 =>  (0x00007ffc07b3b000)
> 	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fe9675aa000)
> 	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe9671e1000)
> 	/lib64/ld-linux-x86-64.so.2 (0x000055e5ce367000)

cichacz:
@hasosoft probably when compiling on lower versions it will work in higher ones, but not vice versa.
Those libs probably need more work, to collect those which are working in most cases, and add just a few to support some quirks (like windows :smile: )
Maybe compiling under libc 2.12 will be sufficient for every other linux (using 2.12+), it would require testing..

hasosoft:
Yes that should work, even so I would advise you to upgrade your glibc since there are some serious vulnerabilities < 2.23 especially for web-facing servers.

See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-7547 and https://www.centos.org/forums/viewtopic.php?t=56467

And while its nice that you're supporting older version such behavior should be discouraged.

cichacz:
I've browsed some threads regarding glibc while looking for solution for myself, and it looks like you can't upgrade it on CentOS6 without breaking anything..
I could upgrade to CentOS7, but i don't like changing something that works well, and is stable even on tiny VPS.

wschoot:
Security related changes will be made to 2.12 as long as it's not EOL (which is not until November 30th, 2020, says https://wiki.centos.org/About/Product

hasosoft:
OK, that's good to know then. It seems like I hadn't read the forum post that well either, they mention the 2.12 update as well.

If some more people on different systems can test the centOS lib then I think it could be made the default just for simplification purposes.

Mr-Groch:
Doesn't work on Synology NAS DS213j - this is 32bit armv71, in log I can see only:

2016-08-08 18:45:50,231 [ search_worker_1][        search][ WARNING] Exception while downloading map: /volume1/homes/groch/PokemonGo-Map/pogom/libencrypt/libencrypt-linux-arm-32.so: internal error

hasosoft:
@Mr-Groch can you try compiling it from source? The link is in the first post.

Mr-Groch:
I have compiled by myself (gcc -shared encrypt.c -o libencrypt.so) and it is working on my platform.

hasosoft:
Would be nice if this one was pulled since it fixes a bug in #27 as well.

irus99:
I believe the same error is happening again on the latest update.

2016-11-13 08:35:58,937 [ search-worker-0][        search][ WARNING] Exception while downloading map: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /usr/local/lib/python2.7/site-packages/pgoapi/lib/libpcrypt-linux-x86-64.so)
2016-11-13 08:35:58,938 [ search-worker-0][        search][   ERROR] Invalid response at 28.606128,-81.490427, abandoning location
2016-11-13 08:35:59,706 [ search-worker-1][        search][    INFO] Searching at 28.605181,-81.489807
2016-11-13 08:35:59,720 [ search-worker-1][        search][ WARNING] Exception while downloading map: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /usr/local/lib/python2.7/site-packages/pgoapi/lib/libpcrypt-linux-x86-64.so)

PR #112, opened 2016-08-08
Fix direction for non english languages

Slivo-fr:
Make the gmap link work for non english languages

## Description
Force EN language on link to allow 'current location' to be recognized

## Motivation and Context
Non english speakers cannot use the get direction link

## How Has This Been Tested?
localy tested 

## Types of changes

## Checklist:


Teelve:
go to "PokemonGo-Map-develop\static\js\map.js" and look for every "Current+Location", and just replace it with the good language (I'm french, so i replaced it with "Ma+Position")

Slivo-fr:
Indeed, but I'd rather provide a permanent working solution for any user :)

Plop:
Yes please.
Another solution could be that it just indicates the marker on the map, but doesn't launch the direction.

Slivo-fr:
Indeed, I just took the old PR code without thinking about it.

Slivo-fr:
Does not work for the gmap app, only in browser. 
There is two solutions left:
- Translate "your location" in every supported language
- As plop suggested, just link to the place and let the user launch the direction

PR #119, opened 2016-08-08
add argument to pause search threads on start

ilendemli:
Adds an -sst argument to pause the search threads on start

## Motivation and Context
Even though you set a start location using the -l parameter, I don't want the script to start scanning at that location immediately. Actually i don't want to pass a -l parameter at all, but I left it mandatory at the moment. After running the script, it will wait for a resume command which can be issued by the web-interface or by a post call to the api

## How Has This Been Tested?
implemented the changes, tried running the script, and made sure that the script resumes after switching the toggle in the web-interface

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -115,7 +115,12 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
     parse_lock = Lock()
 
     # Create a search_worker_thread per account
-    log.info('Starting search worker threads')
+
+    if args.stop_search_threads:
+        log.info('Waiting for start command')
+    else:
+        log.info('Starting search worker threads')
+
     for i, account in enumerate(args.accounts):
         log.debug('Starting search worker thread %d for user %s', i, account['username'])
         t = Thread(target=search_worker_thread,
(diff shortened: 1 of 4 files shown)


chuyskywalker:
Why?

ilendemli:
@chuyskywalker i would like to set the location/coordinates before it starts to scan

off topic: how do i create two pull requests, my push just appeared in this PR

chuyskywalker:
See, I don't really think ya need this though. Sure, you set a `-l` that isn't really what you want and then you go to the UI and it's already scanning somewhere. Well, as soon as you move it, that scan queue is cleared out and the scan loop starts over at the new location. The map has a few extra "scanned area" bubbles that'll go away in a bit, but that's really it.

_(I'm always highly reluctant to add *more* config settings since this thing is already overflowing...)_

JZ6:
-1

Chlodochar:
:-1: for me too. Such an option seems very specific for the use-case. I'd rather see some way to save whether the server is scanning and where it is scanning. That way a server can resume where it left off or keep being paused while being shut down. It seems to serve the same use-case as this PR.

PR #127, opened 2016-08-08
Fix scalingo.json to target the develop branch

Soulou:
As master does not exist anymore, it resulted in an error message


invisiblek:
It doesn't just pull the default branch (develop)?

Soulou:
As the `source` parameter in the URL is explicitely defined, it doesn't.

PR #152, opened 2016-08-08
Nuke Heroku

invisiblek:


invisiblek:
It seems Heroku's IP is blocked. We shouldn't direct users to this anymore.

chuyskywalker:
lgtm

Chlodochar:
Farewell Heroku, you've served me well. This looks good to me :+1:

hannobo:
The map can run through a proxy simply by modifying **Procfile** like this:

`web: export HTTP_PROXY=$PROXY && export HTTPS_PROXY=$PROXY && python runserver.py -a "$AUTH_SERVICE" -u "$USERNAME" -p "$PASSWORD" -l "$LOCATION" -st $STEP_COUNT -H 0.0.0.0 -P $PORT -k $GMAPS_KEY $EXTRA_ARGS`

My map still runs on Heroku perfectly stable and I still think Heroko is the best of the three options. Sooner or later, all common cloud IPs will be blocked anyway and proxy usage is inevitable. -- So I suggest re-adding Heroku at least when official proxy support came to the API, or something like the above is documented.

Chlodochar:
Using a proxy is not that genral. However, I can see people hosting their own Dokku or w/e that is compatible with Heroku's Procfile. Maybe bringing this back for that, but testing would still be preferable.

PR #160, opened 2016-08-08
Add a flag to allow purging stale pokemon from the database

invisiblek:


JZ6:
+1 but why a week lol, a day or even an hour would make more sense for the average user

Chlodochar:
I might be missing something, but the code seems to suggest they are removed after 12 hours, not a week. Other than that, this seems like a useful addition :+1:.

Cronick:
Would this not go against the purpose of the stats page?

Mr-Groch:
"Full stats" page data will be incomplete

Chlodochar:
I don't think so. With this the stats will only show a timeframe of a week. @invisiblek, maybe change the option to accept the number of days Pokemon are kept.

Cronick:
I don't know why, but I don't really experience the same amount of data used on 30 minutes. I have about 220k rows in my database for Pokemons, and it's only about 30-40 mb.

Must be some crazy areas you guys are scanning to find Pokemons that can fill 10 gb of data in 30 mins.

Frozen-byte:
Please don't delete after 7 days, or let the user set the expire time.

TallGuy74:
Would it be possible to leave the default on what it is now (0 = disabled)?

invisiblek:
Possible? Yes.

I think we should be cleaning peoples' db by default though. I feel like its bad practice to let users' databases just fill up forever, even though we KNOW that most of the data expires and is basically useless after the given time.

huanga:
Can we make this configurable? Personally, I have large enough storage, and would much rather capture as much data as possible for future machine learning exercises.

invisiblek:
The current patch is configurable.

invisiblek:
Now defaulting to disabled. Users can set as they see fit.

chuyskywalker:
lgtm

Chlodochar:
Looks good :+1:

Issue #72, opened 2016-08-08
Add option to increase scan point distance


Mantriur:
It would be great to be able to influence the spacing of the scanned map points, or to toggle the distance from the pokemon detection radius to that of gyms and pokestops. That way one could have a separate scanning task keeping a large area of gyms up to date.

TheRealGitCub:
Is what you're describing different from what can be done using `-st`?

JZ6:
1. If by spacing you mean increase the space between each step, have fun missing that 100 iv dragonite in those spaces.

Mantriur:
No, st defines the amount of scan points from the current position. I mean the distance between those.

@JZ6 It's not for scanning pokemons. It's to have a supplemental scanning task keeping a larger area of gyms up to date, because they are reported over a far longer distance than pokemons.

BaIthamel:
This is not an issue it's a feature request. Please don't use the *issues* tab for requests.

TheRealGitCub:
Isn't that one of the purposes of the Issues tab on GitHub and the reason that the enhancements tag is there by default?

chuyskywalker:
I think this is a legitimate, but sorta naunced, request.

Essentially, we can only get 70m worth of pokemons per scan point, but the gyms/pokestops show much, much further away points.

If you could setup a thread to change our algorithm for each step from 70m to, say, 500m, the same `-st` value would cover a much, much larger area of gyms/pokestops.

This _can_ be achieved right now by adding a ton of `-st 1` threads, but that'd be rather tedious.

chuyskywalker:
In fact:

with the normal 0.070 (70m step):

![2016-08-07 22_08_47-pokemon go map](https://cloud.githubusercontent.com/assets/131800/17469850/04d9f87a-5cec-11e6-8d36-91f88223502c.png)

Hack that value over to 0.500 (500m) and FWOOSH:

![2016-08-07 22_10_05-pokemon go map](https://cloud.githubusercontent.com/assets/131800/17469853/11ad0010-5cec-11e6-9f3d-40039bfcff8a.png)

Way, WAY more gyms/pokestops, zero extra server load.

JZ6:
so we

chuyskywalker:
[...]

PR #161, opened 2016-08-08
Added -ss --step-size option for larger scan point distances

Mantriur:
See https://github.com/PokemonGoMap/PokemonGo-Map/issues/72

 -ss STEP_SIZE, --step-size STEP_SIZE
                        Distance between scan points, 70 is default and max
                        for pokemons, use with -np.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -59,7 +59,7 @@ def get_new_coords(init_loc, distance, bearing):
 
 def generate_location_steps(initial_loc, step_count):
 
-    # There's another call to this *not* using args.step_count as second argument, this is for step_size only.
+    # Do not use 'args" to replace the "step_count" parameter above, it does not always equal args.step_count!
     args = get_args()
     
     #Bearing (degrees)
@@ -68,9 +68,14 @@ def generate_location_steps(initial_loc, step_count):
     SOUTH = 180
     WEST = 270
 
-    pulse_radius = args.step_size / 1000.0	# dist between scan points (-ss), meters to km, default is 70.
+    if (args.no_pokemon == False): # scan_size requires -np, 70 is default
+        pulse_radius = 0.07
+    else:
+        pulse_radius = args.step_size / 1000.0	# dist between scan points (-ss), meters to km, default is 70.
+        
     xdist = math.sqrt(3)*pulse_radius   # dist between column centers
     ydist = 3*(pulse_radius/2)          # dist between row centers    
+    
     yield (initial_loc[0], initial_loc[1], 0) #insert initial location
 
     ring = 1
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -59,8 +59,9 @@ def get_new_coords(init_loc, distance, bearing):
 
 def generate_location_steps(initial_loc, step_count):
 
+    # There's another call to this *not* using args.step_count as second argument, this is for step_size only.
     args = get_args()
-
+    
     #Bearing (degrees)
     NORTH = 0
     EAST = 90
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -34,6 +34,7 @@ from pgoapi.exceptions import AuthException
 
 from . import config
 from .models import parse_map
+from .utils import get_args
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Mantriur:
Don't merge this yet, apparently the Default option is purlely informational.

Mantriur:
Fairly certain this works properly now.

karololszak:
FYI in `static/js/map.js` there is a `setupScannedMarker` function that draws the scan radius on map and there's a `radius: 70, // metres` param there.

```
function setupScannedMarker (item) {
  var circleCenter = new google.maps.LatLng(item['latitude'], item['longitude'])

  var marker = new google.maps.Circle({
    map: map,
    clickable: false,
    center: circleCenter,
    radius: 70, // metres
    fillColor: getColorByDate(item['last_modified']),
    strokeWeight: 1
  })

  return marker
}
```

It can be changed to reflect the current scannedMarker. I don't know how you'd have to go about it, maybe a param to the function, maybe a global variable? Ask frontend devs :>

pogom/search.py
```diff
@@ -57,16 +58,19 @@ def get_new_coords(init_loc, distance, bearing):
     return [math.degrees(new_lat), math.degrees(new_lon)]
 
 def generate_location_steps(initial_loc, step_count):
+
+    # There's another call to this *not* using args.step_count as second argument, this is for step_size only.
```
chuyskywalker: > There's another call to this *not* using args.step_count as second argument

What's that got to do with anything? This code comment is confusing...

pogom/utils.py
```diff
@@ -48,6 +48,9 @@ def get_args():
                         help='Location, can be an address or coordinates')
     parser.add_argument('-st', '--step-limit', help='Steps', type=int,
                         default=12)
+    parser.add_argument('-ss', '--step-size',
+                        help='Distance between scan points, 70 is default and max for pokemons, use with -np.',
```
chuyskywalker: > Distance between scan points, 70 is default and max for pokemons, use with -np.

> Distance between scan points, useful for scanning gyms/pokestops at wider range. Does not pickup pokemon more effectively.

Or something like that.

pogom/search.py
Mantriur: The functions second step_count argument is fed from args.step_count further down, it looks like an obvious optimization freebie to lose the second argument and use args.step_count instead. However, generate_location_steps() is used in another place in the code, where it is fed a different value.

pogom/utils.py
Mantriur: Force -ss to 70 unless used with -np? I see no real loss there, it *is* useless for pokemons.

pogom/search.py
```diff
@@ -57,16 +58,24 @@ def get_new_coords(init_loc, distance, bearing):
...
+    args = get_args()
+    
     #Bearing (degrees)
     NORTH = 0
     EAST = 90
     SOUTH = 180
     WEST = 270
 
-    pulse_radius = 0.07                 # km - radius of players heartbeat is 70m
+    if args.no_pokemon: # scan_size requires -np, 70 is default
```
chuyskywalker: This should never be unset, though, right? The args has a default value...

pogom/search.py
```diff
@@ -57,16 +58,24 @@ def get_new_coords(init_loc, distance, bearing):
     return [math.degrees(new_lat), math.degrees(new_lon)]
 
 def generate_location_steps(initial_loc, step_count):
+
+    # Do not use 'args" to replace the "step_count" parameter above, it does not always equal args.step_count!
```
chuyskywalker: When else is it called, I only found one usage.

chuyskywalker: Ahh, in the mock generator -- that's not so important

Mantriur: It's also in util.py insert_mock_data():

`locations = [l for l in generate_location_steps((latitude, longitude), num_pokemon)]`

chuyskywalker:
@karololszak you would specifically NOT want to modify the client side code. This does not scan a larger area *at all*

karololszak:
@chuyskywalker it **does** scan a larger area, an area of 1km^2, for pokestops and gyms. And changing the client-side code to show bigger circles would fill the gaps in between them.

This shouldn't be a set-variable PR, we should go with "scan for pokemons: 70m; scan for gyms and pokestops: 1000m" -- pokemon scan is of fixed value, so gyms/pokestops scan should be fixed as well. There is no point doing 500m circles, it's a waste of time.

Mantriur:
@karololszak It would require the radius to be added to the database, as it can vary within the same dataset. This is really for a single very specific case and imho not worth the hassle.

karololszak:
If you're thinking of a scenario where one worker is doing the big scan (for the gyms and pokestops) and there are other workers doing the pokemons, leaving it as it is would imply that it scanned for pokemons there as well (it's the same size of circle, same color etc).

Do whatever you want with the frontend, I don't really care about it, I just wanted to point it out.

Mantriur:
Well, yeah, if you'd add color coding, so the user can see the type of scan, it'd be kinda neat to have the correct circle size. Though that's outside the scope of what I had planned here. :/

Mantriur:
Not familiar with github: Is there anything left for me to do? I don't see a way to trigger another Landscape test or to view the conflicts.

chuyskywalker:
I'm pretty sure we just told landscape to take a hike, so don't worry about it.

That said, it looks like you need to rebase a bit at this point.

Issue #144, opened 2016-08-08
Geo-locate isn't working


NullEnt1ty:
"Scan follows location" doesn't seem to work. Console output below.

## Expected Behavior

Changing the marker position and scanning the area at my position.

## Current Behavior

Won't change the location at all. Console outputs:

`Uncaught TypeError: Cannot read property 'getPosition' of undefined @ map.min.js:formatted:1448`

which is this line: https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/static/js/map.js#L1903

## Possible Solution

https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/static/js/map.js#L866

Assign returned value to the marker variable. This fixed it at least for me.

`marker = createSearchMarker()`

## Steps to Reproduce (for bugs)

1. Open map.
2. Activate "Scan follows location" and try to locate your position.

## Your Environment

* Version used: https://github.com/PokemonGoMap/PokemonGo-Map/commit/0566c29226a50c19083ac1235ee16a18c2cf13c4
* Environment name and version (e.g. Python 2.7): Python 2.7.9; Webserver: nginx/1.6.2 (https enabled); OS: Debian Jessie
* Operating System and version (desktop or mobile): OS: Windows 8.1; Browser: Chrome 52

snajjar:
On Google Chrome >= 51, Geolocalisation is blocked if the host website doesn't use HTTPS

NullEnt1ty:
Sorry, I didn't made this clear: I'm using HTTPS. Geolocalisation is working fine, but the marker doesn't update its position and therefore the map still scans an old location.

nylocx:
I'm not a JS expert but the current implementation can't work as NullEnt1ty explained. But I think if creating a global variable is intended there is no need to do it outside of the function. Just removing the var before marker in the function will be enough.

rrbailey89:
Is it possible to make this work without requiring https?

nylocx:
Hopefully not from a programmers point of view ;)
https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only
But as the document states the google maps api should be uneffected by this change
https://developers.google.com/maps/documentation/geolocation/intro#overview
And localhost based sites are declared as secure too.

NullEnt1ty:
There might be some browsers which will allow sending your location over HTTP, but I'd recommend to get a free certificate (e.g from https://letsencrypt.org/).

[...]

PR #164, opened 2016-08-08
Fix follow location feature and current-location button

sadimusi:
## Description
Two very small fixes to `app.js` and `map.js`:

When the "follow location" switch was toggled on, the location was not being updated and instead an error would be shown in the console every second.

Clicking the current-location button would also throw an error, caused by a typo in the code responsible for hiding the sidebar when clicking on the map.

## Motivation and Context
Both of these features were not working and throw errors in the browser console.

## How Has This Been Tested?
Only locally on Safari and Chrome

## Types of changes

## Checklist:

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -113,7 +113,7 @@
     if (event.target.matches('a[href="#nav"]')) {
       return
     }
-    if ($stats && event.target.matches('a[href="#stats]')) {
+    if ($stats && event.target.matches('a[href="#stats"]')) {
       return
     }
     $nav.classList.remove('visible')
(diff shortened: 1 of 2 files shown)


nylocx:
As I posted in #144 shouldn't the global declaration of marker go to the createSearchMarker function to avoid strange behaviour if later someone calls the function again and and expects it to actually create something?

sadimusi:
I just did it this way because setting global variables inside functions is generally considered bad practice and harder to read.

nylocx:
Ok, learned something again, I'm still not used to this JS conventions there is so much diversity in code I read and that coming from a c++ guy ;)

fmonts:
please merge

Mr-Groch:
+1

Slivo-fr:
+1, please merge this :)

PR #187, opened 2016-08-08
Optimize grunt

chuyskywalker:
Our grunt setup can be PAINFULLY slow. There are a number of reasons:

 - Loading every npm task for _any_ change
 - Watching a TON of files (1500, originally, due to overly zealous use of `**` matchers)
 - Constantly rebuilding files which have not changed

I've addressed all of these issues in a way that should be, effectively, transparent and backwards compatible. You'll only notice that everything is faster now.

Look at each of the commits for a point-by-point speed ups

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,8 @@
 module.exports = function(grunt) {
 
+  // load plugins as needed instead of up front
+  require('jit-grunt')(grunt);
+
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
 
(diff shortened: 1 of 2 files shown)


Chlodochar:
Tested and seems to be working alright here. However, I don't exactly see a significant boost in speed:

PR-187:
```
$ time grunt build
...
Done, without errors.
grunt build  2.41s user 0.04s system 115% cpu 2.114 total
```

Develop:
```
$ time grunt build
...
Done, without errors.
grunt build  2.65s user 0.08s system 114% cpu 2.377 total
```

chuyskywalker:
The speed diff is really just around `watch` and when changes happen there.

Chlodochar:
Aah ok. At the moment `grunt` without `build` seems to initially not run `lint`, but after saving a file it does. Is that intended? Also, once it does run lint in that state, fails with:
```
Warning: Cannot read property 'execute' of undefined Use --force to continue.
```

chuyskywalker:
`default` calls `build` and `watch`

`build` does not run linting cause....that breaks a lot of stuff.

But, when running `watch` it does invoke linting -- the idea being that as you are developing, your code should damn well be linted correctly.

What file failed on the linter? I thought we had that all fixed...

chuyskywalker:
And, just in case, stop running with `grunt ...` use `npm run grunt -- ...` instead, this way you are guaranteed to be using the grunt for this project.

tkSimon:
Hopefully this also solves #229

chuyskywalker:
Probably not; this still makes/keeps those around, etc. I'm just optimizing the flow/speed here.

PR #191, opened 2016-08-08
use specific version of pgoapi in requirements.txt

Chlodochar:
## Description

Use a specific version of pgoapi in requirements.txt. Currently it is using the master branch of pgoapi, which means that anytime pgoapi pushes a commit, random people will use a different version. API changes of pgoapi will silently result in errors, but only for people who have recently updated using pip.

It seems better to keep everyone on one version and upgrade to a next version once that is compatible and tested.

## How Has This Been Tested?

`pip install -r requirements.txt`

## Types of changes

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ PyMySQL==0.7.5
 flask-cors==2.1.2
 flask-compress==1.3.0
 LatLon==1.0.1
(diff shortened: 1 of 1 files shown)


invisiblek:
+1000000000

PR #201, opened 2016-08-08
added flake8 linting file and applied linting rules

Chlodochar:
## Description

This adds a flake8 configuration file for linting the project. I also solved the issues that flake8 mentioned on the current code.

I have ignored "Line-too-long" issues, because there were a bunch of those and would result in multiple changed lines (harder to see whether it is actually still correct in the diff).

## Motivation and Context

This is partly to move from Landscape to anything else. It annoys me that it is so flaky. At least we're now able to use the command-line to run flake8, but it also opens the doors for other services like SizeCI.
## How Has This Been Tested?

Ran the crawler.

## Types of changes

## Checklist:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+ignore = E501
+exclude = node_modules,venv,.git,docs,Tools
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -8,7 +8,7 @@ from flask import Flask, jsonify, render_template, request
 from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
-from s2sphere import *
+from s2sphere import LatLng
 from pogom.utils import get_args
 from datetime import timedelta
 from collections import OrderedDict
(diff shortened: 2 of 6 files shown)


chuyskywalker:
> C'mon Landscape, gimme some sweet sweet score increase.
> EDIT: Ugh, Landscape is stuck again.

Sums up this pr nicely, lol

chuyskywalker:
So the idea, roughly, would be to include flake8 and run it as part of, say, travis-ci (et. all) instead?

Chlodochar:
@chuyskywalker stage 2 isn't clear yet. However, whatever we will do, we will want to run linting locally as well. Flake8 seems to be the standard for Python, so this is a first step towards deprecating Landscape.

Landscape currently has its own settings for flake8, so it's not trivial to match those settings using commandline.

I would like to try SideCI next. It seems a bit spammy if there already were issues in the code before a PR, but it comments the problems inline in the PR, which seems helpful. Once this PR is merged, there shouldn't be any flake8 issues anymore, so SideCI *could* be used.

invisiblek:
bump. shipit?

chuyskywalker:
Done'zo

PR #212, opened 2016-08-08
Add bounding polygon option for location_generator.py

brettgus:
## Description

Added option so that bounding polygon can be specified.  This allows more flexibility that the beehive shape.

## Motivation and Context
This solves the problem of giving the user more flexibility in the shape of the workers created.  Previously I had to export out all the coordinates, because I was working around objects like lakes.. and only wanting specific neighborhoods included as workers.  

## How Has This Been Tested?
I've tested with multiple sets of data on an Ubuntu 16.04 system.  However, I am unable to test on windows.

## Types of changes

## Checklist:

Example usage, for Milwaukee near the lake:
python location_generator.py -lat 43.036586 -lon -87.912787 -st 3 -lp 8 -polygon '(43.06452,-87.92514),(43.06427,-87.87329),(43.0442,-87.89149),(43.00692,-87.88428),(42.9744,-87.85201),(42.97089,-87.92033),(43.02362,-87.92788)'

Using the --polygon argument will only generate 80 worker processes, whereas without it would create 169

A [tool like this](https://codepen.io/bgus/full/grBdrr) could be hacked together (if there isn't another one out there) for generating the coordinates

### DIFF
--- a/Tools/Hex-Beehive-Generator/Instructions_for_location_generator.txt
+++ b/Tools/Hex-Beehive-Generator/Instructions_for_location_generator.txt
@@ -11,9 +11,9 @@ Experimental Windows support!  Use the --windows flag
 
 
 An example (Linux):
-    python location_generator.py -lat -33.9 -lon 151.1 -st 5 -lp 3 --accounts /home/pika/accounts.csv
+    python location_generator.py -lat -33.9 -lon 151.1 -st 5 -lp 3 --polygon '(-33.895, 151.094),(-33.894, 151.107),(-33.913, 151.099)' --accounts /home/pika/accounts.csv
 An example (Windows):
-    python location_generator.py -lat -33.9 -lon 151.1 -st 5 -lp 3 --accounts C:\Documents\accounts.csv --windows --installdir="C:\Downloads\PokemonGo-Map"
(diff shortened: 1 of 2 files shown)


chuyskywalker:
I'd love to see something like "draw some polygons, get a list of all scan locations need to cover the space" type application.

Granted, we have to change the application a bit to support this (ie, instead of "start at x,y, got outwards z steps" a change to "just here's a list of locations") but it'd be super useful for weirdly shaped areas where a radial search isn't really useful. _(This is me drooling over such a feature to outline my work/home neighborhoods which aren't big circles ;D)_

chuyskywalker:
I'd started working on something like a nice UI for beehive generation over (https://github.com/AHAAAAAAA/PokemonGo-Map/pull/3474) but it didn't get too far.

Dagothur:
Does this generate correct clusters for you?

http://i.imgur.com/019f3Yl.png

Frozen-byte:
@chuyskywalker would be nicer to "draw" an area on the map - and then workers are filling inside this shape.

Dagothur:
Looks like if I go back to the commit before latlon was used I can get a cluster generated that aligns properly.

brettgus:
@Dagothur Yeah, I just noticed that myself, trying to nail down wher ethat is coming from.

brettgus:
@Dagothur the issue should be fixed now with commit dff9bb6

primaxius:
fails checks

PR #220, opened 2016-08-08
Show more details on why importing pgoapi did not work

d33psky:
## Description
Add a stack trace on import pgoapi exceptions.

## Motivation and Context
When importing pgoapi does not work while it *is* properly installed the stack trace will show why the import failed, like this ::

2016-08-08 22:24:54,616 [      MainThread][     runserver][CRITICAL] It seems `pgoapi` is not installed. You must run pip install -r requirements.txt again
  File "./runserver.py", line 30, in <module>
    import pgoapi
  File "/usr/local/lib/python2.7/dist-packages/pgoapi/__init__.py", line 50, in <module>
    from pgoapi.pgoapi import PGoApi
  File "/usr/local/lib/python2.7/dist-packages/pgoapi/pgoapi.py", line 34, in <module>
    from pgoapi.rpc_api import RpcApi
  File "/usr/local/lib/python2.7/dist-packages/pgoapi/rpc_api.py", line 50, in <module>
    from POGOProtos.Networking.Envelopes_pb2 import RequestEnvelope
  File "/usr/local/lib/python2.7/dist-packages/pgoapi/protos/POGOProtos/Networking/Envelopes_pb2.py", line 9, in <module>
    from google.protobuf import symbol_database as _symbol_database
ImportError: cannot import name symbol_database

This issue was actually fixed by removing the systems' version of protobuf ::

sudo apt-get remove python-protobuf

After which "python ./runserver.py --help" works as expected.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -29,7 +29,7 @@ if os.path.isdir(oldpgoapiPath):
 try:
     import pgoapi
 except ImportError:
-    log.critical("It seems `pgoapi` is not installed. You must run pip install -r requirements.txt again")
+    log.critical("It seems `pgoapi` is not installed. You must run pip install -r requirements.txt again", exc_info=True)
     sys.exit(1)
 
 # Assert pgoapi >= 1.1.6 is installed


chuyskywalker:
I don't think a more verbose error would help most users; instead if we could detect that bad version of protobuf and error out in that case with the `apt/yum remove` advice we'd help far more people.

Chlodochar:
@chuyskywalker I don't think you can predict all possible ImportErrors. It might still be good to print the error, but follow it up with some general advice. At the moment the general advice is on top of the stack trace, maybe not the first thing people will see.

primaxius:
Like chuy said... I don't think a more verbose error would help most users

PR #243, opened 2016-08-09
Add cache busting to flask url_for entries to prevent bad js/css

### LINKED ISSUE #169: Error with JS
Recently I clone the repository and I get this errors:

> map.built.js:227 Uncaught ReferenceError: center_lat is not defined

> map.built.js:1082 Uncaught TypeError: Cannot read property 'setMapTypeId' of undefined

> 19 map.built.js:677 Uncaught TypeError: Cannot read property 'getBounds' of undefined

chuyskywalker:
Fixes #169

All urls are now using the flask helper `url_for` and with the cache buster, it will calculate the md5 of the file so that, as the file changes, so too does the URL. Also does it as an early part of the URL so your guaranteed to not get a stale cache from GET params busting (which can be problematic).

All the `lstrip` stuff is to keep the URLs relative since, by default, they are absolute which breaks proxies for a few folks who host the app a directory or two deep.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,3 +22,4 @@ sphinx-autobuild==0.6.0
 recommonmark==0.4.0
 sphinx_rtd_theme==0.1.9
 requests==2.10.0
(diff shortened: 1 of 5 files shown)


invisiblek:
have
my
babies
:heart:

PR #285, opened 2016-08-09
Fix cache busting warning

EdenGottlieb:
Fix warning `ExtDeprecationWarning: Import
ing flask.ext.cache_bust is deprecated, use flask_cache_bust instead.
.format(x=modname), ExtDeprecationWarning, use flask_cache_bust instead.
.format(x=modname), ExtDeprecationWarning`
## Description

Cache busting was implemented in a way that was deprecated. I fixed it by importing the package mentioned in the warning and changing the function name to the one present in the GitHub repo which we are pulling from.
## Motivation and Context

I encountered it, others in the issues section have encountered it
## How Has This Been Tested?

Code runs without error. Not sure if cache busting is actually working though, but there is no reason it won't.
## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -41,7 +41,7 @@ if not hasattr(pgoapi, "__version__") or StrictVersion(pgoapi.__version__) < Str
 from threading import Thread, Event
 from queue import Queue
 from flask_cors import CORS
-from flask.ext import cache_bust
+from flask_cache_bust import init_cache_busting
 
 from pogom import config
 from pogom.app import Pogom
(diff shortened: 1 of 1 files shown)


chuyskywalker:
That's what I get for [following instructions](https://github.com/ChrisTM/Flask-CacheBust), eh?

chuyskywalker:
LGTM

urls' have `/static/<hash>/path` still and load fine. I never got the error previously, so I can't say it's gone away, but there's no new errors either.

EdenGottlieb:
@chuyskywalker It is confusing, I see how it failed you. Nothing worse than an obsolete README!

PR #296, opened 2016-08-09
Only open and close the database once per cycle

invisiblek:


invisiblek:
open, work, work, work, work, work, close
vs
open, work, close, open, work, close, open, work, close, open, work, close, open, work, close

every cycle

Toetje585:
Hi There, 

Does it improve database handling, or whats the motivation behind this?

Good work so far :+1:

Cy4n1d3:
Not sure how the underlying connections are handled, but in general opening connections is a very expensive task, which makes it desirable to keep connections open as long as possible if there is still work in the pipeline.

---

Not directly related to this PR: how are the underlying connections handled? Persistent connections, connection pooling, single connection per task?

Gagaro:
I have a lot of "database locked" error on my Raspberry Pi, this would probably help.

Cy4n1d3:
Are you using SQLite? Depending on the amount of workers you're running, this might *not* help you.
SQLite is file based and not really good at handling concurrency. The RPi itself is also not an I/O powerhouse, which even worsens the problem with SQLite and multiple workers (i.e. concurrent database access).

Dagothur:
Off topic, but, is the pogo map triggering the locks on the database by design? If so, does anyone know why?

Cy4n1d3:
Database locking is needed to keep the ACID principle ([Wiki link if interested](https://en.wikipedia.org/wiki/ACID)).
Simple explanation: to keep the database in a consistent state, each operation needs to be done before another can be started to prevent a number of possible errors and problems.

Gagaro:
Yes I'm using SQLite. I also thought about using [WAL](https://www.sqlite.org/wal.html) but could not test it yet.

Dagothur:
@Cy4n1d3 do you know if the map makes explicit calls to lock? I guess I don't understand how the database portion of what the map does can have any problems.

For example, with mysql, something like this:

INSERT INTO pkmn (id, timestamp, duration, ...) VALUES (123, 28281010991, 900, ...) ON DUPLICATE KEY UPDATE timestamp=28281010991, duration=900

from my perspective would never 'lock' the database and cause other queries to fail. Is my understanding incorrect?

Cy4n1d3:
This indeed depends on locking mechanisms (pessimistic/optimistic), database type and the way transactions and occuring errors are handled.
If you're looking at traditional RDBMS like MySQL, you'll see that consistency is valued over speed.
In a scenario where speed is crucial but data consistency is not a primary concern (often applies to distributed databases in a redundant master/slave scenario) you might let concurrent writes across multiple nodes happen and correct occuring errors later on.

To get to the core of your question: your understanding is incorrect in this case, especially if looking at SQLite databases. SQLite is file based, which means that the following happens on every write: lock file,  execute transaction (including logging, journaling, etc), write file, remove lock from file. Even if this doesn't take long, you'll run into problems if a certain amount of concurrent actions happen (as you can guess). This surely can be optimized to some degree but in general SQLite is not built for heavy usage scenarios.

invisiblek:
so....this is working well here (mysql).....is it breaking sqlite?

DeviantEng:
I'm seeing all kinds of `Lock wait timeout exceeded; try restarting transaction`, and I'm hoping this PR will resolve that.  Thanks @invisiblek, and let's get it merged!  :D

MJ869:
NameError: name 'db' is not defined

PR #304, opened 2016-08-09
Update pgoapi

invisiblek:


GenuineSounds:
Tested: Works fine, no obvious bugs that I can see.

EdenGottlieb:
Are we fixed on a single commit just to make sure it still works if the code breaks on their side?
I thought the whole point of a develop branch is to be bleeding edge without having to worry about breaking it

invisiblek:
I'd prefer to target a specific hash

Chlodochar:
@EdenGottlieb the explicit version will avoid problems where a git checkout of a certain version of pokemongomap will work today, but not tomorrow. The same goes for different people trying the code at different times. It can be extremely confusing when parts are fluctuating. The builds need to be reproduceable by different people at different times.

simarsenault:
Looks like it takes lastest commit:

image

I ran:
```
git pull
npm install
npm run build
```

Chlodochar:
@simarsenault Yes, that's normal behavior. It assumes `dc1b222...` is an commit-hash, which it is, so that's good ;).

simarsenault:
@Chlodochar Oh alright, thanks for the clarification! :smile:

PR #320, opened 2016-08-09
Fixed broken "Get directions" links

e7d:
The "get directions" links where broken, landing on a Google map itinerary starting from nowhere as "Current Location" would not be parsed correctly.

## Description
Replacing "Current Location" with "Current Position" fixes the issue, so the itinerary is built and displayed correctly.

## Motivation and Context
The "get directions" buttons were all broken

## How Has This Been Tested?
Click any "Get directions" button and you land on an invalid itinerary.

Broken (before):
broken

Fixed (after):
fixed

## Types of changes

## Checklist:


invisiblek:
Weird, I don't see this behavior

CoolCanuck:
I don't have this issue.

jaquer:
I believe this is locale-dependent. Notice how Google translated "Current Position" to "Votre Position" (Your location, in French). Maybe #112 is a better fix?

e7d:
You're right @jaquer. Tested with different locales, and the problem does not seem to happen for all of them. This considered, #112 is actually a better fix that will work for everyone.

PR #327, opened 2016-08-09
Update docs to run pip install as user

invisiblek:


chuyskywalker:
Why?

invisiblek:
shouldn't be a need to run this as root, but it looks like some of the scripts for external stuff (travis for instance) are expecting it?

chuyskywalker:
No, it looks like that travis instance just died. I don't know how to force a recheck without pushing something else to the branch. 

https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption--user

Seems safe enough.

Th3Mouk:
This doesn't solve the installation problem.

Chlodochar:
Just to note, this doesn't work so well with `virtualenv`, but it removes the need of virtualenv in my case (needing to use root). It's better than what it was before. :+1: for merging this once rebased.

primaxius:
tested on virtualenv too ..not good.  Someone tried this and messed up his install as he had stuff installed in 2 locations causing conflicts

PR #329, opened 2016-08-09
Retry connecting to database if failed FOREVER

invisiblek:


Infernoman:
Using this and your connect_once database patch. Things going smoothly so far. Great work.

PR #332, opened 2016-08-09
Show more info in the pokemon info

chuenlok:
## Description

When users click on the pokemon icon,
it shows more information about the pokemon, including iv, catch probability, moves, attack, defense and stamina.

Two new parameters: -ed, -si. `-ed` is encounter delay. `si` is to enable scanning the IV and details.

## Motivation and Context

Let users know more about the pokemon before they want to catch that
## How Has This Been Tested?
I checked the DB structure and the information is filled correctly.
Tried on chrome and seeing it is working.

I tested it with only single worker thread. More workers may trigger a IP ban and I am not sure the optimal value of the encounter delay.

## Types of changes

## Checklist:


chuenlok:
The IV seems correct on my Mac but not correct on a ubuntu linux. Not sure if it is related API problem.

Romfrosk:
I've started to use this. and it seems to work with a few errors:

Using Windows: 

IV: shows 0.09 no matter the ADS values

Information does not appear in the lured pokemons

Romfrosk:
Also. Using this makes you account receive a Softban for 15 minutes

chuenlok:
@Romfrosk I think we should use a dummy account for the map scanning? 

The values are from the pgoapi directly. It seems working on mac but not on the other OS. Maybe I will add separated calculating later

aldarund:
Its additional api call. How the api will react to this  regarding scan delay? Someone above mentioned a soft ban - doest it mean account wont be able to find new pokemon for 15min? If so its really bad.
Also since its additional api call so additional load shouldn't it be optional feature?

chuenlok:
As I know, softban only affects catching pokemons and spinning pokestops so even it is soft banned, it should be fine.

I reduced the time to sleep for the scan delay based on the time used on encounter pokemon, but I am not sure whether encountering pokemons will still require same scan time delay. 

any thoughts?

chuenlok:
I will make the one for lured pokemons later.

chuyskywalker:
As much as I'd like to see `iv` for Pokemon, starting an encounter to get that data starts to make this less of a map and more of a bot than I am personally comfortable with.

Romfrosk:
I did some tests on this:

I'm running 2 environments at the moment.

Dev: Ubuntu - 36 workers - unique users
Production: Windows - 300 workers - unique users - publicly avaiable

When running last nights latest dev on my production setup, i had no problems with workers discovering 0 pokemon, 0 pokestops, 0 gyms.
After implementing this pull request into Prod and trying it out for an hour i saw that every now and then my workers start getting no results for 15 minutes, then starting to work again. just to fail again after a random time (5-15 minutes).

reverted back to last nights dev, and restarted my setup. it went fine for 2 hours.
Assuming this had to do with this pull request i set up an Ubuntu Dev Environment to test it out on a smaller scale.
After setting it up and running the map with last nights dev + this pull request on the Ubuntu with 36 new PTC users i saw that they did indeed pull the info correctly, but after 5-15 minutes i started to see the same behaviour as before, with workers not picking up anything.

Main problem was: this affected BOTH environments making me believe it's an IP Softban and not user based.

rrbailey89:
Id like to see something like this added.

snahelou:
+1

marumari:
I agree with @chuyskywalker, plus I really don't like the thought of doubling my API requests.  A lot of my commits have specifically been to keep my requests down.

For people talking about softbans, it's entirely related to the number of API requests from a given IP in a given amount of time. If you use something like this to check IV, you need to half the number of workers to keep under whatever arbitrary threshold that Niantic has instituted.

If this does get merged, it should absolutely be a configurable option and it should be disabled by default. I also see that the PR says it updates the documentation, but I don't see any commits to the docs. Perhaps I'm just missing it?

rrbailey89:
I'd still like to see this added.  But I agree it should be a toggle.

chuenlok:
Thanks for the suggestion. I will set it as optional and add the encounter delay parameter.

chuenlok:
Just added it as optionally.

chuyskywalker:
I suspect your should dedupe `encounter` requests as well. That way if the scanner goes over an are multiple times, a person isn't starting the encounter multiple times for the same, already discovered and "encountered" pokemon.

chuenlok:
it shouldn't reencounter pokemon now.

FlarumHostingCo:
Getting this error..    File "/usr/local/lib/python2.7/dist-packages/pymysql/err.py", line 115, in _check_mysql_exception
    raise InternalError(errno, errorvalue)
InternalError: (1054, u"Unknown column 't1.iv' in 'field list'")


I don't see anything in the PR to add a column to the database

aldarund:
What about cp? Maybe its possible to calculate and display cp in percent of maximum cp of this pokemon on this trainer lvl?

FlarumHostingCo:
getting an error randomally..

pogom/models.py", line 408, in construct_pokemon_dict
    pokemon_info = ecounter_info['wild_pokemon']['pokemon_data']
KeyError: 'wild_pokemon

rrbailey89:
@flarumhostingco you need to flush your database to use this

FlarumHostingCo:
@rrbailey89 Ok,  flushed and restarted both runserver.py and MariaDB and still receive the random error

rrbailey89:
i would like to see this implemented still.

chuyskywalker:
This is quite out of date now, and it doesn't really fit with the ethos of the project, so I'm going to close it.

darkeagl:
Hi I have a problem when I try to use this: I ge the error: iv : pokemon_info['cp_multiplier'] invalid syntax. This means for me that there is no data fetched. I checked the code already 4 times but cannot get behind this. Could you maybe help me, like:
pokemon_info = ecounter_info['wild_pokemon']['pokemon_data']
 ecounter_info = result['responses']['ENCOUNTER']
 result = api.encounter...
and I think I have a problem there because where does ENCOUNTER come from and the rest. Without it everythong worked like a charm. 

I am confused a little bit right now :(

darkelement1987:
How do i merge this with 3.2.0 release?

PR #344, opened 2016-08-09
Fix of parse_map method when GET_MAP_OBJECTS is empty

Th3Mouk:
## Motivation and Context

When I run the server, I received empty GET_MAP_OBJECTS array. Probably there's a problem, on API call or response, but this shouldn't be blocking. The worker must continu to search others locations.

```sh
  File "/Users/Oak/Projects/pokemap/pogom/search.py", line 242, in search_worker_thread
    parse_map(response_dict, step_location)
  File "/Users/Oak/Projects/pokemap/pogom/models.py", line 302, in parse_map
KeyError: 'map_cells'
```

## How Has This Been Tested?
 - OSX El Capitan
 - Python 2.7.10
 - pip 8.1.2 from /Library/Python/2.7/site-packages/pip-8.1.2-py2.7.egg (python 2.7)

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -296,6 +296,9 @@ def parse_map(map_dict, step_location):
     gyms = {}
     scanned = {}
 
+    if len(map_dict['responses']['GET_MAP_OBJECTS']) < 1:
+        return False
+
     cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells']
     for cell in cells:
         if config['parse_pokemon']:


invisiblek:
-1

If you're getting empty responses, something is wrong. Ignoring them is not a fix

Th3Mouk:
Is this possible what a fresh account can be instant ban ? It's probably the reason, and this case must be handled.

invisiblek:
You have accounts that have not accepted the ToS or something then. This is not an error we should ignore like this.

Th3Mouk:
@invisiblek this is working. So I close it thanks, if the case append again I will link this thread.

PR #372, opened 2016-08-10
Make map.js work with wrong local system times

Kruptein:
## Description

I added a function getCurrentDate() that uses getDate but converts it to a UTC aware/correct timestamp.  This function is used everywhere where the current date was used.

## Motivation and Context
This prevents pokemons not showing up due to local clocks having a weird configuration.
Using UTC correctly displays the pokemons and the scan locations.

## How Has This Been Tested?
I have a laptop that dualboots between windows and archlinux.  They generally fight for the system clock which results in my archlinux environment having the clock 2 hours ahead, which caused no pokemons to appear due to the disappearing time being < my "current" time.
The change fixes this bug on my archlinux box and normal working clients before (like my mobile phone) still work after the fix as well.

## Types of changes

## Checklist:

I don't think there is any change of documentation required, the method name getCurrentDate might be better renamed to getUTCDate or something alike.


Kruptein:
Don't merge this yet; Seems there is still something wrong

Kruptein:
I wrongly assumed it also worked on machines with a correct system time due to pokemons appearing, but their disappear time is 2 hours behind so that is just wrong.

In my usecase I have a map on a fixed location with a server in that area so I could pass the currentTime of the server in each XHR request and then determine an appropriate offset, but this might not be a good approach for every usecase.

Any ideas ?

So my current proposal is adding a field in the ajax requests that the server fills in with its current ms since epoch, the client should thus be able to determine if any offset is necessary.

tkSimon:
You'll have to have a correct system clock for the map to work properly

Kruptein:
Yes the times are all perfectly fine if your local system time is correct.  There are just some cases where it's less trivial to adjust your local system time or where you want to set your clock to a different timezone on purpose.

I have a fix on my local server atm that sort of works, it displays the correct minutes left for pokemons.

It's a pretty niche fix, so I probably won't merge it to upstream.

static/js/map.js
```diff
@@ -783,6 +783,11 @@ var Store = {
 // Functions
 //
 
+function getCurrentDate () {
+  var now = new Date();
+  return new Date(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate(), now.getUTCHours(), now.getUTCMinutes(), now.getUTCSeconds());
```
Chlodochar: This converts a UTC date *as-is* to a local-time date. Basically adding a timezone to a date. That is highly confusing for me to understand in what situation this is correct behavior. Could you elaborate?

Kruptein:
That code fixed it for my wrong local time but is wrong in general.  I have a more legit working fix on a local branch, but had to postpone testing due to exams.

As this pr is just wrong I'll just close it for now and create a new one later.

EvilLost:
any update on a new PR?? was unable to find one

PR #377, opened 2016-08-10
Fix example.py indentation

invisiblek:


FrostTheFox:
+1 for example.py

CoolCanuck:
pls. was my idea sir

PR #385, opened 2016-08-10
A simple memoizer to only parse the args once

chuyskywalker:
Previously, every call to `get_args()` would re-parse everything, which is a waste. Now it doesn't.


chuyskywalker:
Sneak sneak sneak

PR #386, opened 2016-08-10
UI wont load if Recent Scanned Location Query is too big

Goldvivi1:
In the UI, if show scanned locations is turned on and the amount of recent scanned locations in the DB are above 15k or 20k the UI wont load (freezes the UI). You will encounter this kind of error if you have a BIG scan area with many workers. There is probably a better way to implement this, i suck with sql queries.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -282,7 +282,9 @@ class ScannedLocation(BaseModel):
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
                  .dicts())
-
+        #UI wont load if this query is too big so cropping the query might Fix it
+        if len(query.values()) > 10000:
+            query = query.objects.order_by('last_modified')[:10000]
         scans = []
         for s in query:
             scans.append(s)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -282,7 +282,6 @@ class ScannedLocation(BaseModel):
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
                  .dicts())
-        #UI wont load if this query is too big so cropping the query might Fix it
         if len(query.values()) > 10000:
             query = query.objects.order_by('last_modified')[:10000]
         scans = []
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -281,7 +281,7 @@ class ScannedLocation(BaseModel):
                         (ScannedLocation.longitude >= swLng) &
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
-                 .order_by(-ScannedLocation.last_modified)[:10000]
+                 .order_by(ScannedLocation.last_modified.desc())[:10000]
                  .dicts())
         scans = []
         for s in query:
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -281,7 +281,7 @@ class ScannedLocation(BaseModel):
                         (ScannedLocation.longitude >= swLng) &
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
-                 .order_by(ScannedLocation.last_modified.desc())[:10000]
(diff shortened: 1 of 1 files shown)


chuyskywalker:
You're right, as it turns out, this isn't a great way to do this. The query itself should have a `limit`, and perhaps `order by` clause applied instead. Cutting the options down _after_ sql fetch is no bueno.

Goldvivi1:
Should i test further to see what would be a good limit? I just assumed it happens if the query is above 10k, so i placed 10k without any math or testing. Didnt check at the time, so i was just assuming with that number.

chuyskywalker:
I don't know, but I suspect we'd need to not only limit the query, but somehow also show on the UI that "This map is showing few items than available because drawing more than NNNN items on a google map is a bad idea."

Google maps does have a way of "clustering" things together that could also be used, but that's getting much further down the complexity path.

invisiblek:
This probably helps a little: https://github.com/PokemonGoMap/PokemonGo-Map/pull/387

Goldvivi1:
I just did a test, query was around 14k and it was already not showing the map. It started to get slow above 11k-12k. I guess my math isnt that bad.

chuyskywalker:
@invisiblek -- nah, this isn't a problem querying the DB, I suspect it's more a problem with trying to display too much on the JS map. I bet you can hit the `raw_data` url just fine, but trying to parse out and layout 10k+ scan circles is probably egregiously bad to do to the map.

Goldvivi1:
thanks to swPoints the query gets cropped on the viewport, but once you zoom out ....

Goldvivi1:
scannedLocation table is already set to add newest location at the bottom of the table.
Instead of an `order by` we could just limit the selection to the bottom rows of the table? Sorting 10k records seems like an exhausting task.

Goldvivi1:
Did some research on django queries. *sighs* Gotta learn someday right? 
`.order_by(-ScannedLocation.last_modified)[:10000]`

invisiblek:
To be honest, it would really sadden me to limit this query like this. There's nothing we can do on the front end to solve this?

Goldvivi1:
@invisiblek Fusion Tables? making the google servers do the processing for us. Does the google API provide town boundaries? when zooming out if the town/city has been scanned completely, just draw a polygon of the whole town/city maybe?

Infernoman:
.order_by(ScannedLocation.last_modified.desc())[:5000]
AttributeError: 'list' object has no attribute 'dicts'

corycomer:
Similar to what @Goldvivi1 was saying, what about using a polygon with only the scanned locs at the edges of the whole area and using a line stroke weight of 2*scanned radius to show the scanned areas?

Infernoman:
File "/home/USER/infernomap/pogom/models.py", line 284, in get_recent
    .order_by(-ScannedLocation.last_modified)[:5000]
AttributeError: 'list' object has no attribute 'dicts'

Goldvivi1:
@infernoman I dont know anything about SQL queries

Infernoman:
Then why are you making a pull request about it !? -_-

Goldvivi1:
@Infernoman bring this problem to light, something is better than nothing.

pogom/models.py
```diff
@@ -281,8 +281,8 @@ def get_recent(swLat, swLng, neLat, neLng):
                         (ScannedLocation.longitude >= swLng) &
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
+                 .order_by(ScannedLocation.last_modified.desc())[:5000]
```
Chlodochar: Please use `.limit(5000)`. `[:5000]` seems to do things in-memory after actually querying the database.

Chlodochar:
At the moment this is a bug, so it needs to be fixed. We can hopefully make the number of scanned locations returned an option in the future with the default set to 5000.

Goldvivi1:
Changed

primaxius:
-1 This is a bug with bad fix.

PR #387, opened 2016-08-10
Turn scanned_id into a PrimaryKeyField

invisiblek:
This is a much smaller field than concatenating the lat and long into a character field and using that for an id.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -6,7 +6,7 @@ import calendar
 import sys
 from peewee import SqliteDatabase, InsertQuery, \
     IntegerField, CharField, DoubleField, BooleanField, \
-    DateTimeField, fn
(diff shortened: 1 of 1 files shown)


invisiblek:
This is going to cause people to have to dump their database (or at the very least their ScannedLocation table).

chuyskywalker:
This is a breaking change; people can't upgrade their way through it blindly.

We don't have an `upgrading.md` for this kind of stuff :/

invisiblek:
Yea, I know. I'll think of something.

invisiblek:
K, this is looking pretty good now.

FWIW: I want to merge this separately: https://github.com/PokemonGoMap/PokemonGo-Map/pull/388

Having the commit that actually makes use of the migration stuff will be handy for others to know how this should work.

PR #388, opened 2016-08-10
Introduce database schema versions

invisiblek:
This essentially does nothing for now other than creating the table we'll use in the future.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -3,6 +3,7 @@
 
 import logging
 import calendar
+import sys
 from peewee import SqliteDatabase, InsertQuery, \
     IntegerField, CharField, DoubleField, BooleanField, \
     DateTimeField, fn
@@ -21,6 +22,8 @@ log = logging.getLogger(__name__)
 
 args = get_args()
 flaskDb = FlaskDB()
+
+db_schema_version = 0
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -290,6 +293,14 @@ class ScannedLocation(BaseModel):
         return scans
 
 
+class Versions(flaskDb.Model):
+    key = CharField()
+    val = IntegerField()
+
+    class Meta:
+        primary_key = False
+
+
 def parse_map(map_dict, step_location):
     pokemons = {}
     pokestops = {}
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -457,3 +469,27 @@ def drop_tables(db):
...
+        db_ver = Versions.get(Versions.key=='schema_version').val
+
+        if db_ver < db_schema_version:
+            # This is where we'll handle migrations in the future
+            log.info("Detected database version %i, updating to %i", db_ver, db_schema_version)
+            Versions.update(val=db_schema_version).where(Versions.key=='schema_version').execute()
+        elif db_ver > db_schema_version:
+            log.error("Your database version (%i) appears to be newer than the code supports (%i).",
+                      db_ver, db_schema_version)
+            log.error("Please downgrade your code base or drop all tables in your database.")
```
Chlodochar: shouldn't this be "*upgrade* your codebase"?

Chlodochar:
+10000. Haven't been able to test this yet, but this should've been in there since the very beginning. Apart from the comment, for looks good :+1:.

pogom/models.py
invisiblek: yea, it was getting late, i'll fixit

invisiblek:
Ok, I think i'm finally happy with this.

pogom/models.py
```diff
@@ -449,11 +460,45 @@ def bulk_upsert(cls, data):
...
     db.connect()
-    db.drop_tables([Pokemon, Pokestop, Gym, ScannedLocation], safe=True)
+    db.drop_tables([Pokemon, Pokestop, Gym, ScannedLocation, Versions], safe=True)
     db.close()
+
+
+def verify_database_schema(db):
+    if not Versions.table_exists():
+        db.create_tables([Versions])
+        InsertQuery(Versions, {Versions.key: 'schema_version', Versions.val: 0}).execute()
```
invisiblek: hmm, the logic is off here i think

GenuineSounds:
In the future will we be distributing updates to the database in the form of incremental sql patches instead of hardcoded python scripts?

invisiblek:
@GenuineSounds I never thought about that. Long-term that might be a good idea.

GenuineSounds:
@invisiblek All of the project I've used long terms always used version to version sql patches and it worked quite well. Of which I'm speaking about https://github.com/mangos/MaNGOS databases and the open source World of Warcraft databases since the beginning basically.

invisiblek:
We'd have to take things into consideration like the fact we're supporting sqlite and mysql databases. Do we need to be concerned with syntax between the two when we're writing straight SQL? Going through peewee abstracts all that for us.

GenuineSounds:
@invisiblek You're right about that... That's an interesting problem. We are going to need an abstraction layer if we want a common patch format. DAMN YOU SQL!!!

invisiblek:
Android does it like this, so it can't be ALL bad =P (Android uses sqlite fwiw)

GenuineSounds:
I'm not well versed in python, I assume the OOP concepts are the same, so I'll speak in those terms.

We have an interface for the patch classes we use and load them dynamically from a folder... by name I suppose. something like
```
interface { string from, to; status patch(db); }
```
Easy to manage updates, either load them by name or load them all and walk the update chain.

Chlodochar:
@GenuineSounds It's a good approach, but you'll still need to automatically choose the right migration. I think for the time being a simple integer for the version with linear migrations is quite sufficient already.

invisiblek:
let's see how this goes...

Issue #179, opened 2016-08-08
Scan optimization: only scan areas that have produced a pokemon in the last 24 hrs


jabberjabber:
When scanning large areas, especially in rural ones, most of the scans result in 0 pokemon. After awhile of not producing anything, these cells should be removed from the scan list, or at least dramatically decrease the scan frequency of these areas.
Based on: https://www.reddit.com/r/pokemongodev/comments/4wpp6z/mappingscanning_and_the_spawnpoint_mmodb_thoughts/

Also, since we know that pokemon show up hourly, we should know exactly when to scan for it instead of searching every several seconds.
Based on: https://www.reddit.com/r/pokemongodev/comments/4wpp6z/mappingscanning_and_the_spawnpoint_mmodb_thoughts/d697o29

sorokyl:
A further optimization would be to only scan each 70m diameter area in known active times (if a scan returns 0 at 22 minutes after the hour) then it is safe to deactive from 7-22.    If you scan again at 28 minutes after the hour and still nothing, increase the dead range from 7-28.

Yossi:
how would this handle lured pokestops?

TallGuy74:
You can make grids with pokestops exempt from all this, and check it every couple of minutes.

Draco18s:
Man, I'd kill for the ability to say "f** scanning this specific zone, I can't get here, ever."
![not ever](http://s10.postimg.org/zeseh6zw9/nope.png)
As you can see, there's a psyduck out there right now. I'd love to have another one (candies, obviously) but between where I am and that spot is a good half mile of literal jungle.  We used to be able to get out there (there's a powerline fire trail along the stream) but we haven't bushhogged the area from our fence over to the stream in a year or more.  I'd swear there's a 5 foot high pile of poison ivy just outside the fence.

khag7:
look into TBTerra's spawnScan to accomplish this.

basically you build a list of spawn points and their spawn times, then only check those points when you need to (once per hour) instead of constant scanning.

Draco18s:
I'll look into that. The git page doesn't offer a whole lot in way of "how to set this up and use it" but if it's based of PokemonGo Map, I should be able to manage it.

sorokyl:
That project isn't nearly at the level of this one as far as features.  I think it's a great enhancement to this project.

sorokyl:
[...]

PR #390, opened 2016-08-10
Add spawnpoint display

Xcelled:
Shows spawnpoints on the map, based on https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into/

## Description
While this does not change scanning behavior, it does display all known spawnpoints on the map for informational purposes. Spawnpoints are highlighted as a 5m circle and can be toggled on/off in options, defaulting to off.

## Motivation and Context
Spawnpoint based scanning is apparently the Next Big Thing (tm), and while I didn't want to implement the whole scan stuff, I thought it'd be informative to see where the spawnpoints actually are, if for no other reason than to laugh at how inefficient beehive is.

## How Has This Been Tested?
Spot-checked on Win10 with Chrome, and Android 6.0.1 with Chrome.
Spawnpoint output was (roughly) compared to the output generated from the SQLite query on the Reddit page.

Pokemon Off:
image

Pokemon On:
image

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation. (Probably, I don't know if the UI options are documented)

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -108,6 +108,9 @@ class Pogom(Flask):
         if request.args.get('appearances', 'false') == 'true':
             d['appearances'] = Pokemon.get_appearances(request.args.get('pokemonid'), request.args.get('last', type=float))
 
+        if request.args.get('spawnpoints', 'false') == 'true':
+            d['spawnpoints'] = Pokemon.get_spawnpoints(swLat, swLng, neLat, neLng)
+
         return jsonify(d)
 
     def loc(self):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -184,6 +184,23 @@ class Pokemon(BaseModel):
             appearances.append(a)
         return appearances
 
+    @classmethod
+    def get_spawnpoints(cls, swLat, swLng, neLat, neLng):
+        query = Pokemon.select(Pokemon.latitude, Pokemon.longitude, Pokemon.spawnpoint_id)
+
+        if None not in (swLat, swLng, neLat, neLng):
+            query = (query
+                     .where((Pokemon.latitude >= swLat) &
+                            (Pokemon.longitude >= swLng) &
+                            (Pokemon.latitude <= neLat) &
+                            (Pokemon.longitude <= neLng)
+                        )
+                     )
+
+        query = query.group_by(Pokemon.spawnpoint_id).dicts()
+
+        return list(query)
+
 
 class Pokestop(BaseModel):
     pokestop_id = CharField(primary_key=True, max_length=50)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -194,7 +194,7 @@ class Pokemon(BaseModel):
                             (Pokemon.longitude >= swLng) &
                             (Pokemon.latitude <= neLat) &
                             (Pokemon.longitude <= neLng)
-                        )
+                            )
                      )
 
         query = query.group_by(Pokemon.spawnpoint_id).dicts()
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -617,7 +617,7 @@ var mapData = {
   pokestops: {},
   lurePokemons: {},
   scanned: {},
-  spawnpoints: {},
+  spawnpoints: {}
 }
 var gymTypes = ['Uncontested', 'Mystic', 'Valor', 'Instinct']
 var audio = new Audio('static/sounds/ding.mp3')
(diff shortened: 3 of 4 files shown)


bbbenji:
Please provide a way to to export spawnpoints.

Xcelled:
@bbbenji That's probably a little out of the scope of this pull request (more of a tool), but the code I added in models.py could easily be reused for such a purpose.

bbbenji:
@Xcelled Got it working, thanks.

invisiblek:
I'm not sure how I feel about this. We're already having performance issues with scanned locations: https://github.com/PokemonGoMap/PokemonGo-Map/pull/386

Wouldn't this particular feature introduce even more performance issues than https://github.com/PokemonGoMap/PokemonGo-Map/pull/386 ?

Infernoman:
This should reduce the queries on the servers both ours and niantic's by only scanning certain spawn points at the intervals they are supposed to spawn, rather than continuously scanning all area's see > https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into

invisiblek:
That patch won't make it into our code base (as-is at least). The idea of skipping scan locations is something I can't agree with.

Now, if it could be improved to be smart enough to prioritize locations and scan those more often, we could talk.

Regardless, its not a fix to what I'm referring to.

Xcelled:
This doesn't skip locations, or impact the scanning at all. It's strictly informational only.

There's no caching of the spawnpoints, primarily because I didn't want to make too many changes to the underlying architecture, since this does not actually implement spawnpoint scanning. This could be something to implement if performance is an issue.

However, I don't think that'll be needed. This, inlike scanned locations, isn't meant to be turned on constantly/over a wide area. That's not particularly useful, as the points are pretty small and get hard to see at too far out of a zoom level. I also noticed some possible inefficiency in the js that could be fixed.

Regardless, if there are performance concerns, conducting profiling is the way to go. I can run some tests.

Xcelled:
I Instrumented UpdateMap in the JS and did a 7 step scan of the Santa Monica pier without the spawnpoint code and timed approximately 1,000 invocations. Then, I added the spawnpoints calls and repeated the process. There were a LOT of spawnpoints, as you can see in this picture:

image

Without the spawnpoints, UpdateMap took an average of 373 ms to run. With it, 378.

Unless there's another method I should look at, the cpu usage doesn't seem terribly affected. Of course, if you zoom out enough, eventually you'll get bogged down by spawn points, but that's true of any marker on the map.

I'd also like to note that at least four people have directly told me this is useful to them, and one requested I send them screenshots of the spawnpoints around them.

cpius:
+1

Let's get the next big thing already!

Goldvivi1:
@xcelled looks cool and useful when zoomed in, when zooming out in the UI they are useless. Maybe using the Viewport SW and NE points to calculate if the spawn points are to be displayed or not.

Xcelled:
@Goldvivi1 that's similar to what's being suggested in #386 for scanned points. My concern with that would be screen size, ie a smartphone screen can only display a small geographical location at a zoom where spawnpoints are useful, but two 24inch monitors could display a much larger one.

I think that for now, it should just be up to the user to turn them on/off as desired. If/when #386 is resolved, hopefully the same fix could be applied here.

schadis:
pls and now make it possible in config that it only scans the whole map like 2 times in a hour for new spawn points and the rest of the time only spawn points shall be scanned +1

Xcelled:
@schadis see #179

Xcelled:
I just wanted to note that, even though the reddit SQL's group by has changed, this pull request does not need to be updated. The reddit group by's purpose is to select unique combinations of spawnpoint coords and times.

This code doesn't calculate or care about the time, which eliminates the need to group on it. spawnpoint_id maps a 1-to-1 to a unique lat/long pair, so grouping on the spawnpoint_id effectively does the same as grouping on the lat/lon pairs.

**TL;DR:** No changes needed.

chuyskywalker:
Awesome. A good step in the direction of being able to change up towards waaaay more efficient tracking.

rjwestman:
I sometimes get spawnpoints, but not for every pokemon on the map.

jcat00uk:
is it feasible to have more info on the spawnpoint displayed, say when you click on a spawnpoint, it brings up the hourly spawntime for that point and maybe the last spawned pokemon at that location

cheers
John

PR #393, opened 2016-08-10
Draw a Poly(s); Get Scan Coords

chuyskywalker:
The direction I'm heading with this is that you can visit a page, draw (or paste) in a set or polygons, and it'll spit out a list of scan locations. You can then feed that into the app and, tada, you have coverage over any shape area you want instead of circluar/hexical spirals.


chuyskywalker:
Where it stands right now:

poly

The covering algo I have in the JS is utter garbage, I need some help getting that to operate on the proper hex grid packing setup.

After that, there's LOTS of UI work to do.

And, of course, figuring out how to get the backend to support injesting a set of coords like this (or perhaps just pushing them into the search overseer? Dunno yet.)

Xcelled:
Looks like you're using squares to cover the area. Hexagons (like the map uses internally) gives you better, less overlapping coverage.

One way I could think to do this (easily) is to generate hexagonal coverage of a circle that circumscribes the drawn area, and then removing any hexagons that fall completely outside of the area. This would let you reuse the existing python code (though you may need to translate it) and should give you a pretty good (but not optimal) packing. Optimal packing of a shape is nontrivial, and an irregular one, even harder.

You may consider putting the covering algo into python, and just having JS post a list of coordinates.

static/js/poly.js
```diff
@@ -0,0 +1,224 @@
...
+      while (curLat < lats.max()) {
+        var pointA = new google.maps.LatLng(curLat, curLng);
+        var pointB = pointA.destinationPoint(90, radiusInKm/2).destinationPoint(0, radiusInKm/2);
+        if (google.maps.geometry.poly.containsLocation(pointA, poly)) {
+          circ(pointA);
+        }
+        if (google.maps.geometry.poly.containsLocation(pointB, poly)) {
+          circ(pointB);
+        }
+        var nextPoint = pointA.destinationPoint(0, radiusInKm);
```
reyqn: `var nextPoint = pointA.destinationPoint(0, radiusInKm*3/2);`

static/js/poly.js
```diff
@@ -0,0 +1,224 @@
...
+        curLat = nextPoint.lat();
+      }
+      curLat = lats.min();
+      var pointA = new google.maps.LatLng(curLat, curLng);
+      curLng = pointA.destinationPoint(90, radiusInKm).lng();
```
reyqn: curLng = pointA.destinationPoint(90, radiusInKm*1.732/2).lng();``

reyqn:
This should help with overlapping

brettgus:
Was working on PR #212 for adding the location_generator.py but will close it out if this takes off because this is a much more elegant of interfacing with it.  

Agree with @Xcelled that the easiest way to handle this is just use the existing algorithm to generate points and discard the points that fall outside the polygon.

tomasznguyen:
I was working on a similar functionality as well (**[see wip](https://tjupa-r1.tjupa.com:8001/test-fill-polygon.html)**).

The workflow is: (a) draw a polygon and (b) click in the polygon to fill it with circles.

What still needs to be done is outputting the circles as e.g. csv and have it import in runserver.py.

I'll try to fix that tomorrow. It would also be nice to support polygons with holes.

static/js/poly.js
chuyskywalker: Perfect, thanks! This is working awesomely.

chuyskywalker:
If my first iteration, I did actually use the server side render. With @reyqn's pointers, though, this works perfectly client side anyway. 

I've just updated this branch with a near fully functioning setup. It's missing the ability to "restore" a set of polygons, but I'm not sure it's worth the hassle at this point.

poly2

JZ6:
+ 1 x 9000

pyr0ball:
+1

rrbailey89:
I tested it and it works great. However the list it generates when copy and pasted is just 1 very long line. Is it possible to have it actually generate a list of coordinates to a file or have it copy paste-able each on a separate line?

chuyskywalker:
It's a JSON string, so the idea would be you'd do something like:

```
# config.ini
location-list = [[big, long], [string, here]]
```
or
```
runserver.py --location-list='[[big, long], [string, here]]'
```

rrbailey89:
I had no idea this was possible. Would I be able to use many accounts in the string and the app automatically does the work on assigning accounts regions to scan?

chuyskywalker:
Yup, that's kind of the idea. Select a bunch of regions in this, feed that coord list and a set of accounts to a config, and let the app figure out the rest. No more "run 5 bajillion different instances in magical super-mecha-mega-turino configuration!" mode.

HusbandoMax:
How do I access this to get the coords like above and how do you format it in console with multiple accounts?

brettgus:
@chuyskywalker Great work!  Quick question.. how do you specify step size?

rrbailey89:
@chuyskywalker Does this work with beehive by any chance or something else that could automatically fill in the string for each line or the list of coordinates pulled from a file and a list of accounts also pulled from a file. 

If not do you think it would be wise to add that functionality to this page some how?

jjbott:
I made some hacky tweaks that provide better(?) results here: https://github.com/jjbott/PokemonGo-Map-1/commit/be54bba8ddd1814cbf6c0fbcc5e300435f482688

It runs the coverage algo many times, shifting the origin a bit each time, to brute force a better set of circles. 

Then, the standard honeycomb isnt going to be optimal no matter what in many (most?) cases. So I made the circles draggable and removable (with right click) so you can tweak the results by hand.

Not going to make a PR because its slow and sketchy. Feel free to steal whatever. :)

brettgus:
@jjbott Haven't had a chance to try it yet but the draggable and removable is an awesome idea.  There isn't an easy way to remove the circles you don't like.

chuyskywalker:
@jjbott -- I dealt with unwanted circles by just subtly adjusting the polygon points until I get what I want. "click to remove" isn't a bad idea, though, and would definitely help for "here's a big area, but screw this lake" setup. Otherwise it can be a bit of a pain to get rid of the lake.

That said, marking specific circles as "deleted" and recovering that is too much of a pain for me to think about right now. 

The latest commit here is me being pretty much wrapped up, for now. You can draw many polygons and manage them (you can't delete, but not really the end of the world) and you can drop, what I am calling, gp-scan markers. (I plan to use these to power statically positioned scanning nodes which just track pokestops and gyms since they have a HUGE radius.) Finally, the entire thing can be recovered by copying out the "simple config" and restoring it later. The 'app config' has no where to go as of this PR, but that's next on my list, along with a LOT of stuff in #179

jjbott:
@chuyskywalker I just dealt with deleting circles by removing them from the map, and making sure the coords list only displays those circles that are still on the map.  

My motivation for removing/dragging is because I can come up with better circle placement that an algorithm can. For example, I picked a small random park, and it came up with http://i.imgur.com/xEMTxq6.png , 13 circles. After some fiddling, I got that down to 9: http://i.imgur.com/PbP1Ygv.png .  Pretty much any area that isnt a giant blob, we can do better manually.

chuyskywalker:
I guess we could drop the "rebuild" aspect. Just use polygons to drop an initial set of circles, allow it to be manipulated, and then "set". Once "set" a polygon's circles would become individually editable and the polygon would be removed.

Recovery would be about the full config; not a set of polygons.

xioxxio:
Where is the wiki on how to use?

rrbailey89:
@xioxxio it's not committed yet. Wiki will probably come following the commit.

Worthy7:
I just forked this but i dont understand, where is the front end to do this?

brettgus:
@Worthy7 Is it though?  That's what it looks like in the code.  However, I cloned this branch and can't get that to load

Worthy7:
Yea it is. 

I just did a load of work on this and have a pretty nice working version now. 

1) The code for creating the front end points is just too beefy. It isn't useful for a large area - it just crashes the browser. I suggest moving the front end calculations, allow a "polygon" variable in the config, and then get the server to parse and calculate the points needed on the backend at startup.

2) I have made an entirely different project for now, solely for finding all of the spawnpoints in a certain area. (it ends when it has finished.) 

3) This is simply a tool for selecting the points on the map. I want the simpleconfig to be able to be passed in on the config.ini to make this easier to use.

phr34k0:
error: unrecognized arguments: --location-list

brettgus:
@phr34k0 That part isn't build out yet

phr34k0:
@brettgus yes i noticed after reading all the pulls and commit. would be awesome if its implemented. much less headaches for a beehive

rrbailey89:
Still looking forward to this guy :)

phr34k0:
@rrbailey89 Meh.. Don't think it'll ever happen. Btw check this http://beehive.bessei-it.eu

Worthy7:
What's the point tbh, we dont need our own project to be able to do this,
we need an additional tool (is ANYONES build of it) we don't need all the
back end madness either just the JS front-end.

paulharman:
Does this fulfill the requirements?  https://github.com/brandonshults/pipoam

shadi1989:
Is someone still working on this? would be a great way to stop scanning lakes/rivers and useless locations

primaxius:
Conflict and seams abandoned :) closed - see alternative https://beccasafan.github.io/pgm-multiloc/

Issue #148, opened 2016-08-08
Notification spam on pokemon despawn


Slivo-fr:
The map often throw multiple notifications when pokemon in the notification list disapear.

## Expected Behavior
Notification only trigger when a pokemon appear.

## Current Behavior
notifications are also triggered when pokemon disapear

## Steps to Reproduce (for bugs)

1.Add some common pokemon in the notification list
2.Activate notification with sound
3.Wait for pokemon despawn

## History
This issue was already there on the last repository, there was a fix in PR that was never merged : https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2131

JZ6:
still waiting on a pr rebase, if you were following closely, perhaps you could do the work of writing the rebase.

Slivo-fr:
I considered this option, but I did not really follow this closely, and I cannot get enough time to do it myself unfortunately.
Maybe re-dev what was done instead of rebasing if it's too complex, I'm not so used to git rebase :/

Hope bringing this up here can help

invisiblek:
Thing is, I'm not really on board with https://github.com/AHAAAAAAA/PokemonGo-Map/pull/2131 and what it did. Fixing this bug was a happy side effect of it, but the UX that was added didn't feel right imo.

I'll look at this next, maybe hack something in.

invisiblek:
After examining the server response, I believe the way lured pokestops are flagged has changed. Also it doesn't appear the pokestop has any info about the lured pokemon anymore. See notes here: https://github.com/PokemonGoMap/PokemonGo-Map/pull/493

invisiblek:
If I'm right, this issue will no longer be an issue. My theory here is that lured pokemon now act the same as regular ones. If that's the case, we should entertain the idea of dropping a sort of 'dedicated' scanner (or force the searcher to scan more often) at lured pokestops.

invisiblek:
I was wrong... 😝 

Need to come up with a way to handle lured pokemon. It seems the scanner's location needs to be really close to the pokestop or you won't get the info for the lured pokemon (unlike how it used to work before). I have ideas, but implementing them isn't exactly trivial...

hhu94:
@invisiblek so are lured pokemon returned as normal pokemon now? Or are they still in the pokestop info?

invisiblek:
[...]

PR #421, opened 2016-08-10
Save encounter_id for lured pokemon

invisiblek:


invisiblek:
i'm hoping this solves https://github.com/PokemonGoMap/PokemonGo-Map/issues/148

PR #461, opened 2016-08-10
Fix #418: Notifications crash Chrome for Android

### LINKED ISSUE #418: Problem when turning on notifications on Android Chrome
When I turn on notifications (either for a pokemon or rarity) and a pokemon is shown on the map that would trigger the notification the whole process stops there. No notification is shown, no other pokemons will be loaded on the map (but the ones that were loaded before the pokemon that raised the notification). It is like the site crashes or something. Removing the notification pokemon from the the notification list and reloading the site helps.

Using: Android Chrome Version 52.0.2743.98 on Android 5.1.1

On Desktop Chrome (Win) it is working

Studio-XP:
## Description
Fixes notifications from crashing the page on Android

## Motivation and Context
With notifications enabled while viewing on your phone, the page would crash, not showing any new Pokemon. This commit prevents the page from crashing, while still alerting a user of a nearby Pokemon by vibrating the phone for one second
https://github.com/PokemonGoMap/PokemonGo-Map/issues/418

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,8 @@
     "countMarkers": true,
     "skel": true,
     "setupPokemonMarker": true,
+    "createjs": true,
+    "Push": true,
 
     "noLabelsStyle": true,
     "darkStyle": true,
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -4,7 +4,7 @@
 import calendar
 import logging
 
-from flask import Flask, jsonify, render_template, request
+from flask import Flask, jsonify, render_template, request, send_from_directory
 from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
@@ -33,6 +33,7 @@ class Pogom(Flask):
         self.route("/search_control", methods=['GET'])(self.get_search_control)
         self.route("/search_control", methods=['POST'])(self.post_search_control)
         self.route("/stats", methods=['GET'])(self.get_stats)
+        self.route("/sw.js", methods=['GET'])(self.static_from_root)
 
     def set_search_control(self, control):
         self.search_control = control
@@ -222,6 +223,9 @@ class Pogom(Flask):
                                gmaps_key=config['GMAPS_KEY'],
                                valid_input=self.get_valid_stat_input()
                                )
+
+    def static_from_root(self):
+        return send_from_directory(self.static_folder+'/js', request.path[1:])
 
 
 class CustomJSONEncoder(JSONEncoder):
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -225,7 +225,7 @@ class Pogom(Flask):
                                )
 
     def static_from_root(self):
-        return send_from_directory(self.static_folder+'/js', request.path[1:])
+        return send_from_directory(self.static_folder + '/js', request.path[1:])
 
 
 class CustomJSONEncoder(JSONEncoder):
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -619,7 +619,7 @@ var mapData = {
   scanned: {}
 }
 var gymTypes = ['Uncontested', 'Mystic', 'Valor', 'Instinct']
-createjs.Sound.registerSound("static/sounds/ding.mp3", "ding")
+createjs.Sound.registerSound('static/sounds/ding.mp3', 'ding')
 var pokemonSprites = {
   normal: {
     columns: 12,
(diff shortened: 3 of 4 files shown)


DiscoTim:
while this will prevent the crashing by wrapping things in a try /catch.
audio.play() which is part of the notifications is also a broken and many tablets don't support vibrate.

there may be better ways, i included these two libraries, https://github.com/CreateJS/SoundJS and https://github.com/Nickersoft/push.js and changes to this apps code was very minimal, and it seems to work as it should across all platforms.

Studio-XP:
audio.play() ins't necessarily broken in the sense that it breaks the page, but it requires user interaction to begin playing the audio and Android doesn't support autostart for several reasons. I haven't spent much time with SoundJS, but I'm curious if/how it handles autostart on mobile.

As you mentioned in the issue though, push.js requires a valid ssl certificate or running locally. If it's running locally, native `new Notification()` will work as expected and I can't see many people going through the effort of getting ssl-proxy setup on nginx.

DiscoTim:
soundjs autostarts and only 3 lines to implement, i should really figure out how to do pr's :)

in map.html....
<script src="https://code.createjs.com/soundjs-0.6.2.min.js"></script>

in the map js...
createjs.Sound.registerSound("static/sounds/ding.mp3", "ding");

then to play...
createjs.Sound.play("ding");

push.js was a little more but not much, should work across any browser, browsing though the source i see different browsers and versions all handle them differently.. of course :).  if you'd like the code i used for that just let me know.

Studio-XP:
Haha, that was easier than I thought. Confirmed working on Android, although a side effect (whether good or bad) is that it'll still ding in the background if I "minimize" Chrome.

Unfortunately, I cannot get sw.js to load up. I'm getting a 500 error when trying to access it directly :/

DiscoTim:
sw.js has to load from root for it to work i used,

this is the start of my app.py for reference.. its the bottom 3 lines that are added:

```
class Pogom(Flask):
    def __init__(self, import_name, **kwargs):
        super(Pogom, self).__init__(import_name, **kwargs)
        compress.init_app(self)
        self.json_encoder = CustomJSONEncoder
        self.route("/", methods=['GET'])(self.fullmap)
        self.route("/raw_data", methods=['GET'])(self.raw_data)
        self.route("/loc", methods=['GET'])(self.loc)
        self.route("/next_loc", methods=['POST'])(self.next_loc)
        self.route("/mobile", methods=['GET'])(self.list_pokemon)
        self.route("/search_control", methods=['GET'])(self.get_search_control)
        self.route("/search_control", methods=['POST'])(self.post_search_control)
        self.route("/stats", methods=['GET'])(self.get_stats)
        self.route("/sw.js", methods=['GET'])(self.static_from_root)

    def static_from_root(self):
        return send_from_directory(self.static_folder+'/js', request.path[1:])
```

then sw.js goes in the \static\js folder and will be accessible from https://mysite.com/sw.js or whichever folder map.html is being served from, mine is https://mysite.com/go/ so it is accessible from https://mysite.com/go/sw.js

DiscoTim:
Push can handle the vibrate as well

```
Push.create(title, {
    body: text,
    icon: icon,
    timeout: 5000,
    onClick: function() {

      window.focus();
      this.close();
      centerMap(lat, lng, 20);

    },
   vibrate: [500, 500]
    });
```

Studio-XP:
I saw that and tried it that way at first, but was unable to get it to work on Chrome or Firefox for Android. I think I read somewhere yesterday after trying it that it doesn't work inside of a notification, but I'll have try and relocate that info.

DiscoTim:
does the demo site vibrate for you? it implements vibrate https://nickersoft.github.io/push.js/

Studio-XP:
No, it doesn't vibrate in either Chrome or Firefox. In fact, it doesn't do anything in Chrome on my end

DiscoTim:
Chrome mobile? you should be getting the system type notifications from android. are you accessing it from https?

Studio-XP:
Yes, Chrome (52.0.2743.98) on Android (Galaxy S6, Android 6.0.1), https, notifications allowed

DiscoTim:
yes i am same, the notifications should be visible if you pull them down from top, you won't see them in the browser.

Studio-XP:
Wow, I feel stupid now... I just checked, and Chrome was set to not show any notifications in Android's app manager

DiscoTim:
no worries :D thought maybe something like that, glad its working.

DiscoTim:
This function could be replaced as well, for some browsers won't like it,

```
$(function () {
  if (!Notification) {
    console.log('could not load notifications');
    return;
  }
  if (Notification.permission !== "granted") {
    Notification.requestPermission();
  }
});
```

with something like..

```
$(function () {
if (!Push.isSupported) {
console.log('could not load notifications'); 
return;
}
Push.Permission.request();
});
```

Issue #456, opened 2016-08-10
Request throttling seems to fail for a period after search has been stopped and restarted


nnpoke:
After the implementation of spreading workers out evenly over the scan delay time then if you stop the search in the location settings UI and later re-enable the search, the worker threads are not throttled for some time.

## Expected Behavior
When disabling and re-enabling search the worker threads should be throttled according to scan delay

## Current Behavior
After re-enabling search a lot of search requests are fired in a very short time which might give you errors saying you have to slow down

## Possible Solution
Make sure workers are throttled when search is activated from the UI

## Steps to Reproduce (for bugs)
1. Disable search from the UI
2. Wait a few minutes
3. Enable search from the UI
4. Notice how fast the search grid is executed. You can also see how fast the console output is generated

## Context
I'm affraid it can cause IP bans

## Your Environment
Server running on Ubuntu and using latest PokemonGo-Map

nnpoke:
The problem is located in search.py:

 sleep_delay_remaining = loop_start_time + (args.scan_delay * 1000) - int(round(time.time() * 1000))
 if sleep_delay_remaining > 0:
     time.sleep(sleep_delay_remaining / 1000)

 loop_start_time += args.scan_delay * 1000

Loop_start_time should be set to current time instead of just adding scan_delay. If you have paused your worker threads for a long time then it will loop without sleeping until loop_start_time has caught up to the actual time.

So my suggestion is to change the last line of code to:
 loop_start_time = int(round(time.time() * 1000))

ocoot:
still have the problem :|

chuyskywalker:
@invisiblek you broke'd it!

invisiblek:
^

Cy4n1d3:
Happens for me aswell, ssh tunnel lost connection and after reconnent all workers are speeding up their efforts quite a bit, which results in heavy throttling :)

RiberoD07:
still having a similar issue not sure if it warrants opening a separate issue for it but when toggling scan follows location or when using the auto locate button near the zoom tools console says thread search paused and nothing happens after that.. I tried toggling it off, refresh page, playing with the other settings, moving the pin on the map and it never started populating again. I had to kill the script and restart. Is there any solution?

nnpoke:
I have solved the problem with this code in pogom/search.py. Find the last line and place the code before that line... Then it will resume threads evenly and reset the loop_start_time.


# Create delay after resume search
[...]

PR #478, opened 2016-08-11
Delay after resume bugfix

taihun81:
Bugfix for issue #456, and added staggered delay when search resumes.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -301,9 +301,6 @@ def check_login(args, account, api, position):
 
     log.debug('Login for account %s successful', account['username'])
 
-    # To avoid first search step resulting 0/0/0
-    time.sleep(args.scan_delay)
-
 
 def map_request(api, position):
     try:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -267,7 +267,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     time.sleep(sleep_delay_remaining / 1000)
                 else:
                     time.sleep(args.scan_delay)
- 
+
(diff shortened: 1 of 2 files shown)


invisiblek:
Looks like a couple very minor complaints from travis-ci. Please fix so we can ship!

pogom/search.py
```diff
@@ -258,6 +265,8 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                 sleep_delay_remaining = loop_start_time + (args.scan_delay * 1000) - int(round(time.time() * 1000))
                 if sleep_delay_remaining > 0:
                     time.sleep(sleep_delay_remaining / 1000)
+                else:
+                    time.sleep(args.scan_delay)
```
invisiblek: So if we missed the window where we should have scanned again, it would wait a full cycle before scanning again?

taihun81: Ooooh...  My bad, I changed the other line earlier but I forgot to change this one... it should have been  sleep_delay_remaining >= 0:

pogom/search.py
```diff
@@ -292,6 +301,9 @@ def check_login(args, account, api, position):
 
     log.debug('Login for account %s successful', account['username'])
 
+    # To avoid first search step resulting 0/0/0
+    time.sleep(args.scan_delay)
```
invisiblek: scan_delay timeouts should be handled elsewhere. if we need a pause after logging in, it should be a standard value

taihun81: This should have been separated PR..  Please ignore and I'm working this out on the other PR request.

PR #484, opened 2016-08-11
Pokemon cries set up as notification sounds

drakeirving:
## Description

This adds functionality so that pokemon notifications with sound active will use the pokemon's cry from the games rather than the previous ding sound. Interacts fine as you add/remove notified pokemon and reload the Store of notified pokemon.

## Motivation and Context

I'm using the map for analysis on spawns. While running the map in the background and waiting for certain pokemon to show up I might get a notification for a different pokemon that I care less about, but the ding doesn't differentiate between pokemon. So I thought I'd set it up to use the pokemon cries, which makes them unique and more importantly recognizable if you already know them.

## How Has This Been Tested?

I wrote the changes on a slightly earlier version and tested on a grunt watch instance until the code style was fine and the changes worked properly. I had arrow functions and other ES6 stuff in before that weren't built/minified properly and had to refactor a bit so it would work. Pulled and retested and it's working as expected.

## Types of changes

- [x] Extra assets

## Checklist:


sLoPPydrive:
Oh yes please! 👍

blindreaper:
Can you please make this optional, 1MB? per device is a lot.

DiscoTim:
awesome! won't work on android chrome though due to audio.play() not supported without user interaction. i'm working on a pr for notifications that uses https://github.com/CreateJS/SoundJS. very easy to setup and use. ie..

```
createjs.Sound.registerSound("static/sounds/123.wav", "123");
....
createjs.Sound.play("123");
```

DiscoTim:
I could be wrong, but won't this error if you are only selecting to be notified by rarity? doesn't look like the id's will get loaded.

SmartPappnase:
I tried it, and it did only work for explicitly selected Pokemon. 

Also, it was not particulary helpfull for me, since those old-school sounds sound mostly the same to me. :( I guess I was no hardcore Pokemon fan than...

DiscoTim:
thanks i suspected that, i haven't listened to them, not a hardcore fan.. just this game. but i really like the idea of different sounds per Pokemon.

marumari:
Why in the world are these PCM and not something compressed?  Mp3 + Mp4 + AAC or VP8/9 + WebM have great support and the files are much, much smaller.

r- as PCM, for sure.

Also, adding these also increases the risk of this repository getting taken down simply for containing copyrighted material.  I know we already have the Pokemon icons, but I don't know if we should be pushing it.

SmartPappnase:
Maybe a better way of distinguishing sounds would be to implement a 2 Tier Pokemon Notification List, which ideally can be turned on / off individually... There are Pokemon for which I would leave almost everything and run outside to catch it... However, certain Pokemon I would like to know about only if I am already out looking for Pokemon...

Btw, is there a way to store the browser settings, or to implement a default Pokemon Notify List on Server start? It is quite anoying to always having to redo this in the browser...

drakeirving:
@blindreaper The only sounds that are loaded by the client are the ones for notified Pokemon, if that's what you're worried about. It does create new Audio objects for the whole list each time you add/remove Pokemon, which could be improved, but it'll only request whatever new resources are needed.

@DiscoTim Entirely possible it errors, the version I developed with didn't have the rarity notification. I'll check this out, but the fix should just be to fallback to the ding, which it probably should do regardless.

@marumari Agreed; only reason they're PCM is because that's how they were recorded, I suppose. I'll convert and change. I'm also (very) mildly concerned about takedown potential, but would it be better to put the files off-site and just leave the naming convention in? I'm not really sure, that seems pretty inconvenient for people wanting to just set up the map.

@SmartPappnase A tiered notification list would be tangential to the core of this contribution and would be something I don't really care to implement myself. Uniqueness and recognizability are the key points here.
As for the browser store, the app already does this automatically (and isn't really something to address in this pull request anyways), so I don't really know what you mean.

## Todo
- [ ] Ding fallback
- [ ] Compress sound files

drakeirving:
Well that's done. I've left the ding fallback be used for the rarity-notified Pokemon (rather than use cries for every one) under the assumption that if people find a Pokemon through rarity that they want further notifications for, they'll put them on the more specific list. Plus that way I don't have to (re)load the cries for every rare Pokemon that pops up.

DiscoTim:
i really like the idea of different sounds per pokemon but i don't want to have to include them all in a list to do it, its such a pain. one of the great things of the rarity list is that you only have to add a few items to get notifications for many items, i only leave out common. others may disagree :)

honestly, i think it would be nice to have a toggle to use standard ding, or cries.

audio.play() doesn't work on chrome android, not your pr specifically, but i really wish this project would move away from it, with something like soundjs.

you don't need to reload the cries for every pokemon that pops up, just something like this called from both list change functions, it also doesn't recreate audio items every time the lists change. it could remove items, but it's only 151 max items :

```
var sounds = { 'default': new Audio('static/sounds/ding.mp3') }

function loadPokemonSounds () {

  $.each(notifiedPokemon, function(key, value){
    if (!(key in sounds)){
      sounds[key] = new Audio('static/sounds/' + key + '.mp3')
    }
  })

  $.each(idToPokemon, function(key, value){     
    if (($.inArray(value['rarity'], notifiedRarity) != -1) && (!(key in sounds)))  {
      sounds[key] = new Audio('static/sounds/' + key + '.mp3');       
    }
  });

  return sounds
}
```

SmartPappnase:
@drakeirving 
I guess my second comment does not really belong here, however, I had to redo the notification settings multiple times despite the caching feature. I don't relly know why, but after a browser crash or restarting the computer it usually lost my settings (no privacy mode). Also I am using different devices and browsers which makes it even more anoying... Ultimately I solved it for me by manipulationg the rarity-list of pokemon to my liking  so I can now very quickly set it up again by just using "notify by rarity". Btw, this also enables me to quickly switch between being notified about everything noteworthy, or just beeing notified about must have Pokemon (for me)...

drakeirving:
As for the previous commit, if we're going to have Pokemon cries for any notification (as @DiscoTim poked towards), it would be better to just load the sounds as-needed instead of preloading as you change around the notifications. It's probably more resistant to complexity in the notification system that way. Now it just loads once and creates one object per, too.

Also personally I think the HTML5 audio incompatibility on Android Chrome sounds like something they should be fixing, not us ;)

DiscoTim:
@drakeirving i agree, probably the best way to do it. there is another pr that introduces a cross browser compatible sound library, hopefully that goes through as well.

DiscoTim:
@drakeirving, probably not the best place to be asking this, but I'm working with audiosprites (http://www.createjs.com/docs/soundjs/classes/AudioSprite.html) do you mind if i use your cries?

primaxius:
-1 think of bandwidth use and does not work on certain devices

PR #498, opened 2016-08-11
Add updated time to gyms

maddhatter:
## Description
Added last update time to gyms.

## Motivation and Context

Since gyms never expire off the map, the team/prestige info can be outdated. This lets you see at glance if the data is recent.

## How Has This Been Tested?

Tested locally using a SQLite database pre-schema versioning, migrations were successfully applied, and the map handles unknown dates cleanly.

https://i.imgur.com/1nGH7fS.png (known update time)
https://i.imgur.com/wAkUmR0.png (unknown)

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -24,7 +24,7 @@ log = logging.getLogger(__name__)
 args = get_args()
 flaskDb = FlaskDB()
 
-db_schema_version = 2
+db_schema_version = 3
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -241,6 +241,7 @@ class Gym(BaseModel):
     latitude = DoubleField()
     longitude = DoubleField()
     last_modified = DateTimeField(index=True)
+    updated_at = DateTimeField(default=datetime.utcnow)
 
     class Meta:
         indexes = ((('latitude', 'longitude'), False),)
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -241,6 +241,7 @@ class Gym(BaseModel):
     latitude = DoubleField()
     longitude = DoubleField()
     last_modified = DateTimeField(index=True)
```
invisiblek: We can't just use this field?

maddhatter: This value gets sent back in the API response, not sure exactly what it is (my guess is the last time something in the gym changed, like prestige level, etc). Whatever it is, it doesn't correlate to the last time it was updated in the database.

invisiblek:
Right, it should change based on the last time something about that gym changed. If we're not updating the database when that happens, that's the real issue here.

maddhatter:
Yeah, but if I move my scanner to a different location, it'll no longer be updating that gym in the database. If I use `last_modified`, I can't easily determine if the gym just hasn't had anything happen to it in awhile, or if I'm not scanning it anymore.

FlarumHostingCo:
@maddhatter I just ran a scan of my neighborhood. took 3 gyms, scanned again..  All gyms on my map still say unknown and all gyms in the DB say NULL.

maddhatter:
Can you enable the Scanned Locations and confirm that those gyms were scanned again? What OS, etc?

jnovack:
I do not like "updatedAt".  Perhaps, "last_scanned" (also keeping with the underscore theme).

Thumbs Up/Down people?

Chlodochar:
@jnovack Javascript does not have the convention to use `_` (see also teamName, teamId, etc). The linter warns about this.

`last_scanned` instead of `updated_at` makes sense to me. `updated_at` implies the same thing as `last_modified`, but `last_scanned` is just the time it was last scanned.

jnovack:
Then `last_scanned` (in DB) and `lastScanned` in Javascript.  Regardless, "Updated at" is not good English.

maddhatter:
Closing in favor of #666

PR #504, opened 2016-08-11
Added scan delay to Beehive-Generator

MastermanT:
## Description

Added scan delay to Beehive-Generator

## Motivation and Context

Needed if server requests to throttle down.

## How Has This Been Tested?

Tested locally (Windows).

## Types of changes

## Checklist:

Needed if server requests to throttle down, default 10, increase/decrease if needed.

### DIFF
--- a/Tools/Hex-Beehive-Generator/location_generator.py
+++ b/Tools/Hex-Beehive-Generator/location_generator.py
@@ -8,6 +8,7 @@ parser = argparse.ArgumentParser()
 parser.add_argument("-lat", "--lat", help="latitude", type=float, required=True)
 parser.add_argument("-lon", "--lon", help="longitude", type=float, required=True)
 parser.add_argument("-st", "--steps", help="steps", default=5, type=int)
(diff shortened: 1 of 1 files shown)


rrbailey89:
Doesn't the current beehive generator already use the SD from the config.ini... so couldn't you simply change that value and restart?

MastermanT:
Yes it does and you could!
But this is also helpful if you need to generate a lot of different scripts with different step sizes etc.
Or if you wish to switch between normal and beehive.

I often use step size 7 and scan delay 5. (for personal use)
But for Beehive I can't use scan delay 5 because it's too much and the server will request to throttle.
Now I could alter the config file generate the script (sh/bat) and then change back, but that seems a bit like a hassle.
Maybe it should be slightly altered so it uses config unless specified otherwise.
I chose scan delay 10 because it's the default in the Pokemon Go Map.

primaxius:
-1 you can specify scandelay in your beehive file by adding -sd arguments to the lines

Issue #462, opened 2016-08-10
First search step of every worker returns 0/0/0


Shobba:
## Expected Behavior
After successful login every worker should find something (spawn, pokestop and/or gym) during its first search step and following steps.

## Current Behavior
After successful login the first search step of every worker (even with just one account) returns 'Upserted 0 pokemon, 0 pokestops, and 0 gyms'. Starts working correctly with its seconds search step.

```
2016-08-11 00:27:51,358 [      MainThread][        models][    INFO] Connecting to local SQLite database
2016-08-11 00:27:51,361 [   search_thread][        search][    INFO] Search overseer starting
2016-08-11 00:27:51,363 [   search_thread][        search][    INFO] Starting search worker threads
2016-08-11 00:27:51,367 [   search_thread][        search][    INFO] New location caught, moving search grid
2016-08-11 00:27:51,370 [ search_worker_0][        search][    INFO] Search step 1 beginning (queue size is 1)
2016-08-11 00:27:51,372 [ search_worker_0][      auth_ptc][    INFO] PTC User Login for: xxxxxxx
2016-08-11 00:27:52,084 [ search_worker_0][      auth_ptc][    INFO] PTC User Login successful.
2016-08-11 00:27:52,084 [ search_worker_0][      auth_ptc][    INFO] Request PTC Access Token...
2016-08-11 00:27:52,224 [ search_worker_0][      auth_ptc][    INFO] PTC Access Token successfully retrieved.
2016-08-11 00:27:53,095 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-11 00:28:02,374 [ search_worker_0][        search][    INFO] Search step 2 beginning (queue size is 59)
2016-08-11 00:28:06,467 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 539 pokestops, and 17 gyms
2016-08-11 00:28:13,376 [ search_worker_0][        search][    INFO] Search step 3 beginning (queue size is 58)
2016-08-11 00:28:17,374 [ search_worker_0][        models][    INFO] Upserted 2 pokemon, 529 pokestops, and 18 gyms
```

Two workers:
```
2016-08-11 00:35:52,750 [      MainThread][        models][    INFO] Connecting to local SQLite database
2016-08-11 00:35:52,754 [   search_thread][        search][    INFO] Search overseer starting
[...]

PR #507, opened 2016-08-11
Workaround for first search step returning 0/0/0 

taihun81:
In response to issue #462 

I'm not sure if this is a valid solution but delaying after initial login does seem to resolve first search steps resulting 0/0/0.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -292,6 +292,9 @@ def check_login(args, account, api, position):
 
     log.debug('Login for account %s successful', account['username'])
 
+    # To avoid first search step resulting 0/0/0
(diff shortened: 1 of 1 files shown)


jsynx2:
Did you modify some .py file for that taihun81 ?

taihun81:
Changed from using scan delay to fixed 10s, however I'm thinking better way is to create an additional arg.  Anyway, when I tested myself, delay of 1-9s will give out 0/0/0 on the first search, and as the delay gets closer to default known scan delay of 10s, I observed reduced occurrence of 0/0/0 so I decided to give 10s for added reliability.

PR #511, opened 2016-08-11
Use lat/long as a composite key for ScannedLocation

invisiblek:


jonathanwcrane:
I would like it if @antiroot looked at this since he identified the bug in the first place.

invisiblek:
-1 this patch sucks, i'll come up with something better

invisiblek:
patch still sucks, but at least it doesn't cause deadlocks and too many connections since moving it down

antiroot:
Better than nothing! My temporary solution in my local repo was to generate a numeric 32bit hash based on the location, but then there's the collision possibility if a database has millions or records (not a problem for me with only ~1500 records). So your patch seems better than my idea in that regard. shame it requires an extra query, but still better than a string index in my opinion

invisiblek:
It should be possible to use a composite key here....looking into it...

invisiblek:
Latest patch works (happiest I've been so far with the database situation on this thing)

I don't like the javascript though. Need to make that a bit more elegant...

antiroot:
I like it, fixes the initial issue and doesn't involve an extra query. I'm curious if floating point precision will be an issue, although if it were then the queries with where clauses using the lat and long fields would have been affected long ago I would assume.

invisiblek:
Yea, plus we're dropkicking the table before we move to this anyway. (I consider the scannedlocation table to be completely, 100% expendable since it it essentially recreated every 30 minutes)

antiroot:
I've been running the most recent patch for a few hours with the clean_database method set to 4 hours and everything seems is resolved and is running stable

jonathanwcrane:
Pretty awesome work IMHO.

@invisiblek can you give me a feel for the plan for this project?  How are milestones determined and are there timeframes?

rrbailey89:
@invisiblek i have merged this patch and I went from seeing this [](http://i.imgur.com/lYxaQIg.png)

to see this http://goo.gl/n4eEKa

antiroot:
@rrbailey89 the patch modifies frontend javascript files so you have to rebuild them in order to see the changes, I'd also recommend clearing your browser cache just to make sure nothing stale is being served from disk.

rrbailey89:
@antiroot thanks that fixed it

PR #513, opened 2016-08-11
Added webhooks.md

ImAnOwl:
I added a article to the docs about webhooks

### DIFF
--- a/docs/extras/webhooks.md
+++ b/docs/extras/webhooks.md
@@ -1,0 +1,54 @@
+# How to Setup Webhooks on Linux with Node.js
+
+## Supported Events
+* Pokemon Spawn
+
+## Requirements
+
+* Node.js (min. v4.4.7)
+* Python (WORKS ONLY WITH 2.7)
+* PiP (min 8.1.2)
+* NPM
+* A linux distribution
+
+## Webserver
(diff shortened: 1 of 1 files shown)


adizam:
Discussing with a few others. My first response is that I believe this needs more information about webhooks in general. This shows how to create your OWN webhook in node.js. It doesn't explain how the project itself spits out data to a webhook (POSTs data).. how to configure it.. or how a project such as yours would suffice for an "inbetween" to kind of "map" data to an alternative shape.

chuyskywalker:
@adizam is right, we can make this much better -- a structure like this would help users a lot:

```
# Webhooks with PokemonGo-Maps
## About webhooks; The What and Why
## Webhooks PokemonGo-Map Can Send
## Example Webhook Receiver Implementations
### Python
### Node
## PokemonGo-Map Webhook Compatible Projects
```

ImAnOwl:
Thank you, Im going to edit this.

jnovack:
I think a bad commit made it's way in here.

+2 for using NodeJS though in your examples.

primaxius:
Declined - needs to be more detailed as specified by https://github.com/PokemonGoMap/PokemonGo-Map/pull/513#issuecomment-239348466

PR #514, opened 2016-08-11
fix undefined variable usage

mattoufoutu:
## Description
An error was introduced by 83fab06f4fea54a58d0b0cebb95bed24551037c7 where a variable is defined in the `else` part of an `if` block, but not in the first one, resulting in the variable being used before it is defined. 

## Motivation and Context
This change is required because it fixes a bug introduced by a recent commit.

## How Has This Been Tested?
After fixing and running the map for about 30min, I don't notice the error anymore and don't notice any error that could have been introduced by this change.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -352,6 +352,7 @@ def parse_map(map_dict, step_location):
                         'active_fort_modifier': f['active_fort_modifier'],
                         'is_lured': True
                     }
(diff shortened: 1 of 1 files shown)


Chlodochar:
Sorry for late reply. This has already been fixed with changes to webhook.

PR #518, opened 2016-08-11
-ss STEP_SIZE Take Three

Mantriur:
Feature Request: https://github.com/PokemonGoMap/PokemonGo-Map/issues/72

First Pull Request: https://github.com/PokemonGoMap/PokemonGo-Map/pull/161

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -31,6 +31,8 @@ from pgoapi.exceptions import AuthException
 
 from .models import parse_map
 
+from .utils import get_args
+
 log = logging.getLogger(__name__)
 
 TIMESTAMP = '\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000'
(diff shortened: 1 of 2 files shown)


imynwulf:
I manually applied your last PR and it's working great.

reyqn:
I don't really see why you would want a pulse radius different from 0.07 (with Pokemon) or 1 (no Pokemon)

SchwarzwaldFalke:
There was a discussion about step_size/pulse_radius on github:
https://www.reddit.com/r/pokemongodev/comments/4wy3h0/psa_if_you_are_scanning_a_large_area_for/

> The big problem comes when you hit around 13~ steps with the 1.0 pule_radius, as your accounts start getting soft-banned for moving too far.(This would happen at probably like 30 steps for your .3 anyways, but yeah)

> I'd love to use a higher value, but in my testing it missed a few even at .35 :-/
The pokestops that I missed were right between who scans, and then it found some in the next rings, so the accounts were not softbanned.

So I guess 0.07 and 1 isn't enough because it depends on your setup which step_size you need.

jnovack:
I personally WANT this. 

I am not scanning for Pokemon, only Gyms status.  Therefore I want a large sweeping step size so that I can cover a good few miles.  

Also, people looking to map points of interest (Pokestops and Gyms) could use this for their database dumps.

Somnium42:
+1
I also like the feature for the exact same reasons jnovack has given.

slideup-benni:
the argument "-ss" is used for spawnpoint-scanning now
https://github.com/PokemonGoMap/PokemonGo-Map/commit/c36d1acb144e96f81c70fecc41ff89ba4ff317cd

jnovack:
And #748 provides an automatic method for deciding scan distance so as to not kill any puppies (read: add ANOTHER variable/option)

primaxius:
Closed as -ss allready used. Not updated in a month.

PR #523, opened 2016-08-11
Simplified and fixed scan retries

Sunakujira1:
## Description
Scan retries had odd delay mechanism and the scan delay was not used after successful recovery after several retries.

## Motivation and Context
Scan retry works more as expected, `-sd 10 -sr 12` should try 10 times with roughly 120 seconds.
Fixing the scanner from bursting several tries over few seconds after recovering from retries.

## How Has This Been Tested?
Local instance, which is now returning less (0 pokemons, 0 stops, 0 gyms) due to no more burst after retries.

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

Removed oddly custom delay between retries, and made it linear. (Ex. `-sd 10 -sr 12` should try up to 12 times with 10 delay each ~= 120 seconds at most.)  
Fixed scan burst issue occurring after successful recovery from retries.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -194,9 +194,6 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
             if args.proxy:
                 api.set_proxy({'http': args.proxy, 'https': args.proxy})
 
-            # Get current time
-            loop_start_time = int(round(time.time() * 1000))
-
             # The forever loop for the searches
             while True:
 
@@ -211,6 +208,9 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                 # The loop to try very hard to scan this step
                 failed_total = 0
                 while True:
+ 
+                    # Get current time
+                    loop_start_time = time.time()
 
                     # After so many attempts, let's get out of here
                     if failed_total >= args.scan_retries:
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -221,11 +220,6 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                         log.error('Search step %d went over max scan_retires; abandoning', step)
                         break
 
-                    # Increase sleep delay between each failed scan
-                    # By default scan_dela=5, scan_retries=5 so
-                    # We'd see timeouts of 5, 10, 15, 20, 25
-                    sleep_time = args.scan_delay * (1 + failed_total)
-
```
chuyskywalker: I don't like removing this -- having some kind of backoff implementation is a polite thing to do for failing looped requests.

Sunakujira1: @chuyskywalker I think that should be a feature of it's own, rather than enforcing it.  
When scan delays used to be 1 second, it made sense to do so. But it being 10 seconds by default, the increment is too much.  
Why wait 20, 30, 40 seconds, when 10 second interval is already very polite?

primaxius:
Declined outdated - Resubmit updated version if needed but in current state declined.

PR #541, opened 2016-08-12
Update pokemon.json

Omikakim:
## Types of changes

## Checklist:

Adjusted the rarity of the Pokémon

Very Common
Common
uncommon
Rare
Very Rare
Special
Epic
Legendary

### DIFF
--- a/static/data/pokemon.json
+++ b/static/data/pokemon.json
@@ -1,7 +1,7 @@
 {
   "1": {
     "name": "Bulbasaur",
-    "rarity": "Uncommon",
+    "rarity": "Common",
     "types": [
       {
         "type": "Grass",
@@ -16,7 +16,7 @@
   },
   "2": {
     "name": "Ivysaur",
-    "rarity": "Rare",
+    "rarity": "Special",
     "types": [
       {
         "type": "Grass",
@@ -31,7 +31,7 @@
   },
   "3": {
     "name": "Venusaur",
-    "rarity": "Very Rare",
+    "rarity": "Epic",
     "types": [
       {
         "type": "Grass",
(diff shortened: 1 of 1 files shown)


FrostTheFox:
I agree with some of these, but some of them have been needlessly changed as well.

SmartPappnase:
I updated this list locally to my personal likeing and restartet the server, however, the changes seam to be ignored in the browser. Does the notify grouping refer to a different file?

Gagaro:
What did you use to determine the rarity ? Your location also affects the pokemon spawning, so rarity will depend on your location and won't be the same for everyone.

I also don't like the term you used. Is special better than rare? Or is rare better than special? Why remove Ultra Rare, which was easy to understand?

I agree that some of the rarities seem off, but I think adding a "Very Common" one would be enough. 8 rarities are a lot.

Cronick:
@SmartPappnase You need to use "npm run build" again, as it makes a min version of the json file.

Slivo-fr:
Agree with gagaro

Romfrosk:
How will this affect my current db of over 1 million spawns

tim95030:
The original rarities were grabbed based on spawn rates from a database. I really don't think adding these weird names makes any sense. I agree some of the things are not quite right from my original implementation. The hope was to eventually use the statistics to maybe run a task to update the rarities based on your location as the rarities are definitely heavily influenced by location. For example people see drowzees all the time in some places. By me there are maybe 1-2 per day. That is not common. I also heavily disagree on the term "special" this has absolutely no meaning. Also you will need to update all of the foreign languages if you make this change.

JZ6:
Biased changes I guess, lets just keep the older labels

SmartPappnase:
@Cronick Thanks a lot!!! now it works.

PR #555, opened 2016-08-12
Fixed lure expiration time

Shobba:
## Description
Calculation of the expiration time for lures was wrong, "timeUntilExpire" to be exact.
screenshot 2016-08-12 14 04 01

## Motivation and Context
Display the correct expiration time

## How Has This Been Tested?
Checked in my own front end

## Types of changes

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1051,9 +1051,9 @@ function gymLabel (teamName, teamId, gymPoints, latitude, longitude) {
   return str
 }
 
-function pokestopLabel (lured, expireTime, latitude, longitude) {
+function pokestopLabel (expireTime, latitude, longitude) {
   var str
-  if (lured) {
+  if (!!expireTime) {
     var expireDate = new Date(expireTime)
 
     str = `
@@ -1193,7 +1193,7 @@ function setupPokestopMarker (item) {
   })
 
   marker.infoWindow = new google.maps.InfoWindow({
-    content: pokestopLabel(!!item['lure_expiration'], item['lure_expiration'], item['latitude'], item['longitude']),
+    content: pokestopLabel(item['lure_expiration'], item['latitude'], item['longitude']),
     disableAutoPan: true
   })
 
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1051,16 +1051,10 @@ function gymLabel (teamName, teamId, gymPoints, latitude, longitude) {
   return str
 }
 
-function pokestopLabel (lured, lastModified, latitude, longitude) {
+function pokestopLabel (lured, expireTime, latitude, longitude) {
   var str
   if (lured) {
-    var lastModifiedDate = new Date(lastModified)
-    var currentDate = new Date()
-
-    var timeUntilExpire = 1800000 - (currentDate.getTime() - lastModifiedDate.getTime())
-
-    var expireDate = new Date(currentDate.getTime() + timeUntilExpire)
-    var expireTime = expireDate.getTime()
+    var expireDate = new Date(expireTime)
 
     str = `
       <div>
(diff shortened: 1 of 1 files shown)


mattskey:
perfect, working spot on in my testing, comparing in game and the map, stops are turning on/off in sync. 
thanks!

invisiblek:
can't we utilize lure_expiration for this?

Shobba:
In previous commits "lure_expiration" was used for lured pokemon (3min timer). Don't know right now if something changed.

static/js/map.js
```diff
@@ -1199,7 +1193,7 @@ function setupPokestopMarker (item) {
   })
 
   marker.infoWindow = new google.maps.InfoWindow({
-    content: pokestopLabel(!!item['lure_expiration'], item['last_modified'], item['latitude'], item['longitude']),
+    content: pokestopLabel(!!item['lure_expiration'], item['lure_expiration'], item['latitude'], item['longitude']),
```
invisiblek: wouldn't really need to send the first argument here since we could do that check in pokestopLabel

invisiblek:
Here's an example of the data you get for a lured pokestop (when your location is more-or-less right on top of it): https://gist.github.com/invisiblek/0ae0f7302b81ef31b0ec403912ee7f81

Here's how we calculate lure_expiration on the back end: https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/models.py#L344-L345

invisiblek:
The real issue here is that we're basing everything on last_modified_timestamp_ms + 30m.

I think we need to be using lure_expires_timestamp_ms from lure_info (which we don't always have).

Shobba:
I see. If we check for lure_info (instead of active_fort_modifier) lure_expiration will be overwritten if the scan is too far away, right?

invisiblek:
Quite honestly, I think the best thing we can do here is make another API call. Check this out: https://gist.github.com/invisiblek/8f47f0ad64b49bfd33efbc184fdae352

That's making a 'FORT_DETAILS' call to Niantic for that fort_id, latitude and longitude.

It would give us the added benefit of getting the pokestop name and image url too =)

Shobba:
I understand the problem, yes. But right now it seems like last_modified_timestamp_ms only changes when somebody puts a lure. So right now it should be fine to take last_modified_timestamp_ms + 30m for the expireTime and check if that's in the past or future. I think that's already somewhere in the code because the frontend map seems to work correctly (pokestops change to lured and back after the lure expired)

invisiblek:
I just used this same patch and watched a lured pokestop. When our map calculated it'd be expired, it didn't in-game. I refreshed the map and it showed like 4 minutes left on it. Waited that out and same thing, wasn't expired in-game. Refreshed again and it showed 8 minutes.

Shobba:
That's strange. I cannot confirm this. I cannot check in-game but if the map says a pokestop is expired it changes, and when I refresh it stays normal, not lured again.
What does a recently lured pokestop show on your map? Timer should start at ~30:00.

SkOODaT:
Im Pretty new to python etc..  but isnt thier some way you could have workers that are close enough pick up the data if its there store it in the database and ignore it if it has an active_fort_modifier untill active_fort_modifier drops and prevent the data from being overwriten ....distance workers wont add anything to the data base but it would be better then having no lured pokemon showing, just a add it if it can if its there type thing

Zoruk:
I made this: https://github.com/keyphact/pgoapi/pull/100 with this one the lurd_info field work well

invisiblek:
on hold pending api update

invisiblek:
https://github.com/PokemonGoMap/PokemonGo-Map/pull/736

SkOODaT:
This seems to be working in obtaining lured pokemon and adding them onto the map https://github.com/Zoruk/PokemonGo-Map/commit/c95f4ac68af4a7ed969cc5d04c891f398b80165b

PR #568, opened 2016-08-12
Removed duplicate results in Pokemon.get_seen

GenuineSounds:
## Description
Removed duplicate pokemon in the results affecting total pokemon count in the full stats page.

This affected the total count making all percentage calculations erroneous and also making the outputted json data larger than it needed to be.

## What This Fixes
EVERYTHING! But more specifically the /stats page.
* Correct total count.
* Total count jumping around sporadically.
* Correct percentage.

## Motivation and Context
This bug was due to sending more than one result per pokemon if there were more than one with the same `lastappeared` time or more precisely `disappear_time`.

## How Has This Been Tested?
Double checked any other statistics and if they would be affected, no problems I could find.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -162,6 +162,7 @@ class Pokemon(BaseModel):
                          pokemon_count_query.c.count)
                  .join(pokemon_count_query, on=(Pokemon.pokemon_id == pokemon_count_query.c.pokemon_id))
                  .where(Pokemon.disappear_time == pokemon_count_query.c.lastappeared)
+                 .group_by(Pokemon.pokemon_id)
                  .dicts()
                  )
         pokemons = []


Infernoman:
+1

Chlodochar:
This needs to be fixed and this seems like the right approach, but I'm not sure whether this fix actually solves the problem.

When there are two pokemon of the same type that both last appeared at the same time (both have `lastappeared == max(lastappeared)`), which latitude and longitude are selected by the query? If you group by pokemon_id, will the latitude and longitude be averaged? Or will both pokemon be returned? For raw SQL I'd say this query would fail, because the columns in the `select` section do not match the columns in the `group by` section.

Chlodochar:
You might want to use `fn.MAX(` around latitude and longitude as well as around `pokemon_count_query.c.count`. group by pokemonid and disappear_time.

primaxius:
Fixed with https://github.com/PokemonGoMap/PokemonGo-Map/pull/988 - closed

PR #581, opened 2016-08-13
Persistent Data of Pokemon Notifications/Exclusion

lix7:
## Description
When restarting the server and browser, the list of pokemons to notify/exclude/rarity are cleared. I've created json file to save those list so you don't have to keep updating the list. 

The file is located in ROOT/static/pokelist.json
It has a few pokemons already in the file to provide an example of how to modify it and also to make people aware of this file. Although this should be discussed if the json file should be empty so people dont think this is some kind of bug and mention this in the documentation. An alternative approach can be to include the location of pokelist.json in the placeholder text of the notify/exclude/rarity text boxes so people know where to look.

The contents of pokelist.json: Exlcludes pidgey, rattata, raticate, drowsee. Nofifies charizard, blastoise, venusaur, mewtwo
{
 "exclude" : [16,20,19,96],
 "notify"  : [3,6,9,150],
 "rarity"  : ["Ultra Rare","Rare"]
}

The order of the pokemon numbers are not important. Currently, this feature only reads data but doesn't write. For example, modifying the list in the browser will not update the pokelist.json file. 

## Motivation and Context
Many have requested for this feature. It also very hard clicking the map on mobile when there too many pokemon on the map.

## How Has This Been Tested?
I've tested this in Firefox and Chrome. Once the webpage loads. The list of pokemon from the file should appear in the corresponding dropbox menus of the Options tab.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1847,6 +1847,18 @@ $(function () {
     $selectRarityNotify.val(Store.get('remember_select_rarity_notify')).trigger('change')
   })
 
+  // Load saved list of pokemon to exclude and notify on map
+  $.getJSON('static/pokelist.json').done(function (data) {
+    Store.set('remember_select_exclude', data.exclude)
+    $selectExclude.val(Store.get('remember_select_exclude')).trigger('change')
+	
+    Store.set('remember_select_notify', data.notify)
+    $selectPokemonNotify.val(Store.get('remember_select_notify')).trigger('change')
+	
+	Store.set('remember_select_rarity_notify', data.rarity)
+    $selectRarityNotify.val(Store.get('remember_select_rarity_notify')).trigger('change')
+  })
+
   // run interval timers to regularly update map and timediffs
   window.setInterval(updateLabelDiffTime, 1000)
   window.setInterval(updateMap, 5000)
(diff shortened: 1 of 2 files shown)


GenuineSounds:
> When restarting the server and browser, the list of pokemons to notify/exclude/rarity are cleared.

There is no problem with the persistent state of the webpages' settings unless it's on your client only. I don't think this is necessary at all, and especially not as a global state for all users that connect to the server. Fix your browser not to delete everything on close or deal with it.

BaIthamel:
Agreed this is not an issue for me, i can clean and reinstall the repo and my browser still holds the settings.

chuyskywalker:
While the above statements are true, there's been a lot of chatter from other places where people do want default lists. But I'd ask for them to be compatible with the `arg`/`config` setup we already have instead of making people code in yet another format for configuration.

CometX:
This is awesome, please keep this current and hope for merge! Would be nice if you could make it work with current config file as well.

CometX:
Does this still work with current Dev branch?

primaxius:
This is an issue with your browser. Not an issue that needs fixing on the servers.

Issue #321, opened 2016-08-09
Searching is slow when using multiple accounts.


RollTideDG:
## Expected Behavior
There is a noticeable delay between initiation of the scan and upserting of the records.

## Current Behavior
The first time through the loop of worker threads it seems to work alright. As it progresses through the queue the time from initiating the scan until the upserting the records takes a really long time. This causes your scan to be really slow even if you have multiple accounts.

Here is an example from my running scanner:

2016-08-09 12:08:55,963 [ search_worker_3][        models][    INFO] Upserted 0 pokemon, 526 pokestops, and 21 gyms
2016-08-09 12:08:56,052 [ search_worker_2][        search][    INFO] Search step 61 beginning (queue size is 30)
2016-08-09 12:09:02,489 [ search_worker_4][        models][    INFO] Upserted 1 pokemon, 514 pokestops, and 18 gyms
2016-08-09 12:09:03,588 [ search_worker_3][        search][    INFO] Search step 62 beginning (queue size is 29)
2016-08-09 12:09:09,174 [ search_worker_5][        models][    INFO] Upserted 2 pokemon, 522 pokestops, and 19 gyms
2016-08-09 12:09:10,063 [ search_worker_4][        search][    INFO] Search step 63 beginning (queue size is 28)
2016-08-09 12:09:15,555 [ search_worker_6][        models][    INFO] Upserted 0 pokemon, 513 pokestops, and 20 gyms
2016-08-09 12:09:16,759 [ search_worker_5][        search][    INFO] Search step 64 beginning (queue size is 27)
2016-08-09 12:09:23,015 [ search_worker_7][        models][    INFO] Upserted 1 pokemon, 554 pokestops, and 21 gyms
2016-08-09 12:09:23,154 [ search_worker_6][        search][    INFO] Search step 65 beginning (queue size is 26)
2016-08-09 12:09:29,785 [ search_worker_8][        models][    INFO] Upserted 2 pokemon, 530 pokestops, and 19 gyms
2016-08-09 12:09:30,576 [ search_worker_7][        search][    INFO] Search step 66 beginning (queue size is 25)
2016-08-09 12:09:36,976 [ search_worker_9][        models][    INFO] Upserted 0 pokemon, 589 pokestops, and 20 gyms
2016-08-09 12:09:37,338 [ search_worker_8][        search][    INFO] Search step 67 beginning (queue size is 24)
[...]

PR #582, opened 2016-08-13
Scan Speed Repair

chuyskywalker:
re: #321

The basic idea here is that this will split searching and parsing into in-memory models, and the inserting of this data, into separate threads.

Discovered in #321, when you have a lot of accounts, an each one needs half a second to "lock" while inserting to the database, you end up locking EVERY search thread. Since all the threads lock, scanning can take _forever_.

This splits out the database into a queue of updates. This is useful for a number of reasons. Firstly, it lets the search workers get on with life and return control without locking out other search workers. Secondly, this gives us serial execution of inserts (like the lock did) but, again, without locking threads that didn't need to be locked.

### ~~This is **VERY** WIP~~ 

I've ripped out webhook reporting while I'm juggling around things. I'll likely bring back webhooks as its own thread _as well_ since it could cause some of the same kind of search-thread delays when it shouldn't.

I need people to give this a try who have more accounts than I do to see if it helps alleviate the scan delays. I've added a TON of debug timings (`-d` on, y'all). I also have some open question about where the `connect_db()` method should go


pogom/models.py
```diff
@@ -378,47 +381,93 @@ def parse_map(map_dict, step_location):
...
-        bulk_upsert(Pokestop, pokestops)
+                # todo: I'm pretty sure that, since this is the only real db
+                #       interaction, we could open the connection outside
+                #       the loop and just keep it indefinitely (unless it
+                #       closes/etc). Reconnecting, essentially, if it bails
+                #       because the loop starts over.
+                #
+                #       However, with sqlite, on my machine, the `conn` step
+                #       is instant, so I'd need someone with mysql to see if
+                #       this matters.
```
j16sdiz: This matters when mysql is not on the same machine

chuyskywalker: Indeed, which is why the todo plea is here. MySQL is much better when you have long lived open connections, but I don't have a mysql setup, so I've not tested it just how bad the connection/close impact of remote sql is. 

Which is why I left the ask in here so someone with that setup could try it and give me some feedback on just how bad the delays is.

hemsleyk: +1 for this issue. Running MySQL (3-way Galera+MariaDB cluster actually) for my univeristy-wide map with lots of accounts and its definitely not scanning as fast as it could be ... I also notice memory and CPU usage slowly rises.

Pretty sure this should be done unconditionally though. Even with SQLite there's no reason to close the file after every write operation, when we expect the next one within 10 seconds. Commercial local-DB apps like Outlook definitely keep their DBs open indefinitely.

hemsleyk:
Preliminary testing is promising ... 20 step area with 20 accounts is more than keeping up with one using 70. Identical SQL database setup (load-balanced Galera+MariaDB cluster on the same servers, but separate docker containers from "production").

chuyskywalker:
Cool, thanks. Looking forward to hearing back.

Chlodochar:
I know this is WIP, but what happened to `send_to_webhook`? It doesn't seem to be called right now.

hemsleyk:
Update: Definitely more efficient. CPU / Mem usage didn't blow up over time either.

chuyskywalker:
@Chlodochar https://github.com/PokemonGoMap/PokemonGo-Map/pull/582/files#diff-541e2b217ba4a7d7859dae2cb443b6b7R350

chuyskywalker:
Which i plan of sussing out this afternoon.

hemsleyk:
Nice. Message me here by the way, got banned on Discord for badmouthing PVP bot devs / users.

chuyskywalker:
@Chlodochar I've fixed webhooks, made them better (in fact) and even made a demo receiver :D

@hemsleyk I've moved the connection phase outside the loop and no longer close the connection. You should give this another go and let me know how it feels.

At this point, I think this is basically ready to go. Splitting all these various concerns out to threads is kinda nice because it lets python do the smart/hard work of scheduling them as efficiently as possible. So yay, threads!

hemsleyk:
Is it still schema ver 3 or ver 4?

Can move the container w/ the accounts back my test image to the testing SQL cluster from production if so. Can run directly on production if the schema ver. is caught up with develop.

jjflipped:
Has anyone tried doing this with #585? I was going to try and parse out if I could merge it, but the files all seem pretty different.

hemsleyk:
Spinning up a test instance now.

hemsleyk:
Test is running on production dataset since db schema seems v4 now on account of the merge from upstream.

Much better than production. This is easily my "healthiest" container. Logs pop scans off the queue much faster than the develop branch.

chuyskywalker:
I'm happy to consider this a success. I'll pull out all the debugging stuff and wrap it up.

hemsleyk:
Whoah, just tried branch on 50 accounts and huge scaling issue. The upsert queue is over 1000. 20 accounts is fine. Its not the SQL cluster because the rest of the docker containers (locations) are keeping up.

chuyskywalker:
I suspect that the db *isn't* actually keeping up on your other containers, but the scan speed is being substantially slowed down by the shared lock across all scan threads.

hemsleyk:
The DB itself isn't crying in its logs and the hosts are happy in terms of CPU/MEM ... I'm using MariaDB/Galera but its supposed to be drop-in mysql, is there a command I can run to see the health of the actual DB instances?

chuyskywalker:
I don't have that many accounts to test with; nor mysql setup. I'll probably need to come up with a mock api if I want to get that level of speed/scale for testing.

hemsleyk:
Interestingly, even with same no. accounts and step size as the one that's working, it still chokes ... could be something with my load-balancing or container scheduling. Both are dense areas.

jjflipped:
@hemsleyk gyms and stops turned off on both?

hemsleyk:
@jjflipped Nope, enabled.

hemsleyk:
Upgraded my whole stack to my test image hemsleyk/pogom and that area still had issues where the others did not ... disabling pokestops fixed it. I now have a separate docker container on step limit 2000 doing just pokestops that I will disable and reclaim the accounts from in a couple days ;)

On that note. It would be nice to have a GYMS_BASIC flag that intentionally returns all gyms as clear, or preferably some form of NO DATA state (if latter, another scanner without the flag would override w/ detailed gym data). I might put in on feathub or fork and DIY.

hemsleyk:
https://radar.valor.science/

Here's my URL. The default location (University of South Florida) is the problepeno. South of there you'll find Ybor which should be worse or at least as bad but it works fine. I think there's some weird issue with pokestops.

chuyskywalker:
So, after working with this a bit more, I'm realizing that there's going to simply be an upper limit to how fast the scan can run, and that adding more accounts will have a diminishing, if not negative, return.

Python, at its core, is single threaded. There's a point where you just run out of CPU mhz to give the whole thing and it can't do no more. The threading here, and removing global locks, is enormously helpful in that it frees up the program to go do more with the cycles it _does_ have, but unless we re-write for multiprocessing (actual multiple processes), there's only so far we can go.

---

In order to really test this, we need to be able to absolutely hammer requests -- using real accounts/ips/proxies to do that SUUUUUUUCKS, so I went ahead and implemented an awesome little `fake-pgo-api.py` app which, combined with the revamped `-m` option, will use the fake pokemon go api server instance instead of a real one. And the fake one you can scan against very, very quickly.

chuyskywalker:
Summary:
 - Remove locking around map parsing
 - Push all db updates into a queue for a single thread to execute through (less contention)
 - Push all webhook updates to queue/thread, but only if enabled
 - Add an example webhook receiver code
 - Add a sufficiently functional pgoapi emulation server for testing

Infernoman:
@chuyskywalker I merged this into my repo, been playing around with it a bit. with a larger number of accounts i needed to greatly increase the number of threads for the db_updater. I also removed the webhooks just for testing. With 404 workers running i matched the number of db_updater threads. and it is running much faster than with the original. but I feel its still having some issues with that number of workers.

chuyskywalker:
> Changed it to 1 DB worker for every 20 workers

What do you mean by that?

Infernoman:
# Thread to process database updates
    for i in range(int(round(len(args.accounts) / 20))):
        log.debug('Starting db-updater worker thread %d', i)
        db_updater_thread = Thread(target=db_updater,
            name='ss_search_worker_{}'.format(i),
            args=(args, db_updates_queue)
        )
        db_updater_thread.daemon = True
        db_updater_thread.name = 'db-updater-{}'.format(i)
        db_updater_thread.start()

Infernoman:
there is probably a better method to this you can also reach me on the discord channel.

chuyskywalker:
I've added options for both db and webhook updater thread counts. We don't really _need_ more threads, per se, but having more gives them a higher chance of getting their turn when fighting against 100's of search threads.

Very few people will need to turn that up, but the option is there now.

PR #585, opened 2016-08-13
added optional spawnpoint scanning

blindreaper:
#608 

Added optional spawnpoint scan
## Description

Pretty much implemented sowoks changes since he didnt want to submit a PR https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into

**If you want to try this:**
Download from https://github.com/blindreaper/PokemonGo-Map/tree/spawnpointscan or if you have git installed:
`git clone -b spawnpointscan https://github.com/blindreaper/PokemonGo-Map.git`
Follow the rest of the instructions here: https://pgm.readthedocs.io/en/develop/basic-install/index.html
While you're there read the rest of the wiki

**Getting started**
**Step 1:** Run normal beehive scanning for an hour* for locations you wish to monitor
**Step 2:** Now run the scanner with -ss mylocation.json, this will dump the coords to the json file for you which is required for spawn point scanning. Optionally use -ne "lat long" -sw "lat long" to only dump locations within the box determined by ne and sw.
**Step 3:** Profit

*hour if you dont get any bans, lockups lag etc. Each "circle" needs to be scanned at least 4 times every 15 mins for the hour, if any fail, you may miss a spawn.

**Example (dedicated server thread + 2 scanner threads):**
python runserver.py -os -l "antartica" -k 321ZA2@#!$XDV && 
python runserver.py -ns -k 321ZA2@#!$XDV -u user1 -p pass1 -u user2 -p pass2 -ss myspawns-newyork.json && 
python runserver.py -ns -k 321ZA2@#!$XDV -u user1 -p pass1 -u user2 -p pass2 -ss myspawns-china.json

Alternatively use the config file because multiple -u and -p is messy.

**Added switches and modifications**
**-st** **x** has no effect, it's hard coded to 1 with -ss.
Location has no effect on dedicated scanner threads (-ns -ss), json file is used for location. Location is still needed for the webserver as a default start location for when the map loads.
Specifying **-ss** **x****.json** (when **x****.json** does not exist) will dump **all** coords and generate **x****.json**.
Specifying **-ne** and **-sw** when generating json will only grab spawns in the bounding rectangle.

## Motivation and Context
Get it into the code, improve/optimize later, add features

## How Has This Been Tested?
My machine

## Types of changes
Added new overseer thread so original is untouched

## Checklist:


WeissJT:
Thanks for implementing this. However it's not working for me.
I'm using your spawnpointscan branch and it seems that is not writing the information to the db. The db is created but stays at 23 kb.
So the map is not showing anything. No visible error in console.
Testing with 900 spawnpoints. On a 3 day old PoGo-Map build + the sowok search.py is working fine.

blindreaper:
Travis complaining about unused variable
`parsed = parse_map(response_dict, step_location)`
so i removed the whole line without thinking

WeissJT:
Thanks! Now it's working fine.

Cy4n1d3:
Will test :)

cris11368:
I can only recommend you also include an option to limit the scan range as well. I have already manually implemented tbterra's changes to my map but I also included this:

https://www.reddit.com/r/pokemongodev/comments/4xa4jp/idea_make_a_definitive_spawn_database_we_can/d6ezral

to limit the scan range because it gets out of hand when you get into the thousands of spawn points over a very large area.

blindreaper:
this pr is for the base implementation. 

once its accepted people can submit pull requests for further optimisation and features.

3kynox:
How to use -ss option please ?

Cloned spawnpointscan branch and get : 

```
2016-08-13 17:19:30,839 [   search_thread][        search][    INFO] Starting search worker threads
Exception in thread search_thread:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/nox/PokemonGo-Map-Unofficial/pogom/search.py", line 227, in search_overseer_thread_ss
    with open('spawns.json') as file:
IOError: [Errno 2] No such file or directory: 'spawns.json'
```

WeissJT:
@3kynox 
You need a spawns.json file created either by https://github.com/TBTerra/spawnScan or exporting a PoGo-Map database to spawns.json using the script provided in this PR / [using this instructions](https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into/d6au9oh)

GraidenLOL:
@WeissJT does it work anymore? I'm getting a lot of retries and downloads failed using the spawnScan.

aldarund:
I think if this spawn.json is needed for this to be working it should be created automatically in case it doesnt exist

WeissJT:
@GraidenLOL 
I used it yesterday and it was working fine, I'll try again in a few hours.

Ardarel:
Hello, I am not sure how to merge this change localy so I downloaded your branch spawnpoitns from your's git. I have few problems. With original code from reddit was CPU usage minimal (load avarge 0.00) With this change is load average: 0,76, 0,60, 0,57..

Second this is I do not see any green circles with "Scanned Locations" on.

theweese:
I have this up and running but I must have something configured wrong.  My steps remaining keeps increasing and the Scanned Locations is only spiraling out from one spot.  Is this normal behavior?

3kynox:
Ok, I had to setup MySQL, then it's time, thanks for help.

cris11368:
@theweese you need to set your step-limit to 1

theweese:
That was it!  Thank you so much.

xordinx:
this working so good.
but its scanning random spawn points..can you make it like it scan first nearby spawn points then it go far

Zerotask:
How do you run it with multiple workers? and how you set the scanning area (like with the beehive generator)

SilverKnight29:
Tested with Ubuntu 14.04.4 LTS using the latest commit 4b39403 and everything works great! Was using an older version of the maps just to maintain support with sowoks reddit post. Please merge this soon!

Smenus:
@Zerotask You don't set the scanning area - it'll scan every spawn point you have in your spawns.json at the time of that spawn. If by multiple workers you mean multiple accounts, then it's no different to normal - the important thing is that you have set step limit to 1.

@xordinx It scans each spawn point in your spawns file one minute after a Pokémon should spawn there. There's no greater control than limiting the area that the spawns in your JSON file cover.

Smenus:
Forgot to mention, I've tried this on macOS and everything is running well. After changing my delay to 11 (I was on 16 from a while back), the queue rarely stays over 0 for long. While I agree that some improvements could be made (and I think the Reddit thread mentions some), this is a good first step.

HasanY1989:
Man this is aswome ! 
But you will need hell lot of workers, fx. if u had a HUGE area, you will need a lot of workers for them to go to spawn point and scan, if i understood correct

Smenus:
@HasanY1989 True, but imagine how many workers you would need if you were doing a honeycomb over that same area! I think the original thread said that it reduced the number of requests by 80% - so you can scan 5x the area with the same workers.

blindreaper:
Added https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into/d6gfer6

vpdragon0:
I have setup the spawns.json and set the -st to 1. However it just scans the initial locations over and over and over again. Anyone get the same problem?

blindreaper:
You need -ss

vpdragon0:
ah thanks!

TallGuy74:
I am seeing some 0 pokemon, 0 pokestops, 0 gyms results if my spawns are relatively far away but close in time. I suspect Niantic has a check for speed between map requests, and if the speed is too high, you don't see anything.

Zerotask:
So i set ss to true, st to and call all worker without a location (lat/lon)?

python runserver.py -os && python runserver.py -ns && python runserver.py -ns

3kynox:
Hello again, looks like a string is used as a number : 

```
2016-08-14 11:54:51,976 [   search_thread][        search][    INFO] Starting search worker threads
2016-08-14 11:54:51,984 [   search_thread][        search][    INFO] Total of 1117 spawns to track
Exception in thread search_thread:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/nox/PokemonGo-Map-Unofficial/pogom/search.py", line 272, in search_overseer_thread_ss
    while timeDif(curSec(), spawns[pos]['time']) < 60:
  File "/home/nox/PokemonGo-Map-Unofficial/pogom/search.py", line 118, in timeDif
    dif = a - b
TypeError: unsupported operand type(s) for -: 'int' and 'unicode'
```

blindreaper:
@Zerotask you need location for both web server and workers. Worker threads will not use the locations but the argument is required. Add -ss to enable spawn point scanning.

@3kynox Yes it looks like your json is broken. With the latest commit, delete spawns.json and it will be made for you.

blinx615:
After implementing this, I seem to be getting a lot of errors regarding the database being locked:
```
`2016-08-14 11:44:26,158 [      Thread-127][           app][   ERROR] Exception o
n /raw_data [GET]
  File "C:\python27\lib\site-packages\flask\app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\python27\lib\site-packages\flask\app.py", line 1641, in full_dispatch
_request
    rv = self.handle_user_exception(e)
  File "C:\python27\lib\site-packages\flask\app.py", line 1544, in handle_user_e
xception
    reraise(exc_type, exc_value, tb)
  File "C:\python27\lib\site-packages\flask\app.py", line 1639, in full_dispatch
_request
    rv = self.dispatch_request()
  File "C:\python27\lib\site-packages\flask\app.py", line 1625, in dispatch_requ
est
    return self.view_functions[rule.endpoint](**req.view_args)
  File "C:\PokemonGo-Map-3.1.0\pogom\app.py", line 90, in raw_data
    d['pokemons'] = Pokemon.get_active(swLat, swLng, neLat, neLng)
  File "C:\PokemonGo-Map-3.1.0\pogom\models.py", line 101, in get_active
    for p in query:
  File "C:\python27\lib\site-packages\peewee.py", line 2967, in __iter__
    return iter(self.execute())
  File "C:\python27\lib\site-packages\peewee.py", line 2960, in execute
    self._qr = ResultWrapper(model_class, self._execute(), query_meta)
  File "C:\python27\lib\site-packages\peewee.py", line 2656, in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "C:\python27\lib\site-packages\peewee.py", line 3492, in execute_sql
    self.commit()
  File "C:\python27\lib\site-packages\peewee.py", line 3316, in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "C:\python27\lib\site-packages\peewee.py", line 3485, in execute_sql
    cursor.execute(sql, params or ())
OperationalError: database is locked`
`````

**Comment by BaIthamel** [comment]
Hold on you just ruined it, we don't want a single spawns.json or you need to add some location based filtering.

**Comment by WeissJT** [comment]
@blinx615 
You need mysql for that amount of workers.
https://pgm.readthedocs.io/en/develop/extras/mysql.html

**Comment by blinx615** [comment]
@WeissJT Aha!  That makes perfect sense.  Will do; thankya.

**Comment by lota7** [comment]
Am I correct when I say this would only be useful and accurate when running the original pogomap over a long period of time (week or so) to map all possible spawnpoints, before implementing this? Since I have noticed new spawnpoints popping up even after 5 days of scanning.

It's often those rare dragonites/snorlax/lapras that spawn in new spawnpoints or spawnpoints that has low spawnrate so it's easy to miss if sample size is too small.

For example yesterday, after 1 week of data collecting, a dragonite spawned in a new location on remote area i'm 100% sure has had no spawns before.

**Comment by WeissJT** [comment]
@lota7 
As far as I know all the spawnpoints are static with different timers (no random or special) and have activity every hour.
That's why spawnScan from TBTerra can scan almost all spawnpoints ( ~98%) in 1 hour if you have enough workers.
See:
https://www.reddit.com/r/pokemongodev/comments/4uuwx0/spawn_myths_debunked_aka_the_truth_about_spawn/ and https://www.reddit.com/r/pokemongodev/comments/4uf1eh/spawnscan_spawn_point_finder

**Comment by justinkwaugh** [comment]
How do you balance the accounts vs number of spawn points.  This is a map of an area I can easily cover with 20 accounts using beehive. Unless the spawning is perfectly staggered, or there is optimization for spawn points near each other, I don't see how I could have enough accounts to cover this all.
screen shot 2016-08-14 at 6 18 17 pm

**Comment by drewreece** [comment]
@justinkwaugh,
Please try running the code.
If 20 accounts can cover that area with beehive you can use far fewer accounts with spawn point scanning. Those spawns are triggered across an hour, not all in one go, knowing when means lots of requests are not required.

Beehive sends requests for empty cells, I get about 50% I'm my area. Spawn scanning doesn't send those requests. That means more Pokémon for fewer server requests.

Whilst I suspect you may get bans for too much teleporting (the spawn scanner isn't optimised) you should cover the same area with less accounts. I don't run 20 accounts or try to map a city, so maybe I am wrong.

**Comment by justinkwaugh** [comment]
@drewreece I just realized that I was using 40 accounts not 2, so it seems a touch more manageable.  That said it seems for an area this large teleport bans would be very likely.  I think a beehive / spawn point combination would be ideal.  A single account could handle a small area with no issue.

**Comment by lopsidedjohnson** [comment]
So, I agree with BaIthamel.  I modified the code as follows:

In utils.py, ADD

`    parser.add_argument('-sf', '--spawnpoint-file',`
`                        help='JSON file to read for spawnpoint scanning',`
`                        default='spawns.json')`

In search.py, CHANGE

`        with open('spawns.json') as file:`

TO:

`        with open(args.spawnpoint_file) as file:`

Then when running the server,  can specify -sf location1.json etc.  Helps limit workers to certain areas to prevent too large of jumps.

**Comment by drewreece** [comment]
@justinkwaugh, it is unclear if you have tried this code or are just speculating on what might work. 

Once you have seen it work you may realise that it only needs a handful of accounts - it really isn't clear to me how you mapped those spawns and if they are de-duplicated for the spawn scanner. It replaces beehive scanning - so I'd suggest you forget about using both, but you probably could via multiple commands.

Anyway this sounds like a speculative support post, so I'd suggest you take it to http://reddit.com/r/pokemongodev/ & I can try explaining without driving the devs here insane. Just link to a post or devs please correct me if I am wrong.

File: runserver.py [hunk 1]
```diff
@@ -163,11 +150,27 @@
     if not args.only_server:
         # Gather the pokemons!
         if not args.mock:
-            log.debug('Starting a real search thread')
-            if not args.spawnpoint_scanning:
-                search_thread = Thread(target=search_overseer_thread, args=(args, new_location_queue, pause_bit, encryption_lib_path))
+            #If running in data gather or old mode
```
blindreaper: #585 

Should it be like this?

>     if not args.only_server:
>         # Gather the pokemons!
>         if not args.mock:
>             # If running in data gather or old mode
>             if not args.spawnpoint_scanning:
>                 log.debug('Starting a real search thread')
>                 search_thread = Thread(target=search_overseer_thread, args=(args, new_location_queue, pause_bit, encryption_lib_path))
>             # if using the new -ss mode
>             else:
>                 if not os.path.isfile(args.spawnpoint_scanning):
>                     log.info('Could not find spawns.json, generating now. This may take up to 60 seconds')
>                     file = 'spawns-' + int(round(time.time() * 1000)) + '.json'
>                     try:
>                         with open(file, 'w+') as file:
>                             spawns = Pokemon.get_all_spawnpoints()
>                             file.write(json.dumps(spawns))
>                             file.close()
>                             log.info('Finished generating ' + file)
>                     except IOError:
>                         log.error("Error writing to " + file + ", exiting")
>                         sys.exit()
>                 else:
>                     search_thread = Thread(target=search_overseer_thread_ss, args=(args, new_location_queue, pause_bit, encryption_lib_path))
>         else:
>             log.debug('Starting a fake search thread')
>             insert_mock_data(position)
>             search_thread = Thread(target=fake_search_loop)

runserver.py
```diff
@@ -163,11 +150,27 @@
...
+            if not args.spawnpoint_scanning: 
+                log.debug('Starting a real search thread')
+                if not args.spawnpoint_scanning:
+                    search_thread = Thread(target=search_overseer_thread, args=(args, new_location_queue, pause_bit, encryption_lib_path))
+                else:
+                    search_thread = Thread(target=search_overseer_thread_ss, args=(args, new_location_queue, pause_bit, encryption_lib_path))
```
blindreaper: #585 

Should it be like this?

I cant see it ever getting to line 159

>     if not args.only_server:
>         # Gather the pokemons!
>         if not args.mock:
>             # If running in data gather or old mode
>             if not args.spawnpoint_scanning:
>                 log.debug('Starting a real search thread')
>                 search_thread = Thread(target=search_overseer_thread, args=(args, new_location_queue, pause_bit, encryption_lib_path))
>             # if using the new -ss mode
>             else:
>                 if not os.path.isfile(args.spawnpoint_scanning):
>                     log.info('Could not find spawns.json, generating now. This may take up to 60 seconds')
>                     file = 'spawns-' + int(round(time.time() * 1000)) + '.json'
>                     try:
>                         with open(file, 'w+') as file:
>                             spawns = Pokemon.get_all_spawnpoints()
>                             file.write(json.dumps(spawns))
>                             file.close()
>                             log.info('Finished generating ' + file)
>                     except IOError:
>                         log.error("Error writing to " + file + ", exiting")
>                         sys.exit()
>                 else:
>                     search_thread = Thread(target=search_overseer_thread_ss, args=(args, new_location_queue, pause_bit, encryption_lib_path))
>         else:
>             log.debug('Starting a fake search thread')
>             insert_mock_data(position)
>             search_thread = Thread(target=fake_search_loop)

turner-curtis:
I am calling runserver.py -l 'xx.xxxx,-xx.xxxx' -ss /opt/PokemonGo-Map/spawns.json using a known good spawns file used with a manually edited version from a few days ago that is still working in another directory. I found this and figured I would pull it and try to get the latest but am seeing some errors. 

2016-08-14 23:23:50,231 [      MainThread][        models][    INFO] Connecting to MySQL database on 127.0.0.1:3306
  File "runserver.py", line 179, in <module>
    search_thread.daemon = True
NameError: name 'search_thread' is not defined

justinkwaugh:
@drewreece As I mentioned in my previous post I'm not so concerned about the number of accounts, but rather about the teleporting, which for an area as large as I am mapping is certainly a problem.  It was still using a single spawns.json previously, but a couple of hours ago it was fixed to allow for multiple ones, which is better, but I believe the code is still generating the spawn.json from a query of the entire db, rather than restricted to the area a worker would normally cover, so it is harder to migrate from beehive for large areas. Perhaps it should be doing what the develop branch is doing for spawnpoint only scanning which is to find the spawnpoints for the current location / step range.

blindreaper:
Should be fixed, BaIthamel/patch-1 broke stuff

blindreaper:
#608

drewreece:
@justinkwaugh, if you don't want to use the entire db generate your own spawns.json.
I made a niave script to do this…
https://gist.github.com/drewreece/4eb4497f57e9f6fb78d14aef27fcc750
Or one of the other scripts that have been posted that make spawns.json.
https://gist.github.com/Xcelled/c5b397dfe2c32e61e7aa06479b32bc34

Or you could use the one that was removed from this issue…
https://github.com/PokemonGoMap/PokemonGo-Map/pull/585/commits/543bc3fc1bab3723fda43a00498486f888a7f286

Provide co-ordinates to get output. I suspect you can run multiple regions via the new spawns file path argument (I have not tested).

@blindreaper - I think that generator script may need to be reinstated - it is useful to isolate areas when the DB has grown out of control (e.g. '_friends_' who decided to tour the planet scanning). Ideally there will be a future feature to makes area selection accessible from the web UI, I'll post a feature request if this ever gets committed.

@justinkwaugh, I can't help with teleporting, I don't try to map huge areas, there are other attempts to minimise or optimise the process, but it is not trivial. Maybe you need to wait if you are concerned by it. Look up the travelling salesman problem if you think you can solve it :)

FYI I disagree with the focus of the current devel branch - it still polls when spawns are not scheduled. I suspect Niantic are more concerned by the terabytes of data pulled by these scanners, I think minimising that is the priority but what do I know.

blindreaper:
@drewreece specifying -ss **x**.json (when **x**.json does not exisst) will generate **x**.json.

drewreece:
@blindreaper, neat :)
Can the new file arg for spawns.json be abused to support multiple processes?

Here is my thinking…
- Generate multiple spawns.json files with limited areas.
- Use a separate run server.py command per file (using server only mode)

This should cause the workers to be restricted to the area in the .json. Could that help reduce the amount of teleporting in the account(s) used on that task? Most people wont need it, but the heavy beehive users scanning the planet might need to use it?

blindreaper:
Yes that's possible with the current implementation but at the moment there's no way to get spawns within an area. My SQLfu is pretty limited.

drewreece:
> but at the moment there's no way to get spawns within an area. My SQLfu is pretty limited.

See the gists above, mine is probably lousy but you can hard code co-ordinates into it. Xcelled uses args for bottom left, top right.  Both parse out duplicate spawns when they trigger at same time & location.

lota7:
@WeissJT Yes I know...but you see it's those 2% that it doesnt catch which are the ones that mostly contain dragonites, snorlax, lapras...since these spawnpoints are rarely activated and wont be caught by TBterra in a small scan period....
As mentioned, I've had a week spawn info, but still am finding new spawnpoints.

So those 2% are the important ones...who cares about the other 98% consisting of only pidgeys and zubats?

blindreaper:
@lota7 The only time that is possible if niantic add spawnpoints. Scanning for 1 hour is enough.

> Conclusion:
> spawn points each have a certain type, that determines their spawn interval and the staytime of the pokemon, that type doesn't change, 1x15, 1x30, 1x45, 1x60 and 2x15 types confirmed.

https://www.reddit.com/r/pokemongodev/comments/4uuwx0/spawn_myths_debunked_aka_the_truth_about_spawn/

There's no once a week spawn or once x days spawn.

But maybe you have discovered something new, you should report your findings to confirm.

pogom/search.py
```diff
@@ -203,18 +231,67 @@ def any_spawnpoints_in_range(coords):
...
+        location = []
+        location.append(spawns[pos]['lat'])
+        location.append(spawns[pos]['lng'])
+        location.append(0)
+        for step, step_location in enumerate(generate_location_steps(location, 1), 1):
+                log.debug('Queueing step %d @ %f/%f/%f', pos, step_location[0], step_location[1], step_location[2])
+                search_args = (step, step_location, spawns[pos]['time'])
+                search_items_queue.put(search_args)
+        pos = (pos + 1) % len(spawns)
+        if pos == 0:
```
TBTerra: why is this section here?

if the search queue is long it can cause many spawns at the beginning of the list to be silently skipped, and if the search queue is short it can cause it to re-scan the end of the list

justinkwaugh:
@blindreaper The current develop branch has exactly a method to get the spawnpoints for an area.  Locally I just combined the two things so now I have a version which doesn't use a json file, it just queries the db for the spawnpoints for the worker when it starts up and then it searches using the method in this branch. All I have to do is modify my beehive.sh to add -ss to the workers and it switches to spawnpoint scanning only within their location/steps. Was there a particular reason people wanted to use a json file? I understand in the case where you have hand tweaked it, but in the basic case I don't see the reason, as the db has all the data anyway.

blindreaper:
@justinkwaugh  @TBTerra It's how the original implementation came, i did not write the code.

WeissJT:
@justinkwaugh 
At least in my case, I prefer the rectangle scan of spawnScan instead of the hexagonal-beehive approach.
It feels more optimized and fast for large areas.

justinkwaugh:
@WeissJT sure, though it shouldn't really be any different optimization wise.  My point was more about making it easy for existing users. If you want to use a different scan method and a custom spawn.json sure, no problem.

justinkwaugh:
Also on a different note, I commented out the actual api calls just to test the spawnpoint lookup and scanning and I noticed that when multiple spawnpoints get queued at once, the worker thread was calling niantic in a rapid-fire way, not delaying between, which will get me soft-banned.  I looked at the current timing/delay code in the worker thread and it doesn't make much sense to me. Why does it not just get the current time after pulling from the queue and use that to determine whether or not it needs to delay after making the call to niantic? I made that change and it was properly delaying 10s between searches

blindreaper:
@justinkwaugh post a patch or pr

tylerdola:
I have a .json file for downtown when I am at work during the day and run a scanner for co-workers. I have one around my house for when I am home in the evenings and weekends. I have prepared a few .json files for the trip I will be taking next week and would like to scan around my hotels. All are fed into the same pogomap server. Now my server db is full of many places that have vast distances in between.  I suppose the proper way is to use a different sql db for each region? Just my comment for why I prefer spawns.json input files @justinkwaugh

rjwestman:
@justinkwaugh you asked how to make it more effective for running it with beehive and differnt workers with different que loads right? Im currently checking the logs of the different workers looking at the que and removing/adding accounts depending on that - im guessing this will result in a good optimization even if its a little bit of leg-work. I kinda dont want to go with only 1 worker and spawn scanning since the accounts might jump a huge distance between scans - which will result in a softban too i guess.

blindreaper:
Specifying -ss x.json (when x.json does not exist) will generate x.json.
Specifying -ne and -sw when generating json will only grab spawns on the bounding rectangle.

justinkwaugh:
I'll commit what I have when I figure out how to git.  Mine will either use the json specified or will grab the spawns from the db for the location / steps

Cy4n1d3:
It would be great to have some basic documentation on usage and available params :)

justinkwaugh:
I'm not sure a PR is correct for this, but here's a link to my branch:

https://github.com/justinkwaugh/PokemonGo-Map/commit/35bc30edf27143599a09dd6ea5a846fad0f09397

blindreaper:
@Cy4n1d3 done, go nuts.
@justinkwaugh which pr?

justinkwaugh:
@blindreaper I apologize, i've not used git / github much.  I forked and made a branch, but it's off of develop not this branch.  I'm not sure the appropriate thing to do

blindreaper:
@justinkwaugh Join the discord, its only a few commands to get started on the branch, im sure somebody will be willing to help: https://discord.gg/PWp2bAm

justinkwaugh:
Alright here is my own PR https://github.com/PokemonGoMap/PokemonGo-Map/pull/700

It is not fully synced with this branch, but the key parts are that in search.py it will use either the specified json or will just grab from the db for the appropriate area.  It is working well for my existing beehive.

pogom/search.py
```diff
@@ -203,18 +231,67 @@ def any_spawnpoints_in_range(coords):
...
+            while not(search_items_queue.empty()):
+                log.info('Search_items_queue not empty. Waiting 10 secconds. Restarting at top of hour')
+                time.sleep(10)
+            log.info('Restarting from top of list and finding current time')
+            pos = SbSearch(spawns, (curSec() + 3540) % 3600)
```
Pomekie: As the list "spawns" is already sorted the next pos returned by SbSearch should be pos+1.

justinkwaugh:
One thing I'm noticing is that I don't seem to find spawns at the time I expect.  Many have only about 10 min left.

lota7:
Then it doesnt look like it works as advertised.
It should never return 0 and if the DB fetches the spawntimers correctly it should always be ~14mins...i'd prefer it to be 14.30  :P

justinkwaugh:
@lota7 I'm not sure what to think. I guess it is possible that by changing that value it is accidentally picking up neighboring spawnpoints a minute early.  I just expected that I would be getting consistent 14 min disappear times with this scanning method, and I am not seeing that at all

justinkwaugh:
@lota7 if you reduce the time for the comparison before it enqueues to 30 from 60 you will get more at 14:30

lota7:
@justinkwaugh I'll try that. But you mentioned that you got a few 10minutes one? 

Regarding 60 to 30 change. is it this in search.py?
while timeDif(curSec(), spawns[pos]['time']) < 60;

Thanks

justinkwaugh:
@lota7 Yes that is the spot to change.  I just had a bug in my query.. I'm
getting all perfect searches now.  Occasionally due to the proximity to
other spawnpoints I get others of varying times (some are even basically a
minute early... 14:59), but there is always one in the center of the search
with the correct disappear time.

pogom/models.py
```diff
@@ -209,16 +210,50 @@ def get_spawnpoints(cls, swLat, swLng, neLat, neLng):
...
+                valid_coord = False
+            if not ne:
+                log.error('Invalid northeast coord, defaulting to dump all')
+                valid_coord = False
+            if valid_coord:
+                query = (query.where(
+                    (Pokemon.latitude >= float(sw.group(1))) &
+                    (Pokemon.longitude >= float(sw.group(2))) &
+                    (Pokemon.latitude <= float(ne.group(1))) &
+                    (Pokemon.longitude <= float(ne.group(2)))
```
brianjob: I think you have the longitudes backwards. Should be:
`
                    
                     (Pokemon.latitude >= float(sw.group(1))) &
                     (Pokemon.longitude <= float(sw.group(2))) &
                     (Pokemon.latitude <= float(ne.group(1))) &
                     (Pokemon.longitude >= float(ne.group(2)))
`

blindreaper:
@justinkwaugh It has been changed to 2710: https://github.com/blindreaper/PokemonGo-Map/blob/d2c3d3c215222651dc29281c98c4fb46849fbd90/pogom/models.py#L249 in https://github.com/blindreaper/PokemonGo-Map/commit/0110e368c0729e6d2b23fe3fbba47087d8785009

pogom/search.py
blindreaper: I'm not sure which line you're referring to.

pogom/search.py
justinkwaugh: Agreed, it works fine without this whole if block

pogom/models.py
drewreece: I don't understand? Why would the longitude of a Pokémon spawn be to the left of (less than) the supplied southwest co-ordinate?

The same is true of the top right co-ordinate, why should it be greater than the supplied one? 
We are defining the area bottom left to top right. 
What results do you see after using this? I'm away from my test machine at present.

All the other queries in models.py follow the old method (as does Xcelled's gist that was refactored here). https://gist.github.com/Xcelled/c5b397dfe2c32e61e7aa06479b32bc34#file-spawnpoint_exporter-py-L57

brianjob: for some reason I was thinking it was north west and south east. disregard my comment.

drewreece: In that case I think d2c3d3c21 needs reverting @blindreaper.

PR #589, opened 2016-08-13
Updated migration code and changed Pokemon PK to be more accurate

Xcelled:
## Description
I refactored the migration code to be a little easier to maintain, as well as being more informative about what it's doing. Now, you just define a migration in a method and add it to the migrations list. Additionally, transactions are used to avoid corrupting a database if a migration fails.

These changes were implemented to facilitate changing the primary key of the Pokemon table from encounter_id to a composite key of encounter_id, spawnpoint_id, and disappear_time, as discussed in #179, which is also in this PR.

## Motivation and Context
To support detecting of spawnpoints from an existing database, as outlined in #179

## How Has This Been Tested?
Tested the migrations several times from 0 to 5 and 4 to 5. No issues. Map still runs, etc.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -24,8 +24,6 @@ log = logging.getLogger(__name__)
 args = get_args()
 flaskDb = FlaskDB()
 
-db_schema_version = 4
-
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
     pass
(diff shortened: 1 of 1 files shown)


Xcelled:
Don't merge this yet, pending possible update.

PR #590, opened 2016-08-13
Only add ssh client, not server to docker image

tkSimon:
Atm openssh server componentens get installed, which is unneccessary, because only client components are needed for git+https deps in requirements.txt.

## Description
Changed `apk add [...] openssh` to `apk add [...] openssh-client`

## Motivation and Context
Docker image size is smaller.
Why have unneeded packages installed?
Security concerns...

## How Has This Been Tested?
Built new docker image, runs like a charm.

## Types of changes
- [X] Bug fix (kind of?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,7 @@ ENTRYPOINT ["python", "./runserver.py", "--host", "0.0.0.0"]
 CMD ["-h"]
 
 # Install required system packages
-RUN apk add --no-cache ca-certificates
-RUN apk add --no-cache bash git openssh
+RUN apk add --no-cache ca-certificates git
 
 COPY requirements.txt /usr/src/app/
 


hemsleyk:
+1

chuyskywalker:
doesn't the `git` package cover this anyway?

tkSimon:
Yup, you're right. While we're at it, bash isn't required either... remove that too?

chuyskywalker:
Yes. I also don't know how/why that got there.

jnovack:
Just make it one line.

```
RUN apk add --no-cache ca-certificates git
```

Ashex:
Giving this a little bump since there's now a published container.

https://hub.docker.com/r/pokemap/pokemongo-map/~/dockerfile/

tkSimon:
That was the plan @jnovack ;)

chuyskywalker:
All wrapped up into #952

Issue #570, opened 2016-08-12
[Feature Request] Add a WAVE_DELAY parameter -wd


LurkerLurks:
## Requested Behavior

Setting WAVE_DELAY (-wd) in seconds will delay the start of the next wave of scanning once the max scan track has been reached. 

For example, you want to scan your immediate area once every 5 minutes and you have 7 accounts setup you can set STEP_LIMIT to 2 and scan your immediate area all at once and then have WAVE_DELAY postpone the next wave for 300 seconds. Alternatively, if you are scanning a larger area The scan would wait until the cooldown has passed before restarting the scan. 

OPTION A: Have the WAVE_DELAY start once the STEP_LIMIT has been reached.

OPTION B: The count for WAVE_DELAY begins right before the first step is scanned. Once STEP_LIMIT has been reached the next step is delayed until the WAVE_DELAY has been reached. If the STEP_LIMIT is not reached after the WAVE_DELAY expires then it simply restarts the scan immediately once the STEP_LIMIT is reached. 

The default value would be 0.

## Context

This would be useful for people who are spawn mapping and people who want to scan all day for nearby notifications but don't need/want a constant scan running in the background.

## Current Behavior

Setting STEP_DELAY only scans an area slowly it does not delay the scanning process. 

## Status: Planning

I am brand new to NodeJS have no formal coding experience. I have some very basic scripting experience but that is it.   I might have some time to study it this weekend and might figure something out in the next week or so. If anyone else knows how implement this quickly, wants mentor me through the process or could point me in the right direction I am all ears.

FrostTheFox:
Please direct feature requests to http://feathub.com/PokemonGoMap/PokemonGo-Map

LurkerLurks:
@FrostTheFox - Implementing Option A ended up being a lot easier than I thought. It looks like a delay was already in place in the last line of the search_overseer_thread of the search.py (~line 176).  I added the --wave-delay entry to the utils.py and it looks like that worked. I have submitted the change in pr #591.

PR #591, opened 2016-08-13
Added a wave_delay for once the step_limit has been reached

LurkerLurks:
## Description

Setting WAVE_DELAY (-wd) in seconds will delay the start of the next wave of scanning once the STEP_LIMIT has been reached.

For example, you want to scan your immediate area once every 5 minutes and you have 7 accounts setup you can set STEP_LIMIT to 3 and scan your immediate area all at once and then have WAVE_DELAY postpone the next wave for 300 seconds.

## Motivation and Context

Built out a solution for my feature request (Issue #570). Sorry it was posted in the wrong place. Let me know if you want it reposted to feathub.

## How Has This Been Tested?

I ran it and looked at the CLI output. There is a delay in the scan once the step limit is hit. I tested various delays: 30, 60 and 120. It worked flawlessly. 

## Types of changes

## Checklist:

I wasn't sure if documentation needed updating. I figured the commandline help text was enough. Let me know if I need to do that. 

This is my first PR to an opensource project. Please be patient with me.


chuyskywalker:
Actually, this isn't working how you think.

The overseer loop is designed to be, basically, continously looping and not pausing for anything. It doesn't get to that `time.sleep(1)` and pause there till the search set is done.

You should actually add this here:
```
         if search_items_queue.empty():
             log.debug('Search queue empty, restarting loop')
```
into
```
         if search_items_queue.empty():
             log.info('Search queue empty, restarting scanners after %d second wave delay', args.wave_delay)
             time.sleep(args.wave_delay)
```

LurkerLurks:
I just tested this it seems to be working ok. I want the delay to start after the first scan so I moved it down alittle bit.  I tried switching between log.info and log.debug and tried using the -d debug flag but I am not seeing the added log message in the console. I am not sure what I am doing wrong in that regard. The pause does seem to be working appropriately through.

chuyskywalker:
This, also, will have unexpected results. It's only working for you because you have a higher number of accounts and a limited set of scan steps. If someone had, say, 200 scan steps, and only a few accounts, the queue will finish well after `wave-delay` does, and they'll be left wondering, "Hey, why didn't wave delay add X seconds after the queue hit zero?"

chuyskywalker:
If you don't want to have it happen on the first scan (which is reasonable) you'll need to add a "isFirstScan" variable to the overseer loop so you can flip that boolean. But the wait will still need to be in the place where I original showed above

LurkerLurks:
That makes sense. I will have some more time to work on it later this weekend. Thanks for walking me through the process.

chuyskywalker:
Of course, thanks for contributing -- I know it can be a hassle to open a PR in somebody else's pool.

LurkerLurks:
Sure, it has been a good learning experience. So does something like this work?

````python
        if search_items_queue.empty():
            if args.wave_delay > 0 & isFirstScan > 0:
                log.debug('Waiting for %d second wave delay', args.wave_delay)
                time.sleep(args.wave_delay)
            log.debug('Search queue empty, restarting loop')
            for step, step_location in enumerate(generate_location_steps(current_location, args.step_limit), 1):
                log.debug('Queueing step %d @ %f/%f/%f', step, step_location[0], step_location[1], step_location[2])
                search_args = (step, step_location)
                search_items_queue.put(search_args)
            isFirstScan = 1
````

I am not sure where/how I should declare the isFirstScan variable. My experience with python is limited to a few afternoons of playing with the beautifulsoup4 library a few months ago. I understand the very basic programming principles but have not developed software before. I don't know why I waited so long to get into it. It is a lot of fun.

**Comment by chuyskywalker** [comment]
You have to set the `isFirstScan = True` way closer to the top of the function so that it's set when the function starts, but is only set once. then it'd be more like

```
if (args.wave_delay > 0 and isFirstScan):
  time.sleep(args.wave_delay)

# queue stuff here
isFirstScan = False
```

LurkerLurks:
The only issue with is is that if you wait for the wave delay to move the position, The pause and scan must run gain before it picks up the new move. I am testing putting the pause before the move check.

LurkerLurks:
I removed the delay after the overscan but there is still an overscan. I am not sure how to avoid that. 

**Overscan Example:** If you have 3 accounts and set the step limit to 7. After the last scan the delay is triggered. During the delay you move the location. Once the pause completes the first 3 scans of the original wave are repeated. Once for each account. The scan then snaps to the new location and starts scanning as usual. 

> This, also, will have unexpected results. It's only working for you because you have a higher number of accounts and a limited set of scan steps. If someone had, say, 200 scan steps, and only a few accounts, the queue will finish well after wave-delay does, and they'll be left wondering, "Hey, why didn't wave  delay add X seconds after the queue hit zero?"

This would be fine/prefered. It was Option B in my original request but seemed more complicated to implement. The goal is being able to set a  standard cadence between scan waves.  This would be a "start a new waves no sooner than 5 minutes apart"  kind of a thing.

primaxius:
Conflicts - resubmit if needed  - Cleaning old PR's

Issue #316, opened 2016-08-09
SSL


FlarumHostingCo:
Please add SSL support to the built in web server.

Arderos:
Put it behind nginx proxy. Far safer choice imo.

Cy4n1d3:
If you're providing the map over public domains, you should be able to set up a reverse proxy and corresponding tls setup yourself.

In general, it's not that easy to generate certificates which are trusted over a wide spread of devices without a domain / static ip.

BaIthamel:
This is not a support forum; please see [the discord #help channel](https://discord.gg/npVqkUn) for assistance.

PR #594, opened 2016-08-13
Add basic support for SSL

ping:
## Description
Add basic support for SSL via 2 new args ``--ssl-certificate`` and ``--ssl-privatekey``

## Motivation and Context
I've read recommendations for users who require https (to enable geolocation for example) to do so via nginx/apache configuration (#316). I run this on an old laptop at home with port forwarding + dynamic dns which allows me to access the map on the run. In my case, a full blown webserver is overkill. With Letsencrypt and their support for dns challenges, it is feasible imo to provide basic SSL support with the built-in webserver.

## How Has This Been Tested?
Tested this in Chrome (OSX, Android), Firefox (OSX). Green lock appears fine. Geolocation features e.g. "Scan follows location" works as expected on mobile and desktop.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -27,6 +27,8 @@
 #host:                  # address to listen on (default 127.0.0.1)
 #port:                  # port to listen on (default 5000)
 #locale:                # pokemon translation
+#ssl-certificate:       # path to ssl certificate
+#ssl-privatekey:        # path to ssl private key
 
 #Uncomment a line when you want to change its default value (Remove # at the beginning)
 #username, password, location and gmaps-key are required
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -120,6 +120,8 @@ def get_args():
                         type=int, default=5)
     parser.add_argument('-wh', '--webhook', help='Define URL(s) to POST webhook information to',
                         nargs='*', default=False, dest='webhooks')
+    parser.add_argument('--ssl-certificate', help='Path to SSL certificate file')
+    parser.add_argument('--ssl-privatekey', help='Path to SSL private key file')
     parser.set_defaults(DEBUG=False)
 
     args = parser.parse_args()
--- a/runserver.py
+++ b/runserver.py
@@ -8,6 +8,7 @@ import logging
 import time
 import re
 import requests
+import ssl
 
 from distutils.version import StrictVersion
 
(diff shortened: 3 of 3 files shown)


chuyskywalker:
An old PR of mine had a better handling for ensuring cert/key exist properly: https://github.com/AHAAAAAAA/PokemonGo-Map/pull/1936/files#diff-6c84a64c790b52a5ff2656d6bb388fefR121

You should steal that.

You'll also fine in the previous repo that the PR was rejected under the rationale that the app remain as simple as possible and that people should add SSL in another layer. That was the opinion of the fork at that time -- I'm open to this change here.

rrbailey89:
This does need to be added. Also you should be able to specify the location of the certs using the config.ini

Chlodochar:
I'm kind of on the fence here. I agree that https should be easily integrated, but then again, manually refreshing a certificate shouldn't be needed anymore, so the next step would be adding ACME support. Virtual hosts seems like a viable option as well, since otherwise port 443 is used purely for PokemonGo-Map. Do we want all of this in the repo, or should we recommend using a stable https server that has all features builtin? (nginx, apache, caddy, etc?)

chuyskywalker:
@rrbailey89 since this is from arg parser, it does work in config.ini (@ping, you should update the config.ini.example to reflect this)

@Chlodochar -- they are not mutually exclusive. We're definitely not going to integrate acme here. But this is such a simple extension, that's already well supported by the framework, it doesn't make sense to _not_ include it. Sure, it's not as perfect an option as others, but this being in place is a nice, simple "make it go" option. Finally, doing this doesn't preclude layering over a port 80 pgomap in anyway (like you can do today, and after this merge).

Chlodochar:
@chuyskywalker I just read #623. Indeed, this should be added.

ping:
Update the config.ini.example with the 2 new args as well as path exists checks for the 2 ssl files.

chuyskywalker:
My SSL cert works on this.

amducious:
just game this a try. Its just what I wanted. got this setup and geo location started right up. thanks

rrbailey89:
@ping is it possible you could do a noon friendly guide for people on how to obtain the certs for those on windows and other systems?

ping:
@rrbailey89 You can get a cert from any issuer that allows you to verify domain ownership via dns or email. Many of the paid providers already do that. 

Letsencrypt supports http (not applicable in this case) and dns verification, but you'll probably have to use a 3rd party client since the official one doesn't do dns-01 just yet (coming soon I think). I used [letsencrypt.sh](https://github.com/lukas2511/letsencrypt.sh) but that requires bash. Fairly straightforward if you're on Mac/Linux but for Windows, you'll need either cygwin or may be use the new bash shell on Win10.

PR #595, opened 2016-08-13
Fix `docker stop` running into a timeout

tkSimon:
`docker stop` running into a timeout, takes 10 unnecessary seconds to complete.
While `Dockerfile` supports `STOPSIGNAL` directive from version 1.9 on, this version is not widely used yet. Adding it would break support for many users.

## Description
Adds wrapper.sh file, to be used by `Dockerfile`.
`wrapper.sh` traps SIGTERM which is used by `docker stop` by default, and sends SIGINT to pyhton. It's not graceful either, but it will make `docker stop` exit in less than a second.

`ENTRYPOINT` and `CMD` directives in `Dockerfile` and `docker run` cli arguments can still be used like before. Just inserted "sh", "wrapper.sh" at the beginning of `ENTRYPOINT`

## Motivation and Context
Reduce `docker stop` time by factor >10. Just very inconvenient to wait 10 seconds for each worker when redeploying new containers.

## How Has This Been Tested?
* `docker run ....`
* `docker stop ...` finishes in <1 sec
* `docker log -tail 10 ...`
* output reads "forwarding SIGTERM as SIGINT" (meaning trap function was actually called)

## Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ EXPOSE 5000
 WORKDIR /usr/src/app
 
 # Set Entrypoint with hard-coded options
-ENTRYPOINT ["python", "./runserver.py", "--host", "0.0.0.0"]
+ENTRYPOINT ["dumb-init", "-r", "15:2", "python", "./runserver.py", "--host", "0.0.0.0"]
 
 # Set default options when container is run without any command line arguments
 CMD ["-h"]
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,3 +25,4 @@ requests==2.10.0
 PySocks==1.5.6
 git+https://github.com/ChrisTM/Flask-CacheBust.git@aa09ad861f104f987928fe9f5107ccfe0e4473c3#egg=flask_cachebust
 protobuf_to_dict==0.1.0
+dumb-init==1.1.3


Ashex:
Docker has native support for changing the stop signal so a wrapper script isn't needed. Just add `STOPSIGNAL SIGINT` to the Dockerfile

tkSimon:
Support for that was added in 1.9, most distributions, including docker hub itself are running older versions. So no support. Adding the `STOPSIGNAL` directive will break `Dockerfile` for many users.

chuyskywalker:
My preference would be to use something more robust, like `dumb-init`, which also has support for [signal rewrites](https://github.com/Yelp/dumb-init#signal-rewriting)

tkSimon:
@chuyskywalker didn't know that. can look into it tomorrow if you want me to

chuyskywalker:
Please do, thanks!

jnovack:
> for many users.

How many users?  1.9 is already 10 months old.

I vote `STOPSIGNAL`.

Ashex:
`STOPSIGNAL` is still a little flakey in 1.10 but is working properly in 1.11. Considering 1.11 was released relatively recently most current distributions only have 1.10 in their repos.

So the question really is, do we care about backwards compatibility _that much_? We're not a stable project pushing releases to distributions, end users are building the project themselves. 

I've been using `STOPSIGNAL` in my containers that are actively used and have had no complaints. That said it doesn't actually work on 1.10.3 so using dumb-init would be nice if it's deemed necessary.

krusmir:
With all due respect, supporting old versions of Docker makes no sense.   This project (and the original one) was released after the release of docker 1.12, why bother supporting old docker versions?  

Even so, most users are using docker 1.9 or greater, so adding STOPSIGNAL won't break their docker builds.

Do the project really want to add bugfixes/workarounds for old versions of docker?  This should be on the end users side to decide if to upgrade or not.

Don't get me wrong, the wrapper.sh  and dumb-init are awesome solutions for supporting legacy docker versions, I just don't see why this project should support that.

tkSimon:
Old? It's not even supported by docker hub. Why force ppl to upgrade their server distro, if they want to host this for their friends?

krusmir:
@tkSimon Can you please explain?  I'm currently using Docker Hub for my images, including the a version of this project with the STOPSIGNAL included and they are all working correctly.

Maybe I'm missing something.  :)

Ashex:
@tkSimon  Can you link to a reference about the lack of support on docker hub and how this impacts users? I was running containers with `STOPSIGNAL` on 1.10 and nothing was catching fire.

tkSimon:
https://github.com/docker/hub-feedback/issues/460
With docker hub running 1.9 i'm assuming this breaks builds on docker hub, as on most users boxes...

jnovack:
IN YOUR FACE!

krusmir:
@tkSimon @jnovack  Not even 1.9, for building images its running on 1.8.x.

tkSimon:
i just noticed, edited my comment

tkSimon:
Also, it's not like there's much effort involved in maintaining backwards compatibility. It's just one line in two files each. What's the big issue here?

krusmir:
@tkSimon You are correct.   I believe this is the best solution.  Here hoping for it to be approved soon. 👍

Ashex:
I setup a test repo [here](https://hub.docker.com/r/ashex/testrepo/) with a dockerfile containing `STOPSIGNAL` and the build was successful. I was also able to run it successfully (on 1.12).

screen shot 2016-08-22 at 21 47 38


Keep in mind that the issue being referenced is **9 months old** discussing 1.8/1.9 when the current release is 1.12. Docker Hub is also a commercial product so it's a little ridiculous to assume they'll just leave on 1.8 for 9 months.

Considering there's no problems building images on docker hub with `STOPSIGNAL` I'm still not seeing a good reason to *not* include it.

krusmir:
@ashex Im doing an automated build with my fork (which includes STOPSIGNAL).  Here's the link to the tested Dockerfile: https://raw.githubusercontent.com/krusmir/PokemonGo-Map/develop/Dockerfile

Want to confirm all works correctly, otherwise, @tkSimon solution would be needed so builds don't break if using docker hub autobuilds.

Ashex:
My point was that I'm not seeing this huge problem that @tkSimon is discussing. Docker Hub builds images from a dockerfile containing `STOPSIGNAL` just fine, the version it builds with is therefore a non-issue from my perspective.

We're not building on a legacy platform, this is 2016 (yes I just said that ), users should be running a current release of Docker. 

Anyways, I'm not going to make a mountain out of a mole hill. dumb-init works, it doesn't add a lot of dependencies so we can use it. `STOPSIGNAL` is nice because it's built-in.

krusmir:
@ashex @tkSimon  For what is worth,  there were no problems using docker hub automated builds.  You can get the repo + image here: https://hub.docker.com/r/krusmir/pokemongo-map-test/

chuyskywalker:
See #952 (tl;dr: `dumb-init` shall be the choice here.)

PR #619, opened 2016-08-13
Fixed the scan follows location feature for installations running under a URL path

jasonsschin:
## Description

Fixed the "Scan Follows Location" feature for installations running under a URL path

## Motivation and Context

The "Scan Follows Location" feature is broken when installed under a URL path. 

Before the fix many errors can be observed in the developer tools console after this feature is turned on.

## How Has This Been Tested?

Tested against own local installation.

N/A

## Types of changes

## Checklist:

… url a path

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1906,13 +1906,13 @@ $(function () {
   window.setInterval(function () {
     if (navigator.geolocation && Store.get('geoLocate')) {
       navigator.geolocation.getCurrentPosition(function (position) {
-        var baseURL = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '')
+        var baseURL = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '') + (location.pathname ? location.pathname : '')
         var lat = position.coords.latitude
         var lon = position.coords.longitude
 
         // the search function makes any small movements cause a loop. Need to increase resolution
         if (getPointDistance(marker.getPosition(), (new google.maps.LatLng(lat, lon))) > 40) {
-          $.post(baseURL + '/next_loc?lat=' + lat + '&lon=' + lon).done(function () {
+          $.post(baseURL + 'next_loc?lat=' + lat + '&lon=' + lon).done(function () {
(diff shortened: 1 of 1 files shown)


Chlodochar:
Won't this cause problems when using `/map.html` or `/statistics.html`?

It might be possible to use something like:
```
var baseUrl = baseUrl.substr(0, baseUrl.lastIndexOf('/')+1);
```
so that the 'filename' is being stripped off.

jasonsschin:
Doesn't seem to effect the stats page or maps page for me.

static/js/map.js
```diff
@@ -1906,13 +1906,13 @@ $(function () {
...
       navigator.geolocation.getCurrentPosition(function (position) {
-        var baseURL = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '')
+        var baseURL = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '') + (location.pathname ? location.pathname : '')
         var lat = position.coords.latitude
         var lon = position.coords.longitude
 
         // the search function makes any small movements cause a loop. Need to increase resolution
         if (getPointDistance(marker.getPosition(), (new google.maps.LatLng(lat, lon))) > 40) {
-          $.post(baseURL + '/next_loc?lat=' + lat + '&lon=' + lon).done(function () {
+          $.post(baseURL + 'next_loc?lat=' + lat + '&lon=' + lon).done(function () {
```
Chlodochar: When you're on `/stats.html`, then `location.pathname` will be `/stats.html`. That means that `baseUrl` is `http://localhost:5000/stats.html`. That means that `baseURL + 'next_loc?lat=' + lat + '&lon=' + lon` will be `http://localhost:5000/stats.htmlnext_loc?lat=000000&lon=000000`, which is incorrect.

Apart from that, `changeSearchLocation` seems to be doing this correctly. Could you verify this? It might be a good idea to use that function instead of fixing this one.

beccasafan:
#1086 has been merged in and contains a fix for this.

Issue #338, opened 2016-08-09
[Bug] Cannot display pokemon popup


Slivo-fr:
Sometime on chrome mobile, clicking on pokemon does not display the popup.

## Expected Behavior
Clicking on the pokemon should display the popup

## Current Behavior
Sometime, on a specific pokemon, on chrome mobile, popup don't show.
A darker square is shown while I click on the pokemon, and the popup does not show up.
Refreshing the page does not solve the problem, and the popup still won't pop up !

## Steps to Reproduce (for bugs)
1.display map on mobile
2.Check pokemon
3.Background chrome to use the app
4.Come back to check map
5.repeat.

## Your Environment
Using a debian hosted map, on chrome mobile latest version.

## Context
Might happens when the same pokemon spawn in the same location multiple times in a row.

NoMotion:
Is there an issue with Chrome on desktop?

Slivo-fr:
Nope,  it often (always)  happens when chrome was in the background for some time. And allais on the pokemon I want to check, maybe the first I try to click

Slivo-fr:
Important point I forgot, adding it to the original thread : the problem is still there if I refresh the webpage !

desource90:
Does this bug only occur for pokemon you've included on your notify list?

Slivo-fr:
Might be the case yeah, didn't think about it

tkSimon:
Probably related to #418

Studio-XP:
Yes, currently notifications break the page on Chrome for Android.

beccasafan:
Caused by #418. Closing in favor of leaving that one open.

PR #623, opened 2016-08-13
Fix #418: Prevent notifications from crashing the site on mobile and make audio work correctly across devices

### LINKED ISSUE #418: Problem when turning on notifications on Android Chrome

Studio-XP:
## Description
Fixes notifications from crashing the page on Android. I'm an ass, so I didn't realize that I shouldn't make pull requests on default branch on my forked version :/ Proper pull request for #461

## Motivation and Context
With notifications enabled while viewing on your phone, the page would crash, not showing any new Pokemon #338 #418. This commit prevents the page from crashing, while enabling mobile users to receive push notifications on their phone.

## How Has This Been Tested?
On both Chrome and Firefox for both desktop and Android.

## Types of changes

## Checklist:


khag7:
Yay! I was just about to install Firefox on my phone to get mobile notifications going but I'll hold off for this to be merged into a release

Studio-XP:
Only catch though is that you need to have this behind an ssl proxy, either Nginx or Apache, for the notifications to work on Chrome. Chrome on Android prevents notifications from an untrusted location (non https)

static/js/map.js
```diff
@@ -1400,7 +1400,7 @@ function loadRawData () {
 
 function processPokemons (i, item) {
   if (!Store.get('showPokemon')) {
-    return false; // in case the checkbox was unchecked in the meantime.
+    return false // in case the checkbox was unchecked in the meantime.
```
Chlodochar: I get why this was done, but it is unrelated to change. `develop` now already has this fixed.

static/js/map.js
```diff
@@ -1448,7 +1448,7 @@ function processPokestops (i, item) {
 
 function processGyms (i, item) {
   if (!Store.get('showGyms')) {
-    return false; // in case the checkbox was unchecked in the meantime.
+    return false // in case the checkbox was unchecked in the meantime.
```
Chlodochar: Same here.

static/js/sw.js
```diff
@@ -0,0 +1 @@
+
```
Chlodochar: What is this doing?

pogom/app.py
```diff
@@ -226,6 +227,9 @@ def get_stats(self):
                                valid_input=self.get_valid_stat_input()
                                )
 
+    def static_from_root(self):
+        return send_from_directory(self.static_folder + '/js', request.path[1:])
+
```
Chlodochar: Why is this needed?

Studio-XP:
So in order for the service worker (mobile notifications) to work, there has to be a blank sw.js file located in the document root. The changes in the app.py loads that file in the document root

Chlodochar:
@Studio-XP That sounds pretty awful, but if that's what needed to get this going I'm all :+1: with this. Could you just add the explanation you just gave to `sw.js` as a comment or before `self.route("/sw.js", methods=['GET'])(self.static_from_root)`. It seems to come out of nowhere at the moment.

With that knowledge, the code looks good :+1:.

Chlodochar:
Tested on Android. Notifications crashed before and with this PR it doesn't anymore. Code makes sense now. :+1: from me. Another contributor must merge.

khag7:
Works for me

DiscoTim:
not only do they not crash, they actually work and the sound works.. :) good job!

Studio-XP:
@DiscoTim Thanks for the help with the 2 libraries

khag7:
I've been playing around with notifications and I've figured out what the deal is with the service worker file (sw.js)

We can use that to display notifications even when the browser is closed. This has pretty cool implications. On mobile, a user could have the browser completely closed but could still receive notifications. Is this something we should look into doing?

Studio-XP:
I tried to look into how to open an address with it, but was unable to find out any info. Let me know what you find. I would love to be able to open the page with it like on desktop

khag7:
my code is kind of a ridiculous mess and I don't have much more time to tweak it today.

So far I have it so that the notifications update every few seconds. This gives me a countdown on the timer and if my location changes the distanc updates as well. So as I walk around outside the nearby pokemon are in my notification bar (with icons!) and if i pull down it updates distance and time every few seconds.

When I implement the action buttons some other stuff breaks, but I'm working on figuring out what the deal is with that and how to fix it.

Once I get all that working, I want to move the code out of the map.js and into the sw.js so that it can run even if the browser is closed. This will probably be the most difficult step and that might be where I get lost. It definitely needs to be a separate PR but I just figured it's worth mentioning here. It involves the user getting an API key for google cloud messaging (GCM) (similar to getting a maps API key) and setting up runserver.py to send out periodic notifications (since the browser isn't open, map.js isn't loaded and isn't constantly refreshing data, the server has to send it to GCM)

Chlodochar:
Sounds cool. Not for this pr but definitely worth looking into. A new issue for this would be appropriate.

khag7:
@Chlodochar Yeah for sure, I'll wait til I have things a bit more "together" before I even open an issue. The only reason I brought it up here is that using a blank sw.js and putting notifications in the map.js is kind of hacky. The serviceworker js is there for a reason, we're skipping it entirely and even though it works we could definitely do better!

DiscoTim:
its a very basic start to working notifications, no service worker means no notifications, or as it is now broken notifications, and a blank service worker is better than none. notifications could be extended forever, and i agree with @khag7 should be separated, the changes this pr makes are minimal and work, and can be easily reverted or extended for when someone pr's a better working solution.

Becold:
What happened to this pull request ? Why is it not merged ?

Chlodochar:
Each PR needs 2 contributors approving it before merging. I have approved it. Another contributor still needs to do so, that's why I added the `awaiting review` label.

BaIthamel:
You need to deal with or at least respond to Chlo's inline comments before this will be merged.

static/js/map.js
Studio-XP: Sorry, didn't notice it was fixed

static/js/map.js
Studio-XP: Same here

static/js/sw.js
Studio-XP: The core of push notifications are serviceworkers. A sw.js file is required in order for them to work. A blank sw.js will introduce basic functionality, but we can add code later on to expand on the push notifications.

pogom/app.py
Studio-XP: This line permits us to load the sw.js file from the document root (localhost:5000/sw.js). sw.js needs to be in the root of the site directory for it to work across the whole site.

Imagine you have define a cookie with a cookie path (directory). Those cookies will only work from that directory and deeper. Similar story here. A sw.js called from a subdirectory will only work from that directory and deeper. So if the sw.js file is located at localhost/example/, it will only work from the /example/ directory and on

Studio-XP:
Sorry, thought I did in the comment under the SSL reference 594.

nunopitbull:
Is this possible to copy paste on top of:
https://github.com/PokemonGoMap/PokemonGo-Map/pull/883

?

If not, how can I apply this without breaking the script above mentioned?

Studio-XP:
@nunopitbull, I don't see any conflicts with the 2 sets of code. You should be able to modify it without any problems :)

nunopitbull:
I mean, i'm a newbie. I just know how to grab on your RAW and copy/paste on the other one. Won't that delete the changes made by the "Lured pokestops" script?

Studio-XP:
Yes, you cannot grab the raw version. Just follow the changes

nunopitbull:
http://prntscr.com/c903of

Studio-XP:
Make sure that the indentation on that line is the same as the one above it. Python is real particular on it

nunopitbull:
Whats a identention? :P

Studio-XP:
Yes. by indentation, i mean using the [ tab ] key. If you highlight the line above, you'll notice that it's not spaces, but instead tabs. Python doesn't use curly brackets `{ }` for statement blocks (if, else, function(), etc...), as javascript/c languages do, but instead relies on proper, consistent indentation. Since the python code on this project uses tabs for indentation, you must use tabs too. Just make sure that there are no spaces before my lines of code, and that you use the [ tab ] key instead. Do the same with the `def static_from_root()` function towards the bottom

nunopitbull:
Same error, I used tab now

Studio-XP:
He's running fine now, was using wrong line spacing in app.py

khag7:
@Studio-XP have you looked at notification actions at all? It's pretty neat, I think it would go well with this feature (as another PR, sure, but along the same lines). See my PR which won't get merged: https://github.com/PokemonGoMap/PokemonGo-Map/pull/866

Studio-XP:
Yes, it is still relevant. I may have to open a new pr due to the changes to map.js a few weeks ago though

Studio-XP:
Somehow this pulled my gym marker pr into it.  I'm gonna open a new clean pr

PR #633, opened 2016-08-13
Added spawnpoints-only mode to beehive

Xcelled:
## Description
Beehive can be quite inefficient, scanning over and over areas that have no spawnpoints. I've implemented this PR to automatically remove areas without spawnpoints from beehive's searches.

## Motivation and Context
Implemented as a proof of concept of filtering a search location based on spawnpoints. This is a useful proof of concept, though, in that it significantly reduces the work needed to scan an area with beehive, once spawnpoints have been collected. It's kind of a hybrid beehive/spawnpoint scanner that's a quick way to get some benefits of spawnpoint scanning.

When using this option, it's possible to reduce the number of accounts or increase the scan delay time and still scan the same area in the same time. Useful if you're being rate-limited or live in a rural area.

Also, generated locations are now cached in a list instead of being re-computed every time.

## How Has This Been Tested?
I ran both in spawnpoints mode and regular mode.

image

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -21,7 +21,7 @@ import logging
 import math
 import random
 import time
-
+import geopy.distance as geopy_distance
 
 from threading import Thread, Lock
 from queue import Queue, Empty
@@ -31,7 +31,7 @@ from pgoapi.utilities import f2i
 from pgoapi import utilities as util
 from pgoapi.exceptions import AuthException
 
-from .models import parse_map
+from .models import parse_map, Pokemon
 
 log = logging.getLogger(__name__)
 
@@ -129,6 +129,8 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
 
     # A place to track the current location
     current_location = False
+    locations = []
+    spawnpoints = set()
 
     # The real work starts here but will halt on pause_bit.set()
     while True:
@@ -161,11 +163,33 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
                 except Empty:
                     pass
 
+            # update our list of coords
+            locations = list(generate_location_steps(current_location, args.step_limit))
+
+            # repopulate our spawn points
+            if args.spawnpoints_only:
+                # We need to get all spawnpoints in range. This is a square 70m * step_limit * 2
+                sp_dist = 0.07 * 2 * args.step_limit
+                log.debug('Spawnpoint search radius: %f', sp_dist)
+                # generate coords of the midpoints of each edge of the square
+                south, west = get_new_coords(current_location, sp_dist, 180), get_new_coords(current_location, sp_dist, 270)
+                north, east = get_new_coords(current_location, sp_dist, 0), get_new_coords(current_location, sp_dist, 90)
+                # Use the midpoints to arrive at the corners
+                log.debug('Searching for spawnpoints between %f, %f and %f, %f', south[0], west[1], north[0], east[1])
+                spawnpoints = set((d['latitude'], d['longitude']) for d in Pokemon.get_spawnpoints(south[0], west[1], north[0], east[1]))
+                if len(spawnpoints) == 0:
+                    log.warning('No spawnpoints found in the specified area! (Did you forget to run a normal scan in this area first?)')
+                any_spawnpoints_in_range = lambda coords: any(geopy_distance.distance(coords, x).meters <= 70 for x in spawnpoints)  # NOQA
+                locations = [x for x in locations if any_spawnpoints_in_range(x)]
+
+            if len(locations) == 0:
+                log.warning('Nothing to scan!')
+
         # If there are no search_items_queue either the loop has finished (or been
         # cleared above) -- either way, time to fill it back up
         if search_items_queue.empty():
             log.debug('Search queue empty, restarting loop')
-            for step, step_location in enumerate(generate_location_steps(current_location, args.step_limit), 1):
+            for step, step_location in enumerate(locations, 1):
                 log.debug('Queueing step %d @ %f/%f/%f', step, step_location[0], step_location[1], step_location[2])
                 search_args = (step, step_location)
                 search_items_queue.put(search_args)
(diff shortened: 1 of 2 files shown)


stlmikey80:
How do I use this?

chuyskywalker:
https://github.com/PokemonGoMap/PokemonGo-Map/issues/179#issuecomment-239650626

danieliser:
Awesome.

stlmikey80:
Very nice, i like it..

chuyskywalker:
Works for me, doesn't impede any existing functionality, heading down the path we're going towards. May be removed, when better more integrated feature set comes along, but functional for now.

2016-08-13 22_03_08-pokemon go map

An example of SF where I only let the regular beehive run for one loop, then activated this mode. Did was what expected.

Xcelled:
Cute graphic :3

pogom/search.py
```diff
@@ -161,11 +163,32 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
...
+                log.debug('Spawnpoint search radius: %f', sp_dist)
+                # generate coords of the midpoints of each edge of the square
+                south, west = get_new_coords(current_location, sp_dist, 180), get_new_coords(current_location, sp_dist, 270)
+                north, east = get_new_coords(current_location, sp_dist, 0), get_new_coords(current_location, sp_dist, 90)
+                # Use the midpoints to arrive at the corners
+                log.debug('Searching for spawnpoints between %f, %f and %f, %f', south[0], west[1], north[0], east[1])
+                spawnpoints = set((d['latitude'], d['longitude']) for d in Pokemon.get_spawnpoints(south[0], west[1], north[0], east[1]))
+                if len(spawnpoints) == 0:
+                    log.warning('No spawnpoints found in the specified area! (Did you forget to run a normal scan in this area first?)')
+                locations = [coords for coords in locations if any(geopy_distance.distance(coords, x).meters <= 70 for x in spawnpoints)]
```
chuyskywalker: This is...a lot of list comprehension and I'm not python-ista enough yet to grok this at a glance. It's working, but confusing to me.

Xcelled: Blame flake! The lambda version was easier to understand.

```python
any_spawnpoints_in_range = lambda coords: any(geopy_distance.distance(coords, x).meters <= 70 for x in spawnpoints)
locations = [x for x in locations if any_spawnpoints_in_range(x)]
```
>any will return True when at least one of the elements is Truthy.
It's also short circuiting, so it stops at the first True.

The argument we pass to any builds a generator (**not** a list) for on demand evaluation. This generator operates on the spawnpoints set and, for each one, returns if it is less than 70 meters away from the given coords. The net result being that that lambda returns true if at least one spawn point is in range of the circle location.

That lambda is used inside a list comprehension that essentially acts as a filter. It returns a list of all locations where that location has a spawnpoint in range.

Flake complained about the lambda, so I had to replace the friendly name with the actual lambda, but it works the same.

chuyskywalker: Huh, well here's why it complained: http://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

_Kinda_ makes sense....I guess. ^shrug^ As with JS, we picked a common standard and just went with it in order to just keep things inline.

Xcelled: Yeah... I agree for larger/reused functions, you definitely want to def them (seewhatIdidthere). But in this case, my purpose was to decompose one line into two for readability, and adding a def would have worsened that by making you track down another function def that would have been just one line used nowhere else... Actually... wow I'm ashamed of myself. I screwed this up in terms of readability. One sec, I'll fix it in a new PR.

Xcelled: Done, see #645. I forgot I could put a def there :<

FrostTheFox:
only one dev +1? niiiice

uparikh113:
Might be missing something really obvious but ran the scan with 0 steps and spawn points only. Got the error that no spawn points were found even though I ran the beehive scan. Do I need to extract out the spawns from the database myself and specify them in a specific file?

drewreece:
@uparikh113, you are misunderstanding what this does - do not use a step size of 1. This code only removes cells without ANY spawnpoints, it doesn't use the timing data at all. Go and use another fork if you want to grab spawns points that are currently spawning. e.g. Sorokyls/ TBTerra versions in #179 or maybe #585.

ovkhan:
How do i enable this mode?
I don't see anything in options ...

vinsilsim:
another legal option would enable a flag to search in all areas with points of spawn.

drewreece:
@ovkhan re-read the options (python runserver.py --help).  If you do not have '--spawnpoints-only' you need to git fetch & pull or check what version you are running.

drewreece:
@vinsilsim, what do you mean? The new '--spawnpoints-only' option sort of does that already & also removes empty points from scans.

Becold:
Does it scan pokestop & gyms too ? It doesnt seems to

vinsilsim:
@drewreece this option makes the search go in all areas with marked respawn? without putting coordinates?

drewreece:
Maybe I'm being dumb - I still don't understand vinsilsim, this is the 'standard beehive pattern' that is defined in your config or command line arguments. It simply excludes cells that have never shown spawns in earlier scans.

Are you talking about a different change? Try the feature to see what it does, pass --spawnpoints-only to the command.

drewreece:
@Becold, I see Pokéstops & gyms using this feature. Lured stops change status, gyms seem to be correct.

You will need to provide more info or raise an issue if you cannot see them, anyway github is not for support so I'm going to stop trying to answer support questions here.

lungbubble:
Just copied those 2 files, and when finally doing --scanpoints-only I got 
AttributeError: type object 'Pokemon' has no attribute 'get_spawnpoints'

Xcelled:
You can't just drop these files in. This has been merged into develop - just pull the latest down.

rrbailey89:
@Xcelled will this scan less often but more effectively .. or will this cycle the scan rotation again faster resulting in scanning the same locations more often?

Xcelled:
The latter. I'd recommend increasing your sd so you don't get banned for teleporting.

rrbailey89:
@Xcelled do you recommend using an -SD larger than 10?

Xcelled:
Yes. How much larger, though, is up to you, and depends on how many spawnpoints you have nearby.

rrbailey89:
@Xcelled [this is the area im covering](http://goo.gl/bFScVP) I suppose ill try 15 or 20 maybe

Xcelled:
Sounds like a good staring point yeah. You can tweak it later if need be.

PR #16, opened 2016-08-07
Add scan-here button

tkSimon:
## Description
Adds a "scan here" Floating Action Button at bottom center of the map. When clicked/touched, will set the current map center as new search location.

## Motivation and Context
When outside hunting Pokemon you often want to check a few nearby places before deciding which way to go. This is way more convenient when you can just move the map one or two kilometers/miles and tap "scan here" instead of having to type in the name of some location.

## How Has This Been Tested?
Tested using docker.

## Types of changes
[X] New feature (non-breaking change which adds functionality)

## Checklist:
[X] My code follows the code style of this project. (I think?)

PR #642, opened 2016-08-14
Remove dynamic location

chuyskywalker:
This tears out the client-side location changing system.

When scanning was much, much faster, it made lots of sense to have a
"pulse" of scans running around you and to reset your scan origin
location frequently without worry.

Now that we're getting closer to the realization that spawed pokemon
are scheduled, rather than randomly discovered, optimizations wiil be
coming down the pipe to greatly increase the scan rate for huge areas.

Additionally, this feature added a lot of confusing complexity for
people that ended up sharing their map with others (who controlled
the search center? Who knows!)

With these things in mind, it no longer makes sense to have the web
app "follow" a single user around, but instead have several relatively
large areas pre-mapped so that a user can roam within those freely
while the application takes care of optimally finding pokemon.
(As opposed to a user having a very small scan area that constantly
pulses around them.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ src/
 _build
 _build_html
 beehive.*
+config/ssl*
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -27,8 +27,6 @@ class Pogom(Flask):
         self.json_encoder = CustomJSONEncoder
         self.route("/", methods=['GET'])(self.fullmap)
         self.route("/raw_data", methods=['GET'])(self.raw_data)
-        self.route("/loc", methods=['GET'])(self.loc)
-        self.route("/next_loc", methods=['POST'])(self.next_loc)
         self.route("/mobile", methods=['GET'])(self.list_pokemon)
         self.route("/search_control", methods=['GET'])(self.get_search_control)
         self.route("/search_control", methods=['POST'])(self.post_search_control)
@@ -36,9 +34,6 @@ class Pogom(Flask):
 
     def set_search_control(self, control):
         self.search_control = control
-
-    def set_location_queue(self, queue):
-        self.location_queue = queue
 
     def set_current_location(self, location):
         self.current_location = location
@@ -63,7 +58,6 @@ class Pogom(Flask):
 
     def fullmap(self):
         args = get_args()
-        fixed_display = "none" if args.fixed_location else "inline"
         search_display = "inline" if args.search_control else "none"
 
         return render_template('map.html',
@@ -71,7 +65,6 @@ class Pogom(Flask):
                                lng=self.current_location[1],
                                gmaps_key=config['GMAPS_KEY'],
                                lang=config['LOCALE'],
-                               is_fixed=fixed_display,
                                search_control=search_display
                                )
 
@@ -112,35 +105,6 @@ class Pogom(Flask):
             d['spawnpoints'] = Pokemon.get_spawnpoints(swLat, swLng, neLat, neLng)
 
         return jsonify(d)
-
-    def loc(self):
-        d = {}
-        d['lat'] = self.current_location[0]
-        d['lng'] = self.current_location[1]
-
-        return jsonify(d)
-
-    def next_loc(self):
-        args = get_args()
-        if args.fixed_location:
-            return 'Location changes are turned off', 403
-        # part of query string
-        if request.args:
-            lat = request.args.get('lat', type=float)
-            lon = request.args.get('lon', type=float)
-        # from post requests
-        if request.form:
-            lat = request.form.get('lat', type=float)
-            lon = request.form.get('lon', type=float)
-
-        if not (lat and lon):
-            log.warning('Invalid next location: %s,%s', lat, lon)
-            return 'bad parameters', 400
-        else:
-            self.location_queue.put((lat, lon, 0))
-            self.set_current_location((lat, lon, 0))
-            log.info('Changing next location: %s,%s', lat, lon)
-            return 'ok'
 
     def list_pokemon(self):
         # todo: check if client is android/iOS/Desktop for geolink, currently
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -109,7 +109,7 @@ def fake_search_loop():
 
 
 # The main search loop that keeps an eye on the over all process
-def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_path):
+def search_overseer_thread(args, pause_bit, encryption_lib_path):
 
     log.info('Search overseer starting')
 
@@ -143,23 +143,6 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
                     pass
             time.sleep(1)
             continue
-
-        # If a new location has been passed to us, get the most recent one
-        if not new_location_queue.empty():
-            log.info('New location caught, moving search grid')
-            try:
-                while True:
-                    current_location = new_location_queue.get_nowait()
-            except Empty:
-                pass
-
-            # We (may) need to clear the search_items_queue
-            if not search_items_queue.empty():
-                try:
-                    while True:
-                        search_items_queue.get_nowait()
-                except Empty:
-                    pass
 
         # If there are no search_items_queue either the loop has finished (or been
         # cleared above) -- either way, time to fill it back up
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -97,9 +97,6 @@ def get_args():
     parser.add_argument('-nsc', '--no-search-control',
                         help='Disables search control',
                         action='store_false', dest='search_control', default=True)
-    parser.add_argument('-fl', '--fixed-location',
-                        help='Hides the search bar for use in shared maps.',
-                        action='store_true', default=False)
     parser.add_argument('-k', '--gmaps-key',
                         help='Google Maps Javascript API Key',
                         required=True)
--- a/runserver.py
+++ b/runserver.py
@@ -142,15 +142,11 @@ if __name__ == '__main__':
     pause_bit = Event()
     pause_bit.clear()
 
-    # Setup the location tracking queue and push the first location on
-    new_location_queue = Queue()
-    new_location_queue.put(position)
-
     if not args.only_server:
         # Gather the pokemons!
         if not args.mock:
             log.debug('Starting a real search thread')
-            search_thread = Thread(target=search_overseer_thread, args=(args, new_location_queue, pause_bit, encryption_lib_path))
+            search_thread = Thread(target=search_overseer_thread, args=(args, pause_bit, encryption_lib_path))
         else:
             log.debug('Starting a fake search thread')
             insert_mock_data(position)
@@ -167,7 +163,7 @@ if __name__ == '__main__':
     init_cache_busting(app)
 
     app.set_search_control(pause_bit)
-    app.set_location_queue(new_location_queue)
+    app.set_current_location(position)
 
     config['ROOT_PATH'] = app.root_path
     config['GMAPS_KEY'] = args.gmaps_key
--- a/runserver.py
+++ b/runserver.py
@@ -13,7 +13,6 @@ import ssl
 from distutils.version import StrictVersion
 
 from threading import Thread, Event
-from queue import Queue
 from flask_cors import CORS
 from flask_cache_bust import init_cache_busting
 
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1,3 +1,12 @@
+/* global google */
+/* global $ */
+/* global Notification */
+/* global navigator */
+/* global location */
+/* global localStorage */
+/* global centerLat */
+/* global centerLng */
+
 //
 // Global map.js variables
 //
@@ -23,7 +32,6 @@ var notifiedRarity = []
 var map
 var rawDataIsLoading = false
 var locationMarker
-var marker
 
 var noLabelsStyle = [{
   featureType: 'poi',
@@ -733,14 +741,6 @@ var StoreOptions = {
     default: false,
     type: StoreTypes.Boolean
   },
-  'geoLocate': {
-    default: false,
-    type: StoreTypes.Boolean
-  },
-  'lockMarker': {
-    default: isTouchDevice(), // default to true if touch device
-    type: StoreTypes.Boolean
-  },
   'startAtUserLocation': {
     default: false,
     type: StoreTypes.Boolean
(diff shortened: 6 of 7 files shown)


Xcelled:
Will the marker follow the user clientside? I've never gotten that working, but also never really tried. Usually, my (painful) method is guessing pokemon positions based on street intersections!

chuyskywalker:
The "hey map, follow the user" was kinda...intermixed here. There is a button in the lower right that can center the map on the user, but you must be serving the site over SSL (for a lot of browsers, geolocation now requires ssl -- which makes sense).

We'll have to bring back an option to 'perma follow' later on. And that would work client side, per user since it's just a matter of where the little "me" icon is located and the `pan` of the map.

Xcelled:
Good to know, thanks. I might go request a feature about that and also changing the marker size/shape/image. There's a spawnpoint that's always covered by mine.

chuyskywalker:
The "follow me" icon is a different one, less in the way. This PR actually removes the obnoxious little google marker.

hemsleyk:
+1 for removing that awful google marker

Worthy7:
May I ask if anyone is working on dynamically scanning areas around the user, but using the spawn points instead of the hexagon? I'm sure it cannot be too many steps to implement (Im still learning)

Slivo-fr:
I'm not sure I understand what you did well, but if it's about removing the 'follow location' toggle and feature, you need to provide another way to allow user to center the scan on their location, like a 'scan here'  button

Worthy7:
OP please can you fix this so that by default, the slider for "Start at user's location" is ON instead of off, or just in general remove that slider and make the map always start on the user. there isn't really any reason that it isn't the default.
I merged your changes into my branch, works fine, but I dont know how to remove that stuff

Moussa:
@Slivo-fr [This PR does that](https://github.com/PokemonGoMap/PokemonGo-Map/pull/16), and it's sorely needed if you're not the only one using a tracker.

DiscoTim:
i understand the need for these changes based on this thread https://github.com/PokemonGoMap/PokemonGo-Map/issues/179, and i think it is an excellent way forward.

it really does need to at least denote, other than you are in the center of the map somewhere, and to a lesser extent track the viewers current location.

Slivo-fr:
I know there is a pr that want to add this kind of scan, what I say is you cannot remove follow location till it's merged. 
Also I'm using this feature on a map I'm the single user, but I scan small range (7steps) and move it where I go when I need using follow location.

chuyskywalker:
#Slivo-fr 

> you need to provide another way to allow user to center the scan on their location, like a 'scan here' button

There's an icon for "location setting" in the lower right of the map. Been there for ages. Still works. This button only centers the *view* on the user, though. It does not change the **_scan_** location.

@Worthy7 

> OP please can you fix this so that by default, the slider for "Start at user's location" is ON instead of off, or just in general remove that slider and make the map always start on the user. there isn't really any reason that it isn't the default.

It's not really my feature, but I suspect that (for a shared map) people didn't want to default users to an empty area or the addition was account for deviceds that don't have geolocation enabled/available in browser.

----

The core of this issue is that dynamically moving the **_scan location_** has become extremely problematic. 

1. We can not scan fast enough to make it efficient to constantly move the scan center. 
2. With spawn point scanning coming down the pipe, it actually requires a **two pass** scan before it becomes efficient (it has to scan for about an hour in a location before you can really use spawn point optimized scans). If we're constantly moving the scan location around, it'll significantly complicate the code behind the optimized scanning system, not to mention it'll probably negate a good portion of it entirely.
3. The "scan follows users" feature has also been strongly problematic when people share the map -- with no true multi user support (which we're _not_ adding), the scan center would get move around arbitrarily and ruin the effectiveness of the scan.

This PR is part of a forward roadmap which includes optimized spawn point scanning and #393 for easier area definition. By locking the scan area down, we make the program operate much better, and with the optimized scanning the locked down scan location isn't a problem because you can cover such a wide enough area that "scan follows user" doesn't really matter anymore.

Slivo-fr:
Exactly the location button on the map only move viewport. I need to keep a way to change the scan location wherever I need to. Activating  and deactivating 'follow location' actually does this

SkOODaT:
Please do not remove this feature my scanner is setup to scan like 5 steps in 10 seconds through multiple accounts it's quick I have web hooks setup to show me distance to Pokemon from the marker location, I'm not greedy just scanning a few blocks and if this gets removed the pin won't follow me to new areas not a good change even with spawn point tracking I think this should be left just in case people want to use it

kevdotcom:
Why remove something you can just deactivate with -fl?

I need this for a project I'm building at the moment, so I'm definitely against removing it!

HasanY1989:
If i underestood this remove the "move around with flag", please do not remove that ! EVen with SpawnTracks spawn point thing..

Moussa:
While I understand the motivation for this change I disagree with the drastic change because it breaks so many use cases. I use my tracker to scan interesting areas in my city, to look for possible nests and such that I can visit. How would I do this in this new world?

DiscoTim:
even still... couldn't the marker just have been disconnected from affecting the scan/backend, and 'Change scan location' becomes 'Change marker location', 'Scan follows location' becomes 'Marker follows location", and 'Start at user's location' basically does the same thing?

chuyskywalker:
> While I understand the motivation for this change I disagree with the drastic change because it breaks so many use cases.

Use cases which aren't tenable for the future of this app; see previous rationales.

> I use my tracker to scan interesting areas in my city, to look for possible nests and such that I can visit. How would I do this in this new world?

Scan the whole city. Not joking, that's how much more effective this will be. So far there's every indication that the spawn points are static and scheduled, so once you have a full list, a handful of accounts can do what previously took dozens. The optimization will be tremendous.

> couldn't the marker just have been disconnected from affecting

*That's already present*, it just doesn't auto-update as you walk around in this PR. That "follow" was tied in with the "and also change the search location" feature. I'll add that auto-follow back in, but it will only affect the map view, it won't change the scan location itself.

---

Y'all need to understand: changing the scan location ruins spawn point scanning. If you move just 10 feet over, we have to (basically) start the scan process over, collecting a regular beehive of scans for 1-2 hours before being able to switch back into spawn-point mode. If we don't do that, by moving the scan location (and not re-beehiving) **you will miss pokemons**. The spawn point optimization only works if you've previously scanned an area -- moving that area around, even a little bit*, would mean needing to re-scan it again.

There is _some_ nuance here, if you are covering a 1KM wide area, and you move 10M to the left, sure, you don't need to re-scan a "brand new" 1KM radius, just the left edge of the new circle -- but now the app has to dynamically figure out what spots qualify for that re-scan and manage that over the next 1-2 hours while it collects enough spawn points to go back into spawn point mode. But you're constantly moving, so it's always juggling "some area is spawn-point scanned, some is not!". It's a coding nightmare. 

It even worse for someone running a small step factor (like 3) (which is one of the use cases being discussed). In that setup, every movement definitely takes you into new territory for which there is no guarantee you have all the spawn points -- you'd be forcing the app into a near permanent, constant full beehive mode around your location. This is the very thing we need to have the app stop doing.

Slivo-fr:
What if I'm traveling to a new city for few hour and just want the map to scan around me for a couple dozen of minutes using the classic mode?

Moussa:
>Scan the whole city. Not joking, that's how much more effective this will be. So far there's every indication that the spawn points are static and scheduled, so once you have a full list, a handful of accounts can do what previously took dozens. The optimization will be tremendous.

I have no doubt spawn point scanning is a massive improvement but I still don't see how it helps my use case. Getting a spawn point map of my (very large) city would take weeks at my current scan speed. And then when I had it how would I 'coax' it to give me an upto date radar of a certain area? A city my size would have spawn points in the hundreds of thousands, I'd still be waiting forever for the tracker to refresh the spawn points I want to eyeball. Even with the most optimistic calculations it would still take me hours to cover my city.

rp000:
> It even worse for someone running a small step factor (like 3) (which is one of the use cases being discussed). In that setup, every movement definitely takes you into new territory for which there is no guarantee you have all the spawn points -- you'd be forcing the app into a near permanent, constant full beehive mode around your location. This is the very thing we need to have the app stop doing.

Hmm, not sure I follow here. I understand the fact if the user moves into new area, and we are in spawnpoint mode, then we do not want to rerun beehive just to catch the spawn points in the new area. That's great.

To do that you want to disallow any scans in areas that we do not have spawn point data for. That's fine too.

However, I don't get the leap from there to "let's not allow users to move locations". Because the user might miss pokemon when they go outside this boundary since spawn points are not complete, and it's a nightmare to make sure that won't happen, so we are gonna pre-define an area, so you will... still miss pokemon when you go outside this boundary?

Now I do see why multiple users moving the marker would be problematic, but that's why we have -fl right? I respect your choice to cater to people running this map as a service, but I am just asking you to consider the little guys that run a small personal scanner with a few accounts. We pop markers down randomly around the city to see if anything interesting is happening but most of the time we don't want to scan the entire city. What's wrong with letting people move markers and only spawnpoint scan a certain radius around that marker, and if you don't have spawnpoint data there, then "too bad, you messed up"?

Xcelled:
The bottom line is that this feature is, at the best, poorly implemented currently. It hammers Niantics servers and is incompatible with the direction the project needs to go.

Therefore it should be removed.

However, it can/will be reimplemented later in a more thoughtful way, especially with regards to spawnpoints.

ilendemli:
Maybe combine those two features into one. I have an app using /loc and /next_loc to set and get current location. With this PR merged, this feature would not work anymore (i am just scanning 3 steps around the current location), maybe make it so that the user can set the location using /next_loc (and also retrieve it) but scan the closest spawnpoints to that location only.

DiscoTim:
@chuyskywalker @Xcelled all spawn point scanning. i get this, and its great.

my concern is removing the location marker, and providing no reference to the users current location, not being able to track the viewers location change, and not allowing the user to set their location for devices that don't have GPS. the app has all this now, it just needs to be dis-connected from the backend, and in doing so the added benefit is that the location marker becomes multi-user friendly.

One of the things i use the location marker for now is distance calculations. how far is a pokemon that spawned from me? what direction is it from me? for example, my notifications all include distance and direction.

chuyskywalker:
@rp000 
>  What's wrong with letting people move markers and only spawnpoint scan a certain radius around that marker, and if you don't have spawnpoint data there, then "too bad, you messed up"?

Because that's terrible experience, sets expectations extremely poorly, and would flood github and the help channel with confused, frustrated users.


@Moussa 
> I have no doubt spawn point scanning is a massive improvement but I still don't see how it helps my use case. 

Your use case is _invalid_. Arbitrary beehive in a "as fast as I can scan" manner is not realistically going to keep working. The fact that it worked when we had free reign of the API and didn't need to worry about being ban-hammered for using the API willy-nilly was pure luck. This is no longer the case.

> Getting a spawn point map of my (very large) city would take weeks at my current scan speed. 
No, it wouldn't. With regular beehive _today_ I can scan about 1/8 of San Francisco in about 20 minutes with 8 accounts. Scanning improvements to suss out just spawn point data will make this even more efficient. Additionally, since spawn points are static, there's a very high chance that we could eventually collect the list centrally and skip most of this entirely (until Niantic changes it, but all indicators I've seen say that's unlikely).

> And then when I had it how would I 'coax' it to give me an up to date radar of a certain area? A city my size would have spawn points in the hundreds of thousands
I doubt that estimate is correct or even close to correct. That SF scan has  ~500 spawn points found (and it's one of the denser areas). Even x8 (which assumes the same density every which is not true) that's approximately 4,000 spawn points. Something like +90% of spawn points only trigger once an hour, so you'd need to scan about 3600 points per hour. Which roughs out to a scan per second. Since each account can only scan once per 10 seconds, you'd need (drum roll please) about 10 accounts.

Like I said, it's realistic to cover a **huge** metropolitan area with a handful of accounts and still catch *every* spawn in that area.

And that number would actually be lower, because a TON of those scan points are within a 70m radius, so you wouldn't need to scan each one individually -- that's just a worst case scenario, in reality it'd be a good deal better.

Moussa:
>I doubt that estimate is correct or even close to correct. That SF scan has ~500 spawn points found (and it's one of the denser areas). Even x8 (which assumes the same density everywhere, which is not true) that's approximately 4,000 spawn points. Something like +90% of spawn points only trigger once an hour, so you'd need to scan about 3600 points per hour. Which roughs out to a scan per second. Since each account can only scan once per 10 seconds, you'd need (drum roll please) about 10 accounts.

I've just checked an area on the very fringes of London and it has ~100 spawn points in an area ~0.4km². I have no idea if spawn point density is greater towards the center of the city but I would be surprised if it wasn't. If we assume it doesn't increase then there are roughly 350,000 spawn points in the city. That doesn't seem to me to be an amount I can realistically scan without an unmanageable (and alarm raising) amount of accounts. But hey, if you think it's possible then I'm happy to wait and see. I just wanted you to understand my use case.

SkOODaT:
what happens if niantic decided to change spawn points or add or remove ones, they may even add more in when they release new generations, i think it be best to leave it as it is but have spawn only scanning if people want to use it

Slivo-fr:
I agree with skoodat make it an option. Let user that are happy with procedural scanning of small area keep going

RatedDS:
I've been using the map since the original version, with a low number of steps just to scan from time to time when I go out to see if there are any interesting Pokemon around,I don't play currently but a friend is using my map for pretty much the same, how does this affect this use case or what should I do to optimize it? I'm confused about all the Web UI settings including this new "Lock marker", I just want the scan to follow the user as always.

Worthy7:
Hi all. After some sleep it's very apparent that I'm wrong and we shouldn't
remove the functionality. Moving forward, what we do need is to clearly
identify the use cases for the main project on this hub.

Personally, I am a fan of the recent branch to allow polygon selection and
thus get scan locations that are a shape defined by the admin. If this was
combined with a more efficient scanner process that was tailored
specifically for catching spawn points - and that's all- then this would be
fantastic. One huge improvement I would like to see is removal of the step
parameter, and instead let the system manage "scan these points for 2
hours" dynamically - even if the number of accounts is lacking to do it all
in one go.

Slivo-fr:
Better keep steps range as an option imo, and allow shape selection / steps adjustment from front. 
I see two major use case :
1 -  I'm running the map on a whole city without ever moving the marker, and can use new system with spawn db & optimized algorithm. 
2 -  I'm using the map for a small area and move it with me without being able to prescan spawns and need to keep using the old scan system.

Slivo-fr:
And, for the second use case, I need to be able to get the scan to follow me and/or quickly set my location as the new scan location

Westacular:
Here's a thought: what if all map cells were forced to align with a pre-defined grid?

- The search location would simply determine _which_ cell is the central one, not the true centre of the search
- The db would store whether a given map cell has complete spawn data
- The search logic could automatically switch between old behaviour and spawn scanning based on whether the cell has complete spawn data:
    * If a map cell has complete spawn data, just do spawn scanning for that region
    * If not, scan it the old fashioned way, but make note of the spawns, and after an uninterrupted hour, flag it as complete and switch to spawn scanning.

This would allow people who use location updating to continue to do so, and the switch to spawn scanning could be smart and automatic.

chuyskywalker:
That idea has a lot merit. It would also allow us to use s2 locations as ids more effectively too. Sharing scanned cells would be a lot more efficient for that as well.

primaxius:
Keep movable scanning as many users run small area scanners they like to move while on the go.

PR #654, opened 2016-08-14
Add server conf option to manually specify encrypt lib location

a-l-e-x-1991:
## Description

Add config option '-el', '--encrypt-lib' to manually specify encrypt lib location

## Motivation and Context

Since PokemonGo-Map relies on a native C encrypt library, it currently works only on a fixed set of systems and archs, for which this library is provided. However, with this change it can be ported to any system that has the needed prerequisites, with an additional step of compiling encrypt lib for the platform and specifying its location with the newly added parameter

## How Has This Been Tested?

This has been tested to work on a MIPSEL router with manually compiled libencrypt.so

## Types of changes

## Checklist:

This allows running server on any system that has necessary prerequisites and compiled encrypt lib available

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -151,6 +151,7 @@ def get_args():
     parser.add_argument('--ssl-privatekey', help='Path to SSL private key file')
     parser.add_argument('-ps', '--print-status', action='store_true',
                         help='Show a status screen instead of log messages. Can switch between status and logs by pressing enter.', default=False)
+    parser.add_argument('-el', '--encrypt-lib', help='Path to encrypt lib to be used instead of the shipped ones')
     parser.set_defaults(DEBUG=False)
 
     args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


pogom/utils.py
```diff
@@ -317,44 +318,47 @@ def send_to_webhook(message_type, message):
...
-def get_encryption_lib_path():
-    # win32 doesn't mean necessarily 32 bits
-    if sys.platform == "win32" or sys.platform == "cygwin":
-        if platform.architecture()[0] == '64bit':
-            lib_name = "encrypt64bit.dll"
-        else:
-            lib_name = "encrypt32bit.dll"
+def get_encryption_lib_path(args):
+    if args.encrypt_lib is not None:
+        lib_path = args.encrypt_lib
```
chuyskywalker: I would add a "file exists" check here with a friendly error message if not present. That way if people fat finger the path name, it doesn't try to load a non-existent file later and do weird stuff.

pogom/utils.py
```diff
@@ -317,44 +318,47 @@ def send_to_webhook(message_type, message):
...
-        raise Exception(err)
+        else:
+            err = "Unexpected/unsupported platform '{}'. If you have encrypt lib compiled for your platform, specify its location with '--encrypt-lib' parameter".format(sys.platform)
+            log.error(err)
+            raise Exception(err)
 
-    lib_path = os.path.join(os.path.dirname(__file__), "libencrypt", lib_name)
+        lib_path = os.path.join(os.path.dirname(__file__), "libencrypt", lib_name)
 
     if not os.path.isfile(lib_path):
```
a-l-e-x-1991: @chuyskywalker 
about your note on line 323

> I would add a "file exists" check here with a friendly error message if not present. That way if people fat finger the path name, it doesn't try to load a non-existent file later and do weird stuff.

Funny how much alike we think in this case :) I almost added that check, but then realized that it's being done already, and I think the message "Could not find %platform% encryption library %path%" is informative enough

a-l-e-x-1991:
@chuyskywalker I've rebased the change, and also placed the check where you wanted it to be, so that now we have 2 different lib not found errors: the old one for automatic mode, and the new one for manual mode: "Could not find manually specified encryption library {}".format(lib_path)

chuyskywalker:
Excellent.

PR #658, opened 2016-08-14
Correct handling spaces on windows in setup scripts

derigel23:


Acidvenom73:
You need the ; after every path. Also, please explain what you're changing, doesn't make sense to people not knowing what it's supposed to do.

derigel23:
First, trailing semicolon is not necessary while it's not critical http://stackoverflow.com/questions/11391390/placing-the-semicolon-in-the-windows-path-environment-variable
Second, quoting in windows variables should be done only once - or in declaration, or in usage.
Second option is most common practice (and in this script too).
Otherwise, it will be double quoting and script  will not run.
That you can check on windows with spacing in path.

primaxius:
Conflicts - closed

PR #670, opened 2016-08-14
Fix basic config example.

NoobDev3:
## Description
basic config example should represent the -l attribute value with double quotes rather than single.

In addition added reminders to make sure you're still in the directory of PokemonGo-Map.

## Motivation and Context
When trying to complete installation and run, additional research required to fully execute.

### DIFF
--- a/docs/basic-install/index.rst
+++ b/docs/basic-install/index.rst
@@ -133,8 +133,8 @@ The most basic config you could use would look something like this:
 
 .. code-block:: bash
 
-  python ./runserver.py -a ptc -u 'USERNAME_HERE' -p 'PASSWORD_HERE' \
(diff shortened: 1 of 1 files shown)


chuyskywalker:
Why? Double quotes cause most shells to jump into a form of interpolation which you don't want for a field like this.

NoobDev3:
@chuyskywalker have you tried to run it using single quotes in windows? it doesn't work.

Actually this PR doesn't completely fix it, because the other params need double quotes as well:

-u
-p
-l
-k

chuyskywalker:
omg windows...why you suck so much. Go ahead and change them all.

NoobDev3:
New commit has the changes for all parameters.

chuyskywalker:
LGTM

PR #677, opened 2016-08-14
Don't keep parse lock after map parse fail

bmtran:
## Description
If there was a map parse fail, it would hold the map parse lock while it waited for a retry. Change it so that it releases the lock before it waits.

## Motivation and Context
This allows other threads to continue while we wait to retry the failed thread.

## How Has This Been Tested?
Ran it on my Windows computer with 10 accounts.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -276,7 +276,6 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                         continue
 
                     # Got the response, lock for parsing and do so (or fail, whatever)
-                    fail_wait = False
                     with parse_lock:
                         try:
                             parse_map(response_dict, step_location)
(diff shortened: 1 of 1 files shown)


justsomedudeonthenet:
Thank you a million times over.  I was trying to figure out why all the workers got behind when one account had issues...sleeping inside of a global lock is definitely a bad idea.

chuyskywalker:
This was a bit poorly implemented -- the sleep triggeres EVERY time now, when it's really only needed on the KeyError.

Won't matter much if I can get #582 added anyway.

sadimusi:
In my opinion this whole section should look like this:
````python
try:
    with parse_lock:
        parse_map(response_dict, step_location)
    log.debug('Search step %s completed', step)
    search_items_queue.task_done()
    break  # All done, get out of the request-retry loop
except KeyError:
    failed_total += 1
    time.sleep(sleep_time)
````

PR #681, opened 2016-08-15
Added velocity limiter

Xcelled:
## Description
Adds support for velocity limiting of worker movements. Eliminates teleporting while scanning, especially with the new spawnpoint-based beehive. When the worker moves to a new location, it will calculate the velocity it 'moved' at. If this is higher than some target velocity, it will delay sufficient time before scanning to 'fake' the target velocity. In effect, it's an adaptive scan delay.

It also prints a warning:

```2016-08-14 20:18:27,793 [ search_worker_0][        search][ WARNING] Worker moving faster (94.37 m/s) than maximum allowed (24.00 m/s). Slowing things down. You need more accounts or a longer scan delay (hint: use -sd option)```

This warning might want to be refactored a bit or even changed to a debug line - I'm open to suggestions.

This WILL slow down scans, even multi-account scans, since every worker is essentially in a car. This could be reduced/alleviated by non-sequential scans like spawnpoints (so workers aren't 'driving' past each other constantly) or else a distance-based method of selecting the next scan area... or even assigning workers pie-slice shaped wedges of the scan area.

## Motivation and Context
Teleporting is an easy way to get banned

## How Has This Been Tested?
Ran it on my system

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -228,8 +228,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
             if args.proxy:
                 api.set_proxy({'http': args.proxy, 'https': args.proxy})
 
-            # Get current time
-            loop_start_time = int(round(time.time() * 1000))
+            last_scan_time, last_scan_location = 0, None
 
             # The forever loop for the searches
             while True:
@@ -238,6 +237,22 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                 step, step_location = search_items_queue.get()
 
                 log.info('Search step %d beginning (queue size is %d)', step, search_items_queue.qsize())
+
+                if args.max_speed and last_scan_location:
+                    elapsed = time.time() - last_scan_time
+                    travel_dist = geopy_distance.distance(last_scan_location, step_location).meters
+                    log.debug('Worker speed: %0.2f m/s', travel_dist / elapsed)
+                    # calculate how long it would take to travel this distance at max speed
+                    min_elapsed = travel_dist / args.max_speed
+
+                    if elapsed < min_elapsed:
+                        # Gotta slow it down
+                        log.warning('Worker moving faster (%0.2f m/s) than maximum allowed (%0.2f m/s). Slowing things down. You need more accounts or a longer scan delay (hint: use -sd option)',
+                                    travel_dist / elapsed, args.max_speed)
+                        time.sleep(min_elapsed - elapsed)
+
+                # Get current time
+                loop_start_time = int(round(time.time() * 1000))
 
                 # Let the api know where we intend to be for this loop
                 api.set_position(*step_location)
@@ -287,13 +302,12 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                             failed_total += 1
                             time.sleep(sleep_time)
 
+                last_scan_time, last_scan_location = time.time(), step_location
                 # If there's any time left between the start time and the time when we should be kicking off the next
                 # loop, hang out until its up.
                 sleep_delay_remaining = loop_start_time + (args.scan_delay * 1000) - int(round(time.time() * 1000))
                 if sleep_delay_remaining > 0:
                     time.sleep(sleep_delay_remaining / 1000)
-
-                loop_start_time += args.scan_delay * 1000
 
         # catch any process exceptions, log them, and continue the thread
         except Exception as e:
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -138,6 +138,8 @@ def get_args():
                         nargs='*', default=False, dest='webhooks')
     parser.add_argument('--ssl-certificate', help='Path to SSL certificate file')
     parser.add_argument('--ssl-privatekey', help='Path to SSL private key file')
+    parser.add_argument('--max-speed', help='Maximum speed a scanner can "travel" between scan locations in meters/second \
+                        (0 to disable). Defaults to 24 (88 km/h or 55 mph).', type=float, default=24)
     parser.set_defaults(DEBUG=False)
 
     args = parser.parse_args()

### CODE REVIEW

**Comment by Xcelled** [comment]
Technically this is a [HELP WANTED] because I don't know the maximum velocity you can go before being banned.

**Comment by chuyskywalker** [comment]
Someone had, at one point, suggested that a good way to "hide" teleportation would be to divide the scan steps by the number of accounts, and have them start that many scan points down the list. Since the beehive setup does a radial outwards, each step is always 140m(ish) away. It's only with multiple accounts which are "leap frogging" over each other would this be a problem. If each thread scanned a sequential portion of the ring, this wouldn't be an issue.

All of this, of course, would be radically different with sp-scanning since the "cover these spots at XX:09" would require a terrible sort of traveling salesman problem.

**Comment by Xcelled** [comment]
@chuyskywalker I thought about that, too, but it'd be terrible to keep all the workers 'in sync' with the staggered scans. One delayed scan and suddenly your two workers are racing again...

Spawnpoints are going to be an inverse salesman because we want him to take the longest. Doubt there's much research on that one... it's also further complicated by multiple salesmen... ouch my head.

**Comment by mew905** [comment]
it'd still be travelling salesman, but it would greatly increase the number of accounts needed to stay within the acceptable parameters, not to mention all the criss-crossing you may as well just do a beehive scan, it'd be more efficient at that point IMO. you couldn't simply ask for which spawn is closest and have it go there, I can guarantee you'd end up with several travelling the same path in under a minute, you'd have to scan the spawn times and all possible paths (travelling salesman) and calculate the best path where a worker could get there in time, but not send a second worker to the same one, nor send it rushing to a different spawn at 45m/s (remember 27 m/s is 100kph, highway speed). It really is an extremely complicated (and probably very CPU intensive) problem.

**Comment by chuyskywalker** [comment]
I'm not sure I follow. While it's a spiral, we could have coordination across all of them relatively easily.

While the full spiral looks like this:
```
queue0: 1---------2---------3---------4---------5---------6---------
```

Programmatically you'd create a queue per thread, instead of the singular one and just check for all queues to be empty before repopulating them.

If you had 4 accounts, you'd get:

```
queue0: 1---------2----
queue1: -----3---------
queue2: 4---------5----
queue3: -----6---------
```

If you start them all at the same time, in theory they'd all finish at the same time. Since they all travel the same distance between each step, they shouldn't get out of sync (by being delayed). And for scan points that fail, we could just skip them.

```
---x-----------
-------x-------
-------[--]-------
---[--]-----[--]------
```

Even with misses or delays, the overall scan wouldn't get that far out sync.

Right now the individual accounts "hop around" a lot because they arbitrarily pull "the next" location off a shared stack.

**Comment by mew905** [comment]
the issue isn't with the spiral scan, the problem lies with the far more efficient spawn point scans, slowing them down to a limited speed means the semi-random spawn points would need to either be pre-calculated, or run in a similar spiral pattern but only including spawn points, but given spawns have their own timers to meet, it'd be hectic, and quite possibly impossible to calculate on a modern home PC, especially when you consider the possibilities are exponential, and the 3 major cities closest to me consist of 1100, 1700, and 8000 spawn points to be met within an hour

**Comment by Xcelled** [comment]
@chuyskywalker oh, yeah, if you nuke the shared stacked you can do that. Unfortunately, if an account gets banned, that worker's area will never get updated. Actually, this might be a good thing to do anyway, to avoid those mysterious holes as well as to move towards more distributed workers. This would in turn help with the DB choking etc.

AND... if we did that.... 
@mew905 Heuristically, we could assign each sp-worker a starting point in the sp-list (ie, worker 0 starts at index 0, worker 1 at 1, etc through worker n). Then, worker 0 gets to pick its next location from the next N locations. It'd choose the one with the lowest velocity to reach from its previous location. Then worker 1 would choose from N-1... etc. Continue in this way, but the order workers "pick" locations is determined by that workers previous velocity, to hopefully avoid a worker being forced to visit two high velocity points in a row.

In fact, maximum velocity being exceeded could trigger a warning/error to add more accounts to reduce the pressure.

If you agree, I think I'll start ripping out that stack...

**Comment by Xcelled** [comment]
As an addon, in a more distributed model, the overseer could replace the entire worker if it gets banned or crashes. A surplus of (idle backup) workers would mean little to no service interruption!

**Comment by mew905** [comment]
As I stated, we'd likely need more accounts for the spawn point scans to be feasible, so... Yes, I agree, having an exceeded velocity warning to add more accounts would be the safest bet without having to precalculate. If you get a testing version, be sure to test it against the traditional spiral scan for the same area. 14m/s is the optimal given each circle (according to chuyskywalker) is 140m apart and you have 10 seconds to get there, but people have differing opinions on the maximum allowed speed, I've driven on the highway at 30m/s while my girlfriend played along the way so I think that's a safe speed. She didn't catch much of anything though. I was thinking about suggesting a backup set of workers that could just "pop in" where and when needed, as long as there's a reasonable time between these workers being used, it shouldn't be an issue

**Comment by chuyskywalker** [comment]
Sounds like a reasonable variant of this optimization.

Also, go fucking figure, you have ask google to solve your TSP, lol:

https://developers.google.com/optimization/routing/tsp

**Comment by mew905** [comment]
bear in mind TSP doesn't have a time constraint, it's simply the fastest way between points without repeating. With the spawn scan, we need to be at a spawn point at a specific time interval, and be able to repeat it every hour

**Comment by Xcelled** [comment]
TSP also visits everything once per salesman. We don't need that. We need all salesmen collectively to visit all points, but optimally, workers should not re-visit points.

**Comment by mew905** [comment]
what if we sorted the spawns by time (since as far as I understand it, it's seconds into the hour) and then spawn the first worker at the closest to the current time, then see if it can get to the next spawn in the sorted list within 14m/s, if it can't, then spawn the next worker, and continue like that?

**Comment by Xcelled** [comment]
That works too. It's essentially the same idea, just shoving the work onto the minimum number of workers, which is a good idea.

**Comment by Xcelled** [comment]
@chuyskywalker I'll say it again, I think we should only accept spawnpoints to scan from an external json file. That way, we can write a helper (in C if needed) to extract the points from the DB/filter them/optimize them/what have you, so we don't need to do that work every time we start the scanner.

**Comment by jccwright** [comment]
@Xcelled, I really like your idea. What about writing a configuration file that has placeholders for datetimeLastUpdate and spawnPointdAddedSinceLastUpdate and run a periodic check for threshold values runSPHelperAfterSPAdded (default 5) and runSPHelperAfterTimeElapsed (default 60 in min). Output from the helper would go to the standalone JSON file you talk about while archiving any one it replaces.

If part of the Helper was an analyzer then appropriate Metadata could also be written into the file regarding minimum workers; estimated maximum velocities; etc.

**Comment by chuyskywalker** [comment]
I'm just realizing that the idea of multiple queues still has teleporting. As soon as you finish the loop, you have to travel back to the starting location. But perhaps that will be not so bad, but a useful place to have the velocity detection/delay in place.

**Comment by chuyskywalker** [comment]
Or you could have each thread take over the next segment each time, that way only one thread (they thread going from the end of the loop to taking the starting segment) would have to teleport delay.

**Comment by jccwright** [comment]
@Chuyskywalker, this can be mitigated by adding a worker whose job it is to "walk" from the end of the spiral back to center during one cycle. You would want to ignore the payloads from this worker so as not to disrupt the visual aesthetic.

**Comment by Xcelled** [comment]
@chuyskywalker No one's been banned for teleporting like that... yet. But yes, that'd be a place where limiting is potentially useful. It doesn't need to scan, either - people routinely turn screens off as they walk. Having another thread "continue on" is going to be much harder to coordinate than having each just walk back to its start point unless you make failed scans simply not retry. (Otherwise, a failed scan will bring two workers closer to overlapping). Better to teleport or delay, I think.

**Comment by chuyskywalker** [comment]
I don't think it's as hard as you suspect, if we had:

```
        run1:           run2:           run3:           run4:
queue0: 1---------2----|-----3---------|4---------5----|-----6---------*
queue1: -----3---------|4---------5----|-----6---------*1---------2----|
queue2: 4---------5----|-----6---------*1---------2----|-----3---------|
queue3: -----6---------*1---------2----|-----3---------|4---------5----|

Each * represents a "teleport" transition in the overall loop for that worker.
The | marks would be synchronization/pause points.
```

Each iteration only one thread has to traverse from the end of the loop back. Every other thread is just continuing along the long path.

Or, as @jccwright has sorta suggested, have a "walking period"

```
        run1:           run2:           run3:           run4:           run5:
queue0: 1---------2----|-----3---------|4---------5----|-----6---------|___walk home___|
queue1: -----3---------|4---------5----|-----6---------|___walk home___|1---------2----|
queue2: 4---------5----|-----6---------|___walk home___|1---------2----|-----3---------|
queue3: -----6---------|___walk home___|1---------2----|-----3---------|4---------5----|
queue4: ___walk home___|1---------2----|-----3---------|4---------5----|-----6---------|
```

Granted there's some math to do to figure out if "walk home" even fits into the correct speed restriction.

Xcelled:
Yes, but what if Q1 hits a snag on several cells and bleeds over into run2? suddenly Q0 and Q1 are scanning 3, 4, 5, etc at the same time...

chuyskywalker:
That's why I said `|` denotes a sync point. You can't bleed over.

Xcelled:
Whoops, I missed that part xD

I'm not a fan of this either... one failing thread would hold up all the others. Retry would take 210 seconds per scan before it gave up? Times N cells? Ouch!

Xcelled:
Actually, failure in general is a problem, especially with spawnpoint scans. Spend too long trying to scan one area and you'll delay the 5 after it... Hopefully there'd be slack time to make this up, but if the delay > slack time, eventually the whole thing will drift, and you'll be teleporting in the meantime.

chuyskywalker:
Or just skip spots that don't scan when they're supposed to and move on with life. Having one or two holes per cycle probably wouldn't be the end of the world, would it?

Xcelled:
No... but another possible idea just hit me. If a scan worker runs into trouble, it sends the coordinator a message and the coordinator relieves it of all work (except the current one it's stuck on) and grabs a new worker to handle the queue. Then, the stuck worker can succeed or die, and either way it goes into the pool of backups and the worst that happens is a hole.

In fact, if there are no backup workers, the worker can be told to get on with life and abandon the current scan, as you said.

rbryson74:
Maybe split the scan area up into a certain level of s2 cells. Then assign each worker ( or small groups) to a specific cell. Any spawnpoint that comes up in that cell can only be assigned to one of those specific workers. 
If you have cells that have more spawn points you could have it assign more workers.

Xcelled:
@rbyrson74 that's a good idea, but you'd need a lot of workers (and thus accounts) and most would be sitting idle.

blindreaper:
How hard would it be to switch to a queue/list sortable by spawn time then distance? Is this possible in python? too expensive on cpu?

The account could grab the closest item or go to sleep (which allows his max travel distance to increase).

DiscoTim:
it seems to me it is more a bus scheduling type problem than a TSP, where workers are buses and spawn points are stops. are https://github.com/google/or-tools of any help? they have python libraries.

Xcelled:
@DiscoTim You're right, that's a better fit for the problem. Google tools even has a section on vehicular scheduling! Neat find, thanks!

Let's move this discussion back to #179, we're really getting away from the PR stuff.

G1K:
Interesting results with this PR.
Delays for this speed is not followed, a worker with a radius of 5 st
`
2016-08-15 22:58:23,988 [ search_worker_0][        search][    INFO] Search step 2 beginning (queue size is 15)
2016-08-15 22:58:23,989 [ search_worker_0][        search][ WARNING] Worker moving faster (939501.44 m/s) than maximum allowed (24.00 m/s). Slowing things down. You need more accounts or a longer scan delay (hint: use -sd option)
2016-08-15 22:58:29,648 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 50 pokestops, and 8 gyms
2016-08-15 22:58:39,050 [ search_worker_0][        search][    INFO] Search step 3 beginning (queue size is 14)`

justsomedudeonthenet:
This is still a good idea - especially when using spawnpoint scanning from some of the other PRs.

However, there is a more efficient way to do it!

Have search overseer create a separate queue for each worker, with a small limit of items (I choose ten).  Instead of adding to a global queue, calculate which worker is closest to the next point that doesn't have a full queue.

This prevents workers from trying to jump all over the map, so even if they need to slow down, they won't be slowed down as badly.

The limited queue means even if a worker dies, you lose at most 10 points.  I have my workers pause for 3 hours if they start getting map errors repeatedly, so their queue will stay full and they won't get assigned new items.

Xcelled:
Agree that there's better ways to do this, some of which have been discussed in this thread. I'm going to close this, then.

PR #684, opened 2016-08-15
New classified pokémon:

Omikakim:
New classified pokémon:

Very Common
Common
uncommon
Rare
Very Rare
Special
Epic
Legendary
unavailable

https://uploaddeimagens.com.br/images/000/667/007/original/Pokemons---Tabela-de-Raridade.jpg

Fontes:
http://www.gamepur.com/guide/23676-pokemon-go-rarity-chart-based-pokemons-location.html
http://www.tecmundo.com.br/pokemon-go/107459-pokemon-go-saiba-categorias-raridade-pokemons.htm

### DIFF
--- a/static/data/pokemon.json
+++ b/static/data/pokemon.json
@@ -1,7 +1,7 @@
 {
   "1": {
     "name": "Bulbasaur",
-    "rarity": "Uncommon",
+    "rarity": "Common",
     "types": [
       {
         "type": "Grass",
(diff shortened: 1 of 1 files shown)


sLoPPydrive:
However you classify Pokémon rarity, it will always be completely wrong for some people. For example I constantly see Pinsirs around my home but never ever encountered a Snorlax.

Chlodochar:
:-1: If there is a solid source for rarity then a PR like this is fine. However the current source seems just as reliable/unreliable as the previous one.

PR #686, opened 2016-08-15
Rarity from statistics

tim95030:
## Description

Added a feature to update rarities based on your collected statistics. As part of this change I also changed the ui to fetch the json from a new endpoint based on the backend. This should allow us to move the pokemon data to the database as well as the locales at some point using the files for initial population only.
## Motivation and Context

To improve the quality of the rarities as rarity is really based on location. Some things are more rare in certain areas then others.
## How Has This Been Tested?

Tested on MacOSX and Chrome.
## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -8,6 +8,8 @@ import gc
 import time
 import geopy
 from collections import namedtuple
+
+from flask import jsonify
 from peewee import SqliteDatabase, InsertQuery, \
     IntegerField, CharField, DoubleField, BooleanField, \
     DateTimeField, fn, DeleteQuery, CompositeKey, FloatField, SQL, TextField
@@ -20,7 +22,7 @@ from base64 import b64encode
 from cache import cache
 
 from . import config
-from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, i8ln
+from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, i8ln, get_pokemon_data
 from .transform import transform_from_wgs_to_gcj, get_new_coords
 from .customLog import printPokemon
 
@@ -678,13 +680,13 @@ class PokemonRarity(BaseModel):
         return None
 
     @staticmethod
-    def update_rarities(duration):
+    def update_rarities():
         num_pokemon = 151
         rarities = []
         rarity_groups = []
         rarities_query = {}
         start = 0
-        seen = Pokemon.get_seen(duration)
+        seen = Pokemon.get_seen(0)
         total_count = seen['total']
         for pokemon in seen['pokemon']:
                 spawn_rate = (pokemon['count'] / float(total_count))
@@ -1205,6 +1207,28 @@ def clean_db_loop(args):
             time.sleep(60)
         except Exception as e:
             log.exception('Exception in clean_db_loop: %s', e)
+
+
+def update_rarity_loop(args):
+    while True:
+        try:
+            timeout = args.rarity_update_interval * 60 * 60
+            PokemonRarity.update_rarities()
+            cache.invalidate(get_all_pokemon, 'all_pokemon')
+            log.info('Pokemon rarities update complete')
+            time.sleep(timeout)
+        except Exception as e:
+            log.exception('Exception in update_rarity_loop: %s', e)
+
+
+@cache.cache('all_pokemon')
+def get_all_pokemon():
+    pokemon = {}
+    for x in xrange(1, 151):
+        pokemon[x] = get_pokemon_data(x)
+        pokemon[x]['rarity'] = PokemonRarity.get_rarity(x) or get_pokemon_rarity(x)
+
+    return jsonify(pokemon)
 
 
 def bulk_upsert(cls, data):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -706,6 +706,12 @@ class PokemonRarity(BaseModel):
         for i, r in enumerate(rarity_groups):
             for pokemon in r:
                 rarities_query[pokemon['pokemon_id']] = {'pokemon_id': pokemon['pokemon_id'], 'rarity': i}
+
+        # Assign ultra rare to any unseen pokemon
+        for pokemon_id in xrange(1, num_pokemon):
+            if not hasattr(rarities_query, pokemon_id):
+                rarities_query[pokemon_id] = {'pokemon_id': pokemon_id, 'rarity': len(PokemonRarity.rarity_types())}
+
         bulk_upsert(PokemonRarity, rarities_query)
 
 
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -679,6 +679,7 @@ class PokemonRarity(BaseModel):
 
     @staticmethod
     def update_rarities(duration):
+        num_pokemon = 151
         rarities = []
         rarity_groups = []
         rarities_query = {}
@@ -692,10 +693,13 @@ class PokemonRarity(BaseModel):
         rarities.sort(key=lambda x: x['spawn_rate'], reverse=True)
         # Split pokemon into groups by rarity using each rarity type percentage as the amount to split
         for rarity_type in PokemonRarity.rarity_types():
-            qty = int(rarity_type.percent * len(rarities) + 1)
+            qty = int(rarity_type.percent * num_pokemon + 1)
             end = start + qty - 1
+            if end > len(rarities):
+                rarity_groups.append(rarities[start:])
+                break
             rarity_groups.append(rarities[start:end])
-            start += qty
+            start += qty - 1
         # Assign a rarity index to each rarity group and setup rarities for DB insertion
         for i, r in enumerate(rarity_groups):
             for pokemon in r:
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -868,8 +868,16 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
                 # Set a value of 15 minutes because currently its unknown but larger than 15.
                 d_t = datetime.utcfromtimestamp((p['last_modified_timestamp_ms'] + 900000) / 1000.0)
 
-            printPokemon(p['pokemon_data']['pokemon_id'], p['latitude'],
-                         p['longitude'], d_t)
+            if args.display_in_console:
+                pokemon_id = p['pokemon_data']['pokemon_id']
+                pokemon_data = {
+                    "id": pokemon_id,
+                    "name": get_pokemon_name(pokemon_id),
+                    "rarity": PokemonRarity.get_rarity(pokemon_id) or get_pokemon_rarity(pokemon_id)
+                }
+
+                printPokemon(pokemon_data, p['latitude'],
+                             p['longitude'], d_t)
 
             # Scan for IVs and moves.
             encounter_result = None
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -17,8 +17,7 @@ from playhouse.shortcuts import RetryOperationalError
 from playhouse.migrate import migrate, MySQLMigrator, SqliteMigrator
 from datetime import datetime, timedelta
 from base64 import b64encode
-from cachetools import TTLCache
-from cachetools import cached
+from cache import cache
 
 from . import config
 from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, i8ln
@@ -29,7 +28,6 @@ log = logging.getLogger(__name__)
 
 args = get_args()
 flaskDb = FlaskDB()
-cache = TTLCache(maxsize=100, ttl=60 * 5)
 
 db_schema_version = 10
 
@@ -190,7 +188,7 @@ class Pokemon(BaseModel):
         return pokemons
 
     @classmethod
-    @cached(cache)
+    @cache.cache('get_seen', expire=60 * 5)
     def get_seen(cls, timediff):
         if timediff:
             timediff = datetime.utcnow() - timediff
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -7,6 +7,7 @@ import sys
 import gc
 import time
 import geopy
+from collections import namedtuple
 from peewee import SqliteDatabase, InsertQuery, \
     IntegerField, CharField, DoubleField, BooleanField, \
     DateTimeField, fn, DeleteQuery, CompositeKey, FloatField, SQL, TextField
@@ -20,7 +21,7 @@ from cachetools import TTLCache
 from cachetools import cached
 
 from . import config
-from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args
+from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, i8ln
 from .transform import transform_from_wgs_to_gcj, get_new_coords
 from .customLog import printPokemon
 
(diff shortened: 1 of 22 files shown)


pogom/models.py
```diff
@@ -309,6 +309,56 @@ def get_recent(swLat, swLng, neLat, neLng):
...
+        for pokemon in seen['pokemon']:
+                spawn_rate = (pokemon['count'] / float(total_count))
+                rarities.append({'pokemon_id': pokemon['pokemon_id'], 'spawn_rate': spawn_rate})
+        rarities.sort(key=lambda x: x['spawn_rate'], reverse=True)
+        for rtype in PokemonRarity.rarity_types():
+            qty = int(rtype[1] * len(rarities) + 1)
+            log.debug("qty: %s", qty)
+            rarity_groups.append(rarities[current:current + qty - 1])
+            log.debug("start: %s   end: %s", current, current + qty - 1)
+            current += qty
```
Chlodochar: I honestly don't understand exactly what is going on here. Would it be possible to make this more easily understandable?
It might already help to make `PokemonRarity.rarity_types()` a namedtuple (so it can be referred to as `name` and `frequency` instead of `[0]` and `[1]`).

pogom/models.py
```diff
@@ -309,6 +309,56 @@ def get_recent(swLat, swLng, neLat, neLng):
...
+
+    @staticmethod
+    def rarity_types():
+
+        return [("Very Common", 0.10),
+                ("Common", 0.18),
+                ("Uncommon", 0.22),
+                ("Rare", 0.22),
+                ("Very Rare", 0.18),
+                ("Ultra Rare", 0.10)]
```
Chlodochar: What are these numbers based on? Would it be bad to just have them all be 0.17 (1/6)?

templates/statistics.html
```diff
@@ -42,6 +42,10 @@
         <header id="header">
             <a href="#nav">Options</a>
             <h1><a href="..">Pokémon Go Map</a></h1>
+            <div id="loading_small" class="loading" style="float: right;display: none;">
+                <img src="{{ url_for('static', filename='images/loading.gif').lstrip('/') }}" alt="Loading"/>
+            </div>
+            <div id="update_rarities" class="raritiesButton" style="float: right;"><span class="label">Update Rarities</span></div>
```
Chlodochar: At the moment it doesn't look like a button. Maybe it's a good idea to just place a button in the sidebar with 'update rarities' instead of the text in the topbar.

Chlodochar:
I like the general approach of this PR. It avoids using rarity definitions from random sources. Code looks mostly good except for the few comments.

templates/statistics.html
tim95030: Agreed. I will clean up the UI for this. I kinda forgot about it once I got going :)

pogom/models.py
tim95030: I did an even split to start with and it seemed not great too many things were very common. My thinking behind this was to allow future work to customize this as they would be very easy to make the percents changeable.

pogom/models.py
tim95030: I apparently forgot to remove my debug messages. But basically the idea here is to assign a rarity number 1-6 to each pokemon in the sorted list based on the percentage required of those pokemon as defined in the tuple. I can probably switch to named tuple without issue for clarity. I worked through several permutations of this area and this was the one that I felt gave the best results. If anyone has a better/cleaner way of assigning the rarity values to the list of pokemon I am very open to ideas.

Here is basically what I am doing here:

   1. Get the spawn rates (percent that pokemon showed up out of the grand total of pokemon found)
   2. Assign this value to each pokemon that was in the list of found pokemon then sort this list by that value from high to low (common to rare)
   3. Iterate through this list and assign an int from 1-6 describing the rarity (1 being more common and 6 being more rare)
   4. Then this is passed to the database. I have not yet decided the best option regarding removing previous entries, or wether all non found pokemon should be given a different rarity tag of legendary/unknown/etc. but for now it replaces the adds new ones and replaces any pre-existing values. If the pokemon isn't in the table at all it reverts to the pokemon.json values. (We may want to update the known legendaries to legendary since it's clear there are a few pokemon that are not findable currently, and possibly something else for the continent based ones like Tauros, Farfetched, etc.)

Chlodochar: It makes sense now. The rarity is determined whether a seen-count is between `current` and `current + qty`. Maybe name it that way `start` and `end`, `from` and `to` or `min` and `max`.

pogom/models.py
Chlodochar: Fair enough.

aldarund:
Nice approach, way better than current hardcoded definitions .

Zerotask:
They should implement it into the dev branch.

Slivo-fr:
Should be an option for people that aren't scanning an area big enough to make pertinent stats

tim95030:
@Slivo-fr What do you mean an option? It is a feature where you have to manually click a button to update the stats. By default it changes nothing.

Slivo-fr:
My bad, forgot what I said

tim95030:
so can we merge this?

rehand:
+1 for merging this PR

BaIthamel:
Serious CPU usage issues reported from this. Please double check your work.

tschroeder-zendesk:
What are you talking about. My query does 2 million records so fast I can't get the leading icon to show.

tschroeder-zendesk:
Plus who? Nobody has reported this to me?

Infernoman:
on my dedicated server with 8/16 cores, it was running all of the threads at 100% load. I am running 10 different webservers with a load of 1k concurrent users. the server only runs the database and the webservers

Infernoman:
after removing your commit the cpu usage went back to normal on my box.

tschroeder-zendesk:
I will look into the rarity display and see if I can add a caching layer.
Everything else about this feature is on demand not constant, but I did
change the rarity lookup function so it's probably during that lookup.
Thank you for the info.

tim95030:
I figured out the issue. It's the fact that I had to move the get all pokemon call to the backend. So now each request is running a loop to figure out the rarity of all 151 pokemon. As such I am probably going to need to load the initial json into a new table of pokemon data and then update the rarities there. As such I am going to need some time to clean this up. In the meantime if anyone has a better idea on this please let me know the issue area is the function on line 241 of app.py which. I suppose one other potential possibility is adding caching to this request so at least it doesn't get hit as often which should alleviate some of the strain. But as such you can do that in nginx or apache which if you are running with 1000 clients you probably should since this application was originally designed as a one user app. Anyways I will see what I can do about this.

tim95030:
@Infernoman Please re-test I have added a caching layer to this request. I watched on my side and the process will spike to about 0.2 seconds load time for the pokemon_data endpoint upon cache clear, but only on one request and then it lowers back down to 0.002 or so. This seems to be where the issue was. I now only process the updated rarities once per time the rarities are updated and only ever on one request. As such the CPU load you saw should no longer exist.

tim95030:
This is ready to be merged

tim95030:
I removed cachetools and used beaker because it is actually threadsafe.

This is rebased yet again. @FrostTheFox @chuyskywalker please merge this. This is rebased yet again. @FrostTheFox @chuyskywalker please merge this.

aldarund:
Is there any perm for who can update the stat? Or anyone can do it anytime? If so not really good

tim95030:
...multi user is so not the original intent of this repo.. sigh... but fine I will see what I can do.

aldarund:
well certainly it is, there a lot of users of this map running this map for their local cities. 
Look how status page implemented. It have a pwd protection. why not just do the same and use same password?

tim95030:
I see it and I'm working on a fix, but the Original intent of the repo was for one user hence the features around moving the search etc. But again no worries I will add it.

tim95030:
I personally think we should just add an admin page and have things like this and status and anything else that is truly the runner based if we are going to move towards multi user, but I don't want to add all that in this PR, so for now I am just going to hack in a password requirement for this.

tim95030:
I have now added password protection to the update.

tim95030:
@FrostTheFox it seems you are the only one merging things lately. As such merge plz?

FrostTheFox:
I'd like if password protection *was* removed again. I'm trying to stop that kind of trend, the status page password protection was added without my comment, and the original stance of running locally still applies. (on another note, sorry you had to code that and I'm here asking for you to remove it :frowning:  )

tim95030:
@FrostTheFox removed password per your request

GamingMaster2000:
I've been using this for about a week or so now. It seems to be working well for me.

DiscoTim:
i really like this idea of this pr, providing its optional and people can still continue using the json if they choose.

tim95030:
There is a button on the stats page that will add the entries to the db tables. It will always resort to the file for entries not found. You can also wipe out the table if you ever want them reset. Later on I can add some more admin like stuff around this, but for now it's a good start.

DiscoTim:
for me that's great, i know some go through and meticulously edit their rarites though. does the button on the stats page mean that if i am sharing with friends they can mess with the rarities in my DB?

tim95030:
Yes currently any external user could do this which is why I had added the password which I was told to remove. If people think it's okay I can add a flag to disable the endpoint so that way at least you can keep others from submitting it the experience won't be amazing as it will simply fail to update, but I think this would be a reasonable solution. Big question is default on or off?

tim95030:
There is now the ability to disable this feature. I defaulted it to enabled. Ready for review.

primaxius:
👎 -1 While I agree that pokemon rarity changes per region, you can simply edit the static/data/pokemon.json to make changes for your region. Also who decides what % is rare...apparently you do and may not be what the general population deams a rare pokemon.

primaxius:
See stats screenshot, I clicked on the update button and then checked the DB... Not all get inserted/updated in the pokemonrarity table.
statsscreen
dbrarity

tim95030:
You guys wait about 2 months and then ask for tons of things. Wtf. I chose the percentages as a base with the intent to add additional features around this fur customization. I implemented the entire rarities in the beginning and nobody cares who gets to choose but thanks for that helpful feedback. You don't like it don't use it shit. Talk about feature creep you ask to implement thing after thing after thing. Just merge this already and then I can actually work on improving it further with customizable rarities and things of that nature. The reason it's on the stats page is to allow you to define a time frame to update the rarities. The only thing I agree with is maybe defaulting it off. This use to be a fun project that I was helping out with lots of new features, but now it takes months to implement anything. And finally that isn't a bug or updated all the Pokemon on your screen that's what the feature does. If you have never seen something it uses the die instead of calculating a rarity for things. This way you can customize the rarity. I do agree adding customization around the percents and adding management makes sense, but only after this is merged let others help out. That's the whole freaking point of open source not this keep everything open for months crap making us since complicated merges when your code is liked, and actually kept up to date like I have.
I an making one last commit to set the default of this isn't merged I'm done helping you guys. I was one of the most active people in the early days, and now thanks to a lack of support and understanding you guys get boring done. I have seen several great features just wither away because you ignore them to favor dumb stuff that moves completely away from the original intent of this project.

tim95030:
Ok. I have realized it was dropping pokemon, so I apologize about that. I switched it so that it bases the rarity off the full set of pokemon so if you have only ever seen 5 pokemon they get classified as very common and the rest use the json file. This way you don't end up with a rare when you run this with only 5 seen. It assumes if you have seen it is more common. I also defaulted it to off. Regarding the other requests can we please leave those as future updates? The feature is off it is almost a no-op now with the exception of the button showing up on the stats page which won't do anything since it's off.

GamingMaster2000:
"-Remove the update button and make it a command line flag that updates it on startup (-ur update rarity)"

Why force a restart of the webserver to update rarities? An update rarity button is a much better way of handling this IMO.

DiscoTim:
Maybe a timer is an option then, that does the update every 12 hours or so? I agree there is no need for a button on the front end and it will just be problematic on shared maps. Once a reasonable size data set is accumulated its not something that will need frequent updating. The only way to view the current rarity levels is to look in the DB, so i can also understand the request for an alternative way to view them. Its a great idea for rarities, and really like to see this get merged.

tschroeder-zendesk:
The multi-user (not officially supported) scenario is fixed via the disabling of the feature. I will look into adding additional viewing and other things in an admin type page and move the button there in the future as quite honestly that's where we need to go for the front end especially since people keep using it as multi user. But for now the button is only on the stats page and is disabled by default don't enable it if you use multi user. You will just have to wait on the rarities in that case. Or run it once click the button then turn it back off. I really am done maintaining this branch it has literally been over 2 months and not that much has changed in it since the beginning.

tim95030:
The multi-user (not officially supported) scenario is fixed via the disabling of the feature. I will look into adding additional viewing and other things in an admin type page and move the button there in the future as quite honestly that's where we need to go for the front end especially since people keep using it as multi user. But for now the button is only on the stats page and is disabled by default don't enable it if you use multi user. You will just have to wait on the rarities in that case. Or run it once click the button then turn it back off. I really am done maintaining this branch it has literally been over 2 months and not that much has changed in it since the beginning.

DiscoTim:
Well, most of that i don't know, have to be worked out with reviewers/maintainers, i understand the frustration though. i personally don't understand the need for the button, i think its something that should be automated anyways. I don't think its so much multi-user maps that aren't supported, but public maps that are not. even still, if the button is disabled why is it still visible, and not work in the way of -nsc for example? and being able to see what your rarities are is a pretty obvious basic immediate need.

DiscoTim:
@tim95030 this seems very reasonable to me :)

tim95030:
Not sure why a button that said disabled when clicked was such a big deal, but whatever. It's now a thread with loop defaulted to 12 hours (customizable though).

tim95030:
@primaxius please approve the changes

primaxius:
Ok I tested this last night and wasnt able to comment. But here goes, I don't believe that if you enable this that it should use the pokemon.json at all. It should include all pokemons in the calculations. Everything not seen technically should end up in ultra rare and be shown as such. I think with this we're trying to get away from static assignments so why would we use them if their not seen yet. Let the update mechanism update them once seen if it need it.

DiscoTim:
@tim95030 i agree with what @primaxius is saying, because you'll end up in a situation where something is ultra rare to you because you have never seen it before, and it will look it up in the json and it could come back as only rare or even common, which is just wrong (for your location). 

I think its expected that someone has a reasonable set of data, (ie. few days worth), before they switch to this, so that the rarities returned are meaningful for them, and in that case anything not seen would be ultra rare (at least at that point in time). that aside, i really do like this pr, and way of handling rarities :)

DiscoTim:
@tim95030 just bumping this, is there any possibility you can do that? would love to start using this.

tim95030:
The reason for the delay is it turns out I missed one reference to rarity in the CustomLog method and it currently is imported in Model... which means I cannot import any part of Model in it so I am trying to figure out a good non circular fix.

primaxius:
@tim95030 I'm gonna test this tonight. Anything I need to know that you can think of?

tim95030:
Not that I know of. I have been using it with no issues. You can set the
frequency in the config. Beyond that it's pretty straight forward. It does
run at the beginning of the launch.

tim95030:
I think everything requested is done now. Let me know if that isn't the case.

tim95030:
@primaxius Are the changes satisfactory?

BrownSlaughter:
I have been looking for something like this, can it be updated/merged?

sebastienvercammen:

PR #693, opened 2016-08-15
[Fix] issue #622, statistics 2x Uncaught TypeError

DiscoTim:
## Description

Fixes two issues with statistics, One on page load, and one when clicking "All Locations" on a specific pokemon. 

## Motivation and Context

https://github.com/PokemonGoMap/PokemonGo-Map/issues/622#issuecomment-239644697

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #694, opened 2016-08-15
fixed wrong redirection in Mozilla Firefox

rehand:
Fixed wrong redirection in Mozilla Firefox

## Description
In Mozilla Firefox on a Desktop PC, the "return false;" caused the browser to display a blank website with "false" on it instead of showing the map overlay.

## How Has This Been Tested?
Tested it locally with Mozilla Firefox 48.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/statistics.js
+++ b/static/js/statistics.js
@@ -89,7 +89,8 @@ function addElement (pokemonId, name) {
   }).appendTo('#seen_' + pokemonId + '_details')
 
   jQuery('<a/>', {
-    href: 'javascript:showOverlay(' + pokemonId + ');',
+    href: '#',
+    onclick: 'showOverlay(' + pokemonId + ')',
     text: 'All Locations'
   }).appendTo('#seen_' + pokemonId + '_details')
 }


DiscoTim:
yes there are two issues here, i pr'd a fix as well: https://github.com/PokemonGoMap/PokemonGo-Map/pull/693

Chlodochar:
So, which one should be used here?

rehand:
In my opinion my solution is more clean, so I'd prefer my PR.

Chlodochar:
Tested both PRs. Sorry, the other PR fixed another long-time issue. I'll have to close this one.

DiscoTim:
whichever one you choose :)

rehand, i would be pressed to find a resource that specifies href='#' is a better solution than href='javascript:void(0);'

That aside, this solution does not address that #nav-accordion is not present in the stats page, while the map.js references it. soo.. 'Uncaught TypeError'. page won't load on FF with it like that.

Chlodochar:
Indeed, closed this PR for #693.

rehand:
DiscoTim you're right, there is no real resource which states, which solution is better. I personally prefer the '#' because then you don't have to add a "javascript:" to your code, which could be more error-prone, but on the other hand my solutions modifies the URL if the applied function in onclick doesn't return 'false'. So both solutions are OK, thanks also for your solution!

PR #700, opened 2016-08-15
Use JSON or just the DB for spawnpoint scanning (for ease of migration for beehive users)

justinkwaugh:
## Description
This is basically https://github.com/PokemonGoMap/PokemonGo-Map/pull/585 but with the ability to not use a json file and instead just grab the appropriate spawn points for the given location / steps.  This allows for a beehive user to just add the -ss param and it will do spawn point scanning, but with no concern for teleport bans.

## Motivation and Context
Teleport bans are an issue with very large spawn point scanning areas.  Also migration from beehive users is not easy.

## How Has This Been Tested?
I've tried it locally with an existing beehive of a local city.  The changes do not affect the normal scanning behavior if the -ss flag is not provided.

## Types of changes

## Checklist:


justsomedudeonthenet:
So this would still require multiple workers, scanning in a beehive pattern, each worker only handling the spawnpoints inside their beehive?

Sounds like a good idea, some of my workers are starting to get banned for teleporting.  But will points that overlap get scanned twice?

justinkwaugh:
@justsomedudeonthenet The only duplicate scans will be points that overlap between beehive sections, not within one. But given how infrequent scans are done, it shouldn't really matter much

justinkwaugh:
Ok, everything is working well now

lota7:
We need something like this more than ever now with the bans that's going on.

Does this also account for accounts being banned? As in, will it assign another worker to it or will the point be skipped entirely?

justinkwaugh:
It doesn't do anything special with bans, but the likelihood is less now, given that the teleporting should not be an issue and with the spawnpoint scanning it is not hitting Niantic's servers very frequently (relatively)

lota7:
@justinkwaugh Thanks for making this Beehive->Spawnpoint transition easy.
One question though, I get this when starting:

```
2016-08-16 00:14:33,056 [ss_search_worker_56.139289_12.940653_0][      auth_ptc][    INFO] PTC User Login successful.
2016-08-16 00:14:33,062 [ss_search_worker_56.139289_12.940653_0][      auth_ptc][    INFO] Request PTC Access Token...
2016-08-16 00:14:33,232 [ss_search_worker_56.139289_12.940653_0][      auth_ptc][    INFO] PTC Access Token successfully retrieved.
2016-08-16 00:14:34,112 [ss_search_worker_56.139289_12.940653_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-16 00:14:34,119 [ss_search_worker_56.139289_12.940653_0][        search][    INFO] Search step 1 completed
```

I have 2 workers running on a small 5step area.
Only the first one logs on initially, but it calls for a scan instantly resulting in 0,0,0. Is this just to initialize the scan process, or was that an intended scan to find a pokemon? If that is so, then every first pokemon in queue will result in 0,0,0 since it logs in and instantly scans it, thus being throttled by niantic.

justinkwaugh:
@lota7 good question, and I'm not certain.  I don't think I am getting that issue, but I will check when I have a chance.  I am also not using more than one worker per 5 step area, so there may be a difference there.

lota7:
@justinkwaugh ah, so 1 is enough for a 5step? That's amazing!
If 1 is enough, then what happens in this scenario.

2 spawns...
• First at 02:20 [mm:ss]
• Second at 02:25  [mm:ss]

Since it needs to pass 10seconds per request, will this automatically queue the second spawn to scan at 02:30 or will it force 02:25 and thus miss the spawn?

Thanks again for this! Tested it on an isolated part of my map where i've documented the spawn times and it works wonders so faron my small runtime!

justinkwaugh:
spawns get queued at the appropriate time, and then the worker will pull them off and search them with a minimum 10 second delay between.  You can't really miss spawns that way, unless you had a ridiculous amount queued.  In some of my 5 step areas I have nearly 200 spawn points and it doesn't miss anything, though some might be with a 13:50 disappear time instead of 14:00 or whatever.

lota7:
That's great.
Guess the only thing left I can think of is ban-handling.
In case they do get banned it auto switchs to the other accs in the pool, so for example if you have 2 accounts and 1 get banned, it wont lose half spawns till you look at the console and notices it.
Any possibility to add that feature?

Otherwise, very good work in making the transiton from beehive to ss scan so easy and without fiddling with json files. I have plenty areas, so having to do the alternate way of having a spawn.json for every rectangle area would be soooo time consuming.

Wonderful work!

lota7:
@justinkwaugh OK, I verified the initial 0,0,0 when starting bug.
I havent tested with 1 worker, but this would be applicable for a larger area requiring 2 workers.

Steps to reproduce:
• 2 workers
• Run scan with -ss. 
• In map you will see it scans an area and in console it returns 0,0,0.
• Start up another scan without -ss on that spawnspot. 
• It will find a pokemon.

Solution:
• Put a timed delay before scanning after having logged in.

Dagothur:
@justinkwaugh can you change the query to use a GROUP BY filter? As it stands the DB grows indefinitely so if you don't plan on purging your DB periodically you are going to end up with a lot of data being sent that could be filtered before it leaves the DB.

Based on this post: 

https://www.reddit.com/r/pokemongodev/comments/4wxteh/i_implemented_tbterras_spawntracker_into/d6b94go 

it looks like using "GROUP BY spawnpoint_id" is sufficient

justinkwaugh:
I'm closing this pull request now that TBTerra has implemented it in his PR

lota7:
@justinkwaugh I am noticing the DB growth . Gone up 20mb in 1 day, when it has been steady the previous days at 60mb.
Could it be related to what @Dagothur  mentioned? Should we hadd a group by filter? Where?

PR #717, opened 2016-08-15
Add append action to webhooks

invisiblek:


Chlodochar:
`default=False` should be `default=[]`

invisiblek:
looks like everything else was designed for this to be used this way, which is nice

Chlodochar:
`nargs='*',` should be removed to make this work. Otherwise python will make a list of lists for `args.webhooks`. However, I do not know what the behavior was before this PR. Maybe `--webhook http://first/ http://second/ http://third/` was possible?

rollin:
> Maybe --webhook http://first/ http://second/ http://third/ was possible?

It's the only way that works now actually.

Chlodochar:
So, deprecate the previous method and use this pr instead? Personally I think `args=*` shouldn't be used at all, as it would be confusing when multiple options start accepting these.

thatsweater:
I tried this out and it doesn't seem to work, PokeAlarm doesn't update as an example either with one or two webhooks.

Chlodochar:
Did you try removing nargs?

thatsweater:
Removing nargs makes a single webhook run, with normal syntax, if i try to run both, either in the way rollin suggested, or separated by commas, or as ["http://first","http://second"] under webhook: in config, program runs but neither webhook is detected.

rollin:
I meant it works this way `-wh wh1 wh2` now, without this PR. Also it works only from the command line, I couldn't set up two webhooks from the config file at all.

Albert62:
@rollin how where you able to do it?
I just tried this, but It doesn't work on command line

justsomedudeonthenet:
Closing because it has merge conflicts and hasn't been updated in 22 days.  If this is still an issue feel free to resubmit an updated PR.

PR #719, opened 2016-08-15
URL escaping of the ampersand (&)

Studio-XP:
## Description
Ampersands (&) should be escaped in urls

## Types of changes

## Checklist:

### DIFF
--- a/templates/statistics.html
+++ b/templates/statistics.html
@@ -70,7 +70,7 @@
             <div class="content">
                 <div id="times_list" style="display: none;"></div>
                 <div id="location_map">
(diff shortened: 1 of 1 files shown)


Chlodochar:
Why is the `&` incorrect? `libraries` is a second parameter right?

Studio-XP:
Yes, it's the second parameter, but when referencing a url from within a page (or anything for that matter), it has to be escaped. Core of the reasoning is that the ampersand is the beginning to html entities, such as &lt; or &quot;

Chlodochar:
Bever realised this was needed, but you're right about it being correct. :+1: on merging this.

PR #723, opened 2016-08-15
Add notifications when a Pokemon is really near

Amoki:
## Description
Add a toggle to enable 'Notify on nearby Pokemon', which notify users when a Pokemon is less than 70 meters away (one scan radius)

## Motivation and Context
I want to be able to have a notification for all nearby Pokemon (no need to move to catch it) without being spammed by a Pidgey 500 meters away.

## How Has This Been Tested?

Tested locally, I tried to play (enable, disable, customize) with all notification settings (including mine) and I have never seen a weird behavior.
I changed my location from the web interface and it works too

image
image

## Types of changes

## Checklist:

## Questions:
I am not an native english speaker, is my phrasing incorrect or can be improved ?

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1187,7 +1187,7 @@ function setupPokemonMarker (item, skipNotification, isBounceDisabled) {
     }
   }
 
-  if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
+  if (!notificationTitle && (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1)) {
     if (!skipNotification) {
       notificationTitle = 'A wild ' + item['pokemon_name'] + ' appeared!'
       notificationMessage = 'Click to load map'
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1173,29 +1173,34 @@ function setupPokemonMarker (item, skipNotification, isBounceDisabled) {
     disableAutoPan: true
   })
 
+  var notificationTitle
+  var notificationMessage
+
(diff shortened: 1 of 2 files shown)


static/js/map.js
```diff
@@ -1166,17 +1173,35 @@ function setupPokemonMarker (item, skipNotification, isBounceDisabled) {
...
+      new google.maps.LatLng(item['latitude'], item['longitude']),
+      locationMarker.getPosition()
+    )
+    if (distance <= scanRadius) {
+      notificationTitle = 'A wild ' + item['pokemon_name'] + ' is right here!'
+      notificationMessage = 'Only ' + Math.round(distance) + ' meters away'
+    }
+  }
+
   if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
```
tim95030: couldn't you add a shortcut here saying if (!notificationTitle && notifiedPokemon....  so that you don't overwrite the nearby message for things that match both filters? Or was that your intent? I feel like all nearby is good, but no need to call the normal notification if a watched pokemon shows up next to you right?

Amoki: Yeah, you're right!
It's fixed now

tim95030:
@chuyskywalker, @BaIthamel, or @invisiblek  Can we merge this?

Darkknightz:
I wonder since radius is 35m, diameter is 70m, when checking if distance < 'scanradius'  (Set to 70) is actually giving the intended effect. Also the encounter is at diameter is 40m, so radius is 20m? (according to PR#1025)
You need to be within 20m for the Pokémon to show in-game.

I have not tested this, but reading the numbers if the intended effect was truly: "I want to be able to have a notification for all nearby Pokemon (no need to move to catch it)" 

Then radius should be 20m, or if radius is halved somewhere in code it should be 40m at the most.

PR #736, opened 2016-08-16
Use our own fork of pgoapi

invisiblek:
* This provides us more control over the api for our project.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ PyMySQL==0.7.5
 flask-cors==2.1.2
 flask-compress==1.3.0
 LatLon==1.0.1
-git+https://github.com/keyphact/pgoapi.git@60096ccfea0a97e5358be1d32283ebbfb853b00a#egg=pgoapi
+git+https://github.com/PokemonGoMap/pgoapi.git#egg=pgoapi
 xxhash
 sphinx==1.4.5
 sphinx-autobuild==0.6.0


invisiblek:
The idea here being we can more easily get updates into the api if we have our own fork. We should continue to submit patches upstream, in fact I'd even recommend we make that a requirement for any pull requests we take here.

chuyskywalker:
I don't have a problem with this, but if keyphact gets back up to date, let's reconverge on that later.

Zoruk:
I just made this little change before compiling the proto files
```
diff --git a/src/POGOProtos/Networking/Envelopes/Signature.proto b/src/POGOProtos/Networking/Envelopes/Signature.proto
index a11e004..c83427f 100644
--- a/src/POGOProtos/Networking/Envelopes/Signature.proto
+++ b/src/POGOProtos/Networking/Envelopes/Signature.proto
@@ -93,7 +93,7 @@ message Signature {
     uint64 location_hash2 = 20; // Location2 hashed (unsigned) - xxHash32
     bytes session_hash = 22; // 16 bytes, unique per session
     uint64 timestamp = 23; // epoch timestamp in ms
-    repeated int64 request_hash = 24; // hashes of each request message in a hashArray signed based on the auth_token or auth_info - xxhash64
+    repeated uint64 request_hash = 24; // hashes of each request message in a hashArray signed based on the auth_token or auth_info - xxhash64
     int64 unknown25 = 25; // for 0.33 its static -8537042734809897855 or 0x898654dd2753a481, generated via xxHash64("\"b8fa9757195897aae92c53dbcf8a60fb3d86d745\"".ToByteArray(), 0x88533787)
 
 }
```

PR #737, opened 2016-08-16
Added possibility to change spawnpoints only scan mode via frontend and speed limit

Mr-Groch:
## Description

Added config['SPAWNPOINTS_ONLY'] controled by frontend - replaced args.spawnpoints_only in search threads. Config vraiable is assigned on app init with args variable, so initial value depends on args variable. UI switch state is valid across all sessions.

EDIT:
Added also speed limit, based on mixed work from @AngryRabbit1 and @Shibby7986 from PR #997, as it's really needed for this spawnpoint only mode to avoid bans. Also made visibility of spawnpoints only mode switch depend on speed-limit arg.

## Motivation and Context

I want to be able to quick scan new area with classic scan mode to search spawnpoints, and then switch to spawnpoints only mode without server restart.

## How Has This Been Tested?

Changing scan mode without changing location, with lcoation change, after location change and before location change.

spawnpoints_only

## Types of changes

## Checklist:


rrbailey89:
This is a good idea

Mr-Groch:
I suggest to merge with #37 - with those 2 controls via frontend we are able to set all important scan options without need of server restart. This is great for individual map usage for those who travel a lot :)

CtrlAltDefeat94:
Can this option be disabled in config?

Mr-Groch:
Yes,  by -fl (fixed location) argument. Is there need for new arg only for this?

CtrlAltDefeat94:
I had forgotten about that one... I think -fl suffices

SlasherKG:
I tried to make these changes to the latest version and while I see in the screen that it 'toggles' the spawnscan setting, it doesn't seem to actually switch from normal search to spawns only.
Am I missing something, or is it broken against the latest commits?

Mr-Groch:
I have merged latest develop branch into my working branch and checked - it is still working as intended. Maybe you are thinking that switch enables - ss mode? It doesn't. This activates - - spawnpoints-only mode introduced in #633. Search workers are restarted like with location change,  so search starts over from marker position

SlasherKG:
Oh, so this will just skip empty cells but continue the regular spiral pattern?

Mr-Groch:
Yes. I think that it will be useless if this would trigger -ss mode. That mode is intended to be fixed location without UI control - I think that enabling -ss arg should disable all location tab. This is for users who often change scan locations, I use that with PR #37 and it works great for me. I;m able to quickly switch scan locations, scan step limit and next to scan only cells with spawnpoints via mobile, without need to connect torough SSH, change args and restart app.

I have also made private branch https://github.com/Mr-Groch/PokemonGo-Map/tree/step_and_search_control with merged this and PR #37.

shiznewski:
Great addition. However i disagree with Mr-Groch in that a triggering -ss mode would be useless.
SS mode is something i use at my house, work, and gf's house. But in order to do it i have to have 3 copies running at all times. If i was able to hit a toggle, that generates the spawns.json for the steps around the marker, and then begins the SS search that would be really helpful and cut down a lot of scans. Also it would help out when i visit a large park. After going a couple times i have all the spawn points and times. If i could toggle -SS to run it would let me know where all them little charmhanders and hiding instantly.

Imoq:
Referenced this on this feature request, although personally I prefer -ss over -spanwpoints-only by far: http://feathub.com/PokemonGoMap/PokemonGo-Map/+83

Mr-Groch:
And what about this PR? Can I get some feedback?

Imoq:
Hey, @Mr-Groch , does it still work with latest commit regarding the new search pattern? I started using your branch (this + possibility to increase/decrease step limit = wow!) a week ago, but promptly stopped after I noticed some more account bans due to "teleporting". I still like spawn location scan way better, was just wondering if this is working with the new pattern.

I hope this PR gets included soon!

Mr-Groch:
@Imoq I'm using this with https://github.com/PokemonGoMap/PokemonGo-Map/pull/997 and --speed-limit 300 --max-speed-limit-sleep 6000 and no bans for a 4 days till now. When I will have some free time I will look, if can switch to ss mode, but this may be tricky to do in runtime, spawnpoint_only mode with speed limit is working for me really good.

Mr-Groch:
Added also speed limit, based on mixed work from @AngryRabbit1 and @Shibby7986 from PR #997, as is really needed for this spawnpoint only mode to avoid bans. Also made visibility of spawnpoints only mode switch depend on speed-limit arg.

averylarry:
I think #995 would really go along with these changes.

Mr-Groch:
With new commits to develop branch, Location & Search Settings needs ordering. I have put all search settings on top, and user location on bottom, now it looks like this:

loc_search_reorganize

NOTE: This pic contains Step limit change input field - it is not part of this PR, screenshot is from my working branch with PR #37 merged.

averylarry:
What happens to the speed limit if time_elapsed is 0?  I'm not good enough with the code to figure that out.  The code in #997 would divide-by-zero.  (This can happen with scan delay 0, which I have been using in conjunction with speed limit because you don't need an arbitrary scan delay when you're dynamically delaying based on speed limit.)
I haven't used your code -- I'm still using the code from 997.

Mr-Groch:
@averylarry Nothing bad, look at last lines of pogom/transform.py:

```python
time_elapsed = end_time - start_time
if speed_limit > 0 and time_elapsed > 0:
  # blabla
else:
  return 0

**Comment by averylarry** [comment]
@Mr-Groch 
right, but I don't know what that means.  If it is returning a speed of 0, then that is wrong.  It should return a speed of a really really big number.  If it is returning a sleep time of 0 seconds, then that is also wrong.  It should return a sleep time of whatever is necessary to get the speed limit because 0 (rounded) seconds have elapsed.

**Comment by Mr-Groch** [comment]
@averylarry It is returning the time worker should sleep to get bellow speed limit. If returns 0, then worker will not sleep addidtional time. This is OK, because if elapsed_time is 0 then there is no way to calculate speed - this situation is only possible on first run, where there was no old location, meaning no move.

**Comment by averylarry** [comment]
@Mr-Groch 
I wouldn't think that is correct.  If there is, in fact, 0 time elapsed and it is not the first run, then you will return 0 which will equate to instant teleport no matter the distance.  Instead, I think it should set the speed to some very very big number like 9999999.9 (approximating instant teleport == infinite speed) and then let the rest of your code determine any speed limit delay needed.

But I probably should shut up unless I can demonstrate that your code (which isn't the same as 997) actually exhibits this problem.

**Comment by averylarry** [comment]
FYI I tried changing the code in transform.py as below.  It works the way I expect even with a time difference of 0 milliseconds.  Further, I think the 0 millisecond thing is most likely due to my system not having sufficient time precision.  I have even seen documentation of the time function list the system's precision capabilities as a limiting factor.
```

def get_speed_sleep(speed_limit, start_location, end_location, start_time, end_time):
    time_elapsed = end_time - start_time

    if speed_limit > 0:
        speed_limit = speed_limit * 1000.0 / 3600.0  # convert to mps to avoid divide by zero errors
        distance = geopy.distance.distance(start_location, end_location).meters
        if time_elapsed > 0:
           speed = distance / (time_elapsed / 1000.0)
        else:
           speed = 9999999.9
    else:
        return 0

    if speed > speed_limit:
        speed_sleep = int(math.ceil(((1000.0 * distance / speed_limit) - time_elapsed) / 1000.0))
        return speed_sleep

    return 0

```

**Comment by Mr-Groch** [comment]
@averylarry using const speed of 9999999.9 means long sleep time even for small distance, but I get the idea. I will implement this for such people that want use scan dalay 0. I'm lazy and don't want do the math - how much speed sleep do you get for 9999999.9 speed?

**Comment by averylarry** [comment]
That's not what happens. It sets the current speed if you don't pause to 9999999.9 mps.  The next few lines then says 9999999.9 mps is too fast (it's greater than speed_limit) and calculates the delay needed to stay under the speed limit.

**Comment by Mr-Groch** [comment]
@averylarry Yes, I know, I want to know average speed sleep for you when you hit that const sleep

**Comment by averylarry** [comment]
Oh -- well I don't have a good answer for that.  It happened very rarely with -sd 0 -sl 65 -msls 120.  The issue only comes up when the queue gets behind -- so it has a location to scan as soon as it's done with the current location.
I used -sd 0 -sl 1 - msls 120 to force the edge case.  In that scenario, I usually had a delay over 100 seconds, but clearly a speed of 1 would do so.
I would say the few times I saw a time difference of 0 milliseconds when using my normal settings the delay was typically 5-15 seconds.  I don't scan a large area so it doesn't take terribly much scan delay to limit the speed to 65 kph.

File: pogom/transform.py [hunk 1]
```diff
@@ -56,104 +56,20 @@ def get_new_coords(init_loc, distance, bearing):
...
+        if time_elapsed > 0:
+            speed = distance / time_elapsed
+        else:
+            speed = 9999999.9  # time_elapsed == 0 means instant teleport - speed is "near" infinite...
     else:
         return 0
 
     if speed > speed_limit:
-        speed_sleep = int(math.ceil(((1000.0 * distance / speed_limit) - time_elapsed) / 1000.0))
+        speed_sleep = int(math.ceil((distance / speed_limit) - time_elapsed))
```
averylarry: I think:
distance is in meters
speed_limit is in meters per second
distance/speed_limit will give seconds
time_elapsed is in milliseconds (right?)
If I'm right, try:
int(math.ceil(distance / speed_limit - time_elapsed/1000.0)

Mr-Groch: Before all times were converted to miliseconds, but no need to do that anymore - all times come now from now()

primaxius:
not everything needs to be on frontend

PR #741, opened 2016-08-16
Stop scanning for Pokemon < 7.5 minutes.

Infernoman:
## Description
 Currently if the queue gets too large it starts scanning for pokemon with only 1 minute left on their timer.

## Motivation and Context
Stops scanning for pokemon with less than 7.5 minutes

## How Has This Been Tested?
Tested my multiple people. Been running for 2 days continuously no crashes. Over a total spawn point count of 34k+

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -270,7 +270,7 @@ def search_overseer_thread_ss(args, new_location_queue, pause_bit, encryption_li
     # find the inital location (spawn thats 60sec old)
     pos = SbSearch(spawns, (curSec() + 3540) % 3600)
     while True:
-        while timeDif(curSec(), spawns[pos]['time']) < 30:
+        while timeDif(curSec(), spawns[pos]['time']) < 10:
             time.sleep(1)
         # make location with a dummy height (seems to be more reliable than 0 height)
         location = [spawns[pos]['lat'], spawns[pos]['lng'], 40.32]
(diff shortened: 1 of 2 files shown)


Infernoman:
Currently it adds pokemon to the queue with 14 minutes left on the timer. If the queue grows too large it would scan a pokemon with only 1 minute left on its timer.

BaIthamel:
Wait so what you're saying is you don't have enough accounts or you should spread your spawnpoints over more workers so you have downtime between scans? You're talking about TBterra right?

aldarund:
Maybe make this time a param that can be passed not a hardcoded value?

Infernoman:
OKAY current course of operations are as follows. The worker checks the spawns.json for the entry. the entry will be added to the queue if there's 14 minutes left on it. IF the queue is SOOO large that it can't make it to this pokemon in 13 minutes. It will skip it. If it reaches the pokemon within 13 minutes it will add this pokemon to the scan with only 1 minute left on its expiry timer.

BaIthamel:
at the very least it needs to be verbose that it's dropping spawns

Infernoman:
@balthamel line 420 log.info('Cant keep up. Skipping')

Infernoman:
Also added the argument as requested and defaulted to 7.5 minutes off of the timer.

justinkwaugh:
You also sneakily reduced the post spawn check time from 60 to 30 (which I approve of)

Infernoman:
@justinkwaugh It could even be a bit lower, I personally only wait 5 seconds, but 30 should be more than enough

TBTerra:
officially +1ing this, changes seems appropriate for the situation (the original algorithm was for data mining, this is more suitable for live mapping

hhu94:
I have personally found that if you don't use the pokestop and gym scanning that much you can disable them to improve speeds and reduce the search queue to 0.

PR #666, opened 2016-08-14
Add Detailed Gym Data

maddhatter:
## Description
Collection detailed information about which Pokemon are in a gym, along with the gym name and description. It's an opt-in feature (using the `-gi` switch), so no additional API hits unless you explicitly want them.

This pairs really well with #748 (already merged) to create a gym-only scanner, and would allow future development on a local gym leaderboard using the information collected (and sent out via webhook).

## Motivation and Context
#605
It's nice to have the option to see detailed gym information about who's in which gyms, the gym's prestige level, the gym's name, etc.

## How Has This Been Tested?
Backend tested on Windows 10 with SQLite and a MariaDB (running on a CentOS VM), in various scanning locations (Mine and @FlarumHostingCo's location, namely)
Frontend tested on Windows 7/10 on Firefox/Chrome, and on Android Chrome

## Screenshots (if appropriate):
2016-08-15 13_32_16-pokemon go map

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #742, opened 2016-08-16
Added fields to pokéstop webhook data

sebastienvercammen:
## Description
Added fields to pokéstop webhook data, changed default behavior to send all data via webhook and added optional params to send only updates (lured pokéstops & pokémon) via webhooks.

## Motivation and Context
Not all data is being sent to webhooks. 3rd party webhooks might not be aware of the data in the scanner's database, so the extra data is welcome.

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -23,6 +23,8 @@
 #gmaps-key:             # your Google Maps API key
 #proxy:                 # Proxy URL e.g. socks5://127.0.0.1:9050
 #webhook:               # webhook URL (including http://)
+#webhook-all-stops:               # send all pokéstops to webhooks, not only lured pokéstops
+#webhook-gyms:               # send gyms to webhooks
 
 # Webserver settings
 #host:                  # address to listen on (default 127.0.0.1)
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -23,8 +23,7 @@
 #gmaps-key:             # your Google Maps API key
 #proxy:                 # Proxy URL e.g. socks5://127.0.0.1:9050
 #webhook:               # webhook URL (including http://)
-#webhook-all-stops:               # send all pokéstops to webhooks, not only lured pokéstops
-#webhook-gyms:               # send gyms to webhooks
+#webhook-updates-only:               # only send updates to webhooks, (excludes gyms & non-lured pokéstops)
 
 # Webserver settings
 #host:                  # address to listen on (default 127.0.0.1)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -430,7 +430,7 @@ def parse_map(map_dict, step_location):
                         'active_fort_modifier': active_fort_modifier
                     }
 
-                    if not args.webhook_all_forts:
+                    if not args.webhook_all_stops:
                         send_to_webhook('pokestop', webhook_data)
                 else:
                     lure_expiration, active_fort_modifier = None, None
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -446,7 +446,7 @@ def parse_map(map_dict, step_location):
                     'active_fort_modifier': active_fort_modifier,
                 }
 
-                # Send all pokéstops to
+                # Send all pokéstops to webhooks
                 if args.webhook_all_stops:
                     # Explicitly set 'webhook_data', in case we want to change the information pushed to webhooks,
                     # similar to above and previous commits.
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -421,9 +421,12 @@ def parse_map(map_dict, step_location):
                         f['last_modified_timestamp_ms'] / 1000.0) + timedelta(minutes=30)
                     active_fort_modifier = f['active_fort_modifier']
                     webhook_data = {
+                        'pokestop_id': f['id'],
+                        'enabled': f['enabled'],
                         'latitude': f['latitude'],
                         'longitude': f['longitude'],
                         'last_modified_time': f['last_modified_timestamp_ms'],
+                        'lure_expiration': lure_expiration,
                         'active_fort_modifier': active_fort_modifier
                     }
                     send_to_webhook('pokestop', webhook_data)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -429,7 +429,9 @@ def parse_map(map_dict, step_location):
                         'lure_expiration': lure_expiration,
                         'active_fort_modifier': active_fort_modifier
                     }
-                    send_to_webhook('pokestop', webhook_data)
+
+                    if not args.webhook_all_forts:
+                        send_to_webhook('pokestop', webhook_data)
                 else:
                     lure_expiration, active_fort_modifier = None, None
 
@@ -444,6 +446,13 @@ def parse_map(map_dict, step_location):
                     'active_fort_modifier': active_fort_modifier,
                 }
 
+                # Send all pokéstops to
+                if args.webhook_all_stops:
+                    # Explicitly set 'webhook_data', in case we want to change the information pushed to webhooks,
+                    # similar to above and previous commits.
+                    webhook_data = pokestops[f['id']]
+                    send_to_webhook('pokestop', webhook_data)
+
             elif config['parse_gyms'] and f.get('type') is None:  # Currently, there are only stops and gyms
                 gyms[f['id']] = {
                     'gym_id': f['id'],
@@ -456,6 +465,13 @@ def parse_map(map_dict, step_location):
                     'last_modified': datetime.utcfromtimestamp(
                         f['last_modified_timestamp_ms'] / 1000.0),
                 }
+
+                # Send gyms to webhooks
+                if args.webhook_gyms:
+                    # Explicitly set 'webhook_data', in case we want to change the information pushed to webhooks,
+                    # similar to above and previous commits.
+                    webhook_data = gyms[f['id']]
+                    send_to_webhook('gym', webhook_data)
 
     pokemons_upserted = 0
     pokestops_upserted = 0
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -430,7 +430,8 @@ def parse_map(map_dict, step_location):
                         'active_fort_modifier': active_fort_modifier
                     }
 
-                    if not args.webhook_all_stops:
+                    # Include lured pokéstops in our updates to webhooks
+                    if args.webhook_updates_only:
                         send_to_webhook('pokestop', webhook_data)
                 else:
                     lure_expiration, active_fort_modifier = None, None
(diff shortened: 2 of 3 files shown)


FrostTheFox:
+1 because seb <3

sebastienvercammen:
Added some extra stuff, so we can optionally enable all data to be sent via webhooks.

jnovack:
I highly disagree with adding ANOTHER environment variable.

The Webhook needs to be intelligent, not the scanner.  If `-wh` is defined, then send pokemon, gym, and stops.  The webhook should be responsible for doing something with the data (after all, you are only sending JSON, it should be formatting it, and filtering it.

sebastienvercammen:
@jnovack I was told not to include all data by default, as they only wanted to push updates (in this case, pokémon and lured pokéstops). Sending everything was considered "too much", so I added it as an option.

jnovack:
Ugh, someone needs their fingers smacked with a ruler.

Every time you add an environment variable, a puppy dies.

invisiblek:
-1
I agree with @jnovack completely here.

sebastienvercammen:
@jnovack @invisiblek Regardless of personal preference and how many puppies have to die, the point is some of us need the extra data sent to our webhooks. And it was @invisiblek who told me not to send everything by default.

So if you want to disagree, provide a reasonable alternative. If you find none, accept this one.

sebastienvercammen:
Moved CLI to config.ini as an alternative to previous suggestions.

JZ6:
Woo a puppy doesn't have to die! @invisiblek

JZ6:
Since configparse needs cli params, lemme think of something else.

sebastienvercammen:
This works... Config items w/o CLI.

jnovack:
Perhaps my statement wasn't clear enough. I had two points I will outline
them simply.

1. **The extra data should ALWAYS be sent.**  It is up to the Web hook
destination to drop or format the data as appropriate. Thus all "extra"
info needs to be sent.

2.  The solution to not having environment variables is not to put them in
a configuration file.  That's WORSE.  **Always send the full data** and you
do not need the option.

Virtual containers on cloud services abhor configuration files.

sebastienvercammen:
-1 to all of you for your lack of interpersonal skills.

Yes, @jnovack, I am fully aware of what you said. If you haven't understood what I said, please re-read my previous comments. My previous implementation isn't what **I** want - it's what was the only possible solution that fit all of the requirements of each of you (which I've received in PM). You should all take the time to talk to each other, as all of you have differing opinions and this GitHub project is not about what you each care most about, it's about developing a proper working platform for our open source users.

The current implementation is how **I** want to do it:

1. Default behavior is now to send all data to webhooks.
2. I follow ConfigArgParse's design choice: all config items have a matching CLI parameter which overwrites its value. To not do so is to ignore what ConfigArgParse was built on, and will only confuse both future developers and the end users. Trying to "hack" past this design choice only makes the code uglier and serves no real purpose besides satisfying some people's personal preferences.
3. `webhook-updates-only`, `False` by default, can be used to change behavior and only send lured pokéstops and pokémon spawns.

If any of you still have negative feedback, please write a 10-page dissertation on the subject before presenting your opinions. Because I understand: you all have your opinions. But they all differ, and they're being destructive to the progress of this pull request, so I will ignore them and stick to my own.

Thank you for taking the time to think before commenting next time. You are not the only people working on this.

jnovack:
> Default behavior is now to send all data to webhooks.

Then we're cool.

mark2s3:
### FYI:
The default setting to send all data to webhooks is a big performance hit. I don't know what powerhouses you are running this on but on my Raspberry Pi 3 the load rises from 0.4 to 2.7 after this change. That's a pretty hefty increase in load.

The practical consequence is that in the course of 30 minutes the notifications I get will have smaller and smaller disappear times. In the beginning it's around 15 minutes and slowly decreases all the way to zero when all times are late and get ignored.

At least this can be disabled with the new setting. But I suppose this will generate lots of support tickets from people who run this on less powerful computers.

Tested with 6ed7f7d4d3f3a3bbe3d639d2318ad381c447c0d9

sebastienvercammen:
@mark2s3 

> At least this can be disabled with the new setting.

There is no issue then.

mark2s3:
> There is no issue then.

If you think that introducing a change that increases load 6-7x by default is not an issue, I don't know what to say.
I hope you are not programming for living.

sebastienvercammen:
@mark2s3 You seem to have entirely missed the context of the conversation and you somehow feel the need to resort to a random insult with no basis.

Please remove yourself from any projects that involve working with other people.

mark2s3:
Sorry, I now read the whole thread and see it was not intended to be the default originally. I did get annoyed by your response of regarding a major increase in load as "no issue" with the default settings. It clearly is if it increases over six times unless the user changes a new setting. I consider this to be a good argument to not enable it as a default.

jnovack:
Curious, third-party here.  What are you doing on this rpi?

How many scanners, how many webhooks.  Give us the details.

mark2s3:
The pi is just for this purpose, running Raspbian.

25 accounts, spawnpoint scan with 800 points and running one webhook. Pretty simple case and never had issues until this default change.

Can you run a load test with this setting enabled and disabled?

sebastienvercammen:
@mark2s3 If it did increase your load by that much, it means the previous blocking implementation of the webhook was slowing down your scanners by at least as much time.

Instead of blocking and slowing down, it now properly continues scanning and simultaneously handles the requests, which can explain the increase in load. The load is the same, it's just not being blocked anymore.

mark2s3:
I mean that the load increase happens if the new setting is disabled (default), and it decreases if it's enabled. The commit is the same.

pyr0ball:
@mark2s3 so basically the issue (as you've described it) is that on a previous commit where the webhook function was basically broken, and not functioning as intended due to blocking, you experienced no increased load, but on the current commit where the webhook function is working as intended (and therefore drawing system resources to perform its function) it's now overloading Your Raspberry Pi.

It sounds to me like the Pi just doesn't have the system resources to handle webhooks functions. Having them enabled while they were blocked amounts to a null string. It's the same as if it was disabled.

mark2s3:
The new default option is to send all data to webhooks. This makes the program slow down to a crawl.
If change the option to send only updates (lured pokéstops & pokémon) it doesn't slow down to a crawl.

Are you saying that the latter is a broken implementation? If it is why is it even an end user option?

mark2s3:
One more time:
When `webhook-updates-only` is not set (the default) I get very high load.
When `webhook-updates-only` is set I do not get high load.

Does `webhook-updates-only` setting switch the broken implementation on and off?

> Tested with 6ed7f7d4d3f3a3bbe3d639d2318ad381c447c0d9

pyr0ball:
@mark2s3 no, you misunderstand. 

On a previous commit the webhooks function receives all data sent all the time, but was working under the same process thread as the scanners. This is fine on multi-threaded processors, but causes "blocking" on single-threaded environments.

Example:
`scan > write db store > read db store > webhook data transmit > syn > ack > scan > [repeat]`

This was not intended behavior as it causes scans to be paused while awaiting the webhook process to complete. The fixed commit breaks the webhook data transmissions off into a new thread so that both can run simultaneously.

Example:
```
Thread1: scan > write db store > scan > write db store > scan > write db store > [repeat]
Thread2: read db > webhook data transmit > syn > ack > [repeat]
```

When using default settings, the amount of data queried and sent is much higher than when using `webhook-updates-only`. Which is why you're seeing a proportional CPU usage increase. While both processes existing on the same thread is less intensive on CPU usage, it's also not doing it's job properly.

mark2s3:
> When using default settings, the amount of data queried and sent is much higher than when using `webhook-updates-only`.

Exactly this. This is what I have tried to say. And this is why I don't think it should be a default setting.

Btw, Raspbery Pi 3 has a pretty good CPU. I've read people use this project even on the first generation Pis. I also run PokemonGo-Map (incl. db) and the webhook on a RAM disk so that can't be a bottleneck.

sebastienvercammen:
From here on out, please only comment on this particular PR (or any issue related to it) if you fully understand the code that has been edited, including how everything below it works (from threading to multiprocessing in general).

It is very unproductive for devs to have to explain themselves to people who can't even understand the basics of threading.

mark2s3:
My issue was informing about significant load increase with the `webhook-updates-only` setting disabled vs. enabled. I never asked lectures about "threading to multiprocessing in general."

sebastienvercammen:
@mark2s3 Please refrain from commenting. You do not understand threading or multiprocessing, and we have tried over and over again to explain the basics, but it's not working out.

We even almost boiled it down to _"when you're sending 5 things, you're sending more than when you're sending only 2"_, but `5 > 2` seems to be difficult to grasp.

I appreciate all contributors to open source very much and I'm even used to talking to people who want to voice their opinions even though they have no technical background, and that's perfectly fine. Wonderful people, who might all have something to add in their own way.

But you're just wrong, and you don't seem to be trying to understand why. You just want to keep being wrong.

BaIthamel:
For god sake i'm not reverting this, it's a good PR and it adds good things well. If you want to argue about how it now does more stuff faster because it's coded better and how that's somehow a BAD thing (i mean wtf stoopid tag here) then sure. But know I'm not reverting it, no matter what you say, this is staying in.

pyr0ball:
I think what he's trying to say is that the default behavior should be sending less data, and use less requirements, however I disagree as the vast majority of users won't experience this kind of bottleneck.

FlarumHostingCo:
@sebastienvercammen  Will this add the name of the PokeStop like say "Frank's Used Cars" to the marker for it on the map?    If so, any way to display it only on lured markers, not all?  If not, any chance of adding it?  :D

sebastienvercammen:
@FlarumHostingCo Nope, that's not what this is for. Check PR #666.

mark2s3:
@pyr0ball 
> I think what he's trying to say is that the default behavior should be sending less data, and use less requirements

Thank you, this is exactly what I mean. It boggles me how others don't get this.

> however I disagree as the vast majority of users won't experience this kind of bottleneck.

That's perfectly fine if nobody else sees a major increase in load when they test the setting on/off.

pyr0ball:
@mark2s3 I'm guessing that even with updates only enabled you'll still see a major increase in cpu usage after this commit when webhooks are enabled sure to the multiple threads. If you could test and let us know we can at least be informed of what the low end system requirements are.

I don't think any of us have random hardware laying around to test on

mark2s3:
>  If you could test and let us know we can at least be informed of what the low end system requirements are.

That's exactly what I have tested. Updates only has low load, sending all data six times higher.

Or did you mean something else? Sorry if I misunderstood.

pyr0ball:
No I think there was some confusion amongst everyone because of where and how you posted. We thought you were asking to revert back to non-multithreaded webhooks implementation because of your cpu getting nuked. We thought this because that's what this pull request is fixing.

mark2s3:
I see that commenting on this PR might have caused that kind of confusion. I commented here because I did not want to spam the issues, so I searched for the appropriate commits/PRs to directly give feedback. Unfortunately there was not a single, clear commit/PR about this, so I decided to comment here where the said option was first introduced.

My only purpose was to say that disabling/enabling the setting significantly increases resource load.

Over and out and thanks for understanding.

PR #748, opened 2016-08-16
Intelligently decide scanning step distance

jnovack:
Intelligently decide scanning step distance

## Description

If a scanner is only doing gyms and/or poke-stops, increase the step distance to 0.9km.

0.9km was specifically chosen because 99.99% of pokestops and gyms are visible within 0.9km and there have been issues with `get_gym_details()` with pokestops and gyms above 0.9km.

## Motivation and Context

If someone wants to only scan gyms and poke-stops, the step distance needs to be manually increased and custom docker containers must be manually built.   This allows the application to decide intelligently the step distance.

I didn't want to add ANOTHER environment variable so I made it intelligent (artificially).

## How Has This Been Tested?

Docker with all 6 permutations of options (`-np`, `-nk`, `-ng`).

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -61,14 +61,14 @@ def get_new_coords(init_loc, distance, bearing):
     return [math.degrees(new_lat), math.degrees(new_lon)]
 
 
-def generate_location_steps(initial_loc, step_count):
+def generate_location_steps(initial_loc, step_count, step_distance):
     # Bearing (degrees)
     NORTH = 0
     EAST = 90
     SOUTH = 180
     WEST = 270
 
-    pulse_radius = 0.07                 # km - radius of players heartbeat is 70m
+    pulse_radius = step_distance            # km - radius of players heartbeat is 70m
     xdist = math.sqrt(3) * pulse_radius   # dist between column centers
     ydist = 3 * (pulse_radius / 2)          # dist between row centers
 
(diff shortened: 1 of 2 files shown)


jnovack:
This obsoletes https://github.com/PokemonGoMap/PokemonGo-Map/pull/518 AND doesn't add ANOTHER environment variable!!

What's not to love?

hemsleyk:
+1

No legitimate reason to scan w/ standard distance when looking for gyms / stops.

chuyskywalker:
Risk of teleport bans is high, but... ^shrug^ run one worker with a relatively high step delay, and it might be ok.

pogom/search.py
```diff
@@ -194,8 +194,16 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
                 except Empty:
                     pass
 
+            # if we are only scanning for pokestops/gyms, then increase step radius to visibility range
+            if args.no_pokemon and not (args.no_pokestops or args.no_gyms):
```
chuyskywalker: I don't follow this logic, if I set `-np -ng` (ie, only scan pokestops) I'll end up with the .07 step radius, but that's not what we want, right? Effectively, if you say `np` (no pokemons) that should be the only trigger we need, right?

jnovack: You are right. Updated.

chuyskywalker:
LGTM

jnovack:
> Risk of teleport bans is high.

You can reduce it to "medium" with the right settings.  `-sd 30` will move you at 82mph (or 36m/sec).

You wouldn't necessarily run a single scanner to sweep entire areas consistently.  You might to only run `-sd 40 -st 16 ` over-night to get a city or even county wide view of all the stops/gyms in the area. 

Then, with a handful of accounts, sit in highly populated areas with `-st 1` to "keep an eye on things".

PR #749, opened 2016-08-16
hotfix to correct times got from db, also doumentation changes

TBTerra:
changes the time that spawns are checked when is -ss mode and are pulling their data from db

## Description
the original commit had the database return the despawn time rather than the (est) spawn time, causing it to not work when -ss was used without a fill, and also causing --dump-spawnpoints to generate an incorrect file

also ive updated the -ss documentation to not have so many formatting/spelling errors

## Motivation and Context
Motivation: fix it before people start blaming me

## How Has This Been Tested?
only tests over the original is that --dump-spawnpoints was used to get the output in human readable format to check that spawn times where not off by 15mins

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -414,7 +414,7 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
                                 log.exception('Search step %s map parsing failed, retrying request in %g seconds. Username: %s', step, sleep_time, account['username'])
                                 failed_total += 1
                         time.sleep(sleep_time)
-                    time.sleep(sleep_time)
+                    time.sleep(args.scan_delay)
                 else:
                     search_items_queue.task_done()
                     log.info('Cant keep up. Skipping')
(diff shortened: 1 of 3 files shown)


pogom/search.py
```diff
@@ -414,7 +414,7 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
                                 log.exception('Search step %s map parsing failed, retrying request in %g seconds. Username: %s', step, sleep_time, account['username'])
                                 failed_total += 1
                         time.sleep(sleep_time)
-                    time.sleep(sleep_time)
+                    time.sleep(args.scan_delay)
```
invisiblek: is this relevant to this change? seems out of place

TBTerra: its not technicaly part of the change noted, i can remove it if you want

invisiblek: i guess i'd prefer it in a separate commit since its unrelated

TBTerra: reverted change

PR #750, opened 2016-08-16
Fix for login delay affecting scan delay.

Infernoman:
## Description
Login delay was affecting the worker threads.

## Motivation and Context
Fixes scan delay in TbTerra
## How Has This Been Tested?
by changing files manually.

## Types of changes

## Checklist:


Infernoman:
Alright so this fixed my issue with 2 accounts but not 400

justinkwaugh:
It does not work the way you say.  There is one thread per account

Infernoman:
even with 1 thread for each account, each thread is being delayed a considerable amount.

justinkwaugh:
There could be an argument for moving the stagger method as in the -ss case the worker thread login is more closely tied to the spawn point time than the starting of the thread. But you put it in a place which doesn't really make sense, and you have removed the ability of it to sleep for scan_delay on a map request failure.

Infernoman:
@justinkwaugh There shouldn't have been a sleep in the while holding a database lock. The database should be unlocked before it goes to sleep on a map failure. Also regarding stagger_thread, It was placed DIRECTLY before the login. and now only occurs right before the login happens.

justinkwaugh:
Yes you don't need to put it before both logins.  This might be better:

```
def check_login(args, account, api, position):

    # Logged in? Enough time left? Cool!
    if api._auth_provider and api._auth_provider._ticket_expire:
        remaining_time = api._auth_provider._ticket_expire / 1000 - time.time()
        if remaining_time > 60:
            log.debug('Credentials remain valid for another %f seconds', remaining_time)
            return

    if args.spawnpoint_scanning:
        time.sleep(random.random() * 5)

    # Try to login (a few times, but don't get stuck here)
    i = 0
```

Infernoman:
@justinkwaugh what you posted will not work, and will constantly add a sleep during check_login() something else i'm trying to avoid. My code is working just fine as it is. you need the stagger_thread in both positions because it checks if there is a proxy, if there is no proxy then it jumps to the else statement, The stagger_thread() check needs to be in both the if and else statements below.
if args.proxy:
else:

justinkwaugh:
Check login returns if there is still enough time prior to where I put the code.  Functionally there is no difference between where I put it and where you did, except that yours was duplicated unecessarily.

Infernoman:
Fixed :) I'm still a bit new to coding, so cleanliness isn't my greatest feature right now. Thank you for the tip. And now i understand why you put it where you did, one more fix incoming.

justinkwaugh:
Also my prior point still stands.  The sleeps previously in search.py were not inside the parse_lock and were after the db usage.  And with your change you have removed any sleep between parse failure retries.

Infernoman:
@justinkwaugh line 416 that was removed is the one your referring to?

justinkwaugh:
Yes

Infernoman:
Fixed.

PR #751, opened 2016-08-16
Print worker and queue status to the screen

justsomedudeonthenet:
Instead of trying to follow the fast scrolling log output, prints the status of the overseer, queue and each worker to the screen.  This makes it easy to see dead workers and growing queue sizes.

## Description
Each worker gets a status queue assigned to it.  The overseer thread periodically calls a function to read from all the status queues and print to the screen.

STDERR should be redirected to a file so it doesn't clutter the screen.

Output is optional, enabled with '-ps', and defaults to off.

## Motivation and Context

- Makes it easy to see dead/erroring out workers
- Easier to follow than the raw log

## How Has This Been Tested?
Tested on Ubuntu on my server.

I expect that this will not work quite as nicely on windows, because of the method used to clear the screen.  It will still work, it just won't clear the screen when it updates.

status

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -154,11 +154,10 @@ def status_printer(threadStatus, search_items_queue):
         # Print the status of each worker, sorted by worker number
         for item in sorted(threadStatus):
             if(threadStatus[item]['type'] == "Worker"):
-                print '{} - {}'.format(item, threadStatus[item]['message'])
                 if 'skip' in threadStatus[item]:
-                    print '\tSuccess: {}\tFailed: {}\tSkipped: {}'.format(threadStatus[item]['success'], threadStatus[item]['fail'], threadStatus[item]['skip'])
+                    print '{} - Success: {}, Failed: {}, Skipped: {} - {}'.format(item, threadStatus[item]['success'], threadStatus[item]['fail'], threadStatus[item]['skip'], threadStatus[item]['message'])
                 else:
-                    print '\tSuccess: {}\tFailed: {}'.format(threadStatus[item]['success'], threadStatus[item]['fail'])
+                    print '{} - Success: {}, Failed: {} - {}'.format(item, threadStatus[item]['success'], threadStatus[item]['fail'], threadStatus[item]['message'])
 
         time.sleep(1)
 
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -139,20 +139,21 @@ def SbSearch(Slist, T):
     return first
 
 # print out the status of each worker
-def print_status(status_queues, worker_status, search_items_queue, overseer_status):
+def print_status(enabled, status_queues, worker_status, search_items_queue, overseer_status):
     # Gather worker status
     for worker, q in status_queues.iteritems():
         try:
             worker_status[worker] = q.get(False)
         except Empty:
             continue
-    # Clear the screen - This, unfortunately, only seems to work on linux. Anyone know a good cross platform way to clear the screen?
-    print(chr(27) + "[2J")
-    # Print the status
-    print 'Queue: {} items'.format(search_items_queue.qsize())
-    print 'Overseer: {}'.format(overseer_status)
-    for worker, message in worker_status.iteritems():
-        print "Worker {} - {}".format(worker, message)
+    if enabled:
+        # Clear the screen - This, unfortunately, only seems to work on linux. Anyone know a good cross platform way to clear the screen?
+        print(chr(27) + "[2J")
+        # Print the status
+        print 'Queue: {} items'.format(search_items_queue.qsize())
+        print 'Overseer: {}'.format(overseer_status)
+        for worker, message in worker_status.iteritems():
+            print "Worker {} - {}".format(worker, message)
 
 # The main search loop that keeps an eye on the over all process
 def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_path):
@@ -193,7 +194,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
                         search_items_queue.get_nowait()
                 except Empty:
                     pass
-            print_status(status_queues, worker_status, search_items_queue, "Scanning is paused")
+            print_status(args.print_status, status_queues, worker_status, search_items_queue, "Scanning is paused")
             time.sleep(1)
             continue
 
@@ -245,12 +246,12 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
             log.debug('Search queue empty, restarting loop')
             for step, step_location in enumerate(locations, 1):
                 log.debug('Queueing step %d @ %f/%f/%f', step, step_location[0], step_location[1], step_location[2])
-                print_status(status_queues, worker_status, search_items_queue, "Queuing next step")
+                print_status(args.print_status, status_queues, worker_status, search_items_queue, "Queuing next step")
                 search_args = (step, step_location)
                 search_items_queue.put(search_args)
         else:
             #   log.info('Search queue processing, %d items left', search_items_queue.qsize())
-            print_status(status_queues, worker_status, search_items_queue, "Processing search queue")
+            print_status(args.print_status, status_queues, worker_status, search_items_queue, "Processing search queue")
 
         # Now we just give a little pause here
         time.sleep(1)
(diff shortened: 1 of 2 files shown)


chuyskywalker:
This feels like a weird use of a Queue since there's really only "the most recent" status.

chuyskywalker:
I also wonder if this would be better/worse as a page like `/status`

ilendemli:
+1 for a /status page

rollin:
/status page should have restricted access then. If I share the map I wouldn't want to share the info about my workers as well.

justsomedudeonthenet:
I am certainly not an expert on python threading, but everything I read recommended using Queues to pass messages between threads, since its difficult to make sure anything else is thead safe.

A /status page would be nice, but would probably update slower, so you wouldn't see a lot of the messages.

mark2s3:
Also don't use print, use the logging module.

chuyskywalker:
> /status page should have restricted access then. If I share the map I wouldn't want to share the info about my workers as well.

Ahh, yeah, that's a good point.

> I am certainly not an expert on python threading, but everything I read recommended using Queues to pass messages between threads, since its difficult to make sure anything else is thead safe.

Yeah, queues are definitely thread safe, but I bet there's a way to pass messages that does end up building up a set of "throw away" values between each redraw.

A little digging reveals that if you made a global dict for messages, preallocating one entry per thread (likely by threadId) then setting those values per thread is threadsafe. (http://stackoverflow.com/questions/1312331/using-a-global-dictionary-with-threads-in-python)

So, 
```
threadStatus = {}
threadStatus[t.threadId()] = ''
```
During the loop and then a 
```
global threadStatus
threadStatus[myId()] = 'doing stuff'
```
Would dramatically simplify this code and is (mostly) thread safe. Safe _enough_ for this use case anyway.

> A /status page would be nice, but would probably update slower, so you wouldn't see a lot of the messages.

^shrug^ We're missing a lot from the queue nature in any case, I suppose. Which raises an interesting idea, if it _was_ a web page, and you kept the queue*, you could keep a running history of N messages per worker so you could see the "now" snap shot, but also have some interactivity to see the last N messages for any given thread in a "click to show more" manner. Which could be super helpful too.

_* Could also be done with the global dict holding arrays of messages too, just sayin'_

justsomedudeonthenet:
@mark2s3 if I use logging, it goes to STDERR.  The whole point is to print it to the screen.  I suppose I could create a second logging instance that logs to STDOUT, if there is really some reason to avoid print.  Is there a reason I'm not aware of why print is bad?

@chuyskywalker I was going on advice from [Python Cookbook](https://www.safaribooksonline.com/library/view/python-cookbook-3rd/9781449357337/ch12s03.html) and [this post on SO](http://stackoverflow.com/questions/14508906/sending-messages-between-class-threads-python).  But a dict would be simpler.  I will try that out tonight and see how it works.  The other methods I have seen involve using locks, and I want to avoid adding more locks that could potentially slow things down.

The problem I have with a /status page is that my workers don't all have a web interface.  I run a separate server only process that does the web interface, and a few workers that do the searching.  So /status wouldn't show anything useful.  If you were running beehive for example, /status would only show the status of the first one.  We would either need some way to gather status from all of the instances onto one page, or each instance would need to run its own webserver on a separate port.

mark2s3:
> Is there a reason I'm not aware of why print is bad?

Mixing print and the logging module is a sign of spaghetti code and this whole repo is already full of it. Please don't create more of it.

> I suppose I could create a second logging instance that logs to STDOUT

Create another handler that goes to STDOUT.

Also this: https://github.com/PokemonGoMap/PokemonGo-Map/pull/751/files#diff-6c43bfb0a183e3412f838ddd49606dbdR152
Please, just don't do that.

pogom/search.py
```diff
@@ -139,22 +139,44 @@ def SbSearch(Slist, T):
...
+def print_status(enabled, status_queues, worker_status, search_items_queue, overseer_status):
+    # Gather worker status
+    for worker, q in status_queues.iteritems():
+        try:
+            worker_status[worker] = q.get(False)
+        except Empty:
+            continue
+    if enabled:
+        # Clear the screen - This, unfortunately, only seems to work on linux. Anyone know a good cross platform way to clear the screen?
+        print(chr(27) + "[2J")
```
Cy4n1d3: What works for me on windows (and should work for unix as well): [see SO](http://stackoverflow.com/a/684344).

Cy4n1d3:
I really like the general idea of this (kinda similar to the way Pokeminer displays information with its main window).
Tested another method to clear the screen I found on SO and it works fine for me, you might wanna have a look at the comment I left.

mark2s3:
@Cy4n1d3 Even that is a hack and increases code debt. The proper and most **maintainable** way to do this is to use the logging module.
With the logging module you can even have it write this stuff on a separate log file and then you can view that.

chuyskywalker:
How does one clear the screen with the log module?

Cy4n1d3:
@mark2s3 I do agree and disagree at the same time. While I do agree that we should resort to using the logger consistently, the scope of this PR requires some information to be displayed on the console and (supposedly) not be logged to a file.
If you're keen on reading through thousands of log lines to find failed workers the logger might suffice (and I am surely not).
In this case, we're trying to get a clean and informative overview of the current status, which requires clearing the console and (possibly) hiding the logger output, which would be done only in case we're using the `-ps` switch.

mark2s3:
The whole idea of clearing the screen is just completely backwards.
Normally the logging module is outputting the INFO lines when the program runs and then this hack clears everything and prints on top of it. WTF? That's simply bad design.
If your goal is to see what the workers do, you can make another logger that logs that info to a file.
Or use the verbose levels with logging.

Also I don't read thousands lines of logs. I filter everything I'm not interested in.

Cy4n1d3:
Well, this should be either logging or status display then.
Many people don't need extensive log files (including me) for their small projects which are personal use only. In this case a simple wrapper function could work imho.
Instead of directly calling the logger / printstatus, call the wrapper which then decides what to do.
The user wants extensive logging down to debug-level? Fine, let him have it.
Desired behaviour is just a simple status overview? Skip logging, update worker status, log the information to console after clearing, done.

mark2s3:
Sorry for being so negative earlier. Had a rough day at work and poured some of it here.
You can do a lot of great stuff with Python's default logging module. E.g. redirecting to multiple places at once (including email) and making your own levels and stages of verbosity. If you use `print`, it will just print to STDOUT and old data just vanishes because you clear the console.
Granted this would need a bigger rewrite than this, but generally it's no good idea to use `print` when you are already using Logging. Unless you want a quick hack. But please, don't merge it for a release version of this project.

justsomedudeonthenet:
@mark2s3 The idea is to use this to get the immediate status of the workers on the screen, and send the logs elsewhere.  I run it as `python runserver.py <other options> -d -ps 2>worker.log`

That way this information goes to the screen, constantly refreshing, and I can look through worker.log if I need more detailed information about what is going on.  This also lets me see what's going on even with full debug logging, so I have a detailed record if there is an error without cluttering up the screen.  

I don't intend to have the logger print to the screen at the same time.  My way to fix that is just redirect STDERR.  I considered adding a switch to log to a file instead of the screen, but logger is configured before arguments are read, and I wasn't sure how to properly change it's settings to log to a file.

If there is a general consensus that print is evil, I can certainly change it to a second logging instance.

@Cy4n1d3 I'm not sure using os.system calls is a great idea.  It would work, but calling system commands is usually bad.  Maybe I should just remove the screen clearing entirely and let it scroll?

BaIthamel:
This is good in theory, it does need to have the old log on screen as well though.

Cy4n1d3:
@justsomedudeonthenet  Yea, you've got a point there. I still like the idea of an easy to comprehend status interface though and right now I can't find a working cross platform solution which does not rely on printing 100 empty lines...
I'd use a secondary logging instance in case the `-ps` switch is set, maybe taking arguments to extend this feature (console/file/all) and satisfy all parties.

@BaIthamel does it? Any particular reason for this?

justsomedudeonthenet:
@Balthamel if you are redirecting it to a log file, just run `tail -f whatever.log` in another window to get your scrolling log output.

justsomedudeonthenet:
@chuyskywalker I have implemented your suggestion of using dicts instead of queues.  Queues were overkill.  With the dict style of passing messages, there is also room to add a few more keys, like how many total successes and failures that worker has had. 

I also made the printing happen in a separate thread, so it can continue to update on a regular basis without having to call it over and over again from the overseer thread.

Thoughts?

justsomedudeonthenet:
I have updated this to be able to switch between pretty status and log output by pressing enter.

justsomedudeonthenet:
New screenshots:

At startup:
screenshot1

After pressing enter:
screenshot2

Pressing enter will keep switching between the two.

Zerotask:
@justsomedudeonthenet this is very nice. To easily swap banned accounts it would be great to see the account used by the worker.

chuyskywalker:
That's coming up in #901

PR #492, opened 2016-08-11
Opinionated login throttling

marumari:
## Description
In #392, somebody has ~180 accounts.  Even with the recent code commits (to spread out logins over scan delay), that can still result in dozens of logins per second.

This commit essentially throttles logins to roughly one per second.  It adds some random jitter to the logins -- .25 seconds in either direction -- so that the load is still spread out evenly.

## Motivation and Context
People are breaking their own stuff and PTC's stuff with overly aggressive logins.

That said, with Niantic's IP-based throttling, I don't know how long mappers with this many accounts have to live anyways.

## How Has This Been Tested?
Works great on my system!

## Screenshots (if appropriate):
Here's what it looks like with 8 accounts and a scan delay of 1:

```
2016-08-11 09:51:20,219 [ search_worker_0][        search][   DEBUG] Delaying thread startup for 0.00 seconds
2016-08-11 09:51:20,244 [ search_worker_1][        search][   DEBUG] Delaying thread startup for 1.13 seconds
2016-08-11 09:51:20,249 [ search_worker_2][        search][   DEBUG] Delaying thread startup for 2.06 seconds
2016-08-11 09:51:20,250 [ search_worker_3][        search][   DEBUG] Delaying thread startup for 3.17 seconds
2016-08-11 09:51:20,250 [ search_worker_4][        search][   DEBUG] Delaying thread startup for 3.77 seconds
2016-08-11 09:51:20,250 [ search_worker_5][        search][   DEBUG] Delaying thread startup for 5.07 seconds
2016-08-11 09:51:20,250 [ search_worker_6][        search][   DEBUG] Delaying thread startup for 5.92 seconds
2016-08-11 09:51:20,266 [ search_worker_7][        search][   DEBUG] Delaying thread startup for 7.12 seconds
```

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #752, opened 2016-08-16
Delay/throttle fix when search resumes, preventing burst of search requests

taihun81:
## Description
Fix for burst of search requests after search resumes, resulting in soft bans.

## Motivation and Context
I have added interval delays to when search queue gets filled up, so that each search thread can only initiate the search at given fixed interval.  That way even if the search resumes after extended period of pause, the search threads won't burst out the request all at once.  **Please comment if the delay interval handled in search queue is a good idea or not**

Spawn point search thread has been left untouched.

## How Has This Been Tested?
Ran, tested repeats of pause and resume for various intervals.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -222,9 +222,9 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
         # Setting delay interval for filling in search queue to throttle down search workers
         if len(args.accounts) > 1:
             if len(args.accounts) > args.scan_delay:  # force ~1 second delay between threads if you have many accounts
-                delay_interval = ((random.random() - .5) / 2)
+                delay_interval = random.random() / 2
             else:
-                delay_interval = (args.scan_delay / len(args.accounts))
+                delay_interval = args.scan_delay / len(args.accounts)
         else:
             delay_interval = args.scan_delay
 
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -295,6 +307,8 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
             if args.proxy:
                 api.set_proxy({'http': args.proxy, 'https': args.proxy})
 
+            api.activate_signature(encryption_lib_path)
```
xrobau: I think this is handled elsewhere now?

Cy4n1d3: It's also missing the proxy authentification thingy.

taihun81: That diff is just pulling API activation out of the loop as it just need to be called once.  Everything else is untouched,

Goldvivi1:
if it helps prevent my account from getting softban everytime i change location, im in on this

lduda007:
Works great for me. Should be merged into develop branch imo.

Cy4n1d3:
So far no problems on my side, thanks :)

Racle:
Works like a charm. Should be merged.

FlarumHostingCo:
I get an error  trying to start run.py 

NameError: global name 'random' is not defined

taihun81:
Sounds like you didn't import random module.  Suspecting you are behind #492 ?

FlarumHostingCo:
Thanks!  Yeah I'm still using old AAHAAAA repo and cherry picking my needs.. Ok I'll try that thanks!

PR #753, opened 2016-08-16
Add new feature: display reachable range for pokemon, pokestop, and gyms

kamykaze:
## Description
This PR adds a new feature/concept of reachable ranges. What's a range? In Pokemon Go, that is the real world area where you need to be standing at in order to catch a pokemon, collect a pokestop, or battle in gym.

This PR adds a visual representation (semi-transparent circles) in the map so you can determine where you need to be in order to interact with certain game objects.

## Motivation and Context

Sometimes you see a pokemon in the game and wonder if you'll be able reach it because it is too deep into a building. If you need to go around the building to get it it may not be worth it. Or maybe you're planning your lunch break but want to find a restaurant where you can battle in a gym while you eat. Or sometimes you just want to find a "sweet spot" where you can collect multiple pokestops at once without having to walk around.

By displaying a 40m radius range, you can visually tell exactly where you need to be to reach any of these.

## How Has This Been Tested?

Implemented and compiled with grunt, tested using Chrome in Mac. Things that were checked:

- toggling of all ranges
- toggling of individual ranges (gym, stops, pokemons)
- refreshing browser with range toggle on and off
- checked that expired pokemons have their range removed along with its marker
- checked for expected markers and ranges after:
  - zooming in/out
  - going in/out of bounds

pokemongomap_reachableranges

## Types of changes

Some functions were changed to pass the marker type in order to display the ranges in different colors. The function calls have been updated properly and there should be graceful fallback values. However, it's possible it may cause conflicts with people changing the same functions

## Checklist:

Only missing mention of the new feature in the main README


justsomedudeonthenet:
I like the idea.  But does it significantly slow down the map, especially when zoomed out?

kamykaze:
Good question. I'm not sure how to test that. It doesn't look like we have anything in place to help measure how long the map takes to load, render, zoom, etc.

To be honest, I don't imagine it would slow things down much since they're just simple circle objects (as opposed to graphic icons), but they do use transparency so that could make things worse.

As far as your scenario goes (zommed out), I suppose I could check for the map zoom and disable the range circles altogether since they're not useful at a zoomed out view.

justsomedudeonthenet:
Disabling the circles when zoomed out past a certain point (might have to play around with it to find out what that point is) would be good.  The map is laggy enough when zoomed out to cover a city, and the circles would be useless at that point.

kamykaze:
I updated the code to automatically remove the range circles when zoomed out (played with the zoom until I got a reasonable zoom level). I also rebased my branch to get the laste commits from 'develop'

kamykaze:
A quick UI update. I realized that showing the scanned locations (specially when you have some old ones that are yellow and red) can cause quite a bit of confusion. I made the range colors a little more noticeable while making pokemon ranges purple.

pokemongomap-recoloredranges

chuyskywalker:
Awesome work; nice touch on the zoom disable factor as well.

static/js/map.js
```diff
@@ -810,6 +815,8 @@ function notifyAboutPokemon (id) { // eslint-disable-line no-unused-vars
 }
 
 function removePokemonMarker (encounterId) { // eslint-disable-line no-unused-vars
```
SkOODaT: this doesnt work right need to have it like this :

    mapData.pokemons[encounterId].marker.setMap(null)
    mapData.pokemons[encounterId].hidden = true
+  mapData.pokemons[encounterId].marker.rangeCircle.setMap(null)
+  delete mapData.pokemons[encounterId].marker.rangeCircle

maddhatter:
Would love to see this get added - sent you a [PR](https://github.com/kamykaze/PokemonGo-Map/pull/1) to resolve the conflicts

static/js/map.js
kamykaze: Thanks for catching that @SkOODaT but that  fix only moves the error down further, allowing you to remove the marker.  I've added an if statement to check if the rangeCircle exists before attempting to remove it.

kamykaze:
@maddhatter Thanks I merged your PR, but by that time there were lots of new commits in develop, so I merged those as well. I also fixed the issue pointed out by SkOODaT.

chuyskywalker:
I'll give this updated version a try tomorrow and probably merge it. Thanks!

chuyskywalker:
Not quite "tomorrow" but didn't forget about'cha!

PR #757, opened 2016-08-17
Fixed miss handling of error when parsing map data

Cor3Zer0:
The method used to handle an error when parsing map data was incorrect. Will now handle error correctly.

## Description
When the search thread was checking if there was an error when parsing the map it was doing it incorrectly and therefor not looping and increasing failed_total counter. Which means people were just getting a traceback with keyerror.

It will now correctly handle the error and loop as was intended 

## Motivation and Context
Code to rerun a scan on an area was being missed if there was an error. This code now runs as intended.

## How Has This Been Tested?
Tested on functional and banned accounts. 
No longer getting traceback when getting a keyerror.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -384,7 +384,7 @@ def parse_map(map_dict, step_location):
         cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells']
     except KeyError:
         return False
-    
+
     for cell in cells:
         if config['parse_pokemon']:
             for p in cell.get('wild_pokemons', []):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -379,7 +379,7 @@ def parse_map(map_dict, step_location):
     gyms = {}
     scanned = {}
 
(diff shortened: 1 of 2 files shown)


mark2s3:
I don't think this is doing what you think it is doing. You are just mixing exception handling and if statements and actually not changing anything but adding more complexity.

`failed_total` increases just fine with the current code. The exception is only printed and does not affect the runnning of the program:
https://docs.python.org/2/library/logging.html#logging.Logger.exception
> Exception info is always added to the logging message.

I agree that printing the exception here is not that useful and actually causes many duplicate issues. Changing it to regular *Logging.error* with "Niantic has banned this account" would be more useful.

Viper2k4:
But if this Error means "Niantic has banned this account". Then the worker should stop instant of try it over and over and over again ^^

PR #765, opened 2016-08-17
Added -ss Compression

Xcelled:
## Description
Added an algorithm from #179. It compresses the spawnpoints list to reduce the number of redundant calls.

Discussion of the algo can be found on #179, but in a nutshell it removes spawnpoints that would be caught by a (configurable) later scan. This 'lag time' defaults to two minutes. For example, if the spawn points are [1, 2, 3], and 1 and 3 are within 70 meters of each other and 3 spawns within two minutes of 1, the algorithm will forgo scanning 1, since its spawn will be caught when 3 is scanned.

Results: You lose up to two minutes on the time of 1, but make less requests.

Aggressive compression will try all permutations of the spawnpoint list to find the optimal compression. However, this takes quite a bit of time for larger spawnpoint lists, roughly `O(n**2)` growth and doesn't add all that much.

With a 2 minute window (the default) on a -st 7 of santa monica pier (note the timestamps):

```
2016-08-17 01:41:44,617 [   search_thread][        search][    INFO] 260 spawnpoints loaded. Compressing spawnpoints...
2016-08-17 01:41:44,706 [   search_thread][        search][    INFO] Total of 213 locations to scan
```

With `aggressive-compress` (again note the timestamps):
```
2016-08-17 01:43:15,934 [   search_thread][        search][    INFO] 260 spawnpoints loaded. Compressing spawnpoints...
2016-08-17 01:43:31,239 [   search_thread][        search][    INFO] Total of 210 locations to scan
```

## Motivation and Context
Heuristically optimizes the list of spawnpoints to further reduce the strain on Niantic (as well as teleporting)

## How Has This Been Tested?
I tested this in a limited fashion, primarily based on debug output. More eyes wouldn't hurt.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -123,6 +123,8 @@ def get_args():
                         action='store_true', default=False)
     parser.add_argument('-ss', '--spawnpoint-scanning',
                         help='Use spawnpoint scanning (instead of hex grid)', nargs='?', const='null.null', default=None)
+    parser.add_argument('--sp-compression', help="Don't scan spawnpoints that spawn up to this many seconds before \
+                        another spawnpoint. This reduces redundant scans. Defaults to 120 (2 minutes)", type=int, default=120)
     parser.add_argument('--dump-spawnpoints', help='dump the spawnpoints from the db to json (only for use with -ss)',
                         action='store_true', default=False)
     parser.add_argument('-pd', '--purge-data',
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -123,8 +123,9 @@ def get_args():
                         action='store_true', default=False)
     parser.add_argument('-ss', '--spawnpoint-scanning',
                         help='Use spawnpoint scanning (instead of hex grid)', nargs='?', const='null.null', default=None)
-    parser.add_argument('--sp-compression', help="Don't scan spawnpoints that spawn up to this many seconds before \
-                        another spawnpoint. This reduces redundant scans. Defaults to 120 (2 minutes)", type=int, default=120)
+    parser.add_argument('--sp-compression', help="The maximum time separation (in seconds) of two spawnpoints for them to be \
+                        compressed. Compression reduces redundant scans. Specify 0 to disable compression. Defaults \
+                        to 120 (2 minutes)", type=int, default=120)
     parser.add_argument('--aggressive-compress', help='Spend extra time compressing the spawnpoints to find a better compression',
                         action='store_true', default=False)
     parser.add_argument('--dump-spawnpoints', help='dump the spawnpoints from the db to json (only for use with -ss)',
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -125,6 +125,8 @@ def get_args():
                         help='Use spawnpoint scanning (instead of hex grid)', nargs='?', const='null.null', default=None)
     parser.add_argument('--sp-compression', help="Don't scan spawnpoints that spawn up to this many seconds before \
                         another spawnpoint. This reduces redundant scans. Defaults to 120 (2 minutes)", type=int, default=120)
+    parser.add_argument('--aggressive-compress', help='Spend extra time compressing the spawnpoints to find a better compression',
+                        action='store_true', default=False)
     parser.add_argument('--dump-spawnpoints', help='dump the spawnpoints from the db to json (only for use with -ss)',
                         action='store_true', default=False)
     parser.add_argument('-pd', '--purge-data',
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -250,6 +250,7 @@ def search_overseer_thread_ss(args, new_location_queue, pause_bit, encryption_li
         t.daemon = True
         t.start()
 
+    log.info('Loading spawnpoints')
     if os.path.isfile(args.spawnpoint_scanning):  # if the spawns file exists use it
         try:
             with open(args.spawnpoint_scanning) as file:
@@ -265,8 +266,34 @@ def search_overseer_thread_ss(args, new_location_queue, pause_bit, encryption_li
     else:  # if spawns file dose not exist use the db
         loc = new_location_queue.get()
         spawns = Pokemon.get_spawnpoints_in_hex(loc, args.step_limit)
-    spawns.sort(key=itemgetter('time'))
-    log.info('Total of %d spawns to track', len(spawns))
+
+    log.info('%d spawnpoints loaded. Compressing spawnpoints...', len(spawns))
+
+    spawns.sort(key=itemgetter('time'), reverse=True)
+    compressed = []
+    while len(spawns) != 0:
+        spawn = spawns.pop(0)
+        compressed.append(spawn)
+        # log.debug('Processing spawn %r', spawn)
+
+        # Remove any other spawns that fall within range of this one
+        redundant = []
+        for sp_to_check in spawns:
+            if timeDif(spawn['time'], sp_to_check['time']) > args.sp_compression:
+                break  # We're already outside the time, no sense continuing
+            dist = geopy_distance.distance((spawn['lat'], spawn['lng']), (sp_to_check['lat'], sp_to_check['lng'])).meters
+            if dist < 70:
+                # log.debug('Removing spawn %d m away: %r', dist, sp_to_check)
+                redundant.append(sp_to_check)
+
+        for x in redundant:
+            spawns.remove(x)
+
+    spawns = compressed
+    spawns.reverse()  # put them back in ascending order
+
+    log.info('Total of %d locations to scan', len(spawns))
+
     # find the inital location (spawn thats 60sec old)
     pos = SbSearch(spawns, (curSec() + 3540) % 3600)
     while True:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -28,6 +28,7 @@ import geopy.distance as geopy_distance
 from operator import itemgetter
 from threading import Thread, Lock
 from queue import Queue, Empty
+from collections import deque
 
 from pgoapi import PGoApi
 from pgoapi.utilities import f2i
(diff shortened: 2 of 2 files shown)


chuyskywalker:
Haven't tried this, but looks really nice. (This is also something, as you mentioned before, where pre-processing would help tons)

Xcelled:
Could whoever merges this make sure to add a credit to https://github.com/Draco18s ? This was his idea.

@chuyskywalker yes. Small steps to get there!

Interestingly, I ran it with a window of 5 minutes, and it only resulted in 30 less calls than two minutes.

slideup-benni:
Is this compatible with https://github.com/PokemonGoMap/PokemonGo-Map/commit/c36d1acb144e96f81c70fecc41ff89ba4ff317cd ?

justinkwaugh:
This seems like an ok change, but it doesn't necessarily reduce teleporting in any reliable way.

Xcelled:
@justinkwaugh it can reduce teleporting. In your case, it would jump from 0 to 1, 1 to 2, and 2 to 3. With compression it'd only jump 0 to 2 and 2 to 3. This is a reduction of one teleport. This becomes highly beneficial if 0 and 2 are close and 1 and 3 are far from them.

Ex if 0 is at 0,0, 1 is at 0,100, 2 is at 0,20, and 3 is at 0,101... without compression it will jump 100 + 80 + 81. With compression, only 20+81 (and over a longer time so velocities are smaller).

The velocity is more important than having an intermediate point, since a jump could be someone turning their phone screen off (and thus not sending positions) as they walk. But if the velocity is too high, cybernetic eyebrows will raise.

This reduction isn't guaranteed, obviously, but it's more likely to happen with fewer points to scan.

In my experience, you will get teleport banned if you jump around with any decent frequency over about 12 hours. A single 300km/h jump every 10 minutes doesn't seem to trigger their ban.

Xcelled:
@slideup-benni yes, I built it on top of that commit.

justinkwaugh:
@Xcelled, In my scenario I was suggesting that 0 and 2 were not close, and 1 and 3 were somewhere in between, reducing the distance between 0 and 2.  However after further thought I do see that because the time between 0 and 2 are constant, the velocity cannot be increased by removing any points in between, and can in fact only be reduced or kept exactly the same as the distance between cannot be increased by removing points.

Regarding frequency, I have a worker which is scanning ~200 spawn points, which results in fairly frequent scans (most are the minimum 10s), and most are in the 100km/h to 300km/h range, and has been running for days.  Just throwing it out there as a data point. 

To some extent I wonder if there is any real point in some of these optimizations, in that they could so easily detect and ban these workers if they wanted by a huge variety of heuristics.

pogom/utils.py
```diff
@@ -123,6 +123,10 @@ def get_args():
                         action='store_true', default=False)
     parser.add_argument('-ss', '--spawnpoint-scanning',
                         help='Use spawnpoint scanning (instead of hex grid)', nargs='?', const='null.null', default=None)
+    parser.add_argument('--sp-compression', help="Don't scan spawnpoints that spawn up to this many seconds before \
+                        another spawnpoint. This reduces redundant scans. Defaults to 120 (2 minutes)", type=int, default=120)
```
tylerdola: If I don't provide the argument, it won't compress, but if I do and leave it blank, it doesn't default to 120
`runserver.py: error: argument --sp-compression: expected one argument`

Xcelled: sp-compression is already on. You only need to specify it if you want to change the time from 2 minutes. I'll change the help text to better reflect this.

tylerdola: I see it now, thanks!

Xcelled:
@justinkwaugh 

>Regarding frequency, I have a worker which is scanning ~200 spawn points, which results in fairly frequent scans (most are the minimum 10s), and most are in the 100km/h to 300km/h range, and has been running for days. Just throwing it out there as a data point.

Maybe it was just coincidence that my workers were banned around the time I started teleporting. I'll keep an eye on them now.

>To some extent I wonder if there is any real point in some of these optimizations, in that they could so easily detect and ban these workers if they wanted by a huge variety of heuristics, especially with spawnpoint scanning. (always querying exactly the coords of spawns and nothing else is pretty easy to notice)

Heuristics can be fooled! The more we make the workers behave like humans, the harder it is for them to detect. We're slowly evolving towards human-esque behavior. The types of heuristics they'd need to run would be pretty intensive, comparing backwards through time on all accounts, maybe even correlating multiple accounts. That could also result in a number of false positives, so I don't think they'll get that aggressive if they have a choice. There are other possible anti-bot measures as well, but they get increasingly disruptive to legitimate gameplay, so likewise I think they'll be avoided.

All these optimizations serve to reduce the strain on Niantic, removing significant pressure on them to fight bots. If the servers aren't being hammered by several thousand beehives constantly, they'll have a much harder time justifying extreme measures to go after map workers.

Zerotask:
What are the benefits to set --aggressive-compress default=True? To force a more human-like scanning method?

Xcelled:
@Zerotask Aggressive compression tries much harder to compress the spawnpoints. The only benefit is possibly a lower number of scans, and thus less stress on Niantic. But it takes a lot of time to do aggressive compression and doesn't produce all that much better of a compression, which is why I left it off by default.

Zerotask:
i saw my load average jumping from 1,5 to 10,0 with Default=True ^^

justinkwaugh:
@Xcelled Heuristics can be fooled indeed.  But I would suggest that spawnpoint scanning is very much moving away from humanesque behavior due to the precision of the scan.  Exact coords at exact times.  I think you overestimate the difficulty in applying heuristics to a specific account over time. You don't compare backwards across time so much as you approach a threshold of detection on an ongoing basis, where outliers reset it.  Across multiple gets trickier yes, but if they are coming from the same IP it gets a little easier. 

I would suggest randomizing the location slightly, randomizing the time slightly, and introducing "fake" scans at random locations along the way. Even so the fact that the account never stops searching is a problem, though you could cycle through accounts on a semi-random basis.  Also the fact that the account never interacts with pokestops or tries to catch pokemon or interacts with a gym is pretty obvious. The list goes on and on.

Xcelled:
Spawnpoint scanning is about as easily detectable as walking in a hexagon. I was referring to how we've added delays and altitude and other similar 'real' factors.

Heuristics would be needed to apply to all accounts all the time to catch anything using a threshold method. That entails a significantly increased amount of CPU and memory usage - a small amount per account multiplied by every account. The resources required would grow in complexity with the heuristics. 

In that, we hold the upper hand. It's much easier (and less costly) for us to add heuristics-busting measures to the workers than it is for niantic to scan for them (you just named several, for example). It's an arms race Niantic can't win.

I've seen far too many companies take the wrong approach to bots/3rd party programs and sadly, the people who end up suffering from the countermeasures are the legitimate players. I'm not saying cheaters should be allowed or anything, but an arms race with them is not the answer.

justsomedudeonthenet:
Would it make sense to just have this as a separate script in /Tools and process the spawns.json file once and save it to a new spawns.json file?

Xcelled:
@justsomedudeonthenet Possibly for huge spawn files, but it took 100ms to process 260, some of which was probably spent doing string interpolation for the logger. For most use cases it doesn't seem worth it to me to make someone go through extra steps that don't take much time and can be done automatically.

An actual spawnpoint optimizer would have a great deal more work to do than this, and that could go in the tools directory.

TBTerra:
Infernoman wont be happy, (he has 34k spawns, n^2 scaling ==>1708sec = nearly half an hour to compress spawns)

having said that, once proper spawn clustering is in, you can compress each cluster individually which should solve the scaling problem somwhat

Xcelled:
That's only n^2 for aggressive compress. It should be closer to n for normal... or he can just turn it off ^^

TBTerra:
@Xcelled i do believe its n^2 for compress, and n^3 for aggressive compress (technically its n^2-2n+1 and n^3-3n^2-3n-1 respectively)

TBTerra:
thinking about it, shouldn't aggressive compress have an early out, if a step of compression made no difference (as all subsequent compresses would also make no difference)

Infernoman:
I can do some testing later today and see how it reacts, I do plan on increasing the number of spawnpoint's so this would be a really interesting feature. BUT does it justify the startup wait time... not so much.

Xcelled:
This isn't made for serious spawnpoints scanners, who will hopefully run their own, much more agressive compressor/cluster on their json file. This is meant for the average user who's scanning a generally small area and can easily afford a second or two of extra time to reduce the workload going forwards.

There are ways I could optomize the code to run faster, but i doubt 95% of users would notice such optimizations.

justinkwaugh:
I still think this has merit, but it would need some tweaking to allow for gps jitter to be used

chuyskywalker:
Since jitter has a fixed radius of 10m, it should be sooooorta easy to reduce the compression radius to 65m (am I thinking right on that?)

Xcelled:
Yes, just change 70 to 65 in the code.

MonjoNL:
There should be an option to export the compressed json so you only have to compress it once and use the compressed file without the need to compress it again.

MonjoNL:
I created a new file named compress.py that compresses my json and stores it using a simplified part of this PR. I can run it as much as I desire until the result is working for me.
It has hardcoded filenames etc. but you get the point. Just replace the filenames and run python compress.py to compress your json.

#!/usr/bin/python
# -*- coding: utf-8 -*-

import logging
import json
import os
import time
import geopy.distance as geopy_distance

from operator import itemgetter
from queue import Queue, Empty
from collections import deque


def timeDif(a, b):
    dif = a - b
    if (dif < -1800):
        dif += 3600
    if (dif > 1800):
        dif -= 3600
    return dif

def compress_spawnpoints(spawns, threshold, early_cutoff):
    compressed = []
    spawns = deque(spawns)  # duplicate so we can modify it
    while len(spawns) != 0:
        spawn = spawns.popleft()
        compressed.append(spawn)
        if len(compressed) >= early_cutoff:
            return None
        redundant = []
        for sp_to_check in spawns:
            if timeDif(spawn['time'], sp_to_check['time']) > threshold:
                break
            dist = geopy_distance.distance((spawn['lat'], spawn['lng']), (sp_to_check['lat'], sp_to_check['lng'])).meters
            if dist < 70:
                redundant.append(sp_to_check)
        for x in redundant:
            spawns.remove(x)
    return compressed


if os.path.isfile('Spawns.json'):  # if the spawns file exists use it
    try:
        with open('Spawns.json') as file:
            try:
                spawns = json.load(file)
            except ValueError:
                log.error(args.spawnpoint_scanning + " is not valid")
            file.close()
    except IOError:
        log.error("Error opening " + args.spawnpoint_scanning)

    spawns.sort(key=itemgetter('time'), reverse=True)
    compressed = compress_spawnpoints(spawns, 120, len(spawns))
    if compressed is not None:
        compressed.sort(key=itemgetter('time'))
        with open('Compressed.json', 'w+') as file:
            file.write(json.dumps(compressed))
            file.close()

jchristi:
I've resolved merge conflicts and resubmitted as #1104 but needs testing

Xcelled:
I'll close this, then.

PR #770, opened 2016-08-17
setup.bat now won't set paths if already set, now also installs npm

Acidvenom73:
## Description

Updated setup.bat to only set paths if not already set. It now also installs npm, and gives a bit more user output.
## Motivation and Context

If the setup was used more than once on a machine it would fill the PATH variable with several copies of the same path. It installs npm because of new dependencies.

## How Has This Been Tested?

Tested with PATH already set and not aswell as with default path to Python and using a custom path.

## Types of changes

## Checklist:

### DIFF
--- a/Tools/Easy-Setup/setup.bat
+++ b/Tools/Easy-Setup/setup.bat
@@ -36,20 +36,20 @@ set /p PATH2= Specify Python path:
 )
 
 
-echo ;%PATH%; | find /C /I "%PATH2%" > temp.txt
-find /c "1" temp.txt
-if %errorlevel% equ 1 goto Notfound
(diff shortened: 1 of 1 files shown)


BaIthamel:
WTF is line 39? echo.txt?

Acidvenom73:
@BaIthamel 

Line 39:
echo ;%PATH%; | find /C /I "%PATH2%" > echo.txt
It echoes the full PATH variable and searches for the python path in there. If it finds it, it makes echo.txt containing the number 1, otherwise the number 0. Then it continues from there depending on what echo.txt contains. After it's done, echo.txt is deleted.

FrostTheFox:
gonna go on the limb here and guess it still works

PR #785, opened 2016-08-17
Minor Performance Fixes

xrobau:
## Description
This consists of two commits - the first slightly reorders the way the query is generated so it puts less load on the database.  The second disables pythons Garbage Collector when building lists. This has little effect on newer python versions, but a dramatic effect on older ones (eg, Ubuntu 14.04)

## Motivation and Context
This was written because of suboptimal queries and performance

## How Has This Been Tested?
Has been running in production for several weeks

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -92,10 +92,13 @@ def get_active(swLat, swLng, neLat, neLng):
...
                      .where((Pokemon.disappear_time > datetime.utcnow()) &
-                            (Pokemon.latitude >= swLat) &
-                            (Pokemon.longitude >= swLng) &
-                            (Pokemon.latitude <= neLat) &
-                            (Pokemon.longitude <= neLng))
+                            (((Pokemon.latitude >= swLat) &
+                              (Pokemon.longitude >= swLng) &
+                              (Pokemon.latitude <= neLat) &
+                              (Pokemon.longitude <= neLng)) |
+                             (Pokemon.pokemon_id << Pokemon.rare_pokemon)
```
xrobau: Whoops, that's for another PR, need to fix.

pogom/models.py
```diff
@@ -183,9 +203,17 @@ def get_appearances(cls, pokemon_id, last_appearance):
                  .order_by(Pokemon.disappear_time.asc())
                  .dicts()
                  )
+
+        # Performance: Disable the garbage collector prior to creating a (potentially) large dict with append()
+        gc.disable()
```
Xcelled: I'd take this out and just do a `return list(query)`. No sense in doing things the hard and slow way for no reason.

pogom/models.py
```diff
@@ -331,10 +365,16 @@ def get_gyms(swLat, swLng, neLat, neLng):
                             (Gym.longitude <= neLng))
                      .dicts())
 
+        # Performance: Disable the garbage collector prior to creating a (potentially) large dict with append()
+        gc.disable()
```
Xcelled: Likewise ^

pogom/models.py
```diff
@@ -358,10 +398,16 @@ def get_recent(swLat, swLng, neLat, neLng):
                         (ScannedLocation.longitude <= neLng))
                  .dicts())
 
+        # Disable the garbage collector prior to creating a (potentially) large dict with append()
+        gc.disable()
```
Xcelled: And again^

pogom/models.py
```diff
@@ -183,9 +203,17 @@ def get_appearances(cls, pokemon_id, last_appearance):
                  .order_by(Pokemon.disappear_time.asc())
                  .dicts()
                  )
+
+        # Performance: Disable the garbage collector prior to creating a (potentially) large dict with append()
+        gc.disable()
+
```
Infernoman: If im not mistaken this one too ^

Xcelled:
Github seems to be messing up the line comments. Anything that uses append with no modifications should be converted to `return list(query)`.

Infernoman:
wow are they ever. will be a +1 from me after the changes @Xcelled spoke about.

pogom/models.py
shivampaw: Aka the first one that was pointed out ^^

Infernoman:
@Xcelled the stats page is going to need to be handled different unless you leave the iteration and addition to a new array.

Xcelled:
@Infernoman I didn't mark any lines that do anything except `query = blah; result = []; for x in query: result.append(x); return result;` Blame github.

xrobau:
Thanks, @Xcelled, I didn't even think of just doing a list(). I'll redo this pull request today!

Romfrosk:
Can't wait until this gets properly added :) 
Any performance boost is welcomed :)

xrobau:
There we go, replaced with list() when I could, and it seemed noticably faster. I didn't do any performance testing against gc.disable()/gc.enable(), because either of them are faster than just a raw append()!

Romfrosk:
Would you need to rebuild or drop any tables for this to be implemented?

rehand:
I'd like to see this PR merged, would increase the overall performance on a Raspberry Pi 1st gen.

Xcelled:
@Romfrosk No.

@rehand +1 me too.

@chuyskywalker this is a small change that shouldn't negatively affect anyone. Can we get it merged quickly?

PR #786, opened 2016-08-17
Moved webhook to grequests for async so it doesn't slow down the scanner

sebastienvercammen:
## Description
Moved webhook to grequests for async so it doesn't slow down the scanner.

## Motivation and Context
Scanner shouldn't wait for a webhook request, it should just continue scanning. grequests uses gevent.

## How Has This Been Tested?
Tested on Win10 Py2.7.11 & Thunderfox tested as well.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -9,7 +9,7 @@ import json
 from datetime import datetime, timedelta
 import logging
 import shutil
-import requests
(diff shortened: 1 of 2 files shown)


FrostTheFox:
can confirm, did test +1

PR #804, opened 2016-08-18
GPS Jitter

jnovack:
## Description

Optionally apply a ±3m jitter (`-j`, `--jitter`) on both Long and Lat when scanning locations.  This calculates to be up to a "as-the-crow-flies" ±9m (avg. 0m) movement on your scanned location.

The added argument is OPTIONAL.  I know that's going to shake some sodas around here.

Goes great with https://github.com/PokemonGoMap/PokemonGo-Map/pull/666

## Motivation and Context

Any time the loop restarts, you are scanning the exact same places.  This applies a small jitter to simulate real time GPS variances.

## How Has This Been Tested?

Yes.  Docker before.  Docker after.

In the *After* screenshots, note:

* the additional `Coordinates` line (enabled with `--debug`).  This is the ACTUAL scanned location.
* to prevent drifting on loops, the queued location has NOT been modified.

Before:
image

After:
image

## Types of changes

## Checklist:
- [X] Killed a puppy (added a new argument)

### DIFF
--- a/example.py
+++ b/example.py
@@ -3,7 +3,6 @@
 # thunderfox : can we put an example.py in 2.0 that just prints "you're a fucking retard"
 # invisiblek wuz here
 # balthemel wuz also 'ere
-# am I doing this right??
 
 # dont make pr to fix pls
 
--- a/example.py
+++ b/example.py
@@ -3,6 +3,7 @@
 # thunderfox : can we put an example.py in 2.0 that just prints "you're a fucking retard"
 # invisiblek wuz here
 # balthemel wuz also 'ere
+# am I doing this right??
 
 # dont make pr to fix pls
 
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -23,6 +23,7 @@ import json
 import os
 import random
 import time
+import geopy
 import geopy.distance as geopy_distance
 
 from operator import itemgetter
@@ -102,6 +103,15 @@ def generate_location_steps(initial_loc, step_count, step_distance):
                     loc = get_new_coords(loc, xdist / 2, EAST)
                 yield (loc[0], loc[1], 0)
         ring += 1
+
+
+# Apply a location jitter
+def jitterLocation(location=None, maxMeters=10):
+    origin = geopy.Point(location[0], location[1])
+    b = random.randint(0, 360)
+    d = math.sqrt(random.random()) * (float(maxMeters) / 1000)
+    destination = geopy_distance.VincentyDistance(kilometers=d).destination(origin, b)
+    return (destination.latitude, destination.longitude, location[2])
 
 
 # gets the current time past the hour
@@ -451,8 +461,17 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     # Ok, let's get started -- check our login status
                     check_login(args, account, api, step_location)
 
+                    # create scan_location to send to the api based off of step_location
+                    if args.jitter:
+                        # if requested, apply jitter to scan_location
+                        scan_location = jitterLocation(step_location)
+                        log.debug("Jittered to: %f/%f/%f", scan_location[0], scan_location[1], scan_location[2])
+                    else:
+                        # Just use the original coordinates
+                        scan_location = step_location
+
                     # Make the actual request (finally!)
-                    response_dict = map_request(api, step_location)
+                    response_dict = map_request(api, scan_location)
 
                     # G'damnit, nothing back. Mark it up, sleep, carry on
                     if not response_dict:
@@ -542,8 +561,16 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
                             break
                         sleep_time = args.scan_delay * (1 + failed_total)
                         check_login(args, account, api, step_location)
+                        # create scan_location to send to the api based off of step_location
+                        if args.jitter:
+                            # if requested, apply jitter to scan_location
+                            scan_location = jitterLocation(step_location)
+                            log.debug("Jittered to: %f/%f/%f", scan_location[0], scan_location[1], scan_location[2])
+                        else:
+                            # Just use the original coordinates
+                            scan_location = step_location
                         # make the map request
-                        response_dict = map_request(api, step_location)
+                        response_dict = map_request(api, scan_location)
                         # check if got anything back
                         if not response_dict:
                             log.error('Search step %d area download failed, retyring request in %g seconds', step, sleep_time)
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -24,7 +24,7 @@ import os
 import random
 import time
 import geopy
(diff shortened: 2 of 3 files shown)


invisiblek:
yay, travis died for no apparent reason...

jnovack:
Yeah, I broke it. :(

How do I get it to re-check?  Do a dummy commit?

invisiblek:
Yea, pretty much. Can push something then forcepush it off the branch or something

invisiblek:
Does this end up storing these "jittered" scan locations in the database too then? That's going to break things for people that use the scanned-only stuff. Would it make sense to still store the "non-jittered" location and only jitter what we send to Niantic?

If we were to do that, I would +1 the idea of turning this on by default.

jnovack:
> Does this end up storing these "jittered" scan locations in the database too then?

No.  I apply the jitter after I pull the stored location.  It's ephemeral to the scan.

invisiblek:
What I'm saying is you're going to be calling parse_map with the step_location containing the jittered values and parse_map will store these value in the database. We should be storing the original values.

rrbailey89:
Im all about appearing more human and I agree with @invisiblek actual real location should be stored in database and only the "jittered" location sent to Niantic via the API.

pogom/search.py
```diff
@@ -340,8 +340,21 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
...
+                        # if requested, apply jitter to scan_location
+                        scan_location = (
+                                        step_location[0] + random.triangular(-0.00003, 0.00003, 0),
+                                        step_location[1] + random.triangular(-0.00003, 0.00003, 0),
+                                        step_location[2]
+                                        )
+                        log.debug("Jittered to: %f/%f/%f", step_location[0], step_location[1], step_location[2])
+                    else:
+                        # Just use the original coordinates
+                        scan_location = (step_location[0], step_location[1], step_location[2])
```
invisiblek: I'd probably just put scan_location = step_location above the if args.jitter and remove the else. I think flake8 will complain that scan_location is undefined or something

justinkwaugh:
step_location is what is passed into parse_map not scan_location, so it should be storing the real location. I'm not sure though why you would want to not store the actual location queried though. That way you can tell what actually happened in retrospect.

jnovack:
k.  All updates applied.

Jittered location only sent in request, not stored in DB.

LGTM!

pogom/search.py
```diff
@@ -340,8 +340,20 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
 
                     api.activate_signature(encryption_lib_path)
 
+                    # create scan_location to send to the api based off of step_location
+                    if args.jitter:
```
invisiblek: no love for search_worker_thread_ss ?

FrostTheFox:
pls no example.py executive only

pogom/search.py
```diff
@@ -340,8 +340,20 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
...
+                        # if requested, apply jitter to scan_location
+                        scan_location = (step_location[0] + random.triangular(-0.00003, 0.00003, 0),
+                                         step_location[1] + random.triangular(-0.00003, 0.00003, 0),
+                                         step_location[2]
+                                         )
```
chuyskywalker: There's a bit of code in the [fake api](https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/fake-pgo-api.py#L54) which utilizes some smarter lat/lng distancing calculations to randomly pick and even spread of points within a given radius of the provided lat/lng value.

example.py
```diff
@@ -3,6 +3,7 @@
 # thunderfox : can we put an example.py in 2.0 that just prints "you're a fucking retard"
 # invisiblek wuz here
 # balthemel wuz also 'ere
+# am I doing this right??
```
FrostTheFox: no, no.

pogom/search.py
jnovack: Would you like me to import that code, or re-port it over into search.py? I don't care either way, I just would like to get it merged.

**@Collaborators** can you guide me to how this should be solved?

* Thumbs Up - My code
* Party Emoji - Import from `fake-pgo-api`
* Love Emoji - Rewrite it in `search.py`

jnovack:
Rebased and added `jitterLocation()`

pogom/search.py
```diff
@@ -110,7 +110,7 @@ def jitterLocation(location=None, maxMeters=10):
     origin = geopy.Point(location[0], location[1])
     b = random.randint(0, 360)
     d = math.sqrt(random.random()) * (float(maxMeters) / 1000)
-    destination = geopy_distance.VincentyDistance(kilometers=d).destination(origin, b)
+    destination = geopy.distance.distance(kilometers=d).destination(origin, b)
```
jnovack: [`geopy.distance.distance` is Vincenty by default.](https://geopy.readthedocs.io/en/1.10.0/#module-geopy.distance)

pogom/search.py
```diff
@@ -542,8 +560,16 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
...
+                        if args.jitter:
+                            # if requested, apply jitter to scan_location
+                            scan_location = jitterLocation(step_location)
+                            log.debug("Jittered to: %f/%f/%f", scan_location[0], scan_location[1], scan_location[2])
+                        else:
+                            # Just use the original coordinates
+                            scan_location = step_location
                         # make the map request
-                        response_dict = map_request(api, step_location)
+                        response_dict = map_request(api, scan_location)
```
chuyskywalker: You should move the `if args.jitter` logic int the `map_request` method so we can avoid duplicate code. (Mostly cause these two overseer methods are spaghetti duplicate nightmares...)

jnovack: wouldn't I have to pass `args` or modify the call to pass `args.jitter`?  #PythonNewbieQuestion

justinkwaugh: Personally I see no reason to make it optional for -ss, there is no downside to doing it as it cannot cause any missed spawns.

justinkwaugh:
Personally I see no reason to make it optional for -ss, there is no downside to doing it as it cannot cause any missed spawns.

justsomedudeonthenet:
There is a reason for making it optional.

It could cause missed spawn points if used in conjunction with -ss Compression from https://github.com/PokemonGoMap/PokemonGo-Map/pull/765

You would only want to use one or the other, using both could put one of the compressed spawnpoint locations out of range of one of the points it is actually trying to scan.

jnovack:
k.  I think I'm ready for **awaiting-review**.

I've been running multiple sessions for hours now.

chuyskywalker:
LGTM

Worthy7:
So just to be clear, the point of this is simply to seem more human to nian?

PR #806, opened 2016-08-18
Output nice error when map download fails and remove worker after 3 f…

springjools:
## Description
i) Catch some exceptions in models.pyL385, raise AuthException. Exceptions needed to catch here because they are first triggered here

ii) In search.py:
- Check if response_dict is of type 'dict'
- Output account name if map download or parse fails
- Remove worker (account name) if repeatedly failing
- Log removed workers in loginerrors.log in root dir

## Motivation and Context
Users are confused with error messages such as "list indices must be integers, not str" etc and spam in #general or #help

## How Has This Been Tested?
Tested on my development branch

## Types of changes

## Checklist:

…ailed attemps. Log removed workers to file 'loginerrors.log'

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -18,6 +18,7 @@ from . import config
 from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, send_to_webhook
 from .transform import transform_from_wgs_to_gcj
 from .customLog import printPokemon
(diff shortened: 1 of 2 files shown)


Becold:
Do it output the username of the account when it fails?

springjools:
Yes!

springjools:
Files changed while I was making PR, will submit new and better.

PR #730, opened 2016-08-16
Decommission faulty accounts

Shibby7986:
Decommission faulty accounts

## Description
If an account produces errors replace it with a fresh account from an accounts buffer

## Motivation and Context
holes in map due to accounts no longer working

## How Has This Been Tested?
work in progress
Have tested the remove_account method and is working with manually passed variables

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #811, opened 2016-08-18
Output nice error when map download fails and remove worker after 3

springjools:
## Description
In search.py:

* Check if response_dict is of type 'dict'
* Output account name if map download or parse fails
* Remove worker (account name) if repeatedly failing
* Log removed workers in loginerrors.log in root dir

## Motivation and Context
Users are confused with error messages such as "list indices must be integers, not str" etc and spam in #general or #help

Example of loginerrors.txt: http://pastebin.com/qaPs4tXq
Example of log output: http://pastebin.com/1q3xm8GQ
## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

…ailed attemps. Log removed workers to file 'loginerrors.log'

### DIFF
--- a/docs/extras/mysql.md
+++ b/docs/extras/mysql.md
@@ -114,7 +114,12 @@ I'd also like to specifically credit Znuff2471 on discord for their great assist
 2. Login to your MySQL DB
    - mysql -p
    - Enter your password if you set one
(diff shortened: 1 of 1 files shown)


CtrlAltDefeat94:
This in combination with #730 would be fantastic :3

PR #812, opened 2016-08-18
Fix Scan delay between scan

edsonschlei:
Stop the thread for scan_delay between each scan.

## Description
Add a time.sleep(args.scan_delay) after each scan.

## Motivation and Context
Avoid the server block the user for too much requests.

## How Has This Been Tested?
Tested local

## Types of changes

## Checklist:

- correct scan delay

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -322,11 +322,6 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                         log.error('Search step %d went over max scan_retires; abandoning', step)
                         break
 
-                    # Increase sleep delay between each failed scan
-                    # By default scan_dela=5, scan_retries=5 so
-                    # We'd see timeouts of 5, 10, 15, 20, 25
-                    sleep_time = args.scan_delay * (1 + failed_total)
-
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -352,7 +352,9 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                         except KeyError:
                             log.exception('Search step %s map parsing failed, retrying request in %g seconds. Username: %s', step, sleep_time, account['username'])
                             failed_total += 1
-                    time.sleep(sleep_time)
+                            time.sleep(sleep_time)
+
+                time.sleep(args.scan_delay)
```
chuyskywalker: This is *directly* in conflict with the code right below here which attempts to ensure the worker only sends another loop at the worker's start time.

justsomedudeonthenet:
This adds a sleep inside the parse_lock, which we just recently managed to remove.  Sleeping inside the lock makes ALL workers hang.

edsonschlei:
The current implementation on the development branch is stopping correctly.

PR #817, opened 2016-08-18
Alternative Scanned Marker

MonjoNL:
Alternative scanned marker visuals

## Description
I didn't like the way the scanned markers are displayed when using the ss mode.
It was to crowded and to much color. I removed the stroke and changed the green->red coloring to a specified color that will fade away. The default color is now blue.
In preparation to give the scanned locations a color based on the worker thread I stored the hue value in the database. The idea is to give the hue value of 360 / number of threads * worker thread number to each scanned location. This will allow monitoring the individual threads.
This is particularly useful in combination of #787

## How Has This Been Tested?
On Mac OSX with Safari
Requires npm build to create map.min.js in order to work

schermafbeelding 2016-08-18 om 22 52 09

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -24,7 +24,7 @@ log = logging.getLogger(__name__)
 args = get_args()
 flaskDb = FlaskDB()
 
-db_schema_version = 5
+db_schema_version = 6
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -342,6 +342,7 @@ class ScannedLocation(BaseModel):
     latitude = DoubleField()
     longitude = DoubleField()
     last_modified = DateTimeField(index=True)
+    hue = IntegerField()
 
     class Meta:
         primary_key = CompositeKey('latitude', 'longitude')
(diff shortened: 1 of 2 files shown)


slideup-benni:
Why you changed the database and add a hue field? is there no other possible solution to get calculate this value?

MonjoNL:
@slideup-benni well you do need to store something for when you update the map.
And because the model doesn't (and shouldn't) know anything about the amount of workers I figured to store the hue value I can calculate in the search.py file and pass on with the parse_map function.
But since I am playing around with the search_overseer_thread_ss method to try to generate smart paths that do not jump across all spawns on top of compressing and delaying spawns I cannot upload this part yet.

But you are right, it should be optional. I will work on that later today.

MonjoNL:
I'm not so sure anymore this is interesting for somebody.
I have been testing with 2 workers and tis results already in a confetti mess.
It helps me with my optimizations but I think it isn't much of a added value for normal users.

I'm putting it on hold for now until it makes more sense or a lot of interest in this idea is present.

schermafbeelding 2016-08-19 om 11 43 53

primaxius:
declined - not up to base and monjo gave up on it

PR #828, opened 2016-08-19
Quick fix to delay being ignored after resume

tai-hun:
This is a quick patch for scan delay being ignored after extended period of search pause.  This needs to be fixed asap otherwise lots of people will get soft ban after resuming the search.. XD

*** I also tried to find a way to resume search gracefully.  Tried putting interval delays when search queue is being built in the first place, tried PR previously - hoped to get some comments on whether that's a good implementation or not - so that we could find a better way to stagger the threads like in the initial start.  What do you all think?

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -408,7 +408,6 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                 # Grab the next thing to search (when available)
                 status['message'] = "Waiting for item from queue"
                 step, step_location = search_items_queue.get()
-
                 status['message'] = "Searching at {},{}".format(step_location[0], step_location[1])
                 log.info('Search step %d beginning (queue size is %d)', step, search_items_queue.qsize())
 
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -402,15 +402,13 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
 
             api.activate_signature(encryption_lib_path)
 
-            # Get current time
-            loop_start_time = int(round(time.time() * 1000))
-
             # The forever loop for the searches
             while True:
 
                 # Grab the next thing to search (when available)
                 status['message'] = "Waiting for item from queue"
                 step, step_location = search_items_queue.get()
+
                 status['message'] = "Searching at {},{}".format(step_location[0], step_location[1])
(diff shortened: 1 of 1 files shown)


Cy4n1d3:
Like the last 2(?) PR's I merged this locally as it's absolutely crucial to prevent bans... thx.

Zentriert:
Search pause .. the one in the GUI? So this won't affect me if I'm only using -ss?

taihun81:
yea, -ss thread doesn't have pause implemented so I didn't touch -ss thread...

Zentriert:
I appreciate it a lot, I just wanted to make sure I wasn't at risk of being banned for this bug. Thank you for your time and contribution!

justsomedudeonthenet:
This, I believe, adds unneeded delays.  Just fixing the one loop_start_time line like in https://github.com/PokemonGoMap/PokemonGo-Map/pull/851 is sufficient.

J-rooft:
I confirm that this fixes #830

Westacular:
The problem with just setting `loop_start_time` like that is that you'll lose the staggered schedule for the threads — they can drift over time, and after a suspend/resume, they'll probably all be going off at the same time.

The way I've fixed this locally is to instead figure out when the _next_ iteration in the staggered schedule should happen, and sleep until then:

```
                 # If there's any time left between the start time and the time when we should be kicking off the next
                 # loop, hang out until its up.
-                sleep_delay_remaining = loop_start_time + (args.scan_delay * 1000) - int(round(time.time() * 1000))
+                current_time = int(round(time.time() * 1000))
+                iterations = (current_time - loop_start_time) // (args.scan_delay * 1000) + 1
+                sleep_delay_remaining = loop_start_time + iterations*(args.scan_delay * 1000) - current_time
                 if sleep_delay_remaining > 0:
-
+                    loop_start_time += iterations * args.scan_delay * 1000
+                else:
+                    log.error('Negative sleep_delay_remaining? start=%g, time=%g, iterations=%g, sleep_delay_remaining=%g', loop_start_time, current_time, iterations, sleep_delay_remaining)
+                    loop_start_time = int(round(time.time() * 1000))
```

(the `else` clause is just there as an error check mostly to make sure my math was right.)

tai-hun:
That looks very nice 👍 I guess all it needs is adding staggered delay on the very first searches after resume...

tai-hun:
this now is closed with new commit #901  :dancer:

PR #851, opened 2016-08-19
Fix incorrect search delay in hex grid worker

### LINKED ISSUE #830: scan delay is not working as expected on slow networks
## Expected Behavior

Setting -sd to 10 should scan with 10 seconds delay every time for each worker.

## Current Behavior

When initial connection takes some time, e.g. 30 seconds, then 3 scans happen immediately for each worker.

1. Find some slow network
2. Use one account to see the result
3. Set sd to some value
4. Run server
5. See the result

## Context

Sometimes initial connection takes up to one minute and after it scanner goes to 6 locations immediately. I guess that it looks suspicious and could result in soft ban.

## Your Environment

* Version used: Latest dev
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): Ubuntu 16.04.1 LTS

## Logs
Logs show that scanner goes to 4 locations at the same time for search_worker_0 even though I have -sd 10
```
2016-08-19 00:57:28,526 [      MainThread][        models][    INFO] Connecting to local SQLite database
2016-08-19 00:57:28,532 [   search_thread][        search][    INFO] Search overseer starting
2016-08-19 00:57:28,533 [   search_thread][        search][    INFO] Starting search worker threads
2016-08-19 00:57:28,533 [   search_thread][        search][    INFO] New location caught, moving search grid
2016-08-19 00:57:28,534 [   search_thread][        search][    INFO] Scan Distance is 0.07 km
2016-08-19 00:57:28,542 [ search_worker_0][        search][    INFO] Search step 1 beginning (queue size is 36)
2016-08-19 00:57:28,542 [ search_worker_0][      auth_ptc][    INFO] PTC User Login for: XXX
2016-08-19 00:57:45,038 [ search_worker_0][      auth_ptc][    INFO] PTC User Login successful.
2016-08-19 00:57:45,039 [ search_worker_0][      auth_ptc][    INFO] Request PTC Access Token...
2016-08-19 00:57:45,706 [ search_worker_0][      auth_ptc][    INFO] PTC Access Token successfully retrieved.
2016-08-19 00:58:07,401 [ search_worker_0][        models][    INFO] Upserted 3 pokemon, 68 pokestops, and 10 gyms
2016-08-19 00:58:07,402 [ search_worker_0][        search][    INFO] Search step 2 beginning (queue size is 35)
2016-08-19 00:58:07,821 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-19 00:58:07,821 [ search_worker_0][        search][    INFO] Search step 3 beginning (queue size is 34)
2016-08-19 00:58:08,217 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-19 00:58:08,218 [ search_worker_0][        search][    INFO] Search step 4 beginning (queue size is 33)
2016-08-19 00:58:08,613 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-19 00:58:08,613 [ search_worker_0][        search][    INFO] Search step 5 beginning (queue size is 32)
2016-08-19 00:58:09,337 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 0 pokestops, and 0 gyms
2016-08-19 00:58:18,551 [ search_worker_0][        search][    INFO] Search step 6 beginning (queue size is 31)
2016-08-19 00:58:19,476 [ search_worker_0][        models][    INFO] Upserted 2 pokemon, 64 pokestops, and 8 gyms
```

justsomedudeonthenet:
## Description
Fix the scan delay in the hex grid worker 

Right now, it assumes that each loop runs for exactly scan_delay seconds, so it just increments the loop_start_time by scan_delay.

Instead, calculate when the previous loop actually finished.  If a loop gets delayed by a few extra seconds, the next loop will still sleep the proper amount of time.

## Motivation and Context
Fixes #830 

## How Has This Been Tested?
Tested briefly on linux.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -487,7 +487,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     status['message'] = "Waiting {} seconds for scan delay".format(sleep_delay_remaining / 1000)
                     time.sleep(sleep_delay_remaining / 1000)
 
-                loop_start_time += args.scan_delay * 1000
+                loop_start_time = int(round(time.time() * 100))
 
         # catch any process exceptions, log them, and continue the thread
         except Exception as e:


taihun81:
Not only need to reset the loop start to current time but it needs to check if sleep delay remaining is < 0.   Otherwise, if delay longer than scan_delay happens(such as extended period of pause, unexpected delay happened in login/search,) loop will run without any delays.   Please see my PR - #828

justsomedudeonthenet:
@taihun81 I don't believe your extra sleep is necessary there.  If sleep_delay_remaining is < 0, then this loop already took long enough, and there is no need to sleep more.  We go slow enough without adding unnecessary delays.

The problem is that sleep_delay_remaining is calculated wrong, because loop_start_time is calculated wrong.  Just this one line fix should prevent it from searching too fast, even after pausing.

With the existing method, if you pause for 50*scan_delay, it would run the next 50 scans with no delay at all.  Just updating loop_start_time to the current time is enough to prevent that.

taihun81:
Let say scan delay is 10s, starting at 0, loop start updates to current time, then you pause the loop for 50s.  scan delay remaining will be calculated 0 + 10 - 50, ends up -40, then loop will pass with delay.

taihun81:
I think the issue is the loop start being updated at the end of loop.  Pause happens when the loops goes back to start and grabbing the queue.  So if the pause is for > scan delay, by the time search is resumed, loop end delay check will ignore delay, making second loop bursting out the searches.  Correct me if I'm wrong... I thought it's pretty simply in the beginning but now I'm keep thinking it over, I'm starting to confuse myself...  XD

taihun81:
or we should just move loop = current time to right after grabbing the queue.

justsomedudeonthenet:
> Let say scan delay is 10s, loop start updates to current time, let say 0, then you pause the loop for 50s. scan delay remaining will be calculated 0 + 10 - (current time + 50) ends up being -40, loop will pass without delay.

Which is fine, the workers have already been sleeping for 50 seconds AFTER finishing their last scan and waiting the appropriate scan delay.  So even if you flipped scanning on and off real fast, it would always wait at least scan delay + a bit.

taihun81:
Delay happens at the end of loop, after API call.  If loop thinks it has slept enough and pass onto api call without delay after coming back from pause, the second loop will start without delays.

justinkwaugh:
I have certainly seen this as well.  If you look at the ss worker thread it does it more rationally I think

pogom/search.py
```diff
@@ -487,7 +487,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     status['message'] = "Waiting {} seconds for scan delay".format(sleep_delay_remaining / 1000)
                     time.sleep(sleep_delay_remaining / 1000)
 
-                loop_start_time += args.scan_delay * 1000
+                loop_start_time = int(round(time.time() * 100))
```
justinkwaugh: Why not just rename it to last search time, and set it when you decide to actually search?  Then it would be correct and also make sense, and wouldn't need setting here.

justsomedudeonthenet:
Closing - fixed in #901

PR #864, opened 2016-08-20
More flexibility with spawnpoint import file

khag7:
## Description
Allow the spawnpoint import file to be a CSV. Allow it to not include lat & lng. Allow the provided filename to be a relative path.

## Motivation and Context
Data portability is useful. Some users have a much easier time dumping some database columns, editing in excel, and saving as CSV. This change will allow those users to skip the step of converting their data to json. Furthermore, we don't need users to provide their lat & lng with each spawnpoint. The spawnpoint_id already encodes that information. We just need spawnpoint_id and time (seconds past the top of the hour). Two columns. That's it!

## How Has This Been Tested?
JSON imports still work. CSV imports work. Tested with both JSON and CSV files of 55,000 spawnpoints in length. About 5 seconds is added to startup for a file this size to convert spawnpoints (no lat & lng provided) to lat & lng.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -388,19 +388,20 @@ def search_overseer_thread_ss(args, new_location_queue, pause_bit, encryption_li
         latlng = LatLng.from_point(Cell(CellId(int('{}00000'.format(spawn['spawnpoint_id']),16))).get_center())
         spawn['lat'] = latlng.lat().degrees
         spawn['lng'] = latlng.lng().degrees
+        spawn['time'] = int(spawn['time'])
     log.info('sorting spawnpoints based on time')
     spawns.sort(key=itemgetter('time'))
     log.info('Total of %d spawns to track', len(spawns))
     # find the inital location (spawn thats 60sec old)
     pos = SbSearch(spawns, (curSec() + 3540) % 3600)
     while True:
-        while timeDif(curSec(), int(spawns[pos]['time'])) < 60:
+        while timeDif(curSec(), spawns[pos]['time']) < 60:
             threadStatus['Overseer']['message'] = "Waiting for spawnpoints {} of {} to spawn at {}".format(pos, len(spawns), spawns[pos]['time'])
             time.sleep(1)
         # make location with a dummy height (seems to be more reliable than 0 height)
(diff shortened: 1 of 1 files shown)


justsomedudeonthenet:
This appears to ignore lat and lng even if provided in the file?

That is unacceptable.

My spawns.json has no spawnpoint IDs, just lat/lng.  People using any of the spawnpoint compression or optimization features of other PRs and tools will have the same issue - we are scanning a specific point, not necessarily a spawnpoint.

I don't particularly see the need for this.  Converting a csv to json is very easy.  But as long as it is not breaking anything else it would be ok.

khag7:
You are correct. And I wasn't sure how that might play. I guess the smarter call here would be to check if lat & lng provided, if not, get that from spawnpoint ID.

My end goal here is to improve data portability. Immagine someone new decides they want to run a scanner, they download our software. Do they really want to do beehive scans for 2 days to collect a large volume of spawnpoints in their area before they can get into spawnpoint scanning? No! Of course not. They want to go to pokemonspawnpoints.com and say "i want all points within 10 miles of my home address and 5 miles of my work address". Boom! They get a flat file with 60 thousand spawnpoints and what time those points should be checked. No need to waste bandwidth with lat and lng if that data is already in the spawnpoint_id. But more importantly is that excel can manipulate CSV and not JSON. I've helped about a dozen friends locally get this up and running. Some of the most challenging things has been sharing spawnpoint data between us because some people used different mapping programs initially. You know what every database software supports well? **Export to CSV.**


I didn't take spawnpoint compression into consideration. Thanks for the advice, I'll figure it out.

justsomedudeonthenet:
Oh I do understand why people want to work with csv files.

But wouldn't it be simpler (and easier to maintain) if you just had a csv_to_json.py script in /Tools?

Do whatever data manipulation you need as csv, then run the script to spit out a spawns.json file.

justsomedudeonthenet:
Ideally, such a script would accept a file with a mixture of spawnpoint id, lat/lng, and both provided.  Use the lat/lng if it is there (to allow for scanning outside the exact spawnpoint for compression) and calculate it if it is not provided.

Even better, make it able to convert json to csv as well - so you can take your json file and edit it in excel again.

primaxius:
Fails and WIP

PR #866, opened 2016-08-20
Auto update notifications with distance & time, add action buttons, move notification logic to a service worker file

khag7:
## Description
Notifications now have distance (from your currentlocation) and time til despawn. This is updated every 20 seconds. So as you move closer, the distance gets less. You can use it like a radar. OR you can just click the action button to initiate navigation immediately.

In addition I moved notification logic to a service worker file. Modern browsers handle notifications generated by a service worker better than notifications served directly from the page. Furthermore, once the logic in the serviceworker file is proven (I think it is, but happy to have others test it) we can use the serviceworker code to serve notifications from the server to the user's device without the need for the browser to be open. This will come as a separate PR, but the ability to do that relies on the notification coming from the serviceworker (not the browser session itself aka map.js) 

## Motivation and Context
Yes, we have webhooks. This is not the same. Notifications that come from our browser aren't sent in the best way possible. Serviceworkers exist to provide a way to send notifications to a users device regardless if they have their browser open or not. This doesn't actually change much, mostly just moving the code from map.js to sw.js, but in future this will be necessary to initiate notifications from the server to the user without an open browser session. Additionally, I did improve notifications by adding "action" buttons. One such action button is "navigate" which, when clicked, leads to navigational directions starting immediately on the user's device. No need to open the map, find the spawn, click it, and click "get directions". Just one step. 

## How Has This Been Tested?
Drove around and used it myself for a few days. Works nice for me.

notifications

## Types of changes

## Checklist:

### DIFF
--- a/docs/basic-install/linux.rst
+++ b/docs/basic-install/linux.rst
@@ -50,9 +50,9 @@ If your output looks as above, you can proceed with installation:
 
 .. code-block:: bash
 
-	pip install -r requirements.txt
-	npm install
-	npm install -g grunt-cli
+	sudo -H pip install -r requirements.txt
+	sudo npm install
+	sudo npm install -g grunt-cli
 	grunt build
 
 troubleshooting:
--- a/docs/basic-install/linux.rst
+++ b/docs/basic-install/linux.rst
@@ -53,7 +53,7 @@ If your output looks as above, you can proceed with installation:
 	sudo -H pip install -r requirements.txt
 	sudo npm install
 	sudo npm install -g grunt-cli
-	grunt build
+	sudo grunt build
 
 troubleshooting:
 	
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -4,7 +4,7 @@
 import calendar
 import logging
 
-from flask import Flask, jsonify, render_template, request
+from flask import Flask, jsonify, render_template, request, send_from_directory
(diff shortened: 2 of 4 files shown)


khag7:
Yeah, I know I didn't pass build check, code style isn't my strong suit, but if someone wants to take a look anyway and let me know what they think as far as the code function, that would be nice. Suggestions appreciated.

Becold:
Run this inside the folder : `npm run lint` ;)

khag7:
thanks.

<-- new guy

Becold:
Not a problem ! We all begin by this way

jnovack:
You don't "need someone else to take a look at it".

Travis tells you what's wrong.   https://travis-ci.org/PokemonGoMap/PokemonGo-Map/builds/153734635#L2066

khag7:
@jnovack I wasn't referring to the errors in Travis, I was referring to the functionality/feature addition. I know the code needs cleaned up, I'll take care of it. But I also was wondering what others thought. For example, chrome only shows the first two actions attached to a notification. I opted for "Navigate" and "Exclude" but there's a bunch of other things we could do there. Maybe someone else has a suggestion or preference. I guess I'm just the kind of guy that likes hearing others feedback!

primaxius:
Fails and WIP

Issue #862, opened 2016-08-20
We're making way too many pokéstop/gym upserts and causing things to break


gabrielsoldani:
## Expected Behavior

Pokéstops/gyms should be more conservatively upserted.
## Current Behavior

Even when pulse_radius is set to 70m in search.py, we call get_cells_id with a 1000m radius. The result is every map request returns pokestóps and gyms outside the 70m radius. This is fine and consistent with how the game actually works, but it means every time we scan two locations next to each other, we insert lots of redundant data in the database.

Since #582 was fixed this actually causes an issue on certain occasions: with many db-threads, we may try to REPLACE pokestops we're already REPLACE'ing in another db-thread, causing a ton of mysql deadlocks and filling all the db upsert queues.
## Possible Solution

We call get_cells_id with the appropriate pulse_radius (70m for pokémon, 900m pokéstop/gym-only since #748 was fixed). This is good because Niantic sends less data to us, but is bad because it's inconsistent with how the official client works.

I've implemented this solution and it seems to work fine.
## Steps to Reproduce (for bugs)
1. Run the map.
## Your Environment

Particularly, to run into the deadlock issue, I was running 5 instances with 15 accounts each on one server (-st 10 -sd 10 --db-threads 10), and a database on a different server with about ~150ms latency between the searchers and the database server.

justsomedudeonthenet:
Does this happen with sqlite, mysql or both?  Presumably you are using mysql since you said its on a different server.

One thing that would help is to stop using 'REPLACE INTO'.  It causes rows to be deleted and readded, which causes all the indexes to be recalculated.  Its a fairly expensive SQL command to use.

gabrielsoldani:
I'm using MySQL.

gabrielsoldani:
Woops

chuyskywalker:
Protip: If you run the main app with `-nk -ng` and a second instance with `-np` the first will scan in 70m steps for just pokemons, and the second will scan in 900m steps for just gyms pokestops.

Second, the db will only deadlock like that if running multiple threads (`--db-threads`) -- maaaybe you could turn that back to 1 for a bit. Otherwise we'll have to add a lock.

It's also the ugly, ugly `REPLACE INTO` rearing its head. If that was `INSERT ... ON DUPLICATE KEY UPDATE` this would be a non-issue. :/

hemsleyk:
+1 since this is #748 aware

c133:
[...]

PR #867, opened 2016-08-20
Limit map_request to cells within step_distance

gabrielsoldani:
## Description
This makes the searchers only request cells within step_distance (70m or 900m, depending on whether we're scanning for pokémon or not).

## Motivation and Context
Previously, we were requesting cells 1000m around each scanned location, even though we're only interested in the 70m around them. This caused a lot of redundant information being requested and received, putting unneeded strain on both ours and Niantic's servers. With enough --db-threads, this also caused MySQL deadlocks when the same pokéstop was updated in two threads at the same time, eventually filling up the db queues (issue #862).

This change significantly reduces the data transferred between workers and Niantic's servers and significantly reduces duplicate database upserts. 

As an added bonus, we no longer get stray pokéstops/gyms outside our scanned locations.

## How Has This Been Tested?
I've tested it on my personal map with beehive.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -339,7 +339,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
             for step, step_location in enumerate(locations, 1):
                 log.debug('Queueing step %d @ %f/%f/%f', step, step_location[0], step_location[1], step_location[2])
                 threadStatus['Overseer']['message'] = "Queuing next step"
-                search_args = (step, step_location)
+                search_args = (step, step_location, step_distance)
                 search_items_queue.put(search_args)
         else:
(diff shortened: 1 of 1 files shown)


jnovack:
Requesting only 1 cell might be a bot-condition.

Consider dropping all non-current cells from processing?

LordNeoZ:
@jnovack What about only requesting the info every x random scanned cells? For st 10 and lower it shouldn't make a difference between that and scanning requesting on every cell.

justj0sh:
Do you mind updating this for the latest develop?

primaxius:
Declined - It's a feature to be able to update the gym/pokestop when a scan happens further away.

PR #868, opened 2016-08-20
Tool to minimize overlapped spawnpoint scans

sf302:
Tool that works with spawnpoint data from MySQL-only databases to cover as many spawnpoints as possible with the fewest scans. Uses heatmaps surrounding each spawnpoint to find a point with the best coverage.

May trip teleport detection with large enough step sizes unless the order in the CSV is optimized either by hand or by something that solves traveling salesman problem and accounts for average speed.

## Description
Added script to Tools directory. Also added an option in utils.py to pass csv path, and added a section in search.py to replace generated locations with data loaded from csv.

## Motivation and Context
Drastically reduces number of scans in spawnpoint scanning modes. For the two locations I was personally running this with, 56 scans covered 204 spawnpoints and 50 scans covered 167 spawnpoints, respectively.

## How Has This Been Tested?
Originally wrote csv file input and this script (in PHP) for personal use weeks ago, ported over to Python today.

optimized

## Types of changes

## Checklist:

### DIFF
--- a/Tools/Spawnpoint-Scan-Optimizer/README.md
+++ b/Tools/Spawnpoint-Scan-Optimizer/README.md
@@ -28,5 +28,6 @@ create function haversine(lat1 double, lng1 double, lat2 double, lng2 double) re
     Full coverage required 56 scan locations
     Cleaning up database
     Coverage approximation via heatmap took 381.390 seconds
+    Optimized scan coordinates saved to spawns_17.627224_138.01003_700m_100s.csv
     $
 
--- a/Tools/Spawnpoint-Scan-Optimizer/README.md
+++ b/Tools/Spawnpoint-Scan-Optimizer/README.md
@@ -12,6 +12,12 @@ Pass the output CSV to runserver.py using the --csv flag to switch to optimized 
 
 Requires numpy to be installed in addition to upstream requirements
 
+Needs the following function created in MySQL as well:
+
+```sql
+create function haversine(lat1 double, lng1 double, lat2 double, lng2 double) returns double deterministic return 12756274 * asin(sqrt(0.5 - cos((lat2 - lat1) * 0.017453292519943295) / 2 + cos(lat1 * 0.017453292519943295) * cos(lat2 * 0.017453292519943295) * (1 - cos((lng2 - lng1) * 0.017453292519943295)) / 2));
+```
+
 ## Example
 
     $ python spawncover.py -lat 17.6272249 -lng 138.0100306 -st 5 -sl 100
--- a/Tools/Spawnpoint-Scan-Optimizer/README.md
+++ b/Tools/Spawnpoint-Scan-Optimizer/README.md
@@ -1,0 +1,26 @@
+# Spawnpoint Optimizer
+
+MySQL-only scan optimizer that uses 40x40 slice heatmaps surrounding each spawnpoint to determine how to cover all spawnpoints within step radius of center with as few scans as possible.
+
+Will potentially cause teleporting unless someone wants to solve the traveling salesman problem for the dataset that this generates.
+
+Pass the output CSV to runserver.py using the --csv flag to switch to optimized spawnpoint instead of spiral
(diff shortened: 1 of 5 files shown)


hklcf:
:+1:

peter-bonanni:
I tested this and the tool works great however ~75% of my accounts I used in the test were softbanned for teleportation so I think it needs some work if being used with multiple accounts. The use case tested is a 2.8km radius with 2,750 spawn points to do an entire city with 300+ accounts through any of ~600 rotating IPs. 

As a data point with 10 heatmap slices it took ~1.5 hours to do the csv generation.

sf302:
>As a data point with 10 heatmap slices it took ~1.5 hours to do the csv generation.

Ick that's painful. Not entirely unexpected though since it's a bruteforce approximation for a set cover problem across thousands of points.

For such a large step radius, it might be worthwhile to use the beehive generator tool to determine your centers using a small leapsize (3 or so?) and then running this tool for each location. You'd then pass the CSV for each beehive center to its respective worker to optimize the scans for each hex in that grid.

JWess:
What do you think about his idea as an alternative:

Iterate through each spawn point. Only scan for that point if it does not fall inside a previous scan's radius. This would not be nearly as optimized as your solution, but it would work for any size set of spawn points, and it would not require any up-front computation. It would prevent what is seen in the attached screenshot.
image

averylarry:
Are you modifying beehive scannning, spawnpoint scanning, or both?

sf302:
Neither. As it stands this just adds a third option. The enhancements made to spawnpoint scanning would have created incompatibilities had I integrated into that.



On Thu, Aug 25, 2016 at 9:37 PM -0700, "averylarry" <notifications@github.com<mailto:notifications@github.com>> wrote:


Are you modifying beehive scannning, spawnpoint scanning, or both?

-

averylarry:
This would not account for spawn time, right?  It's a way of scanning fewer locations but you still cycle through them on a loop.  A better version of --spawnpoints-only.

JWess -- that looks like full spawnpoint scanning -ss.  The reason for the overlaps is that spawnpoint scanning scans each individual spawnpoint WHEN it spawns.  See PR #765 as a first step to reducing spawnpoint scans and there are several PRs talking about clustering and generally optimizing the spawnpoint scanning model.

cecca00:
this may be very useful for gym scanning

cecca00:
Hello, I tried running the script on gym locations, changing input table from pokemon to gym and splat from 70 to 900, and running `python spawncovergym.py --lat 46.1319 --lng  11.22716 -st 3 --slices 100` but I get bad results:
bad
good
red dots are the gyms, yellow circles the output of the script, green circles what i was expecting

gabrielsoldani:
@cecca00 consider modifying the script from #990 which was merged into the develop branch.

cecca00:
@gabrielsoldani this the result running the script you mentioned, it's good but still not optimal, got 8 scan locations instead of 7. This one from @sf302 is supposed to give better results, any idea on why it's not working for me?
tool

gabrielsoldani:
@cecca00 Both algorithms find the local optimum, not the global optimum. #990  with spawn points and time windows the difference is normally very close (~3%), so it's good enough.

If you want the global optimum, shuffle the list of gyms and re-run the algorithm until you're satisfied with the results, or consider using a different algorithm (look into the [geoometric set cover problem](https://en.wikipedia.org/wiki/Geometric_set_cover_problem) if you want to dig deeper).

krummrey:
Why does this still show up as a PR?
Looks like it has been merged?

gabrielsoldani:
@krummrey #990, which serves the same purpose, was the one merged.

krummrey:
Shouldn't this be closed then? Just to make it easier to keep track of relevant PRs?

primaxius:
Closed as alternative was merged https://github.com/PokemonGoMap/PokemonGo-Map/pull/990

PR #874, opened 2016-08-20
Limit the radius of spawnpoint based scanning relative to the search marker and account for search marker movement (aka spawnscan while driving)

khag7:
## Description
Provides a command line argument which allows users to skip scanning a spawnpoint which is greater than X meters from their current search location. This is dynamic and keeps up with the search marker location as it moves.

## Motivation and Context
Spawnpoint based scanning which uses the database or an import file will scan ALL points provided, no matter where the search marker location is. If a user has spawnpoint locations covering a large area, they may travel within that area and wish to have the scan follow their location. This is possible with the regular scanning method, but not with spawnpoint based scanning. This allows a user to set a radius of X meters around them, and any spawnpoints in their list that are too far away will simply be skipped over. Additionally, since the user is moving while using this feature, it makes sense to scan points which might have spawned in the last 10 minutes. Previously this was limited to starting the scan at 1 minute old spawns. 10 minutes might not even be enough, we might want to adjust that further.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -18,6 +18,7 @@
 #no-pokestops:          # disables pokestop scanning (default false)
 #scan-delay:            # default 10
 #step-limit:            # default 12
+#spawnpoint-limit-radius # if using spawnpoint-scanning above, will skip any spawnpoints which are this distance or more away from the search location (default 0 which is unlimited)
 
 # Misc
 #gmaps-key:             # your Google Maps API key
--- a/docs/basic-install/linux.rst
+++ b/docs/basic-install/linux.rst
@@ -29,6 +29,8 @@ Debian's sources lists are out of date and will not fetch the correct versions o
 	ln -s /opt/python/bin/python2.7 /usr/bin/python2.7
 	ln -s /usr/bin/python2.7 /usr/bin/python
 	ln -s /usr/local/bin/python2.7 /usr/local/bin/python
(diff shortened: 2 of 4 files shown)


BaIthamel:
The current method does allow you to set the range of spawnpoints that are scanned. I'm not sure it coves movement yet but that's on the wanted list for us.

khag7:
That's only correct if a user is pulling spawnpoints from the database. If a spawnpoint json file is provided, the whole file is used no matter what. No radius check is done. This PR fixes that. Or maybe I'm doing it wrong?

In terms of defining the radius: I did consider using -st, but I opted for a new variable of -slr so that a user could store their preferred radius for each scanning method. I went back and forth trying to come up with the best solution and this is what I landed on. I'm not 100% opposed to using -st, we'd just take that value * 70 and use that. But its also kind of convenient to store in config.ini my preferred step size for "regular" scanning and also my preferred radius for spawnpoint scanning. Makes switching between the two easier (slightly).

justsomedudeonthenet:
Closing because this has merge conflicts and hasn't been updated in 17 days.  Also because using spawnpoints from the database is already sufficient to give the behaviour you want.

khag7:
Fine by me, I've stopped submitting bug fixes and feature additions here. I have a group of about 2 dozen people sharing spawnpoint json files and this PR allows us to use json files to scan while traveling to new areas. Not sure if anyone else needs that but we like it.

PR #876, opened 2016-08-20
Gitignore config ssl folder

khag7:
Some ssl configurations require intermediate certificates or multiple domains connecting to the same installation. It's easier to have an ssl folder than to assume everyone will have two files named ssl.crt and ssl.key

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ _build_html
 beehive.*
 config/ssl.cert
 config/ssl.key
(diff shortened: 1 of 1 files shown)


Becold:
I think https://github.com/PokemonGoMap/PokemonGo-Map/pull/845 is better

khag7:
ME TOO! I did not see that. That's awesome.

Issue #847, opened 2016-08-19
0 upsert after a re-login (with -ss)


Becold:
## Expected Behavior
Getting pokestop or gyms after a re-login.

## Current Behavior
Got 0/0/0.

## Possible Solution
Maybe flag the spawnpoints and retry it 30 seconds after. If again 0/0/0, flag the spawnpoints + accounts somewhere (in an output file).
Maybe the spawnpoints changed.

## Steps to Reproduce (for bugs)
Run this script : 
`python runserver.py -os -l 'myass'  `
`python runserver.py -ns -ss spawnpoints.json -l 'myass' -sd 15 -ld 11 --db-max_connections 100`
And wait until the script need to re-login the accounts.

My logs : http://puu.sh/qH1fS/d1980dc03b.png

## Context
Guess it doesnt scan pokemons properly.

## Your Environment
* Version used: My last commit is 4a27fd01c2e52809c96adc94ae63cc12a073249a
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): Linux debian 3.16.0-4-686-pae #1 SMP Debian 3.16.7-ckt25-2+deb8u3 (2016-07-02) i686 GNU/Linux

Draag:
Please don't close this or merge it with the other one that mine got swept into, this issue is specifically after RE-authenticating. It's not the same as the other one which was producing 0/0/0 after initially starting the server.

pyr0ball:
I observed similar behavior like this last week, but haven't had a chance to trace the cause yet. Could still be configuration issue, but will need further investigation.

mda8:
had a similar issue and it got merged with #462 too. See #821

BaIthamel:
I'm so confident they're related, I'm pretty sure we don't reauthorize yet we just log in again, so this would be the same as errors after logging in the first time.

Issue #683, opened 2016-08-15
Scan delay ignored after search step download failure


g5bEM6RpSbQF:
When the search_worker reports a 'Search step xx download failure' it waits to retry, then tries to play "catch up" by ignoring the scan delay until it is caught up.  This "catch up" behavior almost always triggers the "Request throttled by server... slow down man".  This makes it easy for Niantic to find software.  (I wouldn't put it past Niantic to intentionally force a download failure, to trigger this behavior, to detect us).

I use a scan delay of 10 seconds.  1 account per runserver instance.

Here's my logs:

2016-08-14 05:49:26,842 [ search_worker_0][        search][    INFO] Search step 10 beginning (queue size is 27)
2016-08-14 05:49:27,353 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 26 pokestops, and 2 gyms
2016-08-14 05:49:36,838 [ search_worker_0][        search][    INFO] Search step 11 beginning (queue size is 26)
2016-08-14 05:49:37,387 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 16 pokestops, and 1 gyms
2016-08-14 05:49:46,842 [ search_worker_0][        search][    INFO] Search step 12 beginning (queue size is 25)
2016-08-14 05:49:47,241 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 18 pokestops, and 2 gyms
2016-08-14 05:49:56,833 [ search_worker_0][        search][    INFO] Search step 13 beginning (queue size is 24)
2016-08-14 05:49:57,477 [ search_worker_0][        models][    INFO] Upserted 0 pokemon, 19 pokestops, and 3 gyms
2016-08-14 05:50:06,842 [ search_worker_0][        search][    INFO] Search step 14 beginning (queue size is 23)
2016-08-14 05:50:12,230 [ search_worker_0][        search][   ERROR] Search step 14 area download failed, retrying request in 10 seconds
2016-08-14 05:50:52,419 [ search_worker_0][        search][   ERROR] Search step 14 area download failed, retrying request in 20 seconds
2016-08-14 05:51:14,360 [ search_worker_0][        models][    INFO] Upserted 1 pokemon, 30 pokestops, and 3 gyms
2016-08-14 05:51:14,360 [ search_worker_0][        search][    INFO] Search step 15 beginning (queue size is 22)
[...]

Issue #607, opened 2016-08-13
Resuming after pausing search thread is causing a burst of requests


ilendemli:
If you pause the searcher and wait a few moments, then start it again will cause the workers to do a burst of request which can cause a ban or throttle the connection.

## Expected Behavior
Starting the workers with delays

## Current Behavior
Resumes all search threads immediatelly

## Steps to Reproduce (for bugs)
1. Run script
2. Let it log in, obtain tokens and search for pokemon
3. Pause it by using the switch in the webinterface
4. Wait a few moments, maybe go make tea or coffee
5. Resume the script using the switch
6. Check the script resuming the working thread instantly

## Context
Caused soft bans or got throttled (returns 0/0/0)

## Your Environment
* Version used: current develop branch, last commit: [f56f3f42efe550495492a0231865467ac96e4305](https://github.com/PokemonGoMap/PokemonGo-Map/commit/f56f3f42efe550495492a0231865467ac96e4305)
* Environment name and version (e.g. Python 2.7): Python 2.7.12
* Operating System and version (desktop or mobile): Ubuntu 16.04.1

jtp10181:
I think this is fixed by https://github.com/PokemonGoMap/PokemonGo-Map/pull/530
Works for me anyway

ProfessorPokes:
I am having a similar issue. If I stop the search partway through my scan then restart it it will gather the completed spawnpoints list and log in my workers, it then speed runs through a list of requests. Either in an effort to catch up to where it was before, or to catch up to the time it took to gather the spawn point locations.

Results in warnings to slow down, but there is nothing I can do.

SlasherKG:
I have the same issue.
I normally pause (by toggling the search switch in the browser).
Then when I get to my new location, I set the location and toggle it back on.
This seems to cause a flood of built up scans that should have been simply discarded instead of run.

Along this same thought, is there a way to start the server but not have it start a search (basically start paused)?

ultrafunkamsterdam:
Maybe that could be a reason for banning too.

For safety reasons, i'm now implementing a little modification to map.js that pauses automatically when moving the marker, and starting when marker is set. Then the amount of pending requests will be reduced.

`
   return google.maps.event.addListener(a, "dragstart", function() {
		searchControl('off');    //turns off search control on dragging start
		console.log('Marker drag start : search=off'); //for testing.
        b = a.getPosition()
    }), google.maps.event.addListener(a, "dragend", function() {
[...]

PR #890, opened 2016-08-20
Fix multiple delay issues.

Infernoman:
## Description
Fix  0,0,0 after login and re auth.
Fix login delay affecting scan delay.
Fix loop_time after the search is paused and restarted.

## Motivation and Context

fixes PR #851, #828, #523, and issues #830, #683, #607, #456, #847, #462

## How Has This Been Tested?

Tested by multiple people. PR taking into account multiple PR's

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -563,6 +563,8 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
             if args.proxy:
                 api.set_proxy({'http': args.proxy, 'https': args.proxy})
             api.activate_signature(encryption_lib_path)
+            # Get current time
+            loop_start_time = int(round(time.time() * 1000))
             # search forever loop
             while True:
                 # Grab the next thing to search (when available)
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -433,8 +433,6 @@ def search_overseer_thread_ss(args, new_location_queue, pause_bit, encryption_li
 
 def search_worker_thread(args, account, search_items_queue, parse_lock, encryption_lib_path, status, dbq, whq):
 
-    stagger_thread(args, account)
-
     log.debug('Search worker thread starting')
 
     # The forever loop for the thread
@@ -539,7 +537,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     status['message'] = "Waiting {} seconds for scan delay".format(sleep_delay_remaining / 1000)
                     time.sleep(sleep_delay_remaining / 1000)
 
-                loop_start_time += args.scan_delay * 1000
+                loop_start_time = int(round(time.time() * 1000))
 
         # catch any process exceptions, log them, and continue the thread
         except Exception as e:
@@ -549,7 +547,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
 
 
 def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encryption_lib_path, status, dbq, whq):
-    stagger_thread(args, account)
+
     log.debug('Search worker ss thread starting')
     status['message'] = "Search worker ss thread starting"
     # forever loop (for catching when the other forever loop fails)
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -565,14 +563,16 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
             if args.proxy:
                 api.set_proxy({'http': args.proxy, 'https': args.proxy})
             api.activate_signature(encryption_lib_path)
+            # Get current time
+            loop_start_time = int(round(time.time() * 1000))
```
justinkwaugh: Why here and not after we decide to actually search after line 574?

pogom/search.py
```diff
@@ -643,6 +649,8 @@ def check_login(args, account, api, position):
             log.debug('Credentials remain valid for another %f seconds', remaining_time)
             return
 
+    stagger_thread(args, account)
```
justinkwaugh: If you have a beehive with worker per process rather than lots of workers for one process, this stagger thread does nothing to make the logins not line up if spawn times line up.

pogom/search.py
```diff
@@ -620,9 +620,15 @@ def search_worker_thread_ss(args, account, search_items_queue, parse_lock, encry
...
-                    time.sleep(sleep_time)
+
+                    # If there's any time left between the start time and the time when we should be kicking off the next
+                    # loop, hang out until its up.
+                    sleep_delay_remaining = loop_start_time + (args.scan_delay * 1000) - int(round(time.time() * 1000))
+                    if sleep_delay_remaining > 0:
+                        status['message'] = "Waiting {} seconds for scan delay".format(sleep_delay_remaining / 1000)
+                        time.sleep(sleep_delay_remaining / 1000)
+
+                    loop_start_time = int(round(time.time() * 1000))
```
justinkwaugh: But... this isn't the loop start time.

Infernoman:
thats the end of the loop. and since we're starting at the beginning of while True: on line 569. and doesnt go back to set the loop time on line 567. Also when the worker threads get paused this affects this scan delay and it needs to be reset at the end of every loop. Hopefully this should answer your other question as well.

justinkwaugh:
Yes, but the start of the loop is not the important part at all.  Who cares when the loop was originally started... it is immaterial. Which is why it is overly confusing when implemented this way and with that variable name.  If you want to keep track of when the last successful scan completed, call it that, and just capture it there. Also all the failure cases are sleeping in their own different ways which combined with this makes it unnecessarily complicated.

I think the main problem is that it is in a sense being calculated backwards in all cases.  We want to delay the correct amount before the next scan, not the correct time after.  What would make more sense is to have a variable that indicated the next allowed scan time. Then every failure case, success case, etc. could simply set that as they see fit for their given reasons, and the next time it decided it wanted to scan, it would check if it was allowed to or not yet, and sleep for the correct amount of time.

pogom/search.py
```diff
@@ -539,7 +537,7 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                     status['message'] = "Waiting {} seconds for scan delay".format(sleep_delay_remaining / 1000)
                     time.sleep(sleep_delay_remaining / 1000)
 
-                loop_start_time += args.scan_delay * 1000
+                loop_start_time = int(round(time.time() * 1000))
```
tai-hun: loop_start_time needs to be reset after getting a new step from the queue.  Otherwise the second round of loop after pause longer than scan delay won't get delayed.  (loop end, loop_start_time = 0:00) --goes back to beginning--> (request new queue, 0:00, but paused) --15s elapsed--> (gets a new step after resume, 0:15) -> api call, done -> (calculate sleep delay remaining, = 0:00 + 0:10 - 0:15, hey it's not greater than 0 so let's pass the loop without delay) -> there comes soft ban.

pogom/search.py
tai-hun: This is fine for the initial login but when search goes on and time comes for re-login, there will be unnecessary additions of delays.

lota7:
@Infernoman , both @justinkwaugh @tai-hun raises valid points and yet you closed this without any changes or comments.

MatthiasDING:
@Infernoman, I'm sorry but why this request was closed?

crckdns:
I'm also wondering exactly about the same right now

lota7:
Guess infernoman is ignoring this issue.

Infernoman:
This is being rebased in another commit, with much more changes that me and a few other developers were working on, sorry to close the PR without notice.

Infernoman:
We took both of your changes into mind while we were working.

lota7:
Thats wonderful to hear!
Greatly awaiting the new commit

Infernoman:
checkout #901

Issue #880, opened 2016-08-20
Unsuccessful login not caught


springjools:
## Expected Behavior

Login with a wrong username should yield a clear error
## Current Behavior

auth_ptc responds with this:

auth_ptc][   ERROR] Could not retrieve token: Your username or password is incorrect

but still you get this from search.py:
 search][    INFO] Login for account asssasa successful
## Possible Solution

I'll work on it, but:

https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/search.py#L656

This does not work. The exception needs to be caught better somehow.
## Steps to Reproduce (for bugs)

Just run the runserver.py like this:
python runserver.py -l "lat lon" -u nonexistentuser1 -p somepasswd
## Your Environment

python 2.7.11
latest dev build: pulled 1 hr ago
win 7 x64

springjools:
Note: I changed L664 to log.info (in case you wonder)

onilton:
@springjools is this still valid?

springjools:
It's from August, searchy.py has been refactored many times since.

Now the output is like this:

2017-03-24 12:26:12,470 [   search-worker-0][        search][   ERROR] Exception in search_worker under account asasa Exception message: TooMa
LoginAttempts('Exceeded login attempts.',).
Traceback (most recent call last):
  File "U:\bin\github\RocketMap\pogom\search.py", line 927, in search_worker_thread
    status['proxy_url'])
  File "U:\bin\github\RocketMap\pogom\account.py", line 60, in check_login
    raise TooManyLoginAttempts('Exceeded login attempts.')
TooManyLoginAttempts: Exceeded login attempts.

Which I think is acceptable.

PR #891, opened 2016-08-20
Fix for checking logged in status and removing invalid accounts

springjools:
## Description
i) in models.py:
* catch KeyError and TypeError exceptions, return result to search.py

ii) in search.py:
* Check for parse errors in parse_map function
* Check login status properly
* Remove dead workers or ones that are not logged in, write log to loginerrors.log

## Motivation and Context
In: Response to: https://github.com/PokemonGoMap/PokemonGo-Map/issues/880
As: Alternative to https://github.com/PokemonGoMap/PokemonGo-Map/pull/806

Currently the check_login https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/search.py#L637 method fails, it will always proceed, no matter if the login was successful or not. This leads to the process that we are sending accounts which are not even logged in to parse the map, and of course the map object received is invalid. Instead of trying 3 times to get the map, we should check if user is valid and logged in.

This needs a change in pgoapi, because the set_authentication method must return the result. It's a simple change though and already posted here: https://github.com/PokemonGoMap/pgoapi/pull/4

With these changes, invalid accounts can be removed at login, and in the case that an account gets banned, that is already logged in, can be handled with the 3 unsuccessful map download method already present in search.py, but instead of catching the exception on search.py it needs to be caught in models.py parse_map, because that's where the KeyError/TypeError occurs.

## How Has This Been Tested?
Tested on my computer with updated pgoapi. Relies on 

Output from console when trying to log on with a wrong account: http://pastebin.com/T5yHhB1a

Output when removing failed workers: http://pastebin.com/T5yHhB1a

## Types of changes

## Checklist:
Requires:
https://github.com/PokemonGoMap/pgoapi/pull/4

Relies on 'set_authentication' in pgoapi.py to return True if login has succeeded, so wait for status of https://github.com/PokemonGoMap/pgoapi/pull/4 before merging

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -23,7 +23,8 @@ import json
 import os
 import random
 import time
-import geopy.distance as geopy_distance
+import geopy
+import geopy.distance
 
 from operator import itemgetter
 from threading import Thread, Lock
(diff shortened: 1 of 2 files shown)


aldarund:
Writing into loginerrors.log not really good. If u launch multiple python process they can write to same file at the same time and so it will be mess up.

springjools:
Hmmm, but the file is opened with the "with" statement

PR #896, opened 2016-08-20
Add docker-compose support to run it as a service.

soleo:
## Description
Add a docker-compose.yml to run PokemonGoMap with docker in one step

## Motivation and Context
Easy deployment for the project.

## How Has This Been Tested?
Tested on Mac OS El Captan v10.11.6 
```
$ docker version
Client:
 Version:      1.12.1-rc1
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   7889dc7
 Built:        Fri Aug 12 18:35:53 2016
 OS/Arch:      darwin/amd64
 Experimental: true

Server:
 Version:      1.12.1-rc1
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   7889dc7
 Built:        Fri Aug 12 18:35:53 2016
 OS/Arch:      linux/amd64
 Experimental: true
$ docker-compose verison
docker-compose version 1.8.0, build f3628c7
docker-py version: 1.9.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.2h  3 May 2016
```

I used ``` docker-compose up -d```, and there is no errors seen in logs

screen shot 2016-08-20 at 5 29 32 pm

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "2"
+
+services:
+  web:
+    image: ashex/pokemongo-map:latest
+    ports:
+      - "5000:5000"
+    environment:
+      - POKEMON_AUTH_SERVICE=ptc
+      - POKEMON_USERNAME=UserName
+      - POKEMON_PASSWORD=Password
+      - POKEMON_LOCATION=Chicago, IL
+      - POKEMON_STEP_LIMIT=5 
+      - POKEMON_GMAPS_KEY=SuperSecret 
+    container_name: docker-pogomap
+  ngrok:
+    image: 'wernight/ngrok'
+    command: ngrok http docker-pogomap:5000 -log stdout -log-level debug
+    depends_on:
+      - web
+    container_name: docker-ngrok
--- a/docs/advanced-install/docker.md
+++ b/docs/advanced-install/docker.md
@@ -68,6 +68,20 @@ docker pull chuyskywalker/pokemongo-map
 ```
 
 Then redo the steps from "First Install" and you'll be on the latest version!
+## Running with docker-compose
+
+Run:
+```
+docker-compose up -d
+```
+
+Once the service is up, use ``docker-compose logs`` to look for ngrok.io URL.
+It should looks something similar with the following.
+
+```
+docker-ngrok | t=2016-08-20T22:00:28+0000 lvl=dbug msg="decoded response" obj=csess id=ee13fdb15b4d clientid=b651117cf5b47b82c2e5f1b20c50cc75 sid=9 resp="&{ClientId:d6ca9f48fcab817a5093614c0faf5179 URL:http://d73667dd.ngrok.io Proto:http Opts:map[Hostname:d73667dd.ngrok.io Auth: Subdomain:] Error: Extra:map[Token:0a4757aeb26f58c4350c178d0edeea2a698abf535f55b93d08314db1a9388ca2]}" err=nil
+```
+
 
 ## Running on docker cloud 
 


Ashex:
Switch the container to the official one :(

https://hub.docker.com/r/pokemap/pokemongo-map/

jnovack:
image

I'm pretty sure you just wanted to show off your sexy command line...

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "2"
+
+services:
+  web:
+    image: ashex/pokemongo-map:latest
```
jnovack: Official repo has been changed.

https://hub.docker.com/r/pokemap/pokemongo-map/

Ashex: Project doesn't have support for environmental variables for configuration yet so it's not directly compatible with this.

jnovack: Meh.

I would rather see an environment variable PR than a `docker-compose` PR confusing docker newbies on which container is most up to date (because the build changes daily).

Ashex: All the "launch on this service" use third party built containers of some sort. The container mentioned is actually mine ;-) latest points to the most recent tagged release and is kept up to date. 

That said, one of the devs had a pr adding variable support on the old project.

soleo: I used @Ashex docker image because others don't work. Perhaps, we should have a stable branch which would already build successfully.

Ashex: @soleo env var support was merged earlier today, take a look at #952 for details. You should be able to use that instead of my container.

hemsleyk:
Should run this with MariaDB or MySQL instead of SQLite.

jorgebsa:
Agree with @hemsleyk, i would go further and suggest that a docker network should be created, and that a MySQL/MariaDB container should be launched in it along with a PoGoMap container. Also, the DB container should have a shared volume from the host, so it's persistent.

Ashex:
What purpose does a docker network serve and where is the benefit (hint, little to nothing)?  Service links are sufficient for communication between containers and you just need to toss in a port mapping to expose ngrok. 


Anyways, this would serve better as an addition to [existing documentation](https://pgm.readthedocs.io/en/develop/advanced-install/docker.html) in the way of a code snippit showing an example of running the container with docker-compose.

jorgebsa:
A docker network handles multiple containers linking better. For example, it would help if you wanted to add a PokéAlarm container to the network.  Also, linking is considered a [legacy feature](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) by docker, network being the current method.

I do agree that simply adding better explanations to the docs on how to achieve this is a better solution. However, if a docker compose feature were to make it to the master, i don't see why not use a docker network, it's a simple as declaring a link to the docker run. Example (omitted many cl options):

```
docker network create pogomap
docker run --name some-mysql --net=pogomap -v /path/to/pogo/mysql/:/var/lib/mysql -e MYSQL_ROOT_PASSWORD=super-secret -d mysql:5.6.32
docker run --name pogomap --net=pogomap -d -p 5000:5000 pokemap/pokemongo-map --db-type mysql --db-name dbname --db-user root --db-pass super-secret --db-host some-mysql --db-port 3306
```

It is definitely not critical

Ashex:
Okay that's actually a pretty valid point regarding links vs networking (Docker Cloud still uses service links and so does Amazon ECS so I simply haven't used the new networks feature). 

A docker-compose file in the project is only useful if we added a launch to docker cloud button which is never going to happen since the AWS IP space is blacklisted.

jnovack:
Not sure how a private network would help, considering you'll still need to expose the Web container, and possibly the DB Container (or it's SSL Ambassador, if you are running distributed) on your "non-docker" network.

jorgebsa:
I've created #1057 , which improves a lot the docker documentation and thus decreased the need of a compose feature. It could still be added later, but i figure the updated documentation should be sufficient for anyone attempting to run the server via docker.

Ashex:
Closing as  #1057 adds the necessary docker documentation.

PR #905, opened 2016-08-21
Removed fake-pgo-api.py

BaIthamel:
Probably got pulled in in some merge or other.


ottowayne:
Look at the file history

> - Add a sufficiently functional pgoapi emulation server for testing
  Since this update is all about contention, I needed to create lots
  of contention. Which means lots of scanning. Which means YER BANNED
  emails from Niantic. In order to get around that, this commit adds
  a "fake" pogo api.

https://github.com/PokemonGoMap/PokemonGo-Map/commit/4a27fd01c2e52809c96adc94ae63cc12a073249a

justsomedudeonthenet:
Please don't remove it! It makes development and testing possible without real accounts.

BaIthamel:
hum, that's a pain yet another thing that needs removing in the release.

j16sdiz:
Move to `contrib/`? We need this for development.

chuyskywalker:
Yes, this should have been in `contrib` just toss it over there (right next to the dummy webhook implementation)

Studio-XP:
From chu's #582. Not sure if pogom/fakePogoApi.py should be moved too or remain in pogom

BaIthamel:
I'll move it over at some point.

PR #908, opened 2016-08-21
Status page

Zoruk:
## Description

Add a web status wich work with multiple worker on same database

## How Has This Been Tested?

Tested with about 50 worker

## Types of changes

## Checklist:

- [x] Add option

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -26,6 +26,7 @@ import time
 import geopy
 import geopy.distance
 
+from datetime import datetime
 from operator import itemgetter
 from threading import Thread
 from queue import Queue, Empty
@@ -35,7 +36,7 @@ from pgoapi.utilities import f2i
 from pgoapi import utilities as util
 from pgoapi.exceptions import AuthException
 
-from .models import parse_map, Pokemon, hex_bounds
+from .models import parse_map, Pokemon, hex_bounds, MainWorker, WorkerStatus
 from .transform import generate_location_steps
 from .fakePogoApi import FakePogoApi
 from .utils import now
@@ -163,6 +164,35 @@ def status_printer(threadStatus, search_items_queue, db_updates_queue, wh_queue)
         time.sleep(1)
 
 
+def worker_status_db_thread(threads_status, name, db_updates_queue):
+    while True:
+        workers = {}
+        overseer = None
+        for status in threads_status.values():
+            if status['type'] == 'Overseer':
+                overseer = {
+                    'worker_name': name,
+                    'message': status['message'],
+                    'method': status['method'],
+                    'last_modified': datetime.utcnow()
+                }
+            if status['type'] == 'Worker':
+                workers[status['user']] = {
+                    'username': status['user'],
+                    'worker_name': name,
+                    'success': status['success'],
+                    'fail': status['fail'],
+                    'no_items': status['noitems'],
+                    'skip': status['skip'],
+                    'last_modified': datetime.utcnow(),
+                    'message': status['message']
+                }
+        if overseer is not None:
+            db_updates_queue.put((MainWorker, {0: overseer}))
+            db_updates_queue.put((WorkerStatus, workers))
+        time.sleep(3)
+
+
 # The main search loop that keeps an eye on the over all process
 def search_overseer_thread(args, method, new_location_queue, pause_bit, encryption_lib_path, db_updates_queue, wh_queue):
 
@@ -182,6 +212,14 @@ def search_overseer_thread(args, method, new_location_queue, pause_bit, encrypti
         t = Thread(target=status_printer,
                    name='status_printer',
                    args=(threadStatus, search_items_queue, db_updates_queue, wh_queue))
+        t.daemon = True
+        t.start()
+
+    if args.status_name is not None:
+        log.info('Starting status database thread')
+        t = Thread(target=worker_status_db_thread,
+                   name='status_worker_db',
+                   args=(threadStatus, args.status_name, db_updates_queue))
         t.daemon = True
         t.start()
 
@@ -327,7 +365,6 @@ def get_hex_location_list(args, current_location):
         locationsZeroed.append(((location[0], location[1], 0), 0, 0))
 
     return locationsZeroed
-
 
 def get_sps_location_list(args, current_location, sps_scan_current):
     locations = []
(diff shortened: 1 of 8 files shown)


justsomedudeonthenet:
I like the idea.  How does it handle multiple separate copies of the map running, like a beehive layout?

Many people were concerned when last talking about it that it should have password protection, as it reveals your usernames.

Zerotask:
Yes, a password protection ist a must have for this.

Zoruk:
@justsomedudeonthenet @Zerotask Password added

aldarund:
So what about multiple copies like a beehive? Does it work with it?

Zoruk:
@aldarund Yep when you lunch the app with -sn "XXX" it will appear with XXX as table header in the status page but each beehive must have a diverent name

subzerofun:
Hey @Zoruk, thanks for the great addon! Would like to test it, but when i use this as my cmd args:

`python runserver.py -a ptc -u 'username' -p 'password' -st 10 -l 12,34 -D data.db -k 'mykey' --status-name 'STATUS_NAME' --status-page-password 'mypassword'`

i can open `localhost:5000/status` and the login page gets rendered, but after i put in the above defined password and click on login, the status page doesn´t get loaded... the browsers just refreshes and changes the url from **/status** to **/status?**. so for some reason i can´t get past the login screen. i tried to bypass the password query by changing your code, but that didn´t work either...

Is there something wrong with my arguments? Do i need to run more than one worker? I´m on OSX, tried Chrome, Firefox and Safari – everywhere with the same result.

When i look at the terminal output everything looks OK:
```
[ search-overseer][        search][    INFO] Search overseer starting
[ search-overseer][        search][    INFO] Starting status database thread
[ search-overseer][        search][    INFO] Starting search worker threads
```

Zoruk:
The login is handled by javascript. The page should not reload

```javascript
$('#password_form').submit(function (event) {
  event.preventDefault()
  statusPagePassword = $('#password').val()
  ...
})
```

ItsCEED:
Hi @Zoruk im kinda newbie, when you gonna make the documentation on how to use it? Also, if i run for example 2 workers x 5 screens(ports) do i get all the info on the main one (5000)?

chuyskywalker:
I'm excited to see this get fleshed out.

Generally speaking, databases really aren't for logging...but...I also think this is more useful to have this feature, than to reject it under demand of a rewrite with a more complex "parent <-- child" api-esque pushing setup.

A quick, 30 second look over; you do need to audit the html page since it's including a boat load of stuff that I'm sure is unused.

I also would not recommend having the data delivered back via `raw_data` -- make a unique end point.

Zoruk:
@chuyskywalker Done

jnovack:
I don't like `/status`.  Everyone knows it is there.  It's just like `/admin`.

I RECOMMEND a basic `--status-page URLENCODED` so I can at least mask it.  Accept any URL Encoded string without `/`.  Examples: `status`, `secret_status`, `my-status-page` `0cba061c25c86424187f08b0344c13d1`

pogom/search.py
```diff
@@ -17,27 +17,27 @@
...
 from pgoapi import utilities as util
 from pgoapi.exceptions import AuthException
+from pgoapi.utilities import f2i
+from queue import Queue, Empty
 
-from .models import parse_map, Pokemon
-from .fakePogoApi import FakePogoApi
 import terminalsize
+from .fakePogoApi import FakePogoApi
+from .models import parse_map, Pokemon, WorkerStatus, MainWorker
```
jnovack: What's with the reorganization?

jnovack:
Secondarily, for the python dummies (myself included), does this work with out-of-process spawned workers?  Someone may have a use-case of runnning multiple searchers on multiple machines, feeding only.  Will this properly track them?

pogom/search.py
Zoruk: Oops i just hit the refractoring of import if you look there is also the same removed

Zoruk:
Yes it's work with multiples worker in multiples process on different machines

Zerotask:
@jnovack you are a grumpy cat. I always read that you dislike things.

jnovack:
Everyone thinks their patch is a work of art and perfect. (Mine included)  So with others, my strength is to come in, apply some happy little trees and make a "good patch" a "GREAT patch".  Make sure that someone considered other circumstances; ensure that a patch does not lead to a deviation of the project goals.  Just move the needle a bit.

The fact is, I rarely comment on PRs I hate, and I mostly comment and post on issues and PRs I *like* so they can be PRs I LOVE. 

Development is hard.  No Pain, No Gain
image

Zerotask:
@Zoruk this tool is really amazing. 
I would suggest to add an accordion (jquery maybe) to switch between different views: a) all a) active b) banned / not working

At the moment I have big trouble with identifying banned accounts (to swap them). This would help a lot.

mew905:
I'd actually like to expand this to encompass the radar as a whole and grab the info from a database. How would one go about it?

maddhatter:
I _love_ this update. I needed this so badly, and this does such a better job than what I was thinking up...

The only thing that would make it better, IMHO, would be if whenever a worker starts up, it clears out old Worker statuses. That way if I notice a banned account and swap it out, I don't have to keep looking at the banned accounts' statuses on the page.

Something like this right at the beginning of `worker_status_db_thread` (before the `while True:`) seems to do it for me:
```python
log.info("Clearing previous statuses for '%s' worker", name)
WorkerStatus.delete().where(WorkerStatus.worker_name == name).execute()
```

static/js/status.js
```diff
@@ -0,0 +1,229 @@
...
+    <div class="status_table" id="table_${hash}">
+      <div class="status_row header">
+        <div class="status_cell">
+          Username
+        </div>
+        <div class="status_cell">
+          Success
+        </div>
+        <div class="status_cell">
+          Faill
```
maddhatter: Looks like there's a typo here

jnovack: Perhaps intentional?  Sarcasm?

maddhatter:
@Zoruk  Would love to see this get merged. Sent you two PRs:

1. [#1](https://github.com/Zoruk/PokemonGo-Map/pull/1) to resolve the conflicts w/ develop
2. [#2](https://github.com/Zoruk/PokemonGo-Map/pull/2) to make the two tiny changes I suggested in my previous comments. (This one has conflicts, but they'll go away once [#1](https://github.com/Zoruk/PokemonGo-Map/pull/1) is merged)

jnovack:
@maddhatter Zoruk has been AFK for 10 days.  Do you want to resubmit this with your fixes?

maddhatter:
I've submitted #1090 to hopefully get this merged.

PR #913, opened 2016-08-21
Fixes to chuyskywalker's branch

justsomedudeonthenet:
## Description
This PR against the drying branch fixes a few issues:

- A race condition in the status screen that caused a crash if it tried to get the length of the worker names before any worker threads had been started
- Saying we were 0 seconds early for a scan and sleeping
- Adds configurable spawnpoint skip time, for people who only want to know about spawns that have at least x seconds left on them
- Captures exceptions in the search worker and makes them sleep if they fail too many times
- Fixes a few status messages
- Adds a sleep after login, which is necessary to prevent the first scan after that from failing
- Avoids sleeping if the worker didn't actually do a scan, when they are skipped for being too late.  This is essential to make sure it dumps too late queue items as quickly as possible and can get caught up again.
- Moves the time adjustment from disappear time to appear time back into models.py - spawns.json files already have the appear time, adjusting their time further would break things.

## Motivation and Context
chuyskywalker's refactoring is much needed and will simplify further development.  This just fixes a few issues so his branch can be merged with develop.

## How Has This Been Tested?
On my linux server, against both the real and fake APIs.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -244,6 +244,10 @@ class Pokemon(BaseModel):
                 if geopy.distance.distance(hl, (sp['lat'], sp['lng'])).meters <= 70:
                     filtered.append(s.pop(idx))
 
+        # Fix time to be appear time instead of disappear time
+        for spawn in filtered:
+            spawn['time'] = (spawn['time'] + 2700) % 3600
+
         return filtered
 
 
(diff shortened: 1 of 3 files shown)


justsomedudeonthenet:
He has made most of these changes himself.  I will make a smaller PR later for some of the less important ones.

Issue #917, opened 2016-08-21
Can't use Webhooks after update


sharklab:
I recently updated PokemonGo_Map but now i'm not able to send a single webhook, i just tried both the config.ini and the -wh "http://requestb.in/xxxx" methods. Is there something new? Here's my call

Python C:\PokemonGo-Map-3.1.0\runserver.py -a ptc -l "lat, long" -wh "http://requestb.in/yslwoeys" -st 50 -sd 5 -u user -p pass

Thanks!

justsomedudeonthenet:
Webhooks are working just fine.

Your 5 second scan delay is probably making scans fail.  Needs to be 10 seconds.

Post debug logs (run with -d) if you still believe there is an issue, but I am using them just fine.

sharklab:
yeah switching the -sd from 5 to 10 did the trick! Thanks

PR #927, opened 2016-08-22
--scan-delay minimum

jnovack:
Provide a minimum for scan delay.

## Description

Requires `--scan-delay` to be a minimum of 10 seconds.

## Motivation and Context

Derpies not reading documentation.  See #917 

## How Has This Been Tested?

It has not. I winged it.  #LivinOnTheEdge

## Types of changes

## Checklist:
- [X] My change requires a change to the ~~documentation~~ *code comments*.
- [X] I have updated the ~~documentation~~ *code comments* accordingly.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -496,8 +496,8 @@ def search_worker_thread(args, account, search_items_queue, parse_lock, encrypti
                         break
 
                     # Increase sleep delay between each failed scan
-                    # By default scan_dela=5, scan_retries=5 so
-                    # We'd see timeouts of 5, 10, 15, 20, 25
+                    # By default scan_delay=10, scan_retries=5 so
+                    # We'd see timeouts of 10, 20, 30, 40, 50
                     sleep_time = args.scan_delay * (1 + failed_total)
 
                     # Ok, let's get started -- check our login status
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -186,6 +186,10 @@ def get_args():
         if (args.step_limit is None):
             errors.append('Missing `step_limit` either as -st/--step-limit or in config')
 
+        if (args.scan_delay < 10):
+            log.warn("--scan-delay is less than 10, setting to 10")
+            args.scan_delay = 10
+
         if args.auth_service is None:
             args.auth_service = ['ptc']
         else:


justsomedudeonthenet:
This is completely unnecessary.

In fact, quite a few people run with a scan delay between 7 and 9, which works perfectly for some.

leclaida:
unnecessary is a bit of a strong word, it is only unnecessary for users that are at least semi-competent. this commit would cut down on the number of "bug reports". perhaps it is better issuing the warning but not resetting the scandelay to 10.

BaIthamel:
It's only unnecessary because some people's trades lock for 2-3 seconds when writing to db.

sebastienvercammen:
+1 to @leclaida 

Add a warning if they go below a certain minimum, but continue anyway after a certain delay (5s - 10s?). _Derpies_ will - hopefully - read the warning and set it to something higher while the rest can keep setting it to whatever they want.

sebastienvercammen:
@jnovack I appreciate the added cynical joke, but _derpies_ exist in ranges rather than being either black or white.

I should have phrased my comment as "a larger percentage of _derpies_ will read the warning, and the number of GitHub issues will be reduced".

jnovack:
Whoa whoa, racist.  I don't see color, I only see Derpie or non-Derpie :p

I heavily disagree that the error message will deter any tickets.  BUT... what do you want to do here?  I can't merge, I have no say; so I am at the mercy of the dev team.

* `log.warn` under <10, or
* Enforce 10 minimum

I chose "enforce" because:

* natural clients do not scan any harder than 10sec,
* lowering the time only elevates the risk of teleporting,
* weighing catering to slower CPU/DB vs client-compliance, I went with client-compliance.

vaurdan:
I use proxies, so sometimes proxies takes 10 seconds to get an answer from the server, which means I need to use a lower delay. I don't think it should be enforced, but a warning is a good idea.

justsomedudeonthenet:
I would go with either warn, or enforce with a switch to bypass it.

BaIthamel:
I think we just need to change the default to 10.

jnovack:
Default is already 10. Closing.

PR #942, opened 2016-08-22
Experimental spawn-point clustering

TBTerra:
Using k++ means clustering to break the big spawn list into a small list covering a small area for each worker. does not eliminate teleporting (needs a simple speed limiter for that) but limits the distance teleported to within the cluster

## Description
Instead of loading the list of spawns and then adding them to a queue shared between all workers (where a worker may be tasked with teleporting across the entire scan area), instead the list of spawns is clustered with each worker handling the entire of one cluster, this means that the worker will only ever have to travel across the cluster (in a city with a cluster of 100spawns, the cluster size is less than 1km)

## Motivation and Context
This is one step towards tackling the issue of accounts being banned for teleporting across the map, it achieves this without having to add excessive delays that a usual speed limiter may have to as well

When  combined with a simple speed limiter it should both minimize teleport bans, whilst keeping a high spawns to workers ratio

## How Has This Been Tested?
The clustering algorithm has been tested on datasets of up to 10k spawns, and the full implementation is being tested by various members of the community

example clustering of 5800 spawns split into 60 clusters (note that there are only 18 colors so some clusters are only differentiated by their letter)

## Types of changes

## Checklist:

this code is not compatible with pr 901

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,18 +12,18 @@ EXPOSE 5000
 WORKDIR /usr/src/app
 
 # Set Entrypoint with hard-coded options
-ENTRYPOINT ["python", "./runserver.py", "--host", "0.0.0.0"]
+ENTRYPOINT ["dumb-init", "-r", "15:2", "python", "./runserver.py", "--host", "0.0.0.0"]
 
 # Set default options when container is run without any command line arguments
 CMD ["-h"]
 
 # Install required system packages
-RUN apk add --no-cache ca-certificates
-RUN apk add --no-cache bash git openssh
+RUN apk add --no-cache ca-certificates git
 
 COPY requirements.txt /usr/src/app/
 
 RUN apk add --no-cache build-base \
+ && pip install --no-cache-dir dumb-init \
  && pip install --no-cache-dir -r requirements.txt \
  && apk del build-base
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -47,10 +47,9 @@ def memoize(function):
 def get_args():
     # fuck PEP8
     configpath = os.path.join(os.path.dirname(__file__), '../config/config.ini')
-    parser = configargparse.ArgParser(default_config_files=[configpath])
+    parser = configargparse.ArgParser(default_config_files=[configpath], auto_env_var_prefix='POGOMAP_')
     parser.add_argument('-a', '--auth-service', type=str.lower, action='append',
-                        help='Auth Services, either one for all accounts or one per account. \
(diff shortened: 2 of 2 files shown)


TBTerra:
as the title suggests this is experimental, it probably has some bugs and is definitely missing useful features. ones to add atm are: simple speed limiter, load pre clustered spawns from file. suggestions are welcome

clvv:
The issue is to figure out exactly what is the banning criteria. Is it speed? Is it distance jumped? or something else entirely ("behaving" like a human player)?

My guess is that there's probably a speed limit and a distance jump limit (i.e. even if you go over speed limit, but jumped distance is small, then there's no penalty). Speed limiting and clustering solves each of the aspect respectively. Now we will just have to see which will result in less bans.

Becold:
Is there any calc to calculate the amounts of accounts per cluster (like beehive setup) or is it using automatically?
And if so, does it use the same accounts per cluster even after a reboot of the server?
> Idea: Maybe we can insert the accounts inside the database like http://feathub.com/PokemonGoMap/PokemonGo-Map/+49

Goldvivi1:
Scanned Locations also needs clustering.

mew905:
looks like it takes your number of workers and splits them up evenly. If this is functional, I'd love to try this

TBTerra:
@Becold the algorithm makes one cluster per account, the tricky bit is working out how many clusters you need, it is also deterministic so if the server is re launched on the same dataset it will return the same clusters in the same order to prevent an account teleporting because its cluster changed

TheNewDarkOne:
Pretty much waiting for this before using -ss again. my accounts survived every ban using a pretty non teleporting beehive but when i switched to -ss they didn't even last 24 hours and every one of them got killed. I watched how it behaved and the teleporting with 6 accounts with a 9 step and over 600 spawn points was pretty bad.

Forbes256:
Is this idea dead? Seems like an incredible improvement.

TheNewDarkOne:
Agreed, and sorely needed. Although I suspect with Niantic stepping up their game now, Terra probably realizes more is needed than this and has gone back to refactoring something better. This is a start but just simply isn't enough. Teleporting needs to be eliminated not just reduced. That's even assuming it gets done in time to reap benefit of it before captcha is enabled, which will set mapping back for a while until the geniuses figure out a way to handle it.

TBTerra:
sorry it got auto closed when i moved the changes to a new 'experemental' branch in my repo, will re-pull shortly

PR #951, opened 2016-08-23
Fix: SS db initial list build delay

rbryson74:
## Description
the PR#901 refactor of the get_spawnpoints_in_hex function included a
double nested loop over the whole db. This reverts to the method used
previously.

## Motivation and Context
The refactored function can take anywhere from 1 minute to reports of 5+ minutes to build the initial list. 

## How Has This Been Tested?
tested on my system 

Does not appear to affect any other code. 

## Types of changes

## Checklist:

the PR#901 refactor of the get_spawnpoints_in_hex function included a
double nested loop over the whole db. This reverts to the method used
previously.


TBTerra:
im currently looking through what was added in 901 and if any further defelopment is to be done on ss mode then its probably that most of that commit will have to be reverted, they are too different in kind to be merged the way they are

rbryson74:
What's great about this function though is that it is only used the one time. The refactor didn't change the incoming variables. And the output is the same format. So the search overseer doesn't even know the difference.

chuyskywalker:
> a double nested loop over the whole db

That's not what's happening at all. 

First; it's not the whole DB (I don't know why you'd think that). It has the same bounding conditions that you reverted back, but coded with geopy instead of a custom rolled bounding distance calc.

Second, it's not a double loop (`n*n`), per-se. It is, at worst, `len(hexLocations) * len(spawnPoints)`. However, practically speaking, it's much fewer because the spawn points list has items removed as they are discovered to be within a hexLocation.

I changed it up like this because:

1. We should be using `geopy` for lat/lng determination. It's a solid library and much easier to work with / understand.
2. We should be filtering the spawns which actually show up in the `--location` hex grid when using `-ss` without a `spawns.json`. 

Speed of generating the list; I can't argue against that to be honest; I don't know one way or the other.

---

This all said: I, personally, think that `-ss` based on db-derived data is going to go the way of the dodo bird pretty soon. There are a TON of SPS optimizations in the works (spawn point compression, optimized bus scheduling, etc) which will make pre-baking the scan list necessary.

rbryson74:
This method can be redone quite easily to use geopy instead of math. But there will be users that will prefer to use the db over a file for any of a number of reasons. I agree that the loop as created in 901 is `len(hexLocations) *( len(spawnPoints)-spawnsfound)` but that still results in the lists being checked repeatedly when 1 pass is sufficient. And by that what I mean is that for spawnpoint scanning based on distance from a central point, strict adherence to the shape of the regular scanning hex is not necessary. A simple circle will work. The only reason to keep it as a hex is when using -ss in beehive with multiple workers to prevent overlap. But that will not really be needed once the clustering and account idling can be implemented.

chuyskywalker:
You make a good point, strictly adhering to the hex covering is probably overkill. I'd be willing to compromise on the idea that when you run in `-ss` mode, the location is just a center point for a circle based on the DB content. Just need to note that in the `utils.py` help attribute for that flag.

chuyskywalker:
If you do that, #1013 would be unnecessary.

rbryson74:
returning to Geopy. setting -ss from db to use a circle and iterate over the list once.

chuyskywalker:
Having checked this out, using all my normal settings, but adding the dump command into the mix...:

`develop`:
```
2016-08-27 05:07:50,245 [      MainThread][     runserver][    INFO] Sawing spawn points to spawns-develop.json
2016-08-27 05:07:51,026 [      MainThread][     runserver][    INFO] Finished exporting spawn points
                     ^^ 781ms 
...
2016-08-27 05:07:51,036 [ search-overseer][        search][   DEBUG] Loading spawn points from json file @ spawns-develop.json
2016-08-27 05:07:51,085 [ search-overseer][        search][    INFO] Total of 161 spawns to track
```

pr:
```
2016-08-27 05:11:06,902 [      MainThread][     runserver][    INFO] Sawing spawn points to spawns-pr.json
2016-08-27 05:11:07,035 [      MainThread][     runserver][    INFO] Finished exporting spawn points
                     ^^ 133ms
...
2016-08-27 05:11:07,041 [ search-overseer][        search][   DEBUG] Loading spawn points from json file @ spawns-pr.json
2016-08-27 05:11:07,044 [ search-overseer][        search][    INFO] Total of 170 spawns to track
```

Now, if I reduce my normal 5 steps, down to three (ie, covering a smaller, but fully known area)...

`develop`:
```
2016-08-27 05:21:37,845 [      MainThread][     runserver][    INFO] Sawing spawn points to spawns-develop.json
2016-08-27 05:21:38,239 [      MainThread][     runserver][    INFO] Finished exporting spawn points
                     ^^ 394ms
2016-08-27 05:21:38,244 [ search-overseer][        search][    INFO] Total of 50 spawns to track
```

pr:
```
2016-08-27 05:19:08,800 [      MainThread][     runserver][    INFO] Sawing spawn points to spawns-pr.json
2016-08-27 05:19:08,915 [      MainThread][     runserver][    INFO] Finished exporting spawn points
                     ^^ 115ms
2016-08-27 05:19:08,921 [ search-overseer][        search][    INFO] Total of 86 spawns to track
```

At the least, it's finding more spawns now (perhaps related to other stated "missing points" bugs?)

However, It's hard to validate because the UI option for "show spawn points" doesn't respect these lists _at all_. In fact, it calls `get_spawnpoints()` instead which only uses the database and actually encompasses the full damn **box** generated region.

It's hard to `:thumbs up:` this without being able to see the real effect.

---

As a side note, please update this comment to accurately reflect the motivation/goal of the code:

```
        # Filter to spawns which actually fall in the hex locations
        # This loop is about as non-pythonic as you can get, I bet.
        # Oh well. Now slightly modified.
```

"Now slightly modified" gives me the same warm fuzzies as git commits that read _"fixed a thing"_ or _"stuff"_.

rbryson74:
Here are my results with dump-spawnpoints between current develop and the PR. These are both at step 20 on a DB with about 65000 points:

Develop:
```
2016-08-27 17:54:19,938 [      MainThread][        models][    INFO] Connecting to MySQL database on 127.0.0.1:3306
2016-08-27 17:54:19,951 [ search-overseer][        search][    INFO] Search overseer starting
2016-08-27 17:54:19,954 [ search-overseer][        search][    INFO] Starting search worker threads
2016-08-27 17:54:19,957 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-08-27 17:54:19,967 [ search-overseer][        search][    INFO] New location caught, moving search grid
2016-08-27 17:54:19,969 [ search-overseer][        models][    INFO] Finding spawn points 20 steps away
2016-08-27 17:54:20,065 [      MainThread][     runserver][    INFO] Web server in SSL mode.
2016-08-27 17:55:19,964 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-08-27 17:56:19,621 [ search-overseer][        search][    INFO] Total of 1969 spawns to track
                ^120 seconds
```

PR:
```
2016-08-27 18:00:15,147 [      MainThread][        models][    INFO] Connecting to MySQL database on 127.0.0.1:3306
2016-08-27 18:00:15,157 [ search-overseer][        search][    INFO] Search overseer starting
2016-08-27 18:00:15,160 [ search-overseer][        search][    INFO] Starting search worker threads
2016-08-27 18:00:15,161 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-08-27 18:00:15,171 [ search-overseer][        search][    INFO] New location caught, moving search grid
2016-08-27 18:00:15,174 [ search-overseer][        models][    INFO] Finding spawn points 20 steps away
2016-08-27 18:00:15,289 [      MainThread][     runserver][    INFO] Web server in SSL mode.
2016-08-27 18:00:16,611 [ search-overseer][        search][    INFO] Total of 2431 spawns to track
                  ^1437ms
```

As far as the spawnpoint display in the UI that is a matter which could be handled but which I think would need its own PR. 

As far as comments and commit messages: I will do a better job with that. I have enough time to do it right.

chuyskywalker:
I'm willing to merge this and see what blows up :)

CometX:
Works as intended. Shaved hours off DB build. Thank you.

rbryson74:
So, separate from the startup delay, there is also an issue with duplicated spawnpoints when the point has more than one disappear time listed (the modified time not the date/time). That issue can be resolved but I believe should be a different PR.

averylarry:
Would it be a little better to use a circle with radius 120m * (steps - 1) + 60m   ??  Or something like that?  I would say the difference is probably negligible for step counts 4 or greater (?).  I just happen to use spawnpoint scanning with step count 2 around my office during the work day.  It covers the 1 block long building I work in.
I'll let better minds decide if the edge case is worth the change.

rbryson74:
@averylarry I was actually working on the exact value when you posted that. the commit for the math is in.

PR #930, opened 2016-08-22
Added correct STOPSIGNAL for docker build.  This fixes long wait time…

krusmir:
Added correct STOPSIGNAL for docker build.  This fixes long wait time…
## Description

Docker containers take a long time to stop.  Specifying the stop signal when invoking docker stop fixes that (in this SIGINT)

One line in the Dockerfile specifying the stop signal to use.  Taken from old pull request on original repo AHAAAAAAA before it was taken down.

## Motivation and Context

Upgrading docker images versions with the images generated by the new docker file takes a long time to resolve the container stops (5 to 20 seconds).  This change makes stopping the running container almost instant.

## How Has This Been Tested?

Yes

With current docker file:
```
docker stop container_name  
```
This takes 5 to 30 seconds depending on how long the container has been running

However, with the pull request modification
```
 docker stop container_name
```
Stops almost instantly.

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #952, opened 2016-08-23
Docker Fixups

chuyskywalker:
Combining a few things:

1. @tkSimon 's work for cleaner exit
2. @tkSimon 's also fixing up package install...mostly
3. Adding ENV hooks for EVERY CONFIG VALUE

Supersedes #930, #595, #590

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ EXPOSE 5000
 WORKDIR /usr/src/app
 
 # Set Entrypoint with hard-coded options
-ENTRYPOINT ["python", "./runserver.py", "--host", "0.0.0.0"]
+ENTRYPOINT ["dumb-init", "-r", "15:2", "python", "./runserver.py", "--host", "0.0.0.0"]
 
 # Set default options when container is run without any command line arguments
 CMD ["-h"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apk add --no-cache ca-certificates git
 COPY requirements.txt /usr/src/app/
 
 RUN apk add --no-cache build-base \
+ && pip install --no-cache-dir dumb-init \
  && pip install --no-cache-dir -r requirements.txt \
  && apk del build-base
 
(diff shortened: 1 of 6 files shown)


Ashex:
You don't need to add env_var for each argument. You can configure a global prefix that is picked up  automatically. I'd also vote for POGOM instead for the prefix ;-)

chuyskywalker:
Damnit, where's the config setup for that?

chuyskywalker:
nm, found it.

Ashex:
I'm a little confused regarding dumb-init. It's in the pip requirements but is called as a command. Any reason it's not being installed via apk?

chuyskywalker:
It's not `apk` instal;able... (is it?)

Ashex:
Yep you're correct, I took a look and a package isn't available. Carry on!

Ashex:
That said, it doesn't really belong in the requirements file since it's only used by the Dockerfile, it should be part of the run statement installing the base deps like git. 

(sorry about the double comments, I'm on my phone heading to the airport)

chuyskywalker:
Agreed. Fixed.

creatingwake:
Thanks @chuyskywalker - this is just what I'd been looking for as I'm running this in a dokku environment.  Big help.
I did try playing around with the code to get the config.ini file loaded instead of using ENV vars but couldn't get it working.  With multiple accounts dropping everything into the $EXTRA_ARGS ENV variable is a PITA.  

Do you know if its possible to replicate in Docker the native python install where the config file is read, but command line (ENV) vars take precedence?

Ashex:
@creatingwake The order of precedence is thus: `command line > environment variables > config file values > defaults`

creatingwake:
@Ashex yes, you would think so but the config.ini file values weren't being used at all on my dokku deployment. I had to replicate everything in ENV vars.  Hence I started hacking Utils.py (unsuccessfully)

Ashex:
That is a bit curious, I've been able to override config.ini settings via argument. I'm not familiar with dokku but perhaps you need to verify the working directory it was running from.

PR #959, opened 2016-08-23
Add optional logging to file

justsomedudeonthenet:
## Description
With the status screen, output can't easily be redirected to a log file anymore.

This adds two parameters:
`-v <optional filename>` will enable debug logging for the main program and pgoapi
`-vv <optional filename>` will enable debug logging for all the imported modules as well

If a filename is specified, log messages will be appended to that file, in addition to being shown on screen.

The `-d` debug parameter no longer does anything, and has been marked as depreciated (but not removed yet, so as not to break anyone's launch scripts.  

References to args.debug have been removed, save one in runserver.py that warns the user the argument is depreciated, and turns on --verbose, so -d can be removed easily at a later date.

## Motivation and Context
Allows saving log messages more easily, helps with troubleshooting.

## How Has This Been Tested?
On my server
## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -360,7 +360,7 @@ def get_sps_location_list(args, current_location, sps_scan_current):
 
     locations.sort(key=itemgetter('time'))
 
-    if args.debug:
+    if args.verbose or args.very_verbose:
         for i in locations:
             sec = i['time'] % 60
             minute = (i['time'] / 60) % 60
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -89,7 +89,6 @@ def get_args():
     parser.add_argument('-c', '--china',
                         help='Coordinates transformer for China',
                         action='store_true')
-    parser.add_argument('-d', '--debug', help='Depreciated, use -v or -vv instead.', action='store_true')
     parser.add_argument('-m', '--mock', type=str,
                         help='Mock mode - point to a fpgo endpoint instead of using the real PogoApi, ec: http://127.0.0.1:9090',
                         default='')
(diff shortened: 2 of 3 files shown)


sebastienvercammen:
+1 on the logging to file, however I'd avoid the extra `--logfile` CLI param.

As an alternative: rework `--debug` to accept an optional parameter `--debug logfile`, with warnings/errors only as default, and add `--verbose` to include all logging.

ottowayne:
> As an alternative: rework --debug to accept an optional parameter --debug logfile, with warnings/errors only as default, and add --verbose to include all logging.

So still another new parameter "--verbose"? Either way I don't like this idea because this is about the regular logs and not about the debug logs. Those should still be optional.

sebastienvercammen:
@ottowayne It's in line with how every properly written software does it. Debugging/logging involves data of different logging levels, and you pick whichever you want to enable. If you want to keep it simple (or you only have 2 specific use cases), you can use `--verbose` as a toggle for what data comes through.

At the very least, `--verbose` is better than adding a param `--logfile`. You could even argue you don't need it at all, considering you can send any output (stdout/stderr) to a file (`>`, `>>`) and that being able to switch to the status screen is _just_ a luxury...

justsomedudeonthenet:
I think the way I want to rework this is to depreciate -d, and add -v and -vv.

-v and -vv will optionally take a filename to log to.

-v  (--verbose) will show debug logs from PokemonGo-Map and pgoapi
-vv (--very-verbose) will show debug logs from those and all other imports we use

The main purpose here is to make logging to a file easy in conjuction with the status screen.  Before you could just redirect output to a file, but that doesn't work with -ps.  Log files are mostly useful for troubleshooting, in which case we want verbose output to begin with.

TachyonRSA:
In the meantime on Linux just use screen and run the command with &>> logfile.log

PR #974, opened 2016-08-24
Override 70m step distance with new argument 'step_distance'

khag7:
## Description
Yes, another command line argument

## Motivation and Context
Recent developments on the data contained within encounter_id shows that it is possible to determine the exact location of a pokemon when seen from up to 200m away (via the nearby list). Although this map has not yet implemented that feature, it is in the works. That feature is independent of this one, thus the separate PR, but I already know that once that feature exists some people will want to only scan 200m at a time (given the account bans, nobody wants to use any more accounts than necessary, and upping the radius from 70 to 200 will let you scan about 8x more area per worker)

## How Has This Been Tested?
Changed scan-distance in config to a few random values, watched scans happen at larger intervals.

## Types of changes

- [x] Yet another command line argument

## Checklist:

- [x] Will update documentation when this variable becomes useful. For now, nobody will need or want to use it.

Feel free to leave this PR sit and wait, but I am hoping to get it at least reviewed so it's ready to go when it becomes useful to be able to adjust the scan distance manually.

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -18,6 +18,7 @@
 #no-pokestops:          # disables pokestop scanning (default false)
 #scan-delay:            # default 10
 #step-limit:            # default 12
+#step-distance:         # default 70
 #min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
 
 # Misc
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -300,12 +300,10 @@ def search_overseer_thread(args, method, new_location_queue, pause_bit, encrypti
 def get_hex_location_list(args, current_location):
     # if we are only scanning for pokestops/gyms, then increase step radius to visibility range
     if args.no_pokemon:
-        step_distance = 0.900
-    else:
-        step_distance = 0.070
+        args.step_distance = 900
 
     # update our list of coords
-    locations = list(generate_location_steps(current_location, args.step_limit, step_distance))
+    locations = list(generate_location_steps(current_location, args.step_limit, args.step.distance / 1000 ))
 
     # In hex "spawns only" mode, filter out scan locations with no history of pokemons
     if args.spawnpoints_only and not args.no_pokemon:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -303,7 +303,7 @@ def get_hex_location_list(args, current_location):
         args.step_distance = 900
 
     # update our list of coords
-    locations = list(generate_location_steps(current_location, args.step_limit, args.step.distance / 1000 ))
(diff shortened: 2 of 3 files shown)


BaIthamel:
I don't see any point implementing this, if we can increase the scanned area size we will by default I expect :)

khag7:
It will be a different scanning method, similar to how spanwpoint scanning is different. It will require a decent amount of information about existing spawnpoints. So a user will have to manually decide that they want to do that.

ottowayne:
I have to go with @Balthamel on this one.

If the mentioned scanning system will be implemented then the scanning distance should be set by that mode which will also be 200m fixed. IMHO there is no need for that config option.

leclaida:
it would not be 200m default because you need complete database to make use of the larger scanning area.

basically,
70m : incomplete database
200m : complete database

hemsleyk:
Sounds like an issue generator for people who don't know what it does.

averylarry:
Are you saying you'll be able to extrapolate which spawnpoint it is from 200m away if, and only if, you already know all (most?) of the nearby spawnpoints?  As in -- (for example) you get information on distance from where you are but not direction so you scan all the known spawnpoints that are that distance away and in the correct timeframe?

If so, then this would perhaps become a default for -ss and only used for -ss?  Or maybe additional code to self-determine the "completeness" of the requested scan area and "intelligently" decide between 70 and 200?

leclaida:
@averylarry yes, you have the gist of it. but you need more than just the spawnpoint info... you need spawnpoint info for every hour for two days. see this reddit thread: https://www.reddit.com/r/pokemongodev/comments/4z9a6d/encounter_ids_not_completely_random_some_bits/

thus it is useless if you do not have a very complete database. though i do agree a better way would be to implement an intelligent design so that the choosing of 200m or 70m is automatic.

khag7:
We would need to collect "nearby" on every step we scan. Analyze each of the nearbys and maybe we can place it on the map, but that will only happen if we know enough history about the spawnpoints in that 200m radius.

But that's all for another PR, I"m about halfway done with it.

Once that feature exists, there will be a lot of users who say "Since I have enough historical information about spawnpoints, I'm confident that I can find most of my pokemon via "nearby", i'd like to try scanning at a larger radius." and they will want to increase from 70 to say maybe 100 at first. Then 120. And continue increasing while decreasing their account usage and checking that they're still getting the right amount of spawns per hour. Then once they make it up to 200 and they still aren't missing anything, they have now decreased account usage by 8x and can use those other 85% of their now unused accounts to go scan other regions.

But all that can't happen unless they have the ability to gradually and manually increase their step size.

If you guys dont want to implement it its no trouble to me. I'm already using it this way. It just seemed like something that would be useful to share; I'm confident you all can manually override your step size but some users aren't as savvy, you know?

PR #979, opened 2016-08-25
Added account rotation

Xcelled:
## Description
This pull request adds the ability to define a number of 'reserve' accounts. These workers are pooled and not assigned to scan. Randomly (about once every hour), or when a worker detects a possible ban, the scanning worker pulls a replacement out of the pool and retires itself into the pool.

This results in accounts being swapped in and out as the scanner runs.

image
image

## Motivation and Context
Workers running continually is a dead giveaway of a scanner. Also, this allows the scanner to more elegantly recover from the loss of an account without a loss of capacity.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -27,7 +27,7 @@ import geopy
 import geopy.distance
 
 from operator import itemgetter
-from threading import Thread
+from threading import Thread, Event
 from queue import Queue, Empty
 
 from pgoapi import PGoApi
@@ -168,6 +168,10 @@ def search_overseer_thread(args, method, new_location_queue, pause_bit, encrypti
 
     log.info('Search overseer starting')
 
+    # Shuffle account order
+    acc_list = list(args.accounts)
+    random.shuffle(acc_list)
+    worker_reserve = Queue()
     search_items_queue = Queue()
     threadStatus = {}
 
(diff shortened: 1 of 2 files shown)


DiscoTim:
this is good. getting consistent bans with one setup every 2 days. running 1 worker, step-limit 2.

jnovack:
I would rather the accounts be in the database. You can still hold X number in reserve, but if you are going this far, just pull accounts from the database.

Suggested columns:

* `name`, `password`
* `enabled` - available in the pool.
* `ready` - not in use.  When a worker uses an account, sets `ready` to `false`. Upon return, sets `ready` to true.
* `process` - `hostname:PID` of running process USING the account.  Great for double-checking if an account has prematurely crashed and not checked back in. 

Benefits:

* Can add/update/remove accounts without restart process.
* Multiple processes can use the same "pool" of accounts.

Christian-X:
I'd love to report a "travelled" distance with each account as well.. Some hints have been made, that a ban by niantic could be issued because of to many km travelled without any levelup/pokemon caught/...

Having a "travelled till ban" measurement in the workers would help to understand whether there really is a connection.

DiscoTim:
@Christian-X thats interesting and does make some sense. not directly related to this pr, but interesting.

Xcelled:
@jnovack that's a much larger scale change and would break a lot of people's setup. If accounts were in the database, any one of multiple workers could pull it out. That becomes a problem when you're running multiple scanners - a worker could suddenly jump a huge number of kilometers.

Also, as of right now, this just works by pausing the thread associated with the account. True account switching is significantly more annoying to make work (I tried it first)

Imo there are much larger architectural changes that should be made first, which will enable features like that to be added more painlessly. (For example, making the workers into a class instead of random methods that get passed everything)

edgarborja:
This idea is great, and the comments make it even better. Here are my thoughts.

@jnovack The accounts being in the DB would be a great improvement, and it would reduce the time it takes to manage the tool a lot.

I would add the following columns: date/time when the account started being used, date/time when the account was disposed.

@Xcelled If the DB contains a column identifying the pid where it's being used, then a 2nd worker will not use that account; maybe we could also store the last location where the account was used, or the start location of the last worker that used it, to prevent the account from traveling too far in the case of a restart.

Xcelled:
@edgarborja if the accounts are tied to PIDs you essentially have the existing system with a larger configuration hassle. The "last scan point/time" column is a neat idea. It could be used to preferentially select workers based on velocity.

I still maintain that these are all things to implement after the searching architecture is cleaned up.

n3cr0cr0w:
PID wouldn't make much sense if your workers are running on multi servers. I would store last scanned area. If it were beehive the configured center location the hive. Column for account status, active, inactive, soft-banned and perma-banned.

FlarumHostingCo:
if I'm running an account 2 active workers and would like to put 2 on reserve to swap out hourly, I use --acc-reserve = 2 ?

Christian-X:
Having the workers in the db could actually lessen the teleport problems:

* Record last_location, last_scan_timestamp, on each scan (maybe cached into a in-mem cache, and only flushed every xx minutes)
(* Record total distance travelled into each account as well)

Whenever a worker selects an account to be used: check distance between account last_scan, and proposed next scan location, divide it by time, put on a speed limit. 
Bark & warn, when no worker is available that meets the criteria - Maybe the worker should "reserve" the account for next use, but be able to switch to a different account.

This would even allow to hot-add accounts to a running scanner...

And even on a restart of the app, no worker would ever teleport, but travel at 30m/s (or so) max.

I'm in favor of adding accounts to the db and draw them from that pool

justsomedudeonthenet:
This method of changing accounts will not work well with the separate scan queue per worker that others are working on.  I prefer #996 (of course, I wrote it, so I am extremely biased).

LordNeoZ:
If the accounts are in a database, there should be some kind of separate database for the accounts. Otherwise, shared maps, beehives and stuff like that could take an account that was just put to rest, not only activating it again, but also teleporting it maybe way far away from the "log out" point.

foltincsaba:
Thanks for this! :D It works great. This was the only thing that solved the freacking softbanning wave that started recently.

(I'm using -st 60 -ss -acc-reserve 40 with 120 accounts)

xioxxio:
Does this work with a beehive, a sh file?

Zerotask:
Independet of this PR, it would be good to just have a csv file with accounts where the workers get accounts and if an account is perma banned, than they remove it and take another one.

Xcelled:
Since #996 is merged, no sense in keeping this.

PR #990, opened 2016-08-25
Add a spawnpoint clustering script to Tools

gabrielsoldani:
## Description
This is a different spawnpoint clustering algorithm than the ones in #868 and #765.

In my computer it compresses 7001 spawnpoints into 4834 clusters in 40 seconds.

## Motivation and Context
#868 is too slow
#765 picks a spawnpoint and removes all spawnpoints near it
We should instead be scanning the centroid of the clusters, so more spawnpoints can be fitted inside each cluster.

## How Has This Been Tested?
It includes a `test` function that verifies all spawnpoints in a cluster are within `radius` meters from its centroid, and the difference between the earliest and latest spawn time in the cluster does not exceed `time_threshold`

## Types of changes

## Checklist:

### DIFF
--- a/Tools/Spawnpoint-Clustering/README.md
+++ b/Tools/Spawnpoint-Clustering/README.md
@@ -1,0 +1,8 @@
+## Usage
+
+```
+python ./cluster.py spawnpoints.json -os spawnpoints.compressed.json -r 70 -t 180
+```
+
+Clusters all spawnpoints in `spawnpoints.json` within 70 meters of eachother and within 180 seconds of spawn time and saves the output to `spawnpoints.compressed.json`
+
--- a/Tools/Spawnpoint-Clustering/cluster.py
+++ b/Tools/Spawnpoint-Clustering/cluster.py
@@ -1,0 +1,179 @@
+import argparse
+import json
+import time
+import random
+
+import utils
+
+class Spawnpoint(object):
+    def __init__(self, data):
+        self.spawnpoint_id = data['spawnpoint_id']
+        try:
+            self.position = (float(data['latitude']), float(data['longitude']))
+        except KeyError:
+            self.position = (float(data['lat']), float(data['lng']))
+        self.time = data['time']
+        
+class Spawncluster(object):
+    def __init__(self, spawnpoint):
+        self._spawnpoints = [spawnpoint]
+        self.centroid = spawnpoint.position
+        self.min_time = spawnpoint.time
+        self.max_time = spawnpoint.time
+        
+    def __getitem__(self, key):
+        return self._spawnpoints[key]
+    
+    def __iter__(self):
+        for x in self._spawnpoints:
+          yield x
+      
+    def __contains__(self, item):
+        return item in self._spawnpoints
+          
+    def __len__(self):
+        return len(self._spawnpoints)
+        
+    def append(self, spawnpoint):
+        # update centroid
+        f = len(self._spawnpoints) / (len(self._spawnpoints) + 1.0)
+        self.centroid = utils.intermediate_point(spawnpoint.position, self.centroid, f)
+        
+        self._spawnpoints.append(spawnpoint)
+        
+        if spawnpoint.time < self.min_time:
+            self.min_time = spawnpoint.time
+            
+        if spawnpoint.time > self.max_time:
+            self.max_time = spawnpoint.time
+            
+    def simulate_centroid(self, spawnpoint):
+        f = len(self._spawnpoints) / (len(self._spawnpoints) + 1.0)
+        new_centroid = utils.intermediate_point(spawnpoint.position, self.centroid, f)
+        
+        return new_centroid
+            
+def cost(spawnpoint, cluster, time_threshold):
+    distance = utils.distance(spawnpoint.position, cluster.centroid)
+
+    min_time = min(cluster.min_time, spawnpoint.time)
+    max_time = max(cluster.max_time, spawnpoint.time)
+
+    if max_time - min_time > time_threshold:
+        return float('inf')
+
+    return distance
+    
+def check_cluster(spawnpoint, cluster, radius, time_threshold):
+    # discard infinite cost or too far away
+    if cost(spawnpoint, cluster, time_threshold) > 2 * radius:
+        return False
+
+    new_centroid = cluster.simulate_centroid(spawnpoint)
+    
+    # we'd be removing ourselves
+    if utils.distance(spawnpoint.position, new_centroid) > radius:
+        return False
+    
+    # we'd be removing x
+    if any(utils.distance(x.position, new_centroid) > radius for x in cluster):
+        return False
+        
+    return True
+    
+            
+def cluster(spawnpoints, radius, time_threshold):
+  clusters = []
+  diameter = 2 * radius
+  
+  for p in spawnpoints:
+    if len(clusters) == 0:
+      clusters.append(Spawncluster(p))
+    else:
+      c = min(clusters, key=lambda x: cost(p, x, time_threshold))
+      
+      if check_cluster(p, c, radius, time_threshold):
+      #if check_cluster_try(p, c, 2*radius, time_threshold):
+        c.append(p)
+      else:
+        c = Spawncluster(p)
+        clusters.append(c)
+        
+  return clusters
+  
+def test(cluster, radius, time_threshold):
+    assert cluster.max_time - cluster.min_time <= time_threshold
+    
+    for p in cluster:
+        assert utils.distance(p.position, cluster.centroid) <= radius
+        assert cluster.min_time <= p.time <= cluster.max_time
+        
+def main(args):
+    radius = args.radius
+    time_threshold = args.time_threshold
+    
+    with open(args.filename, 'r') as f:
+        rows = json.loads(f.read())
+
+    spawnpoints = [Spawnpoint(x) for x in rows]
+      
+    print 'Processing', len(spawnpoints), 'spawnpoints...'
+
+    start_time = time.time()
+    clusters = cluster(spawnpoints, radius, time_threshold)
+    end_time = time.time()
+
+    print 'Completed in {:.2f} seconds.'.format(end_time - start_time)
+    print len(clusters), 'clusters found.'
+    print '{:.2f}% compression achieved.'.format(100.0 * len(clusters) / len(spawnpoints))
+    
+    try:
+        for c in clusters:
+            test(c, radius, time_threshold)
+    except AssertionError:
+        print 'error: something\'s seriously broken.'
+        raise
+
+    # clusters.sort(key=lambda x: len(x))
+
+    if args.output_clusters:
+        rows = []
+        for c in clusters:
+            row = dict()
+            row['spawnpoints'] = [x.__dict__ for x in c]
+            row['lat'] = c.centroid[0]
+            row['lng'] = c.centroid[1]
+            row['min_time'] = c.min_time
+            row['max_time'] = c.max_time
+            rows.append(row)
+     
+        with open(args.output_clusters, 'w') as f:
+            f.write(json.dumps(rows, indent=4, separators=(',', ': ')))
+    
+    if args.output_spawnpoints:    
+        rows = []
+        for c in clusters:
+            row = dict()
+            # pick a random id from a clustered spawnpoint
+            # we should probably not do this
+            row['spawnpoint_id'] = random.choice(c).spawnpoint_id
+            row['lat'] = c.centroid[0]
+            row['lng'] = c.centroid[1]
+            # pick the latest time so earlier spawnpoints have already spawned
+            row['time'] = c.max_time
+            rows.append(row)
+        
+        with open(args.output_spawnpoints, 'w') as f:
+            f.write(json.dumps(rows, indent=4, separators=(',', ': ')))
+          
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Cluster close spawnpoints.')
+    parser.add_argument('filename', help='Your spawnpoints.json file.')
+    parser.add_argument('-os', '--output-spawnpoints', help='The filename to write compressed spawnpoints to.')
+    parser.add_argument('-oc', '--output-clusters', help='The filename to write cluster data to.')
+    parser.add_argument('-r', '--radius', type=float, help='Maximum radius (in meters) where spawnpoints are considered close (defaults to 70).', default=70)
+    parser.add_argument('-t', '--time-threshold', type=float, help='Maximum time threshold (in seconds) to consider when clustering (defaults to 180).', default=180)
+    
+    args = parser.parse_args()
+    
+    main(args)
--- a/Tools/Spawnpoint-Clustering/cluster.py
+++ b/Tools/Spawnpoint-Clustering/cluster.py
@@ -7,12 +7,26 @@ import utils
 
 class Spawnpoint(object):
     def __init__(self, data):
-        self.spawnpoint_id = data['spawnpoint_id']
+        # not needed but useful for debugging
+        self.spawnpoint_id = data.get('spawnpoint_id') or data.get('sid')
+        
         try:
             self.position = (float(data['latitude']), float(data['longitude']))
         except KeyError:
             self.position = (float(data['lat']), float(data['lng']))
+        
         self.time = data['time']
+        
+    def serialize(self):
+        obj = dict()
+
+        if self.spawnpoint_id != None:
+            obj['spawnpoint_id'] = self.spawnpoint_id
+        obj['latitude'] = self.position[0]
+        obj['longitude'] = self.position[1]
+        obj['time'] = self.time
+
+        return obj
         
 class Spawncluster(object):
     def __init__(self, spawnpoint):
@@ -81,7 +95,6 @@ def check_cluster(spawnpoint, cluster, radius, time_threshold):
         
     return True
     
-            
 def cluster(spawnpoints, radius, time_threshold):
   clusters = []
   diameter = 2 * radius
@@ -93,7 +106,6 @@ def cluster(spawnpoints, radius, time_threshold):
       c = min(clusters, key=lambda x: cost(p, x, time_threshold))
       
       if check_cluster(p, c, radius, time_threshold):
-      #if check_cluster_try(p, c, 2*radius, time_threshold):
         c.append(p)
       else:
         c = Spawncluster(p)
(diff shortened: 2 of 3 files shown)


ottowayne:
Haven't analyzed the code in detail but the results do seem nice.
> python cluster.py ../../spawns.json -os spawns_compressed.json -r 70 -t 20
> Processing 1316 spawnpoints...
> Completed in 2.34 seconds.
> 1147 clusters found.
> 87.16% compression achieved.
>
> python cluster.py ../../spawns.json -os spawns_compressed.json -r 70 -t 30
> Processing 1316 spawnpoints...
> Completed in 2.44 seconds.
> 1098 clusters found.
> 83.43% compression achieved.
>
> python cluster.py ../../spawns.json -os spawns_compressed.json -r 70 -t 90
> Processing 1316 spawnpoints...
> Completed in 1.89 seconds.
> 928 clusters found.
> 70.52% compression achieved.
>
> python cluster.py ../../spawns.json -os spawns_compressed.json -r 70 -t 120
> Processing 1316 spawnpoints...
> Completed in 1.80 seconds.
> 871 clusters found.
> 66.19% compression achieved.

MangoScango:
This is definitely nice. With a 3 minute window it took me from 1799 to 1312. Pretty painless optimization. It needed an edit for my spawns.json though. It was generated with TBTerra's spawnScan and used "sid" instead of "spawnpoint_id".

clvv:
A forseeable problem is with `--jitter` turned on. A plausible solution is to just run this upon launch and take jitter option into account.

gabrielsoldani:
My idea was that you could account for jittering with a smaller radius (so, for example, if you're adding a 10m jitter, `-r 60` should work).

justsomedudeonthenet:
Please if possible make it work for json files that don't have the spawnpoint ID.

Depending on what tools were used to create it, the json file could have `spawnpoint_id`, `sid` or nothing.  The fields required by the main program are just `lat`, `lng`, and `time`.

I am super happy to see this as an optional script.  There's no need to put this in the main program.  The people who need this will be willing to go to the extra couple minutes of effort to compress it once.

gabrielsoldani:
@justsomedudeonthenet Fixed :wink:

averylarry:
Just gonna throw this idea out there.
Could we incorporate jitter directly into spawnpoint scanning compression?
Uncompressed spawnpoint scanning could use a jitter of 69.9m.  However, once you compress the spawnpoints, each new scan cluster will have a unique maximum jitter based on how spread-out the grouped spawnpoints happen to be.  If the compression algorithm also calculated maximum jitter, and stored that in the .json, and then the spawnpoint scanning could use that . . .

It's a lot to ask, I'm thinking.  It seems like a good idea from a jitter/randomizing/more-human-looking perspective.  OK -- I'm probably just trying way too hard.

DoktorJ:
How do I run this pull? I've merged it into my local copy, but it looks like it's just operating on the regular spawnpoint scan (so if there's a tight cluster of, say, 5 spawn points it scans the same area 5 times). Do I need to use something other than the `-ss` flag?

gabrielsoldani:
@DoktorJ This patch does not automatically compress spawnpoints when you launch `./runserver.py` with `-ss`. You must run `./Tools/Spawnpoint-Clustering/cluster.py` before and configure your `./runserver.py` instances accordingly.

`./Tools/Spawnpoint-Clustering/README.md` has instructions on how to run it.

All you have to do afterwards is point the searchers to use the newly generated `spawnpoints.compressed.json` file instead of your uncompressed spawnpoints file.

Note that it will still scan the same area multiple times if the `time` between them is bigger than `time_threshold` (defaults to 180 seconds, can be set with the `-t` command-line arg).

topcatzero:
Nicely done!

I noticed a few clusters consisting of a single spawn point. Some of these had spawn points from larger clusters within the supplied distance. Have you considered finding an algorithm with less greedy clusters, resulting in more evenly distributed spawn points? I'm not sure how complex that would be, but I'm thinking it might look less bot-like if the number of scans centered directly on a spawn point are minimized.

beccasafan:
Here are my numbers comparing this to the [compress_spawnpoints.py](https://gist.github.com/justsomedudeonthenet/f1f4f27d4aa90a616b517e8ced949d88) algo: https://gist.github.com/beccasafan/083ad91cbf0cedb07a6983a4d9702e7a

It doesn't look like it's starting from the biggest clusters. It'd increase the computation, but might be a worthwhile optimization? (Or I just missed that while glancing at it?)

beccasafan:
@averylarry yes, both algorithms take in a time and pair that with the location.

averylarry:
@beccasafan I wasn't clear -- I wanted to check that you and others understood that the algorithms account for time.
For instance -- IF topcatzero is simply looking at the scan circle on the map and thinks that every spawnpoint inside the circle is getting scanned -- that is not correctly taking into account that the "cluster" of spawnpoints is not simply physically close but also within a certain timeframe of each other.

I didn't understand what you meant by not starting "from the biggest clusters" (and I still might not).  I thought it maybe referred to the actual scanning starting with the largest clusters, which clearly wouldn't account for time.  Now I think you're meaning the algorithm should start with the largest physical clusters theorizing that there are more potential physical + time clusters in a dense area.  As such, I think that is hinted at in the code:

169            # pick a random id from a clustered spawnpoint
170            # we should probably not do this

Sorry -- I wasn't trying to sound like I was stupid nor was I wanting to be condescending.  ;-)

beccasafan:
@averylarry ah, yeah I'm using the term "cluster" to match both on time and distance. I was just curious if it was sorting by the number of spawnpoints in a cluster before removing them. Maybe it should start from the lowest though...that might eliminate some more that way because you'd end up with less single clusters perhaps? I'd have to think about that much more.

gabrielsoldani:
@beccasafan I'm a little confused about what you mean, since the algorithm never removes any spawnpoint.

The order in which the spawnpoints are clustered hardly matters because, naturally, each cluster's centroid moves a little as more and more spawnpoints are added to it. To test it, run `random.shuffle(spawnpoints)` before the algorithm starts and you'll the difference in compression between all the different orders will never be higher than 2% with real-world data.

shadi1989:
There's an issue with the compression calculation.
72.89% is kept, not compressed. 27.11% is removed so only 27.11% compression.
If it was 72.89% compressen you would keep a lot less spawnpoints.

https://images-2.discordapp.net/.eJwFwdENhCAMANBdGIDWckh1G4IEjUoJ1Phxud3vva95-mVWs6u2sQJsx0jSNztUeizZFpFy5diOYZPcEFVj2u9cdQDhxzkOc1gIg-fAHogI2TFPC6Kb0BPDU88qb7WtFvP7Awf2It0.SU_wQHswbP4XuCdUjXAbW7po9XM.png

averylarry:
Depends on your interpretation.  It is compressed to 72.89% of it's original size.  I don't think it's wrong, but it could be misunderstood.  I would, though, suggest changing the wording and reversing the number so a bigger percentage=better compression.  Something like "removed 25% by compression" or "compressed to 45% fewer scan points".

gabrielsoldani:
I was using the 'compression ratio' that shows when you're using software
like 7-zip and WinRAR, which is what I assume most people are familiar
with. I agree that it could be reworded to avoid confusion, though.

jchristi:
With this option, because its creating a clustered spawn point file to be used as input for pogom, that means that if you want to view spawn points on the map, it will actually show cluster points instead?

averylarry:
No.  Spawnpoints on the map are not affected.  The generated file is a list of scanning locations instead of spawnpoints.  The only gui representation is looking at scanned locations.

justsomedudeonthenet:
I have been using this extensively and it works great.

Obihoernchen:
I get some strange error after some time. I added some debug:

```
spawnpoint_id: 47a8523879f
spawnpoint_position: (52.56669878525873, 13.385848618759656)
cluster_centeroid: (52.49566257972049, 13.422290675560475)
lat1: 0.917461970707  lat2: 0.916222155032  lon1: 0.233627131571  lon2: 0.234263165448
spawnpoint_id: 47a8523879f
spawnpoint_position: (52.56669878525873, 13.385848618759656)
cluster_centeroid: (52.50582956386427, 13.460281795661672)
lat1: 0.917461970707  lat2: 0.91639960238  lon1: 0.233627131571  lon2: 0.234926235581
spawnpoint_id: 47a8523879f
spawnpoint_position: (52.56669878525873, 13.385848618759656)
cluster_centeroid: (52.5666987852587, 13.3858486187597)
lat1: 0.917461970707  lat2: 0.917461970707  lon1: 0.233627131571  lon2: 0.233627131571
  File "cluster.py", line 200, in <module>
    main(args)
  File "cluster.py", line 138, in main
    clusters = cluster(spawnpoints, radius, time_threshold)
  File "cluster.py", line 109, in cluster
    c = min(clusters, key=lambda x: cost(p, x, time_threshold))
  File "cluster.py", line 109, in <lambda>
    c = min(clusters, key=lambda x: cost(p, x, time_threshold))
  File "cluster.py", line 74, in cost
    distance = utils.distance(spawnpoint.position, cluster.centroid)
  File "/home/obi/PokemonGo/PokemonGo-Map/Tools/Spawnpoint-Clustering/utils.py", line 14, in distance
    return acos(sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)) * R
ValueError: math domain error
```
Problem is `(sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)` is > 1
So acos(>1) fails.

Any idea why?

Imoq:
I got 3 dumps made correctly but found an issue on the 4th:

```
root@raspberrypi:~/PokemonGo-Map/spawns# python cluster.py borely-spawns.json -os borely-spawns-compressed.json -r 70 -t 180
Processing 348 spawnpoints...
Completed in 0.83 seconds.
145 clusters found.
41.67% compression achieved.
root@raspberrypi:~/PokemonGo-Map/spawns# python cluster.py perier-spawns.json -os perier-spawns-compressed.json -r 70 -t 180
Processing 486 spawnpoints...
Completed in 1.97 seconds.
192 clusters found.
39.51% compression achieved.
root@raspberrypi:~/PokemonGo-Map/spawns# python cluster.py prefecture-spawns.json -os prefecture-spawns-compressed.json -r 70 -t 180
Processing 376 spawnpoints...
Completed in 1.20 seconds.
144 clusters found.
38.30% compression achieved.
root@raspberrypi:~/PokemonGo-Map/spawns# python cluster.py vieuxport-spawns.json -os vieuxport-spawns-compressed.json -r 70 -t 180
Processing 526 spawnpoints...
  File "cluster.py", line 197, in <module>
    main(args)
  File "cluster.py", line 135, in main
    clusters = cluster(spawnpoints, radius, time_threshold)
  File "cluster.py", line 106, in cluster
    c = min(clusters, key=lambda x: cost(p, x, time_threshold))
  File "cluster.py", line 106, in <lambda>
    c = min(clusters, key=lambda x: cost(p, x, time_threshold))
  File "cluster.py", line 71, in cost
    distance = utils.distance(spawnpoint.position, cluster.centroid)
  File "/root/PokemonGo-Map/spawns/utils.py", line 14, in distance
    return acos(sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)) * R
ValueError: math domain error
root@raspberrypi:~/PokemonGo-Map/spawns#
```

Obihoernchen:
It's the very same error

Obihoernchen:
It looks like `sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)` is greater than 1 sometimes. And acos for more than 1 doesn't work.

If I change the code like this (just a ugly fix which is probably not correct) it works without a failure:
```diff
diff --git a/Tools/Spawnpoint-Clustering/utils.py b/Tools/Spawnpoint-Clustering/utils.py
index a2af21b..a064c61 100644
--- a/Tools/Spawnpoint-Clustering/utils.py
+++ b/Tools/Spawnpoint-Clustering/utils.py
@@ -10,8 +10,12 @@ def distance(pos1, pos2):
     lon1 = radians(pos1[1])
     lat2 = radians(pos2[0])
     lon2 = radians(pos2[1])
+    #print ("acos(sin("+str(lat1)+")*sin("+str(lat2)+") + cos("+str(lat1)+")*cos("+str(lat2)+") * cos("+str(lon2)+"-"+str(lon1)+")) * "+str(R))
+    step=sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)
+    if step > 1:
+        return 0.0
+    return acos(step) * R

-    return acos(sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)) * R

 def intermediate_point(pos1, pos2, f):
     if pos1 == pos2:
@@ -22,8 +26,12 @@ def intermediate_point(pos1, pos2, f):
     lat2 = radians(pos2[0])
     lon2 = radians(pos2[1])

-    delta = acos(sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1))
-
+    step=sin(lat1)*sin(lat2) + cos(lat1)*cos(lat2) * cos(lon2-lon1)
+    if step <= 1:
+        delta = acos(step)
+    else:
+        delta = 0
+
     if delta == 0: # too close
         return pos1 if f < 0.5 else pos2
```

gabrielsoldani:
I suspect it's calling acos(1.000000000000187) or something similar because of floating point precision. Can anyone confirm this is the case?

@Obihoernchen since you can reproduce this issue, can you please print `step` when it's greater than `1` to see if that's the case?

Obihoernchen:
@gabrielsoldani okay that's very interesting I added print step and also print lat1,2 lon1,2.
`step` is printed 1.0 ?! I don't get that because than it shouldn't go to this if/else case but ok xD

What's interesting is that every time `lat1 == lat2 and lon1 == lon2` there is this issue (see below).
It's printing `step` and `lat1 lat2 lon1 lon2` for `distance` and `intermediate_point` function if step > 1.
```
$ python cluster.py -os spawnpoints.clustered.json spawnpoints.json -r 70 -t 840
Processing 18986 spawnpoints...
1.0
1.0
0.917461970707 0.917461970707 0.233627131571 0.233627131571
1.0
1.0
0.917461970707 0.917461970707 0.233627131571 0.233627131571
1.0
1.0
1.0
0.916738666026 0.916738666026 0.233771025827 0.233771025827
1.0
1.0
0.916738666026 0.916738666026 0.233771025827 0.233771025827
1.0
1.0
1.0
0.917402372657 0.917402372657 0.236833507926 0.236833507926
1.0
1.0
0.917402372657 0.917402372657 0.236833507926 0.236833507926
1.0
1.0
1.0
1.0
Completed in 531.63 seconds.
8961 clusters found.
47.20% compression achieved.
```

Maybe we should just do `if step > 1: step = 1.0`?

gabrielsoldani:
@Obihoernchen Likely because pos1 and pos2 are different, but they become equal after the conversion to radians. Thanks for your output. See if https://github.com/gabrielsoldani/PokemonGo-Map/commit/691e016e1f8fc25521a23fc65d395c07f28acfa9 fixes it.


Btw, GitHub experts, do I need to make a separate PR for this?

Obihoernchen:
works fine now thank you 👍

jchristi:
@gabrielsoldani yes, your branch was merged at point of a previous commit. You should submit a new PR, but you can still use the same source branch since its now at a later commit.

thatsweater:
I believe the convention for "percentage compression" is backwards, or at least per what you would expect. I ran a series of compressions, dumping to different files, and the general trend i see here is as follows,

10 seconds:
> Processing 10732 spawnpoints...
Completed in 158.51 seconds.
6175 clusters found.
57.54% compression achieved.

20 seconds:
>Processing 10732 spawnpoints...
Completed in 152.25 seconds.
5656 clusters found.
52.70% compression achieved.

30 seconds:
>Processing 10732 spawnpoints...
Completed in 146.03 seconds.
5334 clusters found.
49.70% compression achieved.

As you can see, even though for 30 seconds there are less clusters as you would expect, it states that a lesser compression was achieved than when running 10 seconds, which is counterintuitive. I would think it would follow 1-(current compression percentage).

Gallifery:
Both the spawnpoint and this cluster technique make the scan points scattered. Current direction of improvement has been trying to maxize the size of area each worker can cover per unit time. However, after applying this technique, I believe the 'teleport' effect becomes a concern and I think it's time to review the problem. So, the fundamental benefit from the spawnpoint scan and this technique is that the number of API calls are reduced to cover a same area, meaning with the same number of accounts, you can scan a larger area. The drawback is that each worker will behave like teleporting from one point to another. Meanwhile, probably for most users, the area to scan is not very big. I guess most of the users just want to scan their neighbourhood or something up to 10 km^2. And for that, the combination of these 2 techniques can assure maybe 15 accouts is enough. Instead of an even larger area, I think it is more meaningful to make the scan behave more like regular player. So, for example, if the user has 5 additional accounts but it is not useful for him to scan a larger area, it makes more sense if he can feed the program with 20 accounts while the program can intelligent assign the search task to each of the 20 worker. By that, I mean techniqeus like each worker should be limited to an area locally and if it has finished scanning that area, it standby at that area as if an player is standing still and wait for new scan requests assigned by the program. Furthermore, also to avoid the teleport problem, the pgogram should be able to calculate a route of the worker to walk/run from one point to another. Of course, this would be slower than teleport, but it would behave more like human. The general idea is to add some redundant account so that each worker can be configured not to work in an aggressive way.

gabrielsoldani:
@Gallifery Defining the order in which spawnpoints get scanned to minimize
teleports is out of the scope of this commit. This is just a single
component of something bigger which could handle the issues you mentioned:

1. Compress the spawnpoints to find the minimum acceptable amount of scan
locations (this algorithm).
2. Partition the compressed spawnpoints in k groups based on their location
so you can assign each group to one of your k workers (k-means clustering).
3. For each of the k groups, calculate the optimal path that visits every
spawnpoint at the appropriate time with the least possible distance walked
(traveling salesmen problem with time windows).

averylarry:
Current improvements are not trying to maximize the area that a worker can cover per unit of time.  It is about getting spawn information as quickly as possible while minimizing account bans.
Teleport has always always been a concern (beehive scanning path changed to a circuit, default scan delay).  This compression helps.  There are numerous other PR's that address additional aspects.  Off the top of my head in increasing complexity: speed limit, clustering (where a given account stays in a certain area instead of flying all over the total scan area), and traveling salesman.

neutron666:
What is this doing ? Is there any documentation about it how to use it right ? All i see is, it's decreasing the number of spawn points i already scanned...how to use it ?!

averylarry:
@neutron666 It's in the tools\spawnpoint-clustering folder.  cluster.py runs it -- and there's a readme.

PR #995, opened 2016-08-25
Experimental spawn-point clustering

TBTerra:
### I'm currently absent from this project due to focussing on my masters degree, by all means use this code for development
# Original post
Using k++ means clustering to break the big spawn list into a small list covering a small area for each worker. does not eliminate teleporting (needs a simple speed limiter for that) but limits the distance teleported to within the cluster

## Description
Instead of loading the list of spawns and then adding them to a queue shared between all workers (where a worker may be tasked with teleporting across the entire scan area), instead the list of spawns is clustered with each worker handling the entire of one cluster, this means that the worker will only ever have to travel across the cluster (in a city with a cluster of 100spawns, the cluster size is less than 1km)

## Motivation and Context
This is one step towards tackling the issue of accounts being banned for teleporting across the map, it achieves this without having to add excessive delays that a usual speed limiter may have to as well

When  combined with a simple speed limiter it should both minimize teleport bans, whilst keeping a high spawns to workers ratio

## How Has This Been Tested?
The clustering algorithm has been tested on datasets of up to 10k spawns, and the full implementation is being tested by various members of the community

example clustering of 5800 spawns split into 60 clusters (note that there are only 18 colors so some clusters are only differentiated by their letter)

## Types of changes

## Checklist:

this code is not compatible with pr 901

### DIFF
--- a/docs/advanced-install/amazon-ecs.md
+++ b/docs/advanced-install/amazon-ecs.md
@@ -12,7 +12,6 @@ Amazon ECS is essentially managed docker allowed you to run multi-container envi
 
 ## Process
 
-
 In the AWS ECS console create a Task Definition with the JSON below. You will need to set the following values:
 
 * `POKEMON_USERNAME` - username for pokemongo
--- a/docs/advanced-install/docker.md
+++ b/docs/advanced-install/docker.md
@@ -12,11 +12,11 @@ For your first install, there's only a few steps. Let's get to it!
 
 ### Start the Map 
 
-You'll need to change the command line params. If you don't know what they are, you can run `docker run --rm chuyskywalker/pokemongo-map -h` and you'll get the full help text. The command below examples out a very basic setup.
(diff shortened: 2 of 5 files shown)


Infernoman:
love you babe <3 glad to see you ALMOST back lol

presidenten:
Do the clustering handle areas that are separated by some space? Like two sides of a lake or if you have points of two towns?

presidenten:
Also, would it be possible to cluster first, and then run this as search algorithm #787 ?

presidenten:
I only get one cluster with 10000+ spawns. Am I missing something?

justinkwaugh:
Yes, this plus #787 would be pretty ideal

justj0sh:
Can someone please make an updated branch of this

primaxius:
Closed as not compatible with current code base. Please resubmit if re-written to be compliant.

thatsweater:
i pray

PR #996, opened 2016-08-26
Allow workers to get a new account when banned or working too long

justsomedudeonthenet:
## Description
Accounts are added to a queue, and workers pull an account from the queue when needed. 

This adds 3 parameters:

`-w/--workers` specified how many search workers to create.  If unspecified, it defaults to the number of accounts, mimicking existing behaviour.

`-asi/--account-search-interval` specifies how long a worker should keep running before switching accounts.  Defaults to 2 hours, but is disabled if -w is not specified.  Once it reaches this time, the worker will switch to a new account, and the old account will be left to rest.  Accounts with too many consecutive failures will also be put to rest.

`-ari/--account-rest-interval` specified how long accounts must rest before being put back in the queue.

Also adds a page to the status screen showing which accounts are on hold, and why they are on hold.

## Motivation and Context
Right now, workers with a banned account simply sit there, doing nothing.  This decreases the number of active workers, and leads to getting behind on scans, and having fewer workers moving faster.

This allows workers to continue working if you have enough accounts.  Rotating accounts regularly may also help avoid bans.

This will also help immensely with any of the upcoming queue-per-worker clustering methods people are working on, since with those if a worker dies, there is an entire region that will not get scanned.  Switching accounts on the fly will solve that.

## How Has This Been Tested?
Ran it against the fake API for a bit, as well as run for several days as my main scanner.
## Types of changes

## Checklist:


ottowayne:
Related to #979 I guess

justsomedudeonthenet:
I hadn't seen 979 when I wrote this.

I prefer my method (but I am of course biased).  His looks to start a thread for every account, and put some to sleep.  If you want to have a lot of spare accounts, that is a lot of extra threads having to run.  Mine just starts the number of threads needed, and doesn't require any extra imports.

His also will conflict with queue per worker type scheduling, where as mine improves it when it comes.

pheiberg:
Nice improvement!

Why is there a need to sleep 15 minutes when getting a new account for the worker (provided that there are other accounts in the account queue)?

justsomedudeonthenet:
The main reason for waiting is to ensure the account isn't used for at least 15 minutes, and in the event that a majority of the accounts get banned it prevents it from flipping through accounts rapidly (potentially causing an IP ban, maybe).

pheiberg:
Ok, but that will cause a worker to be unavailable for 15 minutes instead of just the offending account. It might be a good idea to add a "available from" property attached to the account.

I've been thinking about coding an account pool functionality that takes last scan location, last scan time and account status into account and lets workers request a suitable account for a specific scan location to remedy banned accounts and teleportation issues. But I haven't gotten around to it yet.

justsomedudeonthenet:
@pheiberg That was an excellent idea.  The workers can now continue immediately, and a separate thread checks once in awhile for accounts that have cooled down long enough to be put back into the queue.

xioxxio:
Can I use this with a beehive?

pheiberg:
@xioxxio - yes, but you can't share accounts between different beehives. Each process needs its own unique list of accounts.

xioxxio:
Great what do I need to do.  I create a beehive using accounts.csv.  The beehive only utilizes say 50 accounts for the *.sh file.  Do you just leave the entire accounts.csv list in the beehive directory. Do you need to delete the accounts the hive is using. Please provide more information.

pheiberg:
I haven't used beehive myself, but as I understood it, the tool generates a shell script containing multiple invokations of the runserver.py with the accounts as command line parameters for each. The csv file isn't used by beehive after the tool is run initially to create the script. You either need to change the beehive script to distribute the accounts evenly or hand edit the  script file and add additional accounts and set the appropriate -w parameter.

justsomedudeonthenet:
Updated for latest develop, and added a -ps screen to show which accounts are on hold.

pyr0ball:
I want this merged. can haz pls?

mbg032373:
Make it pretty again:


`# How pretty
  status = '{:10} | {:5} | {:' + str(userlen) + '} | {:7} | {:6} | {:5} | {:7} | {:10}'`

justj0sh:
Pressing F doesn't show accounts on hold just FYI

mbg032373:
Make Hold screen look more like status screen?

```javascript
elif display_type[0] == 'failedaccounts':
            status_text.append('-----------------------------------------')
            status_text.append('Accounts on hold:')
            status_text.append('-----------------------------------------')
            status = '{:' + str(userlen) + '} | {:10} | {:20}'
            status_text.append(status.format('User ID', 'Hold Time', 'Reason'))

            for account in account_failures:
                status_text.append(status.format(account['account']['username'], time.strftime('%H:%M:%S', time.localtime(account['last_fail_time'])), account['reason']))
```

justsomedudeonthenet:
@justj0sh It works for me.  You have to press enter after the F.

justsomedudeonthenet:
@mbg032373 thanks for the suggestion!  Added.

mbg032373:
@justj0sh It works for me as well.  It does not show a list of spare accounts, but it will show any accounts that get put on hold.

mbg032373:
@justsomedudeonthenet I didn't show this, so my bad, but you have to move the `userlen` declaration outside of the `if` statement.

````javascript
# Create a list to hold all the status lines, so they can be printed all at once to reduce flicker
        status_text = []
        userlen = 4

        if display_type[0] == 'workers':
````

justsomedudeonthenet:
@mbg032373 ah, missed that.  Thanks.

mbg032373:
I get the following error with latest dev and latest changes:

From status view:
`Exception in search_worker using account _<username>_. Restarting with fresh account. See logs for details.`

From log view:
`Exception in search_worker under account _<username>_ Exception message: 'proxy'`

n3cr0cr0w:
I prefer accounts being stored in the database like #1037

Talhoo:
I have the exact same problem, using latest dev #1078

Exception in search_worker using account _<username>_. Restarting with fresh account. See logs for details.

From log view:
Exception in search_worker under account _<username>_ Exception message: 'proxy'

mbg032373:
@justsomedudeonthenet New commit works.  I've tested it out using no proxy and with 3 proxies.  Both work as expected.

justsomedudeonthenet:
@mbg032373 thanks for testing!  It seems to be working perfectly now.

Talhoo:
Works perfectly. thx !

I was wondering, if you go through all this trouble to let the workers appear a bit more human like and not let them be online 24/7.  Wouldn't it be better to have the working time (aka -asi) of a worker be a bit randomized ? Right now, the following happens : "X accounts log off at the same time, X accounts log on at the same time". Which probably could look a bit suspicious ?

mbg032373:
@Talhoo @justsomedudeonthenet could do something like this.....

````javascript
                if (args.account_search_interval is not None):
                    min = args.account_search_interval * 0.75
                    max = args.account_search_interval
                    search_interval = random.randint(min, max)
                    if (status['starttime'] <= (now() - search_interval)):
                        status['message'] = 'Account {} is being rotated out to rest.'.format(account['username'])
                        log.info(status['message'])
                        account_failures.append({'account': account, 'last_fail_time': now(), 'reason': 'rest interval'})
                        break
````

FrostTheFox:
blugh to merge or not to merge...

neutron666:
Does this work ? I have 50 accounts, specified 30 workers, i was awaiting that 30 accounts log in, do their work for let's say 1 hour (-asi) and then rotate. All that happens for me is that 30 workers are rotating all the time and no worker starts to search. All i get is:

One account logging in and then waiting to get new account from queue, account xxx is being rotated out to rest, waiting for new account from queue..this goes through all 50 accounts and then the rest time jumps in...

justj0sh:
Can you add something to rotate out when there's too many no items as well? I'm getting random soft bans and the workers just keep continuing which probably extends the ban.

justsomedudeonthenet:
@neutron666 i suspect you set -asi to 1?  It is measured in seconds (read --help, it says so).  Set it to 3600 if you want them to switch every hour.

neutron666:
Oh well :P Yes, now it works, Thanks.

azd2:
Would it be possible to add a solution like #1037 (save accs and their usage to mysql db) for not having to specify a csv-file with accs for each thread?
Would make life A LOT easier when running beehives with many threads.
When using only one thread with many workers this works like a charm.

luannpeixe:
It would be interesting to have the accounts of workers in the database and to be able to delete those that were banned by the status page.

justsomedudeonthenet:
@azd2 I would rather work on getting the beehive running all within one process so we don't need to run multiple copies.

ottowayne:
My server instance always starts with the first accountname empty. It tries to login, gets an error and replaces it with an actual account.

So it does work but I don't know where the empty account is coming from.

azd2:
@justsomedudeonthenet Yes right, i think it's already in progress by chuyskywalker in his pokepath-multiaccount branch. This great idea would indeed solve most of my problems if it finally works. 
You would still need to run multiple processes if you wanted to scan more than one area though, where using db for accs would still help with efficiency.

justsomedudeonthenet:
@azd2 you misunderstand me.  I want to make it possible to scan multiple separate locations within one instance.

azd2:
Uh didn't think that would be ´possible. Thank you, shutting up now and not keeping you from working on it any more :-)

n3cr0cr0w:
I would like something like #1037, I run several servers connected to one database. I don't want to split my accounts into several smaller lists, it would be nice to have only one python window running.

PR #1007, opened 2016-08-26
Jitter by Default, Except on Spawnpoint Scanning

jnovack:
Turn on `--jitter` by default, except with `-ss`

## Description

Jitter all search requests by default, except with spawnpoint scanning.

`--jitter` is kept as a variable, but prints a warning when used, it doesn't actually do anything.

## Motivation and Context

`--jitter` has been optional and working for a week now.

## How Has This Been Tested?

With and without spawnpoint scanning.

## Types of changes

## Checklist:
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.

Changes to the documentation will take place when `--jitter` is removed from the arguments.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -501,7 +501,7 @@ def search_worker_thread(args, account, search_items_queue, pause_bit, encryptio
                 check_login(args, account, api, step_location)
 
                 # Make the actual request (finally!)
-                response_dict = map_request(api, step_location, args.jitter)
+                response_dict = map_request(api, step_location, args.spawnpoint_scanning)
 
                 # G'damnit, nothing back. Mark it up, sleep, carry on
                 if not response_dict:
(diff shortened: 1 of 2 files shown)


slideup-benni:
why not for ss?

justsomedudeonthenet:
I approve of not doing it by default for spawn scanning.  Because many of us are compressing our spawnpoints, and jittering might move them outside of the compressed area that catches a few spawnpoints at once.

averylarry:
I really want it for spawn scanning.  The sole purpose of jitter (I believe) is to look less like a bot.  I would prefer that the compression algorithm take jitter into account -- spawnpoints within N minutes and 60 meters instead of 70 meters.  (I actually changed the code in #765 to do just that.)

A compromise would be to set the jitter for ss much smaller.  1 meter instead of 10 meters.  Or even smaller really -- it just shouldn't be the exact same every time.

averylarry:
Side question -- is the current jitter implementation used for both hex and ss?

rp000:
Having the compression algorithm take jitter into account is a good elegant way. 

Compression isn't in yet though so maybe this PR could go in with jitter enabled by default for everyone, and have compression take jitter into account with the compression PR?

justinkwaugh:
Also, why are we not using the jittered location in the pgoapi.set_position() call?

jnovack:
> Also, why are we not using the jittered location in the pgoapi.set_position() and the check_login() call?

Great question. You are the first to bring it up.

justinkwaugh:
Well I actually asked the same thing in your original pull request comments when you first added jitter, but no one responded :P

jnovack:
> Well I actually asked the same thing in your original pull request comments when you first added jitter...

If the devs thought it would be useful, I have to assume they would have decided to push for it.

primaxius:
Declined - Jitter works fine as it is, give the option the server operator.

PR #1009, opened 2016-08-26
Don't Allow Duplicate Gym Members

maddhatter:
## Description
If multiple threads/processes update a gym at the same time, it is possible for duplicate gym members to be inserted into the database. Making the changes to the `GymMember` table should prevent threads/processes from stomping on each other.

## Motivation and Context
Duplicate gym members were reported in the Discord and on #666.

## How Has This Been Tested?
I don't have any issues running this, but I didn't have duplicate values to begin with.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -740,13 +740,16 @@ def parse_gyms(args, gym_responses, wh_update_queue):
     if len(trainers):
         bulk_upsert(Trainer, trainers)
 
-    # get rid of all the gym members, we're going to insert new records
-    if len(gym_details):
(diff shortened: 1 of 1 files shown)


foltincsaba:
I'm running it with -st 60 -ss -gi -sd 10 with 80 accounts in a city area.

Sometimes (after a few mins) the whole search "collapses"; each worker fails after one another and then never comes back.

I figured that the problem could be that when a worker finds a lot of gyms (>6) it doesn't have enough time to scan each one within 10 secs. So the only solution is to increase -sd.

But this is not optimal.

djv2:
does not work. was just doing some scanning and still get dupes

Issue #809, opened 2016-08-18
Scanning With Multiple Accounts


jnovack:
Scanning with multiple accounts should properly spiral to reduce "teleporting".
## Expected Behavior

Spirals should disperse out properly during scanning.

Two Scanners:
![image](https://cloud.githubusercontent.com/assets/387006/17781234/82713d2c-653c-11e6-9d3f-73b338e8192e.png)

Three Scanners:
![image](https://cloud.githubusercontent.com/assets/387006/17781863/0ac212e4-653f-11e6-9ab4-dd2f7b5e4eb0.png)
## Current Behavior

Spirals play "follow the leader" and teleport.
![image](https://cloud.githubusercontent.com/assets/387006/17781595/e464e87a-653d-11e6-8a49-0eefb7853dcc.png)
## Possible Solution

An algorithm or static mapping will have to be supplied to figure out how many spirals should be used.

Need someone with a Math degree or who was a Mathlete.
## Steps to Reproduce (for bugs)
1.  Run 2+ accounts for scanning.
## Your Environment
- Version used:  7729aab526913db84b6e967c47e8c1c13828e125
- Environment name and version (e.g. Python 2.7):  Dockerfile
- Operating System and version (desktop or mobile):  Dockerfile

ItsCEED:
i was about to write down the same issue but you did it in details! Good job and i hope fix will be published soon

irus99:
I second that

jnovack:
*Newbies, adding nothing to the conversation is unhelpful.  Just click the "reaction button", add a "thumbs up", and sit and wait for someone else to do the heavy lifting.*

When we generate the location queue, this is how it generates it.

![image](https://cloud.githubusercontent.com/assets/387006/17783520/69f70f9c-6546-11e6-853f-96bc0773ff36.png)

The brute force way would be to create static arrays based on the amount of scanners.  But that is U.G.L.Y, you ain't got no alibi, you ugly; hey hey, you ugly!

```python
scanner = [];
scanner.append([null]);
scanner.append([null], [1,2,3,4,5,6,7,8,9,10...]);
scanner.append([null], [1,2,3,4,14,15,16,17,18,19], [5,6,7,8,9,10,11,12,13]);
scanner.append([null], [1,2,10,11,12,13,14,15], [3,4,5,16,17,18,19], [6,7,8,9,23,24,25,26,27]);
```

Then, grab `scanner[number_of_scanners][my_number]` and iterate.

[...]

PR #1010, opened 2016-08-26
Hex Scan: Pokepath Mod

chuyskywalker:
See #809; this is only part 1.

Old:

old

New:

new

New Really Big:

big

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -425,6 +425,7 @@ class ScannedLocation(BaseModel):
                         (ScannedLocation.longitude >= swLng) &
                         (ScannedLocation.latitude <= neLat) &
                         (ScannedLocation.longitude <= neLng))
+                 .order_by(ScannedLocation.last_modified.asc())
                  .dicts())
 
         return list(query)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -266,7 +266,7 @@ class Pokemon(BaseModel):
         # This loop is about as non-pythonic as you can get, I bet.
         # Oh well.
         filtered = []
-        hex_locations = list(generate_location_steps(center, steps, 0.07))
+        hex_locations = generate_location_steps(center, steps, 0.07)
         for hl in hex_locations:
             for idx, sp in enumerate(s):
                 if geopy.distance.distance(hl, (sp['lat'], sp['lng'])).meters <= 70:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -311,7 +311,7 @@ def get_hex_location_list(args, current_location):
         step_distance = 0.070
 
     # update our list of coords
-    locations = list(generate_location_steps(current_location, args.step_limit, step_distance))
+    locations = generate_location_steps(current_location, args.step_limit, step_distance)
 
     # In hex "spawns only" mode, filter out scan locations with no history of pokemons
     if args.spawnpoints_only and not args.no_pokemon:
--- a/pogom/transform.py
+++ b/pogom/transform.py
@@ -63,35 +63,71 @@ def generate_location_steps(initial_loc, step_count, step_distance):
     SOUTH = 180
     WEST = 270
 
-    pulse_radius = step_distance            # km - radius of players heartbeat is 70m
-    xdist = math.sqrt(3) * pulse_radius   # dist between column centers
-    ydist = 3 * (pulse_radius / 2)          # dist between row centers
+    pulse_radius = step_distance         # km - radius of players heartbeat is 70m
+    xdist = math.sqrt(3) * pulse_radius  # dist between column centers
+    ydist = 3 * (pulse_radius / 2)       # dist between row centers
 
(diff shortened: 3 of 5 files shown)


Mr-Groch:
Don't like this as default... Waiting for scan all locations north of my position to scan nearby circle at south? I'm using map with often location changes, for personal use and few friends who also travel...

chuyskywalker:
Yeah, I noticed that too; feels a bit odd, but logically doesn't matter. As long as each scan point gets covered once per loop, it doesn't really matter which order that is.

averylarry:
Hard to think that is much of an issue if you have -st 2 or 3.  Any more steps than that and you should maybe be looking at multiple workers plus it might not be very efficient for location changes (and you risk teleports causing account bans).

Part 2 of this (I presume) relates to starting different workers at different points along the path (staying synchronized).  Thus another worker would be starting close to the end of the path, which is south of your location.

But from a very strict closest-to-me first scan philosophy with 1 worker this is worse.  Unless maybe the starting point is adjusted to 7 (I think) circles along the path before the exact center.  Then the very beginning of the scan would catch all of the 2 step radius before scanning anything on the 3rd step radius.  If that makes any sense.  (Though we might just be trying too hard at this point.)

chuyskywalker:
Yes, the next part will be to split workers up to start at different points in the path, but that's a fairly significant changes (queue per worker thread instead of one larger shared queue).

Location changing always risks teleporting -- only a per-workers rate limiter can ever stop that. (which can/should be covered in a different 

> closest-to-me first scan philosophy with 1 worker this is worse

Again, no it's not. Your perspective is merely presentational. It doesn't matter if you scan "loop center out" or **any other pattern**. All that matters is that each point gets scanned in less than < 15 minutes. If you've got too much `-st` to cover that correctly in center-out, you'll have the same problem.

averylarry:
No no no -- I understand.  I'm just saying if you don't scan a whole city and you don't do spawnpoint scanning and all you want is to know what you can walk to immediately -- say you're driving around and decide on a whim to stop at a park.  In that scenario -- at the very least you'll *feel* better if you move the marker to your location and it strictly spiral scans outward from your location and you know what's closest first before finding things farther away.  It makes a difference for the first 30 seconds or a minute or 2 only.

I wasn't clear in that I think it's a miniscule edge usage case that shouldn't really be considered -- I just wanted to acknowledge what I thought was a theoretical accuracy while basically saying it doesn't matter at all.

chuyskywalker:
I see what you mean, but I'm inclined write it off as an edge case. We kinda cater to two types: scanning a huge damn city/area with fixed location turned on, or scanning a tight area that can follow a user like a personal scanner. In cases where it's used as the personal scanner, the step distance is usually much tighter such that this pattern isn't a huge problem.

Regardless, it would technically be a reduction in functionality, but one I'm willing to make for the tradeoff we'll get (in a few PRs) with offset, non-teleport scanning.

chuyskywalker:
> Unless maybe the starting point is adjusted to 7 (I think) circles along the path before the exact center. 

Actually, I kinda missed that statement, and it makes good sense. I may try to work that in so that if the center moves, when it rescans you at least get some of the central area before it moves on to the general path.

chuyskywalker:
Works awesome:

pokepath

jnovack:
I am not against the "non-central" start.  I like it.

MonjoNL:
I like this PR too.
Have a look at a part of #787 for an assigning individual locations to each worker example. Although this is slightly different. The start positions should be spread out the path but all workers will need all locations in their queue to complete the path.

chuyskywalker:
@MonjoNL that's exactly what the next pr after this will do.

I've let this simmer enough, in it goes :)

rrbailey89:
Is this a switch or command line argument?

chuyskywalker:
It's the default.

rrbailey89:
Good to go

Silvanus0:
Actually wouldn't it be easier to just have one path, but the option to start half the workers in reverse order? That way you still do the center first?

NGPriest:
Is there a limit of workers that can do this? :S

OverlordPhaethon:
@NGPriest Chuy stated above, in response to Monjo, that the next PR will allow multiple workers. If you try it now, your workers will follow the new path sequentially (worker 1 pos 1, worker 2 pos 2...), and depending on how large your step size is and how many workers you dedicate, they will cover great distances in a short amount of time (teleporting). That being said, I tried it anyway and it's amazingly fast, if you're willing to sacrifice accounts.

Love this method. Soon we will have no more multiple instance beehive for fear of teleporting, just one large step size to dump workers in and maybe one smaller to follow you outside your boundaries. Great work everyone.

NGPriest:
@OverlordPhaethon Sorry about that, didn't read most of the above, just skimmed and missed it.
Thanks for answering :D

springjools:
I like this. The less obvious movement probably makes it harder to flag as a bot, but maybe I'm thinking it from our perspective and not from theirs

Issue #939, opened 2016-08-22
-ss not "finding" all the spawnpoints -- is it spawnpoint compression?


averylarry:
## Expected Behavior

Beehive scan shows n spawnpoints.  Using -ss should equally find n spawnpoints.
## Current Behavior

Using -ss "finds" fewer spawnpoints than exist in beehive.
## Steps to Reproduce (for bugs)

Is spawnpoint scanning compression on by default for -ss scanning?  #765 doesn't seem to have made it into dev branch.
In a small area, after clearing the database, I ran a -st 2 beehive for 2 hours:
runserver.py -l "my location" -st 2 -j -nk -ng

![capture](https://cloud.githubusercontent.com/assets/21130007/17869241/ff7f9fe6-6877-11e6-8f8b-a87c58f4e122.JPG)

There are 26 spawnpoints inside the beehive.  If you are conservative with edges, there are still at least 23 spawn points inside the beehive.  (Note the spawnpoints outside the scanned locations are from running -ss for a short period of time.)

When I then start a -ss session (without dumping a .json file), I get 17 spawn points to track:
runserver.py -l "my location" -st 2 -j -nk -ng
. . .
2016-08-22 14:07:18,375 [ search-overseer][        models][    INFO] Finding spawn points 2 steps away
2016-08-22 14:07:18,405 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-08-22 14:07:18,469 [ search-overseer][        search][    INFO] Total of 17 spawns to track

If I increase to -st 3 (I read somewhere to +1 the steps for edge cases), I still only get 21 spawnpoints:
runserver.py -l "my location" -st 3 -j -nk -ng -ss
. . .
2016-08-22 14:11:52,065 [ search-overseer][        models][    INFO] Finding spawn points 3 steps away
2016-08-22 14:11:52,158 [ search-overseer][        search][    INFO] Total of 21 spawns to track

If I dump to a .json file it shows 17 or 21 depending on the step size, matching the above.

If not compression, could it be related to jitter?  I don't really know how since increasing to 3 steps should eliminate that as a possibility.  Plus it is consistent.  I always get 17 or 21.  I even tried -st 54 and it still only returned 21.
## Your Environment

* Version used: 0448f93
* Operating System and version (desktop or mobile): win7

averylarry:
I did specify the same location each time -- I just formatted it wrong for displaying here.

MonjoNL:
[...]

PR #901, opened 2016-08-21
Search Refactor

chuyskywalker:
Ok. So.

- `search.py` has a boat load of duplication in it
- The two search modes are incongruent with one another (one populates the queue in a go, the other hold the loop open) which leads to...
- The two search modes have different feature sets (for example, sps doesn't support pausing)
- Sleep timers are a confused mess
- The spawn point search code is woefully under documented (in-line)
- There is a lot of "magic" geolocation manipulation code present

I've spent a bucket load of time here to produce...

- The pairs of overseer and worker search methods have been merged
- The method for looping through points has been aligned in both cases
- SPS mode now supports pausing and location changes correctly
- The logic behind spawnpoint scanning has been better documented inline and explained
- `scan_delay` now works in a more expected manner: after a thread finishes _everything_, it will sleep for the exact number of seconds set
- Cleaned up the `-ps` display a bit (much prettys)
- Removed a lot of the custom _(copy/pasted?)_ geolocation coordinate code and replaced it with geopy instead (much better, way easier to read)
- Moving some methods to more sensible files (transform.py got a few new neighbors)

This is a pretty big refactor, but to keep feature set in parity, it's needed. Adding SPS was really good, but to keep the app sane, we need to address the **massive** tech debt it incurred to just shove it in.

Issue #855, opened 2016-08-19
disappear_time for 30 minute spawns is 15 instead of 30


hagehg:
I have a spawnpoint in my area where Pokemon will remain for 30 minutes every hour. (from ~xx:45:57 to xx:15:57)

During the first 15 minutes of the spawn, each scan will reset the disappear time to 15 minutes into the future and overwrite the row in the database.

For example:

| encounter_id | pokemon_id | disappear_time |
| --- | --- | --- |
| MTY3MDczMTc5ODYzMzU5OTYxNDE= | 54 | 2016-08-19 22:02:41 |

will be overwritten with:

| encounter_id | pokemon_id | disappear_time |
| --- | --- | --- |
| MTY3MDczMTc5ODYzMzU5OTYxNDE= | 54 | 2016-08-19 22:06:15 |

Starting from xx:00:57 each scan will have the correct disappear_time xx:15:57 untill the pokemon eventually disappears.
## Expected Behavior

30 minute spawns should have the correct disappear_time from the moment they first appear
## Current Behavior

30 minute spawns are reset to disappear 15 minutes into the future with each scan
## Steps to Reproduce (for bugs)

1. Identify a 30 minute spawn
2. Scan it repeatedly
3. Watch the disappear_time in your database change constantly
## Your Environment

- Version used: dev branch from ~24h ago
- Environment name and version (e.g. Python 2.7): 2.7.12
- Operating System and version (desktop or mobile): Ubuntu 16.04 LTS

GreenSlime96:
This same issue affects me too and causes trouble especially when using spawnpoint-scanning.

koljanowak:
I see the same issue.

pyr0ball:
@chuyskywalker could you take a look at this?

GreenSlime96:
@pyr0ball it's a known bug.

matpag:
yeah, same problem for me

Saprophator:
Here is a useful thread with some data to the different spawn times and their behavior[reddit](https://www.reddit.com/r/pokemongodev/comments/4yzqc2/spawn_point_types_clearing_confusion/)

matpag:
@Saprophator really usefull link

Xcelled:
[...]

PR #986, opened 2016-08-25
Fix finding spawnpoints when using -ss

averylarry:
## Description

Changes how the spawnpoint finder iterates through the list of spawnpoints to eliminate skipping some.
## Motivation and Context

Fixes #939
## How Has This Been Tested?

Ran -ss with --dump-spawnpoints and compared the resulting .json files using the orig. code and the new code.  Kept it small having done a beehive scan with -sd 2.  The new code finds more spawnpoints and the count matches what is visually seen on the gui.  The new code also did not have any duplicated spawnpoints.
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

(Hope I'm doing this right.)

PR #1013, opened 2016-08-26
Fix missing spawns when exporting

MonjoNL:
Fix for reported issue #939
~~Fix for calculating the remaining time when time_tile_hidden was very large.~~ #855

## Description
Since the search refactor #901 exporting spawn points into a json file was exporting far less spawn points on the database than before this PR.
As example a previous build exported 186 spawns where the current build exports 134.
This fix finds 180 spawns again, not exactly the same as before, but this is probably more accurate.
The problem was caused by deleting entries from a list while still in a for loop that loops thru that same list.
This results in skipping items and ultimately missing spawns.

## Motivation and Context
If you didn't exported a spawn point file yet you don't want to miss a lot of spawn points.
Thats the whole idea of this project.

## How Has This Been Tested?
Exported spawns on the same database with the same location and amount of steps with different builds. And before and after this patch.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -279,7 +279,7 @@ class Pokemon(BaseModel):
             # todo: this DOES NOT ACCOUNT for pokemons that appear sooner and live longer, but you'll _always_ have at least 15 minutes, so it works well enough
             location['time'] = (location['time'] + 2700) % 3600
 
-        #remove dublicates because when a spawn falls into more than 1 hex location it is listed multiple times
+        # remove dublicates because when a spawn falls into more than 1 hex location it is listed multiple times
         noDupes = []
         [noDupes.append(i) for i in filtered if not noDupes.count(i)]
         return noDupes
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -505,16 +505,15 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue):
     for cell in cells:
         if config['parse_pokemon']:
             for p in cell.get('wild_pokemons', []):
-
-                time_till_hidden = p['time_till_hidden_ms']
-                # Correct misterious time_till_hidden value by removing extra bit in front of the Int32 that indicates this is a long spawn
-                # 2147483647 is the maximum value of an Int32 and can be read since this is read as a Int64
-                if time_till_hidden > 2147483647:
-                    time_till_hidden = time_till_hidden - 2147483647
-
-                d_t = datetime.utcfromtimestamp(
-                    (p['last_modified_timestamp_ms'] +
-                     time_till_hidden) / 1000.0)
+                # time_till_hidden_ms was overflowing causing a negative integer.
+                # It was also returning a value above 3.6M ms.
+                if 0 < p['time_till_hidden_ms'] < 3600000:
+                    d_t = datetime.utcfromtimestamp(
+                        (p['last_modified_timestamp_ms'] +
+                         p['time_till_hidden_ms']) / 1000.0)
+                else:
+                    # Set a value of 15 minutes because currently its unknown but larger than 15.
+                    d_t = datetime.utcfromtimestamp((p['last_modified_timestamp_ms'] + 900000) / 1000.0)
 
                 printPokemon(p['pokemon_data']['pokemon_id'], p['latitude'],
                              p['longitude'], d_t)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -510,11 +510,11 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue):
                 # Correct misterious time_till_hidden value by removing extra bit in front of the Int32 that indicates this is a long spawn
                 # 2147483647 is the maximum value of an Int32 and can be read since this is read as a Int64
                 if time_till_hidden > 2147483647:
-                     time_till_hidden = time_till_hidden - 2147483647
+                    time_till_hidden = time_till_hidden - 2147483647
 
                 d_t = datetime.utcfromtimestamp(
                     (p['last_modified_timestamp_ms'] +
-                    time_till_hidden) / 1000.0)
+                     time_till_hidden) / 1000.0)
 
                 printPokemon(p['pokemon_data']['pokemon_id'], p['latitude'],
                              p['longitude'], d_t)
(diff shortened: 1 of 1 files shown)


averylarry:
Probably better than #986.  The issue is also eliminated by #951.

MonjoNL:
@averylarry Yes, #986 doesn't make any difference to the current build in my case.
And #951 reverts #901 for spawn file generation and is the original code.
But I like the way #901 intended to rewrite it, so I prefer this PR.

MonjoNL:
Sorry, I intended to create a separated PR for the 959+ hour thing.
That one is even more interesting :)

MonjoNL:
Wait guys, found another instance that was not working fine.
So I guess this is not working correct after all in all cases.
Maybe we still need to do the current between 0 and 3600000 check in addition to be safe.
Really thought I figured it out.

busyhare25:
done some local testing and has worked out the right ms for my the one example so far, waiting for more large numbers to spawn so i can test again. had to create a github account to comment :(

                # time_till_hidden_ms was overflowing causing a negative integer.
                # It was also returning a value above 3.6M ms.
                if 0 < p['time_till_hidden_ms'] < 3600000:
                    log.info('Time_Till_Hidden_Normal: %d ms , PokeID: %s @ %f , %f',p['time_till_hidden_ms'],p['pokemon_data']['pokemon_id'], p['latitude'],p['longitude'])
                    d_t = datetime.utcfromtimestamp(
                        (p['last_modified_timestamp_ms'] +
                         p['time_till_hidden_ms']) / 1000.0)
                else:
                    # Set a value of 15 minutes because currently its unknown but larger than 15.
                    time_till_hidden = (p['time_till_hidden_ms'] & 0x3FFFFF) + ((p['time_till_hidden_ms'] >> 22)*1000)
                    log.info('Time_Till_Hidden_Large: %d ms, using %d ms, PokeID: %s @ %f , %f',p['time_till_hidden_ms'],time_till_hidden,p['pokemon_data']['pokemon_id'], p['latitude'],p['longitude'])
                    d_t = datetime.utcfromtimestamp((p['last_modified_timestamp_ms'] + time_till_hidden) / 1000.0)


2016-08-27 19:53:06,919 Time_Till_Hidden_Large: 882195811 ms, using 1391971 ms, PokeID: 116 @coords
2016-08-27 20:01:30,610 Time_Till_Hidden_Large: 881692108 ms, using 888268 ms, PokeID: 116 @coords
2016-08-27 20:04:17,540 Time_Till_Hidden_Large: 881525180 ms, using 721340 ms, PokeID: 116 @coords
2016-08-27 20:06:31,574 Time_Till_Hidden_Normal: 793577 ms , PokeID: 116 @coords

MonjoNL:
In a new test I strip everything above 24 bits instead above 32 bits for the 595+ hour bug.
Looks promising, wil test it for a day or two.

                if time_till_hidden > 2147483647:
                    time_till_hidden = time_till_hidden - 2147483647
                if time_till_hidden > 1073741823:
                    time_till_hidden = time_till_hidden - 1073741823
                if time_till_hidden > 536870911:
                    time_till_hidden = time_till_hidden - 536870911
                if time_till_hidden > 268435455:
                    time_till_hidden = time_till_hidden - 268435455
                if time_till_hidden > 134217727:
                    time_till_hidden = time_till_hidden - 134217727
                if time_till_hidden > 67108863:
                    time_till_hidden = time_till_hidden - 67108863
                if time_till_hidden > 33554431:
                    time_till_hidden = time_till_hidden - 33554431
                if time_till_hidden > 16777215:
                    time_till_hidden = time_till_hidden - 16777215
                if time_till_hidden > 8388607:
                    time_till_hidden = time_till_hidden - 8388607

busyhare25:
why not just use the bitwise & operator? 
time_till_hidden = time_till_hidden & 0xFFFFFF
https://wiki.python.org/moin/BitwiseOperators

MonjoNL:
Yes thats much better. Sorry for the repost.
I posted at another PR first so copied my first peace of code to the correct topic.
Thanks for the feedback and testing this as well.

busyhare25:
to expand on the line i was using:
(p['time_till_hidden_ms'] & 0x1FFFFF) 
takes the right 21bits and uses that as the new timestamp, but it's short by ~210seconds
((p['time_till_hidden_ms'] >> 22)*1000)
taking the first 8bits (typically 11010001 or 11010010) as decimal gives us our ~210 seconds so *1000 to give ms and add it to the first number. could use static +210000 but as i've only had a few of these spawns not sure how much it would vary and +/-1s isn't that big of a error margin.
using more than the right most 21bits has given me time to despawn of over 1hour if bit #9 = 1 which so far is very wrong.

MonjoNL:
Have some observations when using 24 bits.

I saw one spawn jump from aprox. 45 minutes left to aprox. 5 minutes left after refreshing (2^21 ms) and an other from aprox. 30 to aprox. 3 minutes left (2^20 ms)

I guess those two bits and possibly the 2^22 bit are meaning something else as well.
21 bits are enough to cover for almost 17,5 minutes.
This bit fix seems to be legit when scanning within that 17,5 minutes left , but getting over 21 bits is not correct yet.
But I feel that we are going into the right direction here.

busyhare25:
i've not seen a 45min spawn yet but 21bits (0x1FFFFF) can give upto 34.95251666666667minutes
if you count the leading 8 bits (11 0100 11) as (210seconds) you get about 38.5mins max view on a spawn
if you use 9 leading bits as seconds to add you can get about 422seconds or about 41.98mins max
at least that's with the few >15min spawns that i've observed in my small scan area thus far. is it confirmed 45min spawns are real? if we assume 15&30min spawns only, then the bit mangling i've come up with works farily well so far accross a few spawns.
it would be nice if someone with a larger # of >15min spawn sightings had logged the raw timestamp+disappear ms so we can run simulations over it with the insights above.

j16sdiz:
@busyhare25 45 minutes and 60 minutes spawns are real.
Try `340408bcadf` at  `22.4239935135441,114.220520803699`, this is a 60 minutes spawn.

MonjoNL:
Please delete, not needed anymore since #951.

@busyhare25 lets start a new issue to keep track of the bit correction to find the real disappear time for long spawns.

Issue #1015, opened 2016-08-26
minor, "Pokemon Go Map" link is broken on stats page if site is hosted in a sub dir


DiscoTim:
The "Pokemon Go Map" link at the top of the stats page (beside options) will take you back to the site root, instead of the sub dir the site is hosted in. i believe the href needs to be href="#". it works properly from the main map page.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Environment name and version (e.g. Python 2.7):
* Operating System and version (desktop or mobile):

BaIthamel:
Your issue template is incomplete. Please document to the level of an asthmatic bee who's lost his inhaler  on midsummer's day!

BaIthamel:
*sigh* okay, more reasonable reply. If you know where the issue is please make a PR, should take you 10 seconds. would take us a little longer to find the line you're talking about.

DiscoTim:
yes i should, i am fairly new to git and didn't want to mess up my repository. but i guess i need to learn. i was just thinking if someone is messing in there they could hit it, its 1 character :)

pyr0ball:
Derp sorry accidental click on the close

PR #1019, opened 2016-08-27
Fix statistics page home link to work with sites running in subdirectories

DiscoTim:
This PR correct a minor issue where the "Pokemon Go Map" header link on the statistics page redirects to the site root, as opposed to the current sub-directory.
## Description

Statistics.html line 44, characters 45 and 46 where changed from '..' to './'
## Motivation and Context

This change is required so that sites running in sub-directories will correctly redirect to map homepage. 

https://github.com/PokemonGoMap/PokemonGo-Map/issues/1015
## How Has This Been Tested?

I clicked on the link and as expected was redirected to the site homepage.

## Types of changes

## Checklist:

### DIFF
--- a/templates/statistics.html
+++ b/templates/statistics.html
@@ -41,7 +41,7 @@
         <!-- Header -->
         <header id="header">
             <a href="#nav">Options</a>
-            <h1><a href="#">Pokémon Go Map</a></h1>
+            <h1><a href="./">Pokémon Go Map</a></h1>
         </header>
         <nav id="nav">
             {% for name, item in valid_input.items() %}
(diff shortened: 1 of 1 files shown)


jaquer:
Are you sure this is the right solution? I don't get sent back to the homepage this way, it just stays on the stats page. I believe the correct href is `./` or even just `.`, per this thread at SO: https://stackoverflow.com/questions/296873/basic-html-how-to-set-relative-path-to-current-folder

DiscoTim:
@jaquer you're right, I've updated it to ./

4RK4N:
whats keeping this from being merged? also this shouls be applied to the /status page top link

DiscoTim:
i am really not sure.

its leaked on to the status page now? it wasn't around when this pr was done, i'll check and update that in this pr when i have a sec, if no one else does first.

PR #1025, opened 2016-08-27
The range for a pokestop and gym is 40 meters, but search and pokemon ranges are 70 meters.

beccasafan:
## Description

Fix the range circle radius numbers.

## Motivation and Context

[#753](https://github.com/PokemonGoMap/PokemonGo-Map/pull/753) added range circles for the markers, but set them all to 40 meters. Pokemon and search markers should be 70 meters while pokestops are gyms are correct at 40 meters.

## How Has This Been Tested?

Local

image

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1246,6 +1246,7 @@ function addRangeCircle (marker, map, type, teamId) {
   var range
   var circleColor
 
+  // handle each type of marker and be explicit about the range circle attributes
   switch (type) {
     case 'search':
       circleColor = '#333333'
@@ -1253,7 +1254,7 @@ function addRangeCircle (marker, map, type, teamId) {
       break
     case 'pokemon':
       circleColor = '#C233F2'
-      range = 70
+      range = 40 // pokemon appear at 40m and then you can move away. still have to be 40m close to see it though, so ignore the further disappear distance
       break
     case 'pokestop':
       circleColor = '#3EB0FF'
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1240,20 +1240,36 @@ function addRangeCircle (marker, map, type, teamId) {
   var targetmap = null
   var circleCenter = new google.maps.LatLng(marker.position.lat(), marker.position.lng())
   var gymColors = ['#999999', '#0051CF', '#FF260E', '#FECC23'] // 'Uncontested', 'Mystic', 'Valor', 'Instinct']
-  var circleColor = '#cccccc'
   var teamColor = gymColors[0]
   if (teamId) teamColor = gymColors[teamId]
 
-  if (type === 'search') circleColor = '#333333'
-  if (type === 'pokemon') circleColor = '#C233F2'
-  if (type === 'pokestop') circleColor = '#3EB0FF'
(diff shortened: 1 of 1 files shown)


chuyskywalker:
@kamykaze review please :)

kamykaze:
Can someone confirm that pokemon show up as far as 70m? I've read it somewhere a while ago that it was 40m, but never bothered to do a real world test on that. I can check it out tomorrow by catching some pokemon when it's day time.

Assuming I'm misinformed and pokemon can be caught at 70m instead of 40m, then the search marker range is not really useful, because it was supposed to be a convenient range to show which objects you can interact with. If they're not all the same distance, then that range is not accurate unless we make it two circles (one with 40m and one with 70m)

As far as the code goes though, it looks good and we can merge it once we confirm that there are two range distances.

Infernoman:
from what i know the encounter range is 50m. and catching range is 70m. so you can be passing the pokemon, when you get within a 50m radius it appears. and when you leave the 70m radius it disappears

beccasafan:
Well boo then. I blame the discord channel for lying to me! 😉

So should we use 40m for pokestop/gym, 50m for search..and 50m or 70m (??) for pokemon?

gabrielsoldani:
Here's the thing about the Pokéstop range: the spin range is 40m, but the lured encounter range is 70m. Which range do we want to show? I'd personally want to show both, but I know people's opinion on it will differ.

The search range is 70m, we've been using it for ages. There's also a despawn range, bigger than 70m. If you enter the search range, the Pokémon appears. It will then disappear if you leave the despawn range.

chuyskywalker:
A pokestop should show the 40m range; the range is about where you need to be to interact with the thing. The lured pokemon will have their own ranges.

kamykaze:
Interesting discussion...If there are different ranges for the pokemon (encounter and catch/despawning), then I would think the more important one is the encounter one, since if you don't get that close, you can't catch the pokemon.

However, it would not be too hard to show both ranges for each pokemon. We can create two circles for the same pokemon, with the bigger one being shown as it currently is, and the smaller one having only a visible border, but no background.

I'll do some ground work as I find pokemon today to test the real ranges and then we can decide on what to do.

beccasafan:
Pending confirmation of the numbers, I would vote for just showing the encounter range. I think a second circle would just serve to confuse things and add noise to the map. (Not that my vote counts for anything! 😂)

chuyskywalker:
Agree with the 40m range. Like with Pokestops, the radius should represent where you need to be to interact with it. The whole point being "do i need to go a street or not?" Or "if I stand right...here, I can get all XYZ without moving".

beccasafan:
Ok, so back to what @kamykaze had, but removed the search marker...

kamykaze:
I'm fine with that. The search marker was added last minute and wasn't really thought out.

I'm actually working on adding a new draggable marker (independent of the search and current locations) that you can drag around and position anywhere in the map to find the "sweet" spots that allows you to trigger multiple pokestops, or multiple spawnpoints.

In the mean time, we can merge this PR as I do prefer the refactoring you did for the different range types. It makes updating them in the future a lot easier.

PR #1029, opened 2016-08-27
We use spaces round these here parts!

chuyskywalker:
Literally:

```
find . -type f -name '*.scss' -exec bash -c 'expand -t 2 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;
```

Because the scss files were a damn mess.

### DIFF
--- a/static/sass/base/_page.scss
+++ b/static/sass/base/_page.scss
@@ -1,32 +1,32 @@
 /* Basic */
 
-	// MSIE: Required for IEMobile.
-		@-ms-viewport {
-			width: device-width;
-		}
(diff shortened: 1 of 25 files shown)


chuyskywalker:
For a fun way to view this change: https://github.com/PokemonGoMap/PokemonGo-Map/pull/1029/files?w=0

sebastienvercammen:
+1, there's never a better topic to argue about than tabs vs spaces

chuyskywalker:
We actually managed to get decent consensus on spaces for python + js, so yeah. Just getting the SASS files inline.

PR #1037, opened 2016-08-28
Decommission faulty accounts 2

Shibby7986:
Adds PoGo accounts into the Db and disables accounts that are suspected to be banned replacing them with a working account from the DB

## Description
adds 3 args
-na for selecting the number of accounts to use.(default is 1)
-ru for remove account from the database.
the values of the standard -u -p and -a are reworked to add users to the database.
-cu clears account usage.

## Accounts usage
There's no longer a need for user or password args on each run they are only needed for adding accounts to the DB. On each run, accounts are retrieved from the DB as required.
 Using -u and -p will add that account to the DB.
re-adding an existing account will just reset its properties.

Account sessions are used so that accounts are only used by one worker at a time.
When an account is deactivated it is replaced with a new account on the fly.
In event that there are no enough accounts the thread will wait for more accounts to be added and once added it will get an account and continue its scanning without any user input or restart.

## Error Handling
Included is better handling of map_phrase errors
Accounts are deactivated after the API returns no map objects or returns a NotLoggedInException.
Users were experiencing map_parse Key errors that are caused by the API not returning any map objects and the key being empty, generally, this is because the account was banned. 
Users would also sometimes get map_parse type errors caused by the API sending NotLoggedInException instead of the scan results. and this error was being passed to the map_parse method instead of map objects. this was caused by Invalid login attempts or an unactivated account.
both of these are now detected before map parse is called to stop the exception and traceback. now instead of these python exceptions, it displays a friendly error and adds one to the fail count.

Once the specified number of failures is reached (-mf flag to change from default) the account is deactivated and replaced.

As suggested by @justsomedudeonthenet deactivation lasts 2 hours and then the accounts are reactivated in case of false positives.

## Motivation and Context
This feature overcomes the issue of holes in the map due to faulty accounts and results in no more tedious reconfiguring of behive.bat or config.ini to replace non-working accounts.

## How Has This Been Tested?
Testing on windows 10,
I filled the database using both command line args and the config.ini
The application runs without issue and accounts are replaced as expected
I tried a few known non-working accounts and they were replaced and set to de-active
resetting accounts in_use status causes the previous users to move on to a new account once the session value is changed.  

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/multi-account.md
+++ b/docs/extras/multi-account.md
@@ -1,11 +1,15 @@
-# Using Multiple Accounts
+# Adding Accounts
 
-PokemonGo-Map supports using multiple accounts to run a worker with multiple threads.
-
+PokemonGo-Map stores your accounts for use in its database. 
+Before you start you will need to load the database with your accounts.
 
 ## Using Command Line Arguments:
 
-To use multiple accounts when running from the command line, you must specify multiple -u and -p values.
+To add an account use the -u -p and -a when you launch runserver.py
+
+Example: `python runserver.py -u thunderfox01 -p abracadabra` 
+
+To add multiple accounts when running from the command line, you must specify multiple -u and -p values.
 
 Example: `python runserver.py -u thunderfox01 -u thunderfox02 -p thunderfox01 -p thunderfox02`
 
@@ -46,3 +50,9 @@ auth-service: [ptc, ptc, google]
 username: [thunderfox01, thunderfox02, thunderfox03@gmail.com]
 password: [password01, password02, password03]
 ```
+## Using multiple accounts
+
+PokemonGo-Map supports using multiple accounts to run a worker with multiple threads.
+to specify the number of accounts use the -na flag
+
+Example: `python runserver.py -na 3`
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -584,6 +584,7 @@ class PoGoAccount(BaseModel):
                 use_account(account['username'], account['session'])
             if i == count - 1:
                 break
+
         return accounts
 
     @staticmethod
@@ -643,7 +644,7 @@ def update_use_account(username):
 def remove_accounts():
     if args.remove_user is not None:
         for user in args.remove_user:
-            log.info('Removing {} from the db'.format(username))
+            log.info('Removing {} from the db'.format(user))
             PoGoAccount.delete().where(PoGoAccount.username == user).execute()
 
 
@@ -1029,7 +1030,6 @@ def clean_db_loop(args):
                      .where((PoGoAccount.time_deactivated < (datetime.utcnow() - timedelta(minutes=120))) &
                             (PoGoAccount.active == 0)))
             query.execute()
-
             
             # Reset usage on idle accounts
             query = (PoGoAccount
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -583,8 +583,7 @@ def insert_accounts():
         log.info("Checking " + account['username'])
         try:
             query = PoGoAccount.create(username=account['username'], password=account['password'], auth_service=account['auth_service'])
-            query.execute()
-            log.info("added" + account['username'])
+            log.info("Added " + account['username'])
         except:
             log.info(account['username'] + " already exists reseting password and status")
             query = PoGoAccount.update(password=account['password'], auth_service=account['auth_service'], active=True).where(PoGoAccount.username == account['username'])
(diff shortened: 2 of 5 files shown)


justsomedudeonthenet:
You mark all the accounts to not be in use when a new webserver is started???

Meaning if anyone restarts an instance with the webserver, but not some of their no server instances, it will try using accounts twice and cause all sorts of problems.

It would probably work if running a single beehive script, since only one copy runs the server and it starts first.  But many people are running multiple servers, separate webserver processes, etc.

Deactivating accounts forever isn't a great idea either.  The point of waiting 2 hours and trying again was because soft bans for moving too fast will cause those errors as well, and the account will start working again after awhile.

I like the idea, but I think it needs some work.

Shibby7986:
@justsomedudeonthenet I figure you would have a separate DB per web server.
I can easily change it to a separate arg for resetting the usage. instead of using -ns

justsomedudeonthenet:
@Shibby7986 Why on earth would you run separate databases?  I certainly want them to share the same data!

Shibby7986:
@justsomedudeonthenet because the web servers would be separate geographical areas and you would want separate stats

Shibby7986:
@justsomedudeonthenet dude don't softbans just give 0,0,0 and not an error?
I can set the accounts to re-activate after a timer if that's the case.

CalcMan:
@Shibby7986 if you want geographic area specific stats, then do a query on the database with 'WHERE longitude BETWEEN ### AND ### AND latitude BETWEEN ### AND ###;'

add the proper web and python plumbing and make it a new PR.

Shibby7986:
@CalcMan cheers, but that's not what I'm after at all. I'm making it so accounts are rotated out when faulty :P.

Shibby7986:
Well... I'm happy with it now. :)
let me know what you guys think.

CtrlAltDefeat94:
+1 from me. Can't wait to use this on my live env :3

CalcMan:
Consider 1 single database with many runserver.py instances, some scanners only, some scanners with web interfaces that the user can move around, and stop scanning by flipping the 'Search' switch (sets the pause bit). Do paused scans still consume an account? (perhaps they should). But more importantly, if a server scanning in location X turns off (maybe the runserver is shutdown), and someone else grabs that now free account, it might end up teleporting, how can we prevent this? is it reasonable for your new accounts table to also include a 'last scanned time' and 'last scanned location', then we could use math to make sure an account moving from point A to point B hasn't exceeded some speed limit threshold value.

n3cr0cr0w:
This has made managing my setup so much easier, I'm sure we can think of more things to store in the account table. Number of empty results, number of failed attempts?

I do agree that this plus a speed limit check would solve just about all my issues. For now I'm running this PR.

IT Rocks. 4620 Accounts in DB, 116 with active=0 and counting.

Shibby7986:
ha :) yeah guys, that's exactly what I did next add a speed limiter.
It also replaces accounts if they have to wait too long due to speed/distance
it's setup on the live branch of my fork you can try that out.
https://github.com/Shibby7986/PokemonGo-Map

Shibby7986:
@CalcMan the last commit will allow the accounts to be reused when a worker thread is closed/crashes/paused.
essentially if an account does not make a scan for 3 mins it is set back to available

erikssonk:
@Shibby7986 Just to give you an heads up on the proxy handling so they also can rotate between workers. Also some other checks on the proxy that will temporarily disable that proxy and spread the load between other proxies available.

KoffeinKaio:
Trying this atm. Noticed a strange behaviour: Some accounts are getting "Invalid response, abandonding location" the counter goes up to 5 and then resets, accounts used (in web and cli status page) seem to be the same, proxy also wont change. (using tor socks, so I dont know if its a faulty account or proxy)

CalcMan:
@Shibby7986 my setup might be different then a lot of others, my 'personal scanner accounts' (those where user can move a marker) are located potentially hundreds (100s) of miles from each other, as well as the server only scans. But this means that if user A and user B pause scans then user B starts up again but grabs the accounts user A was using, then the account might end up moving over 500miles in a 3 minute jump. I don't know the backend ban limits, but that seems like it might set them off.

CalcMan:
Did this PR go dark?

CtrlAltDefeat94:
Would be a real shame if it were :(

jorgebsa:
This won't be merged as-is (even if rebased) as it would be incompatible with the current account cycling methods AND the issue that has been raised about each account's previous locations and how it could easily teleport around. Unless a way to control the teleporting is implemented along with this change, do not expect it to be merged.

primaxius:
Incompatible with current system and would need rebase/rewrite

PR #1067, opened 2016-08-31
Read accounts from CSV file with -ac/--accountcsv parameter

justsomedudeonthenet:
## Description
This allows a CSV file containing accounts to be specified, instead of **or in addition to** multiple -a/-u/-p parameters

Each line must contain the auth type, username and password.  Additional fields after that are ignored and can be left there.

eg:
```
ptc,username1,password1,00.00,00.00,dovahkiin1
ptc,username2,password1,00.00,00.00,dovahkiin2
```

## Motivation and Context
Adding hundreds of accounts from the command line is painful.  CSV files are painless.  Many already have their accounts in CSV files in the correct format.

## How Has This Been Tested?
Tested various valid and invalid combinations of csv files and command line account parameters.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -207,7 +207,7 @@ def get_args():
         if (args.step_limit is None):
             errors.append('Missing `step_limit` either as -st/--step-limit or in config')
 
-        if args.auth_service is None:
+        if len(args.auth_service) is None:
             args.auth_service = ['ptc']
         else:
             num_auths = len(args.auth_service)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -46,12 +46,13 @@ def get_args():
     # fuck PEP8
     configpath = os.path.join(os.path.dirname(__file__), '../config/config.ini')
     parser = configargparse.ArgParser(default_config_files=[configpath], auto_env_var_prefix='POGOMAP_')
-    parser.add_argument('-a', '--auth-service', type=str.lower, action='append',
+    parser.add_argument('-a', '--auth-service', type=str.lower, action='append', default=[],
                         help='Auth Services, either one for all accounts or one per account: ptc or google. Defaults all to ptc.')
-    parser.add_argument('-u', '--username', action='append',
+    parser.add_argument('-u', '--username', action='append', default=[],
                         help='Usernames, one per account.')
-    parser.add_argument('-p', '--password', action='append',
+    parser.add_argument('-p', '--password', action='append', default=[],
                         help='Passwords, either single one for all accounts or one per account.')
+    parser.add_argument('-ac', '--accountcsv')
     parser.add_argument('-l', '--location', type=parse_unicode,
                         help='Location, can be an address or coordinates')
     parser.add_argument('-j', '--jitter', help='Apply random -9m to +9m jitter to location',
@@ -169,22 +170,37 @@ def get_args():
             print(sys.argv[0] + ": error: arguments -l/--location is required")
             sys.exit(1)
     else:
+        # If using a CSV file, add the data into the username,password and auth_service arguments.
+        # CSV file should have lines like "ptc,username,password".  Additional fields after that are ignored.
+        if(args.accountcsv is not None):
+            account_fh = open(args.accountcsv, 'r')
+            account_fields = [line.split(",") for line in account_fh]
+            account_fh.close()
+            try:
+                for line in account_fields:
+                    args.auth_service.append(line[0].strip())
+                    args.username.append(line[1].strip())
+                    args.password.append(line[2].strip())
+            except:
+                print(sys.argv[0] + ": Error parsing CSV file. Lines must be in the format '<method>,<username>,<password>'. Additional fields after those are ignored.")
+                sys.exit(1)
+
         errors = []
 
         num_auths = 1
         num_usernames = 0
         num_passwords = 0
 
-        if (args.username is None):
-            errors.append('Missing `username` either as -u/--username or in config')
+        if (len(args.username) == 0):
+            errors.append('Missing `username` either as -u/--username, csv file, or in config')
         else:
             num_usernames = len(args.username)
 
         if (args.location is None):
             errors.append('Missing `location` either as -l/--location or in config')
 
-        if (args.password is None):
-            errors.append('Missing `password` either as -p/--password or in config')
+        if (len(args.password) == 0):
+            errors.append('Missing `password` either as -p/--password, csv file, or in config')
         else:
             num_passwords = len(args.password)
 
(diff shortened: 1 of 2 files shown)


xioxxio:
why not use the additional fields like long and lat?

justsomedudeonthenet:
@xioxxio use them for what?  They are irrelevant to the scanner.  I just mention them so people know that if you have such a CSV file, you can just drop it in place, no editing necessary.

Zerotask:
I guess 99 % off all accounts used by people here are ptc. So I would add the default case ptc. That means if you dont add ptc, then it assumes that it's a ptc account.

pogom/utils.py
```diff
@@ -169,29 +170,44 @@ def get_args():
             print(sys.argv[0] + ": error: arguments -l/--location is required")
             sys.exit(1)
     else:
+        # If using a CSV file, add the data into the username,password and auth_service arguments.
+        # CSV file should have lines like "ptc,username,password".  Additional fields after that are ignored.
+        if(args.accountcsv is not None):
+            account_fh = open(args.accountcsv, 'r')
```
chuyskywalker: Generally it's more "pythonic" to use `with open():` clauses for this kind of thing.

pogom/utils.py
```diff
@@ -169,29 +170,44 @@ def get_args():
...
+            account_fields = [line.split(",") for line in account_fh]
+            account_fh.close()
+            try:
+                for line in account_fields:
+                    args.auth_service.append(line[0].strip())
+                    args.username.append(line[1].strip())
+                    args.password.append(line[2].strip())
+            except:
+                print(sys.argv[0] + ": Error parsing CSV file. Lines must be in the format '<method>,<username>,<password>'. Additional fields after those are ignored.")
+                sys.exit(1)
```
chuyskywalker: There is a distinct lack or error handling around this area (but also so in others, I think) where it doesn't check that the fields are valid, structurally. For example:

```
fakeauth,,
nothing,user,
,,oops
```

etc.

pogom/utils.py
```diff
@@ -169,29 +170,44 @@ def get_args():
...
         errors = []
 
         num_auths = 1
         num_usernames = 0
         num_passwords = 0
 
-        if (args.username is None):
-            errors.append('Missing `username` either as -u/--username or in config')
+        if (len(args.username) == 0):
+            errors.append('Missing `username` either as -u/--username, csv file, or in config')
```
chuyskywalker: "csv file using -ac"

pogom/utils.py
```diff
@@ -46,12 +46,13 @@ def get_args():
...
-    parser.add_argument('-a', '--auth-service', type=str.lower, action='append',
+    parser.add_argument('-a', '--auth-service', type=str.lower, action='append', default=[],
                         help='Auth Services, either one for all accounts or one per account: ptc or google. Defaults all to ptc.')
-    parser.add_argument('-u', '--username', action='append',
+    parser.add_argument('-u', '--username', action='append', default=[],
                         help='Usernames, one per account.')
-    parser.add_argument('-p', '--password', action='append',
+    parser.add_argument('-p', '--password', action='append', default=[],
                         help='Passwords, either single one for all accounts or one per account.')
+    parser.add_argument('-ac', '--accountcsv')
```
chuyskywalker: Needs a `help` attribute.

chuyskywalker:
Overall this looks fine, but I haven't tested it out for a lot of the weird combinations that people can end up using. Think about configs from ENV variables, mixed with cli args, mixed with ac, mixed with config files. And all the various combinations there-in.

It's one of those times where some really solid unit testing would help a ton.

justsomedudeonthenet:
@chuyskywalker thanks for the detailed review.  I will try to get the issues you pointed out addressed tomorrow.

I have tested it with a combination of accounts from csv, config and command line and haven't found any issues thus far.

justsomedudeonthenet:
@chuyskywalker I have updated it with significantly better error handling and more pythonic file opening.

It now ignore blank lines or lines that start with '#', so you can have comments in the file if desired.
Any line that doesn't contain the right number of fields, or has blank fields, will print an error with the line number and exit.

And it has a help message now.

chuyskywalker:
Code wise this LGTM, I haven't had a chance to actually try it. If someone else can verify before I get to it, I'll merge it.

FrostTheFox:
blugh, i might just merge this

FrostTheFox:
eh what can go wrong lets do it

springjools:
How do you specify the csv file?

Talhoo:
@springjools It's in title. -ac/--accountcsv . And you can always find arguments with "python runserver --help" or looking through utils.py

springjools:
It's good to have informative titles, but it would be even better to also include it in the description . The help argument only works after client has pulled from git.

Have a nice day.

springjools:
Please remember to also update this page with the new -ac switch.

https://github.com/PokemonGoMap/PokemonGo-Map/wiki/Command-Line-Arguments

gtuxyco:
i have 40 accounts in beehive sh
Can i use the same account csv file for all instance?
just replace -u and -p in each line by -AC?

n3cr0cr0w:
Do not use one file for multiple instances, one file per instance.

subzerofun:
so if i want to scan multiple locations at once, i still have to start up multiple instances?
i thought this was also an solution to get rid of all the command line clutter.

isn´t there a solution where i can just put in all my scan locations in one line, like:
```python
python runserver.py -ac myaccounts.csv -k 'somekey' -l 'long1,lat1' -l 'long2,lat2' -l 'long3,lat3'
```
and let one instance handle all the worker requests, account rotation, etc. ?

or could someone give me an example how i would need to setup my cmd line args to use one  `accounts.csv`  for multiple locations?

so if have 50 accounts in one csv file i still have to fire up multiple instances and also can´t use the same csv for every instance?

sorry if these are too much questions at once, but i have searched the docs for this, but unfortunately there is no info on how to combine the csv feature with multiple locations. handling multiple instances (or multiple scan locations) still seems more complicated than it needs to be.

subzerofun:
or ~ i´m just too stupid to figure it out myself 😶

gtuxyco:
+1 stupod here :D.

redcl0udcoc:
+1 stupid! How can this work with beehive? would like for all instances to be able to use a single csv to queue accounts that have a problem.

justsomedudeonthenet:
Having the accounts shared between parts of a beehive is not yet possible.  I am working on it :)

n3cr0cr0w:
Tell me it is accounts in the database, I don't want to give that up now that I have tasted it.

PR #644, opened 2016-08-14
Multiple proxies

erikssonk:
## Description
* Add support for multiple proxies via --proxy 
* Add --proxy-timeout - Sets the timeout for checking proxy, default is 5 seconds
* Add --proxy-display - Use with -ps Adds information about which proxy is used for each worker, default is index
		
		full = display full proxy addres
		index = display the current index for that proxy

At start it will check the entire proxy list for non-working proxies, this includes those who take to long or can't connect to Niantic via the proxy.
When that happens, the proxy will be removed from the list and only working proxies will be used.

If proxy was configured but no proxies as working, it will exit the program with an error that no proxies is working.

Each worker is assigned a specific proxy server to use.

The proxy servers are spread out evenly over the accounts.

## Motivation and Context

I wanted to spread my network load to avoid ip banning and not hammer Niantic's system with X amount accounts from the same ip

## How Has This Been Tested?

This has been tested locally with mixed configurations in config.ini without a problem, traffic goes over select proxy server for that worker.

## config.ini example

		proxy: [socks5://localhost:8123,socks5://localhost:8124] # List of proxy address
		proxy-timeout: 3 # Overrides the default timeout value of 5 seconds and uses 3 seconds instead
		proxy-display: full # Overrides the default value of index and prints full proxy address when using -ps

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1073, opened 2016-08-31
Add support for multiple proxy servers.

mik9:
## Description

User will be able to specify multiple proxy servers.

## Motivation and Context

I have 2 motivation for this change:
1. When map works under one IP server side is able to identify that "this accounts is used in map, not on real device".
2. For now many Tor IP gets banned and sometimes map may stop working because of it. To resume map we need to restart Tor proxy and map. With this change only threads using banned IP will be paused and other may continue work.

## How Has This Been Tested?

Tested with docker with 10 and 1 proxy servers values in config.ini

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -206,11 +206,12 @@ def search_overseer_thread(args, method, new_location_queue, pause_bit, encrypti
             'user': account['username']
         }
 
+        proxy = args.proxy[i % len(args.proxy)] if args.proxy is not None else None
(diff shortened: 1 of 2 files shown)


erikssonk:
@mik9 Have a look at https://github.com/PokemonGoMap/PokemonGo-Map/pull/644

xioxxio:
You are wasting your time this has already been done and working well.

PR #1078, opened 2016-08-31
Spawnpoints rendered as images instead of drawn circles

rbryson74:
## Description
Renders the spawnpoints as images. A series of dynamically chosen images is provided. 

## Motivation and Context
The current method for displaying spawnpoints on the map is to use the Circle drawing function. When the number of spawnpoints grows to several thousand many browsers will start to lag or even crash. 
This change performs better on many systems. High powered computers should be able to visualize tens of thousands of spawnpoints now. 

## How Has This Been Tested?
I have tested this on my map in Chrome. A couple of other users have pulled it to test as well. 

Imgur

## Types of changes

## Checklist:


justsomedudeonthenet:
This is beautiful.  Makes the spawnpoints look nicer, and massively reduces lag.

chuyskywalker:
I'm curious if you could do the same thing, but by ...

I was going to suggest https://developer.mozilla.org/en-US/docs/Web/CSS/filter#hue-rotate()_2, but it turns out IE has absolutely zero support for that.

I'd still like to see a sprite for all these icons instead of a folder full of them, but that's a bit nit-picky.

rbryson74:
I can look into adding them all to a sprite to pull from.

rbryson74:
I should be able to get the spawnpoint icons merged to a sprite. It may take me a little while.

chuyskywalker:
Works perfectly as is, the sprite would just be a nice optimization, but I'll be happy to take a PR for that later and get this in now as is. Tested it myself with a TON of points and it looks great.

Darkknightz:
Is it possible this feature makes my MySQL go 100% cpu usage when I turn on spawn points on my map now? When I turn it on while zoomed out on my town ~3200 spawn points the map becomes unusable.

NoobDev3:
Any explanation for the color coding?

PR #1084, opened 2016-08-31
Fix gevent db

aldarund:
Gevent cant be used with blocking db queries. With blocking anything every other request will wait for a long one to finish. So one long db query -> every other requests are waiting for it to finish.


aldarund:
i vote against using gevent with current state. A proper review of whole code is needed if u really want to switch to gevent and coroutines. Code need to be checked for a blocking operations. You cant have any blocking operations with geven. I cant guarantee that there no other blocking things in the code, that wasnt patched, and i really think they exist.  If need replacement for flask - uwsgi\gunicorn. Not a gevent

chuyskywalker:
Pretty sure that turning the monkey patch on, back in the original pr for this, horribly broke things for a few people.

tai-hun:
@chuyskywalker I think that is why this PR has inserted the monkey patch in the very beginning.

shadi1989:
Editing the runserver.py with "from gevent import monkey; monkey.patch_all() # noqa" on line3 doesn't solve this issue. won't load the workers within the instance and won't start the webserver.

aldarund:
@chuyskywalker you just cant use gevent without monkey patch. There are blocking operations. Like long db query. For example spawn point stat. While a query executing - everything other wont work, no other request will be served until this one is finished.  And this goes to any request. Basically it will be a single threaded server that is blocked by anything. And even with monkey patching for proper work a code review should be done for any other blocking code and operations that are not patched.
@shadi1989 what issue? If it not working at all for u from begin - its a total different issue.

Romfrosk2:
i found that doing the following change makes the code pass:

PR #1033, opened 2016-08-27
Have Get Directions use actual current location instead of blue dot PoGoMap location

pir2:
## Description
Change the URL in the Get Directions link in `map.js` to use actual current location

## Motivation and Context
Using "Current+Location" actually sends the blue dot location of the Pokemon Go Maps.  Using My Location force Google Maps to actually grab the user's current location and use it as the starting point

## How Has This Been Tested?
On my own map

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #992, opened 2016-08-25
Get Directions Not Working


KyleBoyer:
The get directions link, must not be populated correctly with the GPS coordinate of that given Pokemon - I clicked get directions on a Pokemon that was right outside my drive way, and it was trying to take me 15 min away(not even close to my location). I tried to do the same with a different location Pokemon, and it mapped out to the same location as the first time! I believe these links may work for only one Pokemon on the map, not the one selected.

jtp10181:
I tested on 3 different pokemon on my map, and it is working fine on my end.

I am running: https://github.com/PokemonGoMap/PokemonGo-Map/commit/c49cca443ff87d0b0e440c4d30d73f1d7df2d8b3

Using Google CHROME browser.

ultrafunkamsterdam:
@KyleBoyer, you need to set your location by pushing the location button in the bottom right of your map, before starting the directions link.

Slivo-fr:
I think we shouldn't start direction, and simply open pokemon coordinates, user will be able to start direction himself.
I don't think giving the coordinates will do the trick, as it might only suggest to review the path, without starting direction

PR #933, opened 2016-08-22
Notification Improvements

DiscoTim:
## Description
Fixes notification bugs on mobile and desktop and includes a number of new features.
-sound packs
-voice (text to speech) notifications
-notifications based on distance
-functional service worker on mobile

## How Has This Been Tested?

Basic table of supported/tested features:

Windows:

Chrome: Notifications, Sound, Voice
Firefox: Notifications, Sound, Voice (with about:config, media.webspeech.syth.enabled set to true)
Edge: Notifications (must have the anniversary update installed), Sound, Voice
 
Android: 

Chrome: Notifications (With SSL, ie. site uses https), Sound, Voice
Firefox: Notifications, Sound,  (Voice not supported by Firefox on Android yet)

IOS (browser notifications are not supported by iOS):

All: Sound, Voice

On Firefox the setting in about:config , media.webspeech.syth.enabled must be changed to true for voice notifications to work. They will work with no changes on chrome.

Chrome/Android push notifications require SSL (HTTPS) to work, this is a requirement of chrome. Sound and Voice notifications will work without SSL.

iOS/All - iOS does not support local web notifications. Sound and Voice will work.

## Screenshots (if appropriate):
notifications

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1086, opened 2016-09-01
'My location' - Selectable/Updating marker representing users current location

DiscoTim:
This PR adds a selectable and updating location marker representing the users current location.

## Description

The "My Location" marker is a selectable icon from the same list that the search marker uses, including "Hidden". It can be toggled to follow your location in realtime, or it can be dragged to a desired position. It is separate from and has no effect on the scan marker or scanning. It is completely client side and utilizes the existing locationMarker.

This PR should also fix the bug with not being able to reliably retrieve directions. Directions are based on the 'My Location' marker (visible or not) and are queried with definitive start and end locations.

Fixes bug for search follows location url path for sites running in subdirectories.
Fixes bug for searchMarker will be null on /stats and other pages.

Additionally, some of the wording within the location settings was changed for consistency/clarity.

## Motivation and Context

Currently there is no way to track/update your location within the map, and the scan marker is confusing in that many people assume it is related to your location. With spawn point scanning and larger areas being covered this provides a way to determine/set your location in relation to pokemon..etc..

https://github.com/PokemonGoMap/PokemonGo-Map/issues/992

https://github.com/PokemonGoMap/PokemonGo-Map/pull/1033

https://github.com/PokemonGoMap/PokemonGo-Map/pull/619

## How Has This Been Tested?

 This has been tested on Windows and Android, Firefox and Chrome with both geolocation enabled and disable. (Chrome requires SSL for geolocation outside of localhost)

nav1
location2
location3

## Types of changes

## Checklist:


Shibby7986:
cool, I'll try this one out.
I might even have a play with adding sharing the location with other users at some point

Cubkyle:
This works perfectly with your #933. If you could add some sort of persistence to it so that I don't have to set the location every time I reopen the page or refresh, it would be perfect.

Shibby7986:
Working fine for me.

DiscoTim:
@Cubkyle just to clarify, are you not using geolocation, "Follow my location" option? you are dragging the marker to set it?

I've updated it so the location is persistent.

Cubkyle:
@DiscoTim Ah, that's correct. I should have been clearer.

DiscoTim:
@Cubkyle okay great, no problem. It should be doing exactly what you want now. thanks for pointing that out.

Cubkyle:
Persistence works, and in my testing I found it was per-user, as expected. Thanks!

nunopitbull:
I have a question:
I'm running the -nsc and -fl because people were changing their location on map and making my bots get banned. Does this still works with those flags?

DiscoTim:
@nunopitbull I've updated with a new commit, and yes it will work with those command switches, and should with any for that matter. 

I think I've fixed the long standing bug with directions as well. I didn't realize that was an existing bug, thanks for pointing that out.

Romfrosk:
i also wonder how i can reactivate that people can have their map follow their location without doing searches.

i'm running -SS and don't want them to try something funny :p 

will test this PR and see how it goes.

DiscoTim:
@Romfrosk2  this or will do that with the  -nsc and -fl  switches, and they would turn on 'follow my location' , its completely client side, it doesn't effect the search. Each client will have their own location marker. Please let us know how the testing goes.

Shibby7986:
sweet, :) I will turn back on the -nsc :P
my web server instance was not doing any searches anyway but still the icon and options are not needed. for me so its cleaner now

static/js/map.js
```diff
@@ -2230,18 +2284,29 @@ $(function () {
...
-            map.panTo(center)
-            searchMarker.setPosition(center)
+        var lng = position.coords.longitude
+        var center = new google.maps.LatLng(lat, lng)
+
+        if (Store.get('geoLocate')) {
+          var baseURL = location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '')
+          // the search function makes any small movements cause a loop. Need to increase resolution
+          if (getPointDistance(searchMarker.getPosition(), center) > 40) {
+            $.post(baseURL + '/next_loc?lat=' + lat + '&lon=' + lng).done(function () {
```
jaquer: This is broken for installations running in a subdir. See #619

DiscoTim: @jaquer really surprised this hasn't been fixed yet. it should be fixed now. caught another bug here where the searchMarker will be null on stats and other pages, but this function will still be running.

justsomedudeonthenet:
This is an excellent feature I never even realized I wanted!

Tested and is working for me.

PR #1091, opened 2016-09-01
Decrease Number of Upserted Rows

maddhatter:
## Description
[SQLite has a default max of 999 variables in a command](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber), going over that limit causes errors. The "widest" table in the database is `gympokemon` with 17 columns. 999/17 = 58.76 - so only 58 parameterized `gympokemon` records can be inserted at once. Rounding down to add some margin of safety, I've updated the "step" size of `bulk_upsert` to 50 rows at a time.

## Motivation and Context

[Comment on #666](https://github.com/PokemonGoMap/PokemonGo-Map/pull/666#issuecomment-243953614) where user is hitting this issue.

## How Has This Been Tested?

Locally.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -839,7 +839,13 @@ def clean_db_loop(args):
 def bulk_upsert(cls, data):
     num_rows = len(data.values())
     i = 0
-    step = 50
+
+    if args.db_type == 'mysql':
+        step = 120
+    else:
+        # SQLite has a default max number of parameters of 999,
+        # so we need to limit how many rows we insert for it.
+        step = 50
 
     while i < num_rows:
         log.debug('Inserting items %d to %d', i, min(i + step, num_rows))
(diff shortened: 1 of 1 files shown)


karololszak:
I am on the other hand hitting error about "db queue too large" with MySQL... I increased db threads but increasing this variable could be better (less inserts, but larger ones).

So, maybe:

a) Do an if switch (if sqlite then 50, leave 120 otherwise)
b) Do it as a param / config setting

maddhatter:
I don't think that'll have a noticable impact on MySQL - most of it's issues are related to the fact all of the upserts are using `REPLACE INTO` which is a super expensive query.

Becold:
It will have a impact on big scale

maddhatter:
I added a conditional for if it's SQLite or MySQL

chuyskywalker:
LGTM

Guile3home:
awesome!!! no warning issue after the conditional is added up to now!!! thank you!!

PR #1097, opened 2016-09-02
Fix for some cases of 0,0,0 returns

rbryson74:
## Description

With High scan delays, and single threads, an auth token refresh could result in significant gaps in the scan data. 

## Motivation and Context

This fix sets a specific delay after the auth token is refreshed and removes a extra set_location call that occurred only for auth token refreshes. 

## How Has This Been Tested?

I have only tested this on my system. I was able to clear all the cases of 0,0,0 on auth refresh though I was still seeing very rarely 0,0,0 after the first token again. 

## Types of changes

## Checklist:

Some users reporting 0,0,0 returns on subsequent auth token refreshes
after the first.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -491,14 +491,17 @@ def search_worker_thread(args, account, search_items_queue, pause_bit, encryptio
                     # No sleep here; we've not done anything worth sleeping for. Plus we clearly need to catch up!
                     continue
 
-                status['message'] = 'Searching at {:6f},{:6f}'.format(step_location[0], step_location[1])
-                log.info(status['message'])
-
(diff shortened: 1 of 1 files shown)


aldarund:
Why scan delay was hardcoded to 20?

rbryson74:
That line was using scan_delay but it wasn't actually a scan delay. That was the wait time after a new token was received. If you had a very high scan delay set (for whatever reason, doesn't really matter) the token refresh would add a much longer delay than it really needed. I found that 20 seconds helps to give it enough time to prevent most of the 0,0,0 responses that get through.

PR #1100, opened 2016-09-02
element.matches polyfill for app.js

DiscoTim:
This PR adds a element.matches polyfill to app.js 
## Motivation and Context

This change is required for browsers that do not support element.matches, which app.js uses.

For example, when clicking anywhere, Microsoft Edge throws the exception:

SCRIPT438: Object doesn't support property or method 'matches'
app.min.js (5,1660)

which is line..

Line: 127
if (event.target.matches('a[href="#nav"]')) {

This error prevents the Nav and Stats from closing, except by clicking directly on the button.

https://github.com/PokemonGoMap/PokemonGo-Map/issues/1212
## How Has This Been Tested?

tested on windows ff/chrome/edge android/ff/chrome and ios/safari and the matches function normally.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -2,6 +2,20 @@
   'use strict'
 
   // Methods/polyfills.
+  if (!Element.prototype.matches) {
+    Element.prototype.matches =
+      Element.prototype.matchesSelector ||
+      Element.prototype.mozMatchesSelector ||
+      Element.prototype.msMatchesSelector ||
+      Element.prototype.oMatchesSelector ||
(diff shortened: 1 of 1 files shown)


Romfrosk:
Looks like this will make the map work in my Tesla. if so. i own you a case of beer or whatever you drink at the pub :+1:

DiscoTim:
haha, nah its solid. it's too bad babel doesn't have it included, and surprising its missing from edge.

PR #1090, opened 2016-09-01
Status Page (#908 Resubmission)

maddhatter:
## Description
Resubmitting #908 with merge conflict resolutions, and two minor fixes:
1. Fix typo
2. Workers clear all their old statuses when they start

## Motivation and Context
#908 has not been being updated

## How Has This Been Tested?

See #908

## Screenshots (if appropriate):
Example Login Page
Example Status Page

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #1102, opened 2016-09-02
Add proxy information on /status page

erikssonk:
## Description

Adds information about which worker is using which proxy on the /status page
## Motivation and Context

As #644 and #1090 was merge into develop. I wanted to see which proxy was used by which worker on /status page
## How Has This Been Tested?

Viewing of status page has been tested in FF, Chrome, Safari on ios 9.3.5

Testing that db is updated has been done locally

status-page
## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1274,4 +1274,3 @@ def database_migrate(db, old_ver):
         migrate(
             migrator.add_column('workerstatus', 'proxy', CharField(null=True, default=""))
         )
-        


slideup-benni:
Can you put the proxy-col after skipped? I think the Username is more important then that?

erikssonk:
@slideup-benni Sure, it can be moved. The reason for me putting it first is that is directly linked to the account being used and makes it easier to separate that from information about the worker and it's progress. For me, this feels more natural to read.

erikssonk:
@slideup-benni Tested your way and looks ok, it's pushed to the PR

mbg032373:
If proxy_display is set to `index`, the /status pages shows index 0 as "No" instead of "0".  If proxy_display is set to `full`, the correct proxy address is displayed.

erikssonk:
@mbg032373 thanks for the heads up on that one, I'll fix that pronto

slideup-benni:
@erikssonk what is the difference between "False" and ""? or is this a bug?
spectacle e12521

jchristi:
Is the db schema migration automatic or is there a command I have to run? Nothing in the docs about this.

erikssonk:
@slideup-benni That's a bug and I'll look into that now.

@jchristi It's automated, it will add a field name proxy to the table workerstatus

slideup-benni:
@erikssonk after a while all blank fields filled with "False". I think this was, because i upgraded to your branch, but the default of the Database is blank

pogom/models.py
```diff
@@ -982,5 +983,10 @@ def database_migrate(db, old_ver):
     if old_ver < 7:
         migrate(
             migrator.drop_column('gymdetails', 'description'),
-            migrator.add_column('gymdetails', 'description', TextField(null=True, default=""))
+            migrator.add_column('gymdetails', 'description', TextField(null=True, default="")),
```
slideup-benni: why you add a comma?

erikssonk:
@slideup-benni Odd, i'm looking into this now and see if can reproduce this. Seems something has changes within the latest merge from develop.

pogom/models.py
erikssonk: My mistake, shouldn't be a comma there. It's going to be removed.

erikssonk:
@slideup-benni Could you please check if my commit fixes the unwanted behavior in status page for proxy?

slideup-benni:
I'm getting a "0" now
screenshot_20160906_192316

erikssonk:
@slideup-benni Do you have console view too? Does that work properly? Are you able to supply me with the proxy setting?

slideup-benni:
@erikssonk No i have no console view, because i'm using a beehive. but maybe this will help you:
In my database only "No" and "0" stands in the proxy col

erikssonk:
@slideup-benni Are you on discord? Would like to chat with you instead so I can solve this.

slideup-benni:
@erikssonk yes. I'm shugo1110

slideup-benni:
We fixed all my problems. From my side of view we can merge it.

erikssonk:
@slideup-benni Thanks for the help with resolving this issue.

BestYohman:
Awesome!  I was about to ask for this in a feature request.  Anyway on your next project you could add pausing each scan?  You do such great work!

djv2:
thoughts on this error when worker starts?
`[status_worker_db][        search][    INFO] Clearing previous statuses for 'workername' worker
Exception in thread status_worker_db:
  File "search.py", line 198, in worker_status_db_thread
    'proxy': str(status['proxy_display']),
KeyError: 'proxy_display'
`

erikssonk:
@djv2 Never encountered that error before. Seems that proxy_display is missing from status variable. 
Took a look in pogom/search.py and that row (In this PR) is status_text.appen() for console on line 198.
Would like to have a look at your pogom/search.py to see the difference between my pogom/search.py och yours. But from just looking at your error message and looking at the file, I'd say that modifications has been made to your's pogom/search.py and that change is what causing this.

djv2:
@erikssonk since it doesn't seem like i can get it to merge with my personal repo, my question is - is it possible without the use of this https://github.com/PokemonGoMap/PokemonGo-Map/pull/1102 feature to see which accounts are using what proxy? is there a file log or any type of display at all?

erikssonk:
@djv2 I don't think you can have that info on the status page without this PR. In console you can see the current proxy being used buy which account with --ps

jchristi:
@FrostTheFox this seems like a fairly non-controversial change and a nice enhancement. Any chance we can get this merged sooner rather than later? Thanks.

pogom/search.py
```diff
@@ -311,7 +313,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, encryption_lib_p
         if args.proxy:
             proxy_display = proxy_url = args.proxy[i % len(args.proxy)]
             if args.proxy_display.upper() != 'FULL':
```
jchristi: Can you consider making 'FULL' the default and have 'INDEX' be optional? This was very confusing for me and for a long time thought the code was buggy. I don't see the value in showing the index (unless your proxies are extremely long URL?).

jchristi: Furthermore, can we allow this to be specified in `config.ini`? The less CLI args I have to specify for every process the better.

erikssonk: All these settings can be set via config.ini, example below
`proxy-timeout: 30`
`proxy-display: full`

I set proxy-display to default index when I coded #644 and now there's another pt #1235 that are in the works. I think it's best to change it in that PR.

jchristi: Good to know its configurable.

obsess: Super excited, and hope this gets merged! Great addition.

erikssonk:
@FrostTheFox @primaxius Any updates on your end with this PR?

Issue #1001, opened 2016-08-26
Times show Pokemon despawn in Stats page and not spawn time.


EmpireRoman:
When you select all locations in Full Stats, the times listed for a spawn is when the pokemon would despawn and not when it would spawn.

## Expected Behavior
If I select a spawn point, it shows spawn time.

## Current Behavior
If I select a spawn point, it shows despawn time.

## Possible Solution
I checked the code in Models.py and from what I see it is not subtracting the 15 minutes off of the times. I am learning as I go, so I might be off of my rocker.

## Steps to Reproduce (for bugs)
1. Click Stats
2. Click Full Stats
3. Select Rattata All Locations
4.Hover spawn point, notice that times match despawn time on live map, not spawn time.

## Context
I am trying to produce pictures of spawn times for 1 species of Pokemon, but I have to subtract 15 minutes for it to be correct.

## Your Environment

* Version used: 3.1.0
* Environment name and version (e.g. Python 2.7): Python 2.7.12
* Operating System and version (desktop or mobile): Windows 10 Build 10586

Thank you for your help and all that you do.

BaIthamel:
Please go educate yourself on the issues we work with on this project, and the wonderful contributions from tbterra then come back and close this.

CoolCanuck:
EmpireRoman, don't mind BaIthamel. They're rude to every single person without reason. Not just you :) 
You can see from any of his comments/closures that his comments are not beneficial to anyone's mental health and he couldn't care less

PR #1103, opened 2016-09-02
Optimize stat heatmap

aldarund:
## Description
Dynamically load appearances times when its requested
Calculate number of pokemon on spawnpoint in python (mysql actually), not js. So no long arrays of data passed from backend, only lat long count spawnpoint_id.

## Motivation and Context
Because loading all location slow and transfer a lot of data.
## How Has This Been Tested?
my city map

## Types of changes

## Checklist:


aldarund:
Also fixes https://github.com/PokemonGoMap/PokemonGo-Map/issues/1001

Zerotask:
I have 1.5 mio finds and I'm using MariaDB 10.1 and I don't notice any speed improvements.

It should bundle the requests.
image

aldarund:
@Zerotask um. This pr  has nothing to do with page load.This is optimization for all location button. And it certainly dont do 142requests. There like 2 json requests to backend that are repeated every few seconds. 142 requests are for the images on the stat page.

Zerotask:
oh the images, yeah right. I just hoped that this PR reduces the page load because, as you wrote, you load the appearances first when a heat map is requested. Maybe you want to look after my problem and how to reduce the page load

aldarund:
@Zerotask hm  i was sure i wrote answer here, but seems like i didnt post it. I guess the main load time for stat page is a long db query that is counting all pokemons. For 142 request it should use same spite as main map.   But this pr begin to work when a all locations link clicked not before.

justsomedudeonthenet:
**This makes heatmaps usable again**

Trying to load rattata heatmap for year without this patch:  Tries to transfer 20MB **every few seconds**

With this patch:  Loads a few KB.  Incredibly fast, doesn't kill my internet, doesn't freeze my browser.

Merge this now pl0x!

j16sdiz:
this is incompatible with sql_mode=only_full_group_by

erikssonk:
Had the same as @j16sdiz needed to remove only_full_group_by to get this working.

aldarund:
@j16sdiz @erikssonk will make a pr with fix today, good catch

PR #1109, opened 2016-09-03
Add colons to config.ini.example

averylarry:
## Description

Added a couple colons to config.ini.example
## Motivation and Context

I think the colons are supposed to be there.
## How Has This Been Tested?

Well, it's a text file, so I looked at it.   ;-)
I added the colons and uncommented the lines and the settings worked.
## Types of changes

- [X] Consistency/typo

## Checklist:

I was setting the new parameters for the status page and noticed the colons were missing.  At least I assume they are supposed to be there.

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -21,7 +21,7 @@
 #step-limit:            # default 12
 #gym-info:              # enables detailed gym info collection (default false)
 #min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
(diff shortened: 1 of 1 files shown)


SlasherKG:
Hmm, I think those only work with a space, not a colon (i.e. "status-name status1" and not "status-name:status1").
At least I could only get it to work that way.

averylarry:
I definitely have both settings working with a colon.  I'm on windows 7.
It also works for me with a space instead of a colon.

PR #1115, opened 2016-09-03
Read usernames and passwords from file

springjools:
## Description
Provide a means for reading login credentials from file. It works like this: let the username field begin with 'file:' and it will use the rest of that field as the path to your credentials file. The works both as python runserver.py -u file:somepath/accounts.csv as well as from config.ini. The path can be absolute or relative (to runserver.py).

## Motivation and Context
It is the belief of the author that this is a feature that has been requested and that will be welcomed

## How Has This Been Tested?
Tested with powershell on windows 7. Not tested with unix/linux, it would be nice if someone could test that part. Mostly how the dashes in filename paths work. But I assume that the user can form the filename path to such a string that it works on any system.

readfromfile

## Types of changes

## Checklist:

Will update the documentation in a timely manner 

when it is needed and topical.

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -15,7 +15,6 @@ from . import config
 
 log = logging.getLogger(__name__)
 
-
 def parse_unicode(bytestring):
     decoded_string = bytestring.decode(sys.getfilesystemencoding())
     return decoded_string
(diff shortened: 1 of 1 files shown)


springjools:
Forgot to mention that the only requirement of the accounts file is that it is comma separated, and that each line begins like this:
auth-service,username,password

Each line can contain more stuff. It is read as line and separated by commas, and the first 3 fields are used.

springjools:
untitled
In case the file cannot be read...

Albert62:
THANK YOU!

aldarund:
Please add a error handling for not proper files

justsomedudeonthenet:
#1067 already handles CSV files, with proper error checking.

FrostTheFox:
#1067 was merged, replaces this.

PR #1122, opened 2016-09-04
Add args.optimize_workers: a simple worker minimizer.

sfjuocekr:
Introducing a simple, yet effective, worker minimization algorithm.

## Description
Adds, stops or starts workers when required based on scan-delay.

## Motivation and Context
Keeps requests to the server and used accounts at a minimum.
Recycles accounts upon stopping.

## How Has This Been Tested?
Tested and tweaked over the past few days, no accounts have been banned.
With roughly 5000 spawn points spread over town it keeps around 14 accounts active with 10 second scan-delay, it swings between 12 and 18 workers.
No fails or skips.

## Types of changes

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -471,10 +471,10 @@ def start_worker(wh_queue, db_updates_queue, encryption_lib_path, pause_bit, sea
         }
 
         t = Thread(target=search_worker_thread,
-                       name='search-worker-{}'.format(worker_index),
-                       args=(args, account_queue, account_failures, search_items_queue, pause_bit,
-                             encryption_lib_path, threadStatus[workerId],
-                             db_updates_queue, wh_queue))
+                   name='search-worker-{}'.format(worker_index),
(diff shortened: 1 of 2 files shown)


justsomedudeonthenet:
I really don't see any advantage to doing this.  Having more accounts actively scanning means they can be idle longer between scans, which means less bans for moving too fast.

What is the advantage to this?

averylarry:
I would think the benefit could be in having spare unused accounts that can be called into action as needed -- when other accounts get banned.  Overall maintaining fewer accounts.

For the record, though -- I don't really see much advantage either.

sfjuocekr:
There might be no advantage, but if you for example set your scan delay higher it will use more accounts and if you set it to less it will use less; but it will always wiggle around scan delay behind using a minimum of accounts.

Over the past week the only accounts that were banned are those which ONLY accepted the ToS!

primaxius:
declined - needs rebase and kinda silly

Issue #1149, opened 2016-09-05
[Feature Request] Webhook backend


adosikas:
As this app sends everything a worker would find via webhooks (except scanned areas but those aren't really neccessary), it would be nice to have the option of a built-in webhook _server_ that recieves pokemon/gym/fort data instead of relying on local workers or mysql.

This would be especially useful, as cloud hosting providers are banned from the api and setting up a publicly acessible mysql is not supported everywhere.  
One could run workers somewhere else, like at home or at a place with limited bandwidth, but still have the advantages of cloud hosting, such as fast response times and scalability.

I hacked up a quick test for pokemon only, however i am not too comfortable with the used frameworks and data structures to make a proper (high quality) pull request myself.

~~Diff: https://gist.github.com/adosikas/b658460f76aad5266a94b6e090226dd9~~

Edit: Figured out how it works (i think) and made a pull request, see #1161

sebastienvercammen:
It's not the map's job to be on the receiving end of the webhooks. Closed.

PR #1161, opened 2016-09-05
Add webhook server

adosikas:
## Description
* Add simple listener for webhooks on a path defined by --webhook-server-path, e.g. webhook_secretstring
* Relay db_updates_queue to the app, as the web thread needs to access it when recieving data
* Added setting to example config

## Motivation and Context
As mentioned in #1149 the abitity to have a cloud hosted instance that relies on data sent via webhooks instead of own workers (ip ranges banned by api) or a publicly accessible mysql database (not possible on some hosts) is quite useful.

## How Has This Been Tested?
I ran this for a few hours, both local and on a heroku instance and it works just as expected.
Sending data were two unmodified instances running locally, behind a IPv4 CGNAT (DS-Lite connection).

## Types of changes

## Checklist:

## Notes:

* I am not too comfortable with the used datastructures, so I hope I converted the webhook data correctly and hope someone can verify or correct that.
* Not sure if adding to the example config counts as updating the documentation or if there is something else.
* As one webhook is sent per object, it causes a lot of traffic on both ends, maybe bundling all results of a map_parse into a single webhook would make sense
* As anyone with the webhook url would be able to add objects I decied to make it configurable, instead of breaking backwards compatibility by adding a password or something alike


pogom/models.py
```diff
@@ -691,6 +691,43 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue):
         'gyms': gyms,
     }
 
+def add_webhook(wh_type, wh_message, db_queue):
+    if wh_type=='pokemon':
+        pokemons = { 1: {
+		   'encounter_id': b64decode(wh_message['encounter_id']),
```
j16sdiz: don't store the b64decoded version. 
other parts of the map use encoded one.

justj0sh:
Or you could just stick the database on AWS and make a read only user?

primaxius:
Just setup a mysql server on a cheap vps.

PR #1165, opened 2016-09-05
Refactor search.py, pull out scheduling into separate class

justsomedudeonthenet:
## Description
This moves all the scheduling decisions (hex search, spawnscan, hex search for spawnpoints only) out of search_overseer/search_worker and into a separate schedulers class.

No functionality changes.

## Motivation and Context
search.py is still a mess.  There are quite a few more efficient scheduling mechanisms people are working on, this starts paving the way for them to just implement a new scheduler class instead of wedging it into search.py

This simplifies search.py, and makes it easier to add future improvements like queue-per worker, worker groups, etc.

This is phase 1.  Once this is in, phase 2 will be one queue per worker (to allow more flexible scheduling).  Phase 3 will be multiple locations in a single python process.

## How Has This Been Tested?
Have tested against fake API in all three scan modes.  Still doing more testing.  There are still some bugs.  But I want feedback on this approach.

## Types of changes

## Checklist:


pogom/schedulers.py
```diff
@@ -0,0 +1,366 @@
...
+        self.empty_queues()
+
+    # Function to empty all queues in the queues list
+    def empty_queues(self):
+        for queue in self.queues:
+            if not queue.empty():
+                try:
+                    while True:
+                        queue.get_nowait()
+                except Empty:
```
aldarund: afaik after each get there should task_done called. 
Something like this look better
while not queue.empty():
    try:
        queue.get_nowait()
    except Empty:
        continue
    q.task_done()

pyr0ball:
will clone and test using real-world scans and existing database. Feedback incoming

pogom/schedulers.py
justsomedudeonthenet: True...it doesn't really matter, but that is the proper way to do it, so I'll fix that.

Issue #1136, opened 2016-09-04
-ss is looking simultaneously at the same spawn point with 2 workers


Imoq:
I cover 3 small areas (-st 5) with -ss and I only turn the scan on (start the workers) when I will be around those areas. I use 2 workers for -st 5 area since I like to grab gym info as well (-gi). Since last night, I started noticing that both workers go to the exact same spawn point (exact coordinates) so, they seem to be duplicating each other's work.

I ran several tests before opening the issue, sometimes when I start the process they both attempt to go to the same point right away, and other times one worker attempts to go to one point, the second to another but after few iterations they both end up going to exactly the same one at the end.
## Expected Behavior

Each worker to go to a different spawn point!
## Current Behavior

Both workers go to the same spawn point, finding the same Pokémon and being inefficient by duplicating the work.
## Possible Solution

Make next worker to go to a different spawn point if there is one worker already on one spawn point.
## Steps to Reproduce (for bugs)

1. Start process with at least 2 accounts. 
2. Look at the log for a bit and realize both accounts go to the same spawn point.
## Context

My parameters are: "-l "43.288218, 5.380372" -st 5 -sd 10 -dc -ns -ss -gi -sn prefecture -ac accounts/prefecture.csv -w 2 -asi 10800". **PLEASE NOTICE** this behavior started before using -ac -w -asi. It most likely started yesterday.

I have some log here to provide more insight into the issue. On this example, both workers started on different spawn points, but they converged to the same time quickly. I have logs with other examples when the workers attempt to start on the exact same one since the beginning.

```
2016-09-04 19:22:50,958 [      MainThread][     runserver][    INFO] Parsed location is: 43.2804/5.3839/28.6173 (lat/lng/alt)
2016-09-04 19:22:50,985 [      MainThread][        models][    INFO] Connecting to MySQL database on localhost:3306
2016-09-04 19:22:51,079 [ search-overseer][        search][    INFO] Search overseer starting
2016-09-04 19:22:51,084 [ search-overseer][        search][    INFO] Starting account recycler thread
2016-09-04 19:22:51,093 [ search-overseer][        search][    INFO] Starting status database thread
2016-09-04 19:22:51,096 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-09-04 19:22:51,099 [status_worker_db][        search][    INFO] Clearing previous statuses for 'perrier' worker
[...]

Issue #1075, opened 2016-08-31
Duplicate spawnpoint scans due to different latitude & longitude roundings


hagehg:
I have older entries in my pokemon table where the latitude was stored with 12 decimals (12 numbers after the . in case decimals is not the right word here). 

Since late at night on August 18th (presumably I pulled a new version of the map then) I get 13 decimals instead.

Example:
![decimals](https://cloud.githubusercontent.com/assets/20759246/18128837/0d641580-6f88-11e6-97df-1af84ac65f3d.png)

When doing spawnscanning, my scanners will regard the 12 and 13 decimal entries of the same spawnpoint as two different locations, resulting in two scanners scanning the same location twice simultaniously:

```
2016-08-31 13:43:12,684 [ search-worker-0][        search][    INFO] Early for xx.xxxxx3,xx.xxxxx6; waiting 11s...
2016-08-31 13:43:13,144 [ search-worker-1][        search][    INFO] Early for xx.xxxxx3,xx.xxxxx6; waiting 10s...
2016-08-31 13:43:23,248 [ search-worker-1][        search][    INFO] Searching at xx.xxxxx3,xx.xxxxx6
2016-08-31 13:43:23,788 [ search-worker-0][        search][    INFO] Searching at xx.xxxxx3,xx.xxxxx6
2016-08-31 13:43:35,043 [ search-worker-1][        models][    INFO] Parsing found 4 pokemons, 26 pokestops, and 8 gyms
2016-08-31 13:43:35,387 [ search-worker-0][        models][    INFO] Parsing found 4 pokemons, 26 pokestops, and 8 gyms
```

I have 10 of these cases in my database (8 with latitude, 2 with longitude). Depending on how many one has, this could be a big waste of resources.
## Expected Behavior

The scanners should realize they're scanning the same spawnpoint regardless of decimals.
## Current Behavior

Two scanners scan the same spawnpoint at the same time
## Possible Solution

Maybe use spawnpoint_id in addition to latitude and longitude somehow
## Steps to Reproduce (for bugs)

1. Run this query on your datase :
   `SELECT spawnpoint_id, latitude, char_length(latitude)-3 AS decimals FROM pokemon GROUP BY spawnpoint_id, decimals;`
2. Look for spawnpoint_ids that appear more then once in the results (I did that by copying the data to Excel and used it's 'Highlight duplicate entries' feature)
## Your Environment

- Version used: dev branch from yesterday
[...]

PR #1179, opened 2016-09-06
Error #1136 fixed

ragreener1:
Stopped using distinct on columns as causes errors if switching from SQLite to MySQL
## Description

Very minor fix to solve bug when changing database - possible performance impact but not significanr
## Motivation and Context

This fixed error #1136 
## How Has This Been Tested?

Tested - fix worked as expected
## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -274,11 +274,8 @@ class Pokemon(BaseModel):
                              (Pokemon.longitude >= w) &
                              (Pokemon.longitude <= e)
                              ))
-        # Sqlite doesn't support distinct on columns
-        if args.db_type == 'mysql':
-            query = query.distinct(Pokemon.spawnpoint_id)
-        else:
-            query = query.group_by(Pokemon.spawnpoint_id)
+
+        query = query.group_by(Pokemon.spawnpoint_id)
 
         s = list(query.dicts())
 


justsomedudeonthenet:
I don't see any appreciable speed difference between the two select statements...in fact, they should return identical results.

Why does this fix the issue?

averylarry:
I don't think anyone ever figured out why it didn't work correctly.  If I remember correctly it was discovered by trial-and-error that the group by syntax appeared to return correct results.  I don't even know if the results were verified.

My suspicion, though, is that using distinct on 2 fields isn't doing what people think it should.  Group by using 2 fields should work the way people think it should.

justsomedudeonthenet:
Wait...its doing distinct on 2 fields?  I didn't catch that.  Which fields?

tiagomartines11:
`GROUP BY` returns the expected results, it only groups whatever is specified on the GROUP BY clause.
`DISTINCT` returns wrong results, as it only considers as duplicate rows those that have all the same data (e.g., a duplicate spawnpoint_id might show up because they have different lat/log due to rounding mistakes)

For MySQL, the distinct query was introduce based on two reasons:
1) MySQL 5.7+ will complain/fail to execute that query, because it enforces sql_mode=only_full_group_by (newbie users had difficulties to disable/understand that).
2) It may have significant performance impact depending on your setup.

I believe we should stick to the DISTINCT query somehow, it seems more appropriate given that we are not doing any aggregation operations. I suggested an alternative to this issue on #1075, but that only tackles the lat/lon situation.

averylarry:
Let's define disappear seconds as the number of seconds past the hour that a given spawnpoint disappears.  (disappear time's minutes * 60 plus disappear time's seconds).
In my data, many spawnpoints have multiple disappear seconds.  I'm not talking about multiples of 15 minutes either (for the 2x15, 30, 45, 60 spawns).  Some of them are only a second or 2 apart.
So the first obstacle -- that does NOT seem correct.
But moving on -- the entire premise of the select statement seems wrong to me (noting fully that I don't know any coding languages but I am decent with SQL queries).  It is supposed to return lat, lon, disappear seconds.  Now lat,lon is unique to each spawnpoint_id.  But disappear seconds isn't.  From a somewhat pure SQL syntax I would do this:
select latitude, longitude, disappear seconds
from pokemon
group by latitude,longitude,spawnpoint_id,disappear seconds

In SQL, this query is invalid:
select latitude,longitude,disappear seconds
from pokemon
group by spawnpoint_id

and so is this:
select latitude,longitude,disappear seconds, distinct(spawnpoint_id)
from pokemon

So I don't fully understand how sqlite and mysql decide what data to throw away.  Nor do I understand why we're throwing away data.
Still noting that I might totally misunderstand the code.

averylarry:
#1075 is a separate issue.  That is bad (or at least inconsistent) data in the database, which (in my opinion) should not be fixed by changing how we query the database.  The bad data should be fixed.

justsomedudeonthenet:
Has anyone tested this on MySQL 5.7, with ONLY_FULL_GROUP_BY turned on?

beccasafan:
The duplicates are currently needed due to the way the map handles (well, doesn't handle) the [non-15m spawns](https://www.reddit.com/r/pokemongodev/comments/4yzqc2/spawn_point_types_clearing_confusion/).

When a spawn comes back with an invalid TTH, the code forces that down to 15m and this is why there are multiple times for the same spawnpoint.

Let's use a 1x30 spawnpoint as an example. This spawn is active from 0m-30m. It has a valid TTH from 15-30m.

If you scan this spawnpoint between 15m and 30m, it will have a valid TTH and every instance will have the same mm:ss disappear time.

However, if you scan this spawnpoint between 0m and 15m, it will have an invalid TTH. Let's say this scan happens at 2m. The code will force that TTH to 15m and thus the pokemon is magically active from 2m until 17m.

The spawnpoint now has times from 2m to 17m, and from 15m to 30m in the database.

If you remove the duplicates, then spawnpoint scanning might pick up that location for 2m to 17. Or it might pick it up from 15m to 30m. It's non-deterministic.

The "duplicates" are kept here so that when looking at the spawnpoint data and doing spawnpoint scanning, that spawnpoint will be scanned at these multiple times.

For another way to handle it, check out the get_spawnpoints method. This is the method used to get the data when you turn on the spawnpoint circles in the ui, and the data that shows in the info window when you hover/click on a spawnpoint circle. The logic here does handle the multiple times for a single spawnpoint. The logic in this method will return a single time for every spawnpoint by figuring out which one is the "real" one - which one has been seen the most. Since everything in the final 15m will have the same spawn time, that works out to be the "chosen" one.

If you hover over a spawnpoint that isn't a regular 15m spawn, it says "May appear as early as ...". This is because with the current handling of these other types, we know it's different, but we don't know what type it is. Still a cop out, but it's something.

Long story short, we kinda need the "duplicates" right now because they kind of aren't duplicates and we want to scan that point at the expanded times to look for a spawned pokemon.

*Whew!*

sebastienvercammen:

PR #1199, opened 2016-09-07
Allow only <username>,<password>  or only <username> in -ac CSV

Darkknightz:
## Description

As long as auth-service has been defined once (or defaults to PTC) this allows for only specifying username and password in CSV.

Also allows for only specifying 'username' in CSV if password is specified once. (either in config.ini or -p)

Also slightly changed error handling to account for 2/3 fields and optimized code.

## Motivation and Context

1. When using only PTC accounts this would not need to be specified in the CSV.
2. When using same password for all accounts this allows for only specifying a list of usernames and specify password only once (in config.ini or with -p ).

## How Has This Been Tested?

Locally

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -204,7 +204,7 @@ def get_args():
                     csv_input.append('<ptc/gmail>,<username>,<password>')
 
                     # If the number of fields is differend this is not a CSV
-                    if num_fields != line.count(',') + 1: 
+                    if num_fields != line.count(',') + 1:
                         print(sys.argv[0] + ": Error parsing CSV file on line " + str(num) + ". Your file started with the following input, '" + csv_input[num_fields] + "' but now you gave us '" + csv_input[line.count(',') + 1] + "'.")
                         sys.exit(1)
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -225,8 +225,8 @@ def get_args():
 
                     # If the number of fields is three then assume this is "ptc,username,password". As requested..
                     if num_fields == 3:
-                        # If field length is not longer then 0 something is wrong!
-                        if len(fields[0]) > 0:
+                        # If field 0 is not ptc or gmail something is wrong!
+                        if fields[0].lower() == 'ptc' or fields[0].lower() == 'gmail':
                             args.auth_service.append(fields[0])
                         else:
                             field_error = True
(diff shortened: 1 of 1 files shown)


justsomedudeonthenet:
My opinion:  It's not hard to generate a proper CSV file.  Tools I have seen to make accounts already spit it out in the proper format.

Edit it in excel if you want.  Or use search/replace to put in the proper fields.

I would rather catch potential errors by being strict about the parsing of usernames, than take a badly formatted file and try to guess what the person meant.

Darkknightz:
Seems like you're taking this personal, I'm just adding to your code which was great. Take it as a compliment I'm adding to and improving what you wrote.

But first of all it didn't catch 'method,username,'
As being empty password because of the newline symbol. So the len(field[2]) would never be 0.

Second of all I hate having more in a file then needed. If I set my password in the config  file and method defaults to ptc why force me to add those.

I feel a little stalked, every pull/issue I made had a comment from you :)

Ps: first time touching Python and Github

justsomedudeonthenet:
Not personal at all.  I welcome improvements :)

I just prefer that the csv files be properly formatted with the same number of columns in each row.

It is a CSV file.  Put a header on the top line and see how it reads:

Auth Method, Username, Password
ptc,frank,secret
bob, hunter2

See the problem?  hunter2 is in the username column now.

Yes, we can guess what was intended.  But we shouldn't.  Having everything in the right columns makes things like editing it in excel (or any other tool) easier.

Darkknightz:
Point taken, accepting multiple formats of proper csv's now (user only/user,pass/method,user,pass)

justsomedudeonthenet:
I like this method a lot better!  So you would specify auth method (and maybe password in config.ini or command like then, and get the usernames (and maybe passwords) from the csv, right?

Darkknightz:
Indeed, and because auth method defaults to PTC you can skip it all together when you use all PTC accounts. Which I assume 99% of all users will.

justsomedudeonthenet:
Alright, changing me :-1: to a :+1:
I will look over the code and test it in a bit.

djv2:
@Darkknightz can you have a mix of user,pass/method,user,pass in the csv? or do you have to select 1 or the other?

Darkknightz:
@djv2 at this point it's one or the other to stick to true csv

mbg032373:
This change sucks.  Now I have to actually edit what I get from the ptc account gen script?  Sigh.

Is there a way to turn this horrible change off and have it work the way it did before?

k4n30:
@Darkknightz & @FrostTheFox 

Yeah, this really needs to put back the option to use default output from ptc account gen script.  That's what the majority of users are going to use, so dropping support for some obscure use case where someone doesn't want to enter all the fields is ludicrous

Darkknightz:
I see the error of my ways, i have uploaded new PR #1277. This allows for more than 3 fields in the CSV where the extra fields are ignored.

PR #1209, opened 2016-09-07
Fix math domain error in the spawnpoint clustering script

gabrielsoldani:
## Description
This adds a check to prevent us from calculating `acos` of something greater than 1.

## Motivation and Context
Fixes an issue that can very occasionally occur when we're calculating the distance between points `a` and `b` and `a != b`, but `radians(a) == radians(b)` due to floating point accuracy problems.

## How Has This Been Tested?
User @Obihoernchen in #990 tested it and it fixes the issue.

## Types of changes

## Checklist:


ChipWolf:
What's the difference between compressed spawnpoints and cluster data?

PR #1218, opened 2016-09-08
Wrap "expire lures on Pstops during db cleanup" with a "do we parse Pstops" check

karololszak:
## Description
If we don't parse pokestops, then we don't need to expire old lures. Simple as that.

## Motivation and Context
Why it bothers me? I have a huge db of pokestops (and gyms) and sadly this query runs too many times and produces no meaningful results, because they all are already lure-expired...

## How Has This Been Tested?
It wasn't tested. Travis to the rescue! ;)

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -886,10 +886,11 @@ def clean_db_loop(args):
             query.execute()
 
             # Remove active modifier from expired lured pokestops
-            query = (Pokestop
-                     .update(lure_expiration=None)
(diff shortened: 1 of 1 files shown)


aldarund:
Not really good. There should be option to disable db clean completely. Because right it will clean on every launched workers, which is no good. And when that option will be added - this pr will become a trouble, because cleandb can be run on one instance like -os, but other will still run pokestops

karololszak:
There is a "disable db clean" PR already merged, so I close this one.

PR #1220, opened 2016-09-08
Add Seconds Behind Counter for Console/Status Page

TheBlueGreen:
## Description
Added a seconds behind counter to the worker scan message. Makes it easier to see at a glace how far your workers are behind in their scanning without have status-names for each worker. View-able on console and status page. Requested on Discord.

## Motivation and Context
Makes it easier to see at a glace how far your workers are behind in their scanning without have status-names for each worker. Requested on Discord.

## How Has This Been Tested?
Tested on my personal server and by the user who requested this change on Discord.

Screenshot

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -633,7 +633,11 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
                     search_items_queue.task_done()
                     status['skip'] += 1
                     # it is slightly silly to put this in status['message'] since it'll be overwritten very shortly after. Oh well.
-                    status['message'] = 'Too late for location {:6f},{:6f}; skipping'.format(step_location[0], step_location[1])
+                    seconds_behind = now() - appears
+                    if seconds_behind > 10:
+                        status['message'] = 'Searching at {:6f},{:6f}. [Seconds behind: {:.0f}]'.format(step_location[0], step_location[1], seconds_behind)
+                    else:
+                        status['message'] = 'Searching at {:6f},{:6f}. [ON TIME]'.format(step_location[0], step_location[1])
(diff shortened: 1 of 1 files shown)


beccasafan:
This feels a bit misleading, because it's the instance that's behind, as a combination of the accounts that are running in that instance, right? It isn't that that single account is behind?

Aside from that, this also changes the logic/ranges of displaying the behind status with "on time", but doesn't update the MainWorker. So now the individual account would say "on time" but the main worker would say it's 8s behind, for instance?

Also, it looks like it will never show that a worker is ahead, it would always show on time?

(Note I didn't pull down the PR, just glancing at the code so just yell if I read something wrong!)

stevenpratt:
This is just saying that when the scan takes place it is more than 10 seconds after the spawn happened, if not it is considered on time. If it is 8 seconds behind that is considered on time as the threshold used the previously implemented logic already uses a 10 second buffer. 

If the main worker(instance) is say 20 seconds behind than the scan will take place roughly 20 seconds as well (slightly off because the logic takes place in two separate places but due to the way the main worker behind check works it was unavoidable without larger changes). With that said the worker status time behind is also different from the actual logic for checking if the scan is too late, so even though my implementation is capable of being slightly off, the current implementation already has that same issue it just isn't visualized.

I agree it isn't perfect but it is still much easier to determine if your workers are keeping up with the amount of spawns it has access to. If you think I should reword the actual message to something along the lines of "Scan took place X seconds after the pokemon has spawned.", it would be more accurate to what is being displayed but also more verbose. Please let me know what you think, thanks.

PR #1234, opened 2016-09-09
Support loading proxies from file

Badiboy:
Allows defining proxyfile with the list of proxies (cli or config).
Disable proxy check result output if more the 10 proxies configured.
Move all proxy checking from 'runserver.py' to 'proxy.py'.


Badiboy:
New version of changes pulled.

PR #1255, opened 2016-09-10
Add class method .getsize() to schedule

springjools:
## Description
Add a method called .getsize() to the BaseScheduler class, which allows the user to easily retrieve the number of total points that is to be looped through in each pass. This is useful for status printing and future proofing.

## How Has This Been Tested?
Tested on local machine. Is a 3 line change anyway.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -57,6 +57,7 @@ class BaseScheduler(object):
         self.status = status
         self.args = args
         self.scan_location = False
+        self.size = 0
 
     # schedule function fills the queues with data
     def schedule(self):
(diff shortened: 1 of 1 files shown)


justsomedudeonthenet:
Love it!

Two tiny nitpicks:

- Forgot to add it to HexSearch.  In fact, if you add it to HexSearch's schedule function, you don't need it in HexSearchSpawnpoint at all.
- Initial value should be None, so you can tell if the location list hasn't been generated yet.

justsomedudeonthenet:
Looks good to me.

PR #1259, opened 2016-09-10
Introducing single-user: search mode (aka Scheduler) and step limit via frontend + speed limit

Mr-Groch:
## Description
Ability to change search mode (scheduler) and step limit in runtime via web page. This is mainly inherited from PR #737 and #37 - I have not closed that PR's, because mybe contributors decide to use those independent. Difference is that visibility of search mode select box and step limit input field depends here on -su param.

Also added speed limit to avoid teleports on spawnpoint only search and location change

## Motivation and Context

The same as PR #737 and #37

## How Has This Been Tested?

The same as PR #737 and #37 

search_mode

img_20160910_213354

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -22,6 +22,8 @@
 #gym-info:              # enables detailed gym info collection (default false)
 #min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
 #status-name:           # enables writing status updates to the database - if you use multiple processes, each needs a unique value
+#speed-limit:           # if next scan would cause the account to move above specified speed in kmph in a straight line, sleep until such time that it could reasonably move that distance
+#max-speed-limit-sleep: # maximum time in seconds to sleep when trying to stay under the speed limit
 
 # Misc
 #gmaps-key:             # your Google Maps API key
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -24,6 +24,7 @@
 #status-name:           # enables writing status updates to the database - if you use multiple processes, each needs a unique value
 #speed-limit:           # if next scan would cause the account to move above specified speed in kmph in a straight line, sleep until such time that it could reasonably move that distance
 #max-speed-limit-sleep: # maximum time in seconds to sleep when trying to stay under the speed limit
+#single-user-control:   # enables controlling search params for individual use
 
 # Misc
 #gmaps-key:             # your Google Maps API key
--- a/pogom/__init__.py
+++ b/pogom/__init__.py
@@ -6,5 +6,6 @@ config = {
     'LOCALES_DIR': 'static/dist/locales',
     'ROOT_PATH': '',
     'DATA_DIR': 'static/dist/data',
-    'GMAPS_KEY': None
+    'GMAPS_KEY': None,
+    'SCHEDULER': ''
 }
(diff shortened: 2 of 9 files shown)


aldarund:
But that would mean that any use who use the map would be able to change it. Not really good

Mr-Groch:
@aldarund It can be disabled. If someone use fixed location - it will not show. Also adding -nsm parameter will disable just this select box

aldarund:
Why its enabled by default then and need switch to be disabled instead of switch to be enabled?

4RK4N:
u could rework this to integrate into the alrdy password protecred /status page. and with some more PR's like this one it could be a fully featured admin frontend :D

Mr-Groch:
@aldarund For the same reason search (pause) switch is enabled for default. I think that main purporse of this map is individual use, but if contributors wants to disable this by default, I will make it disabled. Waiting for more feedback :)

jaquer:
I'm really loving your work @Mr-Groch, but @aldarund has a good point. I would switch the logic around and enable the feature only if `search-mode-control` argument is true. I know I'm not a contributor, but this is a core functionality change.

SlasherKG:
This and the step control in the web gui will be awesome! 
It's a pain to have to RDP to my PC every time I need to change steps and type of scan.

I hope these get merged!

Mr-Groch:
OK, I have disabled it by default. Added -su arg (--single-user-control) to enable search mode select box and step limit input field. I wish to use that arg also for current other search controlls like search pause, search location following user - what do you think?

Spiritvn:
I think you should move this to the Search Mode and Search Step Limit to Status Page which requires password, it should be for backend than frontend. 

I would like to have the speed limit merge, combine its with on demand mode here: https://github.com/PokemonGoMap/PokemonGo-Map/pull/1271 i can have some seperated workers to scan follow my friends location and lower down the BAN chances

Mr-Groch:
@Spiritvn It is now not enabled by the default. You must specify -su paramter to enable this. I want also cover search pause switch and search follows user location switch with -su parameter, but need more feedback about that.

Spiritvn:
I know it's disabled by default, but i agreed with @4RK4N , if you enable it, it's still on the place where other users could change it and affect to the others. The map page should be related to "FrontEnd" than that's backend function, status page is more for backend so it's better structure, it's my personal opinion though.

Really love to have speed limit goes with scan follow me :)

Mr-Groch:
Param full name is --singe-user-control (su like superuser ;) ) and my idea is that only users who want to have this kind of control will enable it. I'm sure there is many people that use this individually. And that why I want also to pull search pause control and search follow user into this new mode. This PR is not to add admin controls behind password protected control panel - just to make lot of users life easier :) 

Admin backend is also good idea, but I would like to see it as seperate PR.

Mr-Groch:
Like the commit message says, I have removed -nsc arg, make search control part of single-user mode (-su). One parameter less ;) After some thinking - I will leave location changing as is - fixed location enabled/disabled is better place to handle that.

primaxius:
Not everything needs to be in the frontend.

Mr-Groch:
@primaxius This is optional feature - only those ho wants this will enable it, so why closed?

CalcMan:
Reopen please, these features are very useful for single user maps, and they are designed to be on only when requested, why not let folks use what they want?

SenSachen:
I too would enable this optional feature if it were implemented.

Issue #1034, opened 2016-08-27
db-cleaner overkill


Stupefy:
## Expected Behavior
would expect the database to be cleaned once by one thread

## Current Behavior
database cleanup happens at almost the same time for every thread.  this causes issues with 100+ worker threads and accessing the database via a single web server.

## Possible Solution
add option to disable database cleanup for workers.

## Steps to Reproduce (for bugs)

1. run 100+ workers.. 
2. log in.. and look at the map..
3.
4.

## Context

i get the following very frequently with a web service running
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\flask\app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Python27\lib\site-packages\flask\app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Python27\lib\site-packages\flask\app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Python27\lib\site-packages\flask\app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Python27\lib\site-packages\flask\app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "C:\PokemonGoMap\Build\PokemonGo-Map\pogom\app.py", line 90, in raw_data
    d['pokemons'] = Pokemon.get_active(swLat, swLng, neLat, neLng)
  File "C:\PokemonGoMap\Build\PokemonGo-Map\pogom\models.py", line 107, in get_active
    for p in query:
  File "C:\Python27\lib\site-packages\peewee.py", line 2967, in __iter__
    return iter(self.execute())
  File "C:\Python27\lib\site-packages\peewee.py", line 2960, in execute
    self._qr = ResultWrapper(model_class, self._execute(), query_meta)
  File "C:\Python27\lib\site-packages\peewee.py", line 2656, in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "C:\Python27\lib\site-packages\peewee.py", line 3492, in execute_sql
    self.commit()
  File "C:\Python27\lib\site-packages\peewee.py", line 3316, in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "C:\Python27\lib\site-packages\peewee.py", line 3485, in execute_sql
    cursor.execute(sql, params or ())
OperationalError: database is locked
[...]

PR #1260, opened 2016-09-10
disable clean db option

aldarund:
With beehive every worker run a clean db loop in a separate thread. It increase opened connections to db  twice. Also it cause a lot of deadlock issues when multiple workers run multiple db clean loops. And its not really needed. One db clean loop for one database more than enough.
So to disable it i added option --disable-clean.

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -156,6 +156,8 @@ def get_args():
                         nargs='*', default=False, dest='webhooks')
     parser.add_argument('-gi', '--gym-info', help='Get all details about gyms (causes an additional API hit for every gym)',
                         action='store_true', default=False)
+    parser.add_argument('--disable-clean', help='Disable clean db loop',
+                        action='store_true', default=False)
     parser.add_argument('--webhook-updates-only', help='Only send updates (pokémon & lured pokéstops)',
                         action='store_true', default=False)
     parser.add_argument('--wh-threads', help='Number of webhook threads; increase if the webhook queue falls behind',
--- a/runserver.py
+++ b/runserver.py
@@ -219,9 +219,10 @@ def main():
         t.start()
 
     # db clearner; really only need one ever
-    t = Thread(target=clean_db_loop, name='db-cleaner', args=(args,))
-    t.daemon = True
-    t.start()
+    if not args.disable_clean:
+        t = Thread(target=clean_db_loop, name='db-cleaner', args=(args,))
+        t.daemon = True
+        t.start()
 
     # WH Updates
     wh_updates_queue = Queue()


gabrielsoldani:
+1. This fixes issue #1034.

antiremy:
If we only need it for one thread, why not make it an --enable-clean argument? That way you only need to add it to one command versus adding --disable-clean to a bunch

kbtbc:
Perhaps enable db clean by default, esp with the -os flag, since in beehive it usually has no workers.  Any -ns threads, disabled by default.  (Do you even need an extra flag?)

gabrielsoldani:
You do, because you can run more than one --only-server instance for load
balancing, and you'd still want only one of them to do the clean up
procedures.

In my opinion, a disable flag is the best option right now. If we ever
split the web server from the search we could more intelligently handle
this. In the best case scenario the webserver would only ever need SELECT
permissions.

aldarund:
I will change pr to following.
1) In instances with -ns clean will be disabled.
2) In instances where server is run it will be enabled by default and --disable-clean will disable it.
Any objections?

gabrielsoldani:
I disagree. People might not want to run the web server (for data
gathering, web hooks, different front-end, etc).

Make it the other way around (--only-server never cleans up, otherwise
clean unless --disable-clean is set). The web server should not handle
clean up, this is not a task for the front-end.

aldarund:
The other way around will mean adding this flag to every search worker, which is very verbose.

aldarund:
Usual setup is when  there 1 server with multiple workers, or a serve with a worker. When there no server at all is quite uncommon.

gabrielsoldani:
Uncommon but possible (in fact it's my setup, since I use my own front-end).

This change (where --no-server has no clean up by default with no way to
enable) would make me and other users run a web server just for clean up,
which makes no sense.

Let me rephrase that: **having our web servers, which are supposed to serve
web pages and data, to clean up after our workers, is a bad design
decision. Having them being the _only_ component which can clean up after
our workers is a _terrible_ design decision.**

Right now we can have many web servers (including zero) and many searchers
(including zero). Just keep the commit exactly like it is right now
(everything runs the clean up unless --disable-clean is set) and it's an
upgrade. If we require a webserver to do clean ups, it's a downgrade.

aldarund:
Well, then best option would be to have cleanup disabled by default, and need to be enabled with flag where its needed.

justsomedudeonthenet:
My opinion:

- Disabled by default is a **bad** idea.  People won't enable it.  Then they will run into issues.
- Enabled only for -os is also kind of bad.  I'd rather the webserver instance focus on web serving.  Some people run multiple -os, which defeats the purpose of this.  Some people run none and just use webhooks.

My vote would be to have it enabled by default on all instances, with the switch to disable it.  It doesn't break any existing behaviour, and people who know what they are doing can make their own decisions on which instances should run the cleanup.

So +1 merge as is please!

Issue #1232, opened 2016-09-09
accounts go into cooldown because of database lock exception


marnikvde:
**Description**
I understand that there are some reasons to put an account in cooldown mode, but there is also a "general exception" reason to put accounts in cooldown, which is probably not needed. All my accounts go in cooldown because the database seems to be locked.

Some info: I'm running this in docker on a raspberry pi, and sqlite is probably not the best approach, but it looks like a general issue.

**Possible Solution**
Be more granular on the kind of exception, and exclude this database lock of the cooldown

**Your Environment**
Version: Not sure and it's not in the log, I did a git clone on sept 7 or 8.
Environment: RaspBerryPi, running inside a docker container

logs:

```
2016-09-09 00:27:14,181 [search-worker-16][        search][    INFO] Waiting to get new account from the queue
2016-09-09 00:27:14,242 [    db-updater-0][        models][ WARNING] DB queue is > 50 (@52); try increasing --db-threads
2016-09-09 00:27:14,273 [ search-worker-9][        search][   ERROR] Exception in search_worker under account <edited> Exception message: database is locked
2016-09-09 00:27:14,274 [ search-worker-9][        search][    INFO] Waiting to get new account from the queue
2016-09-09 00:27:14,275 [ search-worker-0][        search][   ERROR] Exception in search_worker under account <edited> Exception message: database is locked
2016-09-09 00:27:14,276 [ search-worker-0][        search][    INFO] Waiting to get new account from the queue
2016-09-09 00:27:14,298 [    db-updater-0][        models][ WARNING] DB queue is > 50 (@51); try increasing --db-threads
```

```
2016-09-09 00:27:37,491 [account-recycler][        search][    INFO] Account recycler running. Checking status of 11 accounts
2016-09-09 00:27:37,495 [account-recycler][        search][    INFO] Account <edited> needs to cool off for 7177 seconds due to exception
2016-09-09 00:27:37,525 [account-recycler][        search][    INFO] Account <edited> needs to cool off for 7177 seconds due to exception
```

lastrolo:
I have the same issue. Accounts going in to cool down because of DB locks

Shentang:
Same here. It's kind of a blocker, unable to use the map at all.

lastrolo:
[...]

PR #1267, opened 2016-09-11
Account database lock exception fixed

Diblo:
## Description

This fix resolves the problem with SQLite and multi threading which put the accounts in cool down mode due to database lock.
## Motivation and Context

The problem arose when the function search_worker_thread could not read from the database because of the writing process. The solution was using "Write-Ahead Logging", so that the read and write process do not influence each other  - https://www.sqlite.org/wal.html
This fix also resolves the issues #1232
## How Has This Been Tested?

This has been tested with 77 workers.
`runserver.py -ns -st 100 -ss --jitter -w 77 -asi 3900 -ari 3500 -msl 875 -ac usernames.csv -l "55.402360, 10.386895" -k [gmaps-key]`
The accounts stopped going into cool down due to database lock.
## Types of changes
## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -60,7 +60,12 @@ def init_database(app):
             stale_timeout=300)
     else:
         log.info('Connecting to local SQLite database')
-        db = SqliteExtDatabase(args.db, pragmas=(('journal_mode', 'WAL'), ('mmap_size', 1024 * 1024 * 32), ('cache_size', 10000), ('journal_size_limit', 1024 * 1024 * 4),))
+        db = SqliteExtDatabase(args.db,
+                               pragmas=(
+                                   ('journal_mode', 'WAL'),
+                                   ('mmap_size', 1024 * 1024 * 32),
+                                   ('cache_size', 10000),
+                                   ('journal_size_limit', 1024 * 1024 * 4),))
 
     app.config['DATABASE'] = db
     flaskDb.init_app(app)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -10,7 +10,7 @@ import gc
 import time
 import geopy
 import math
-from peewee import SqliteDatabase, InsertQuery, \
+from peewee import InsertQuery, \
     Check, CompositeKey, ForeignKeyField, \
     IntegerField, CharField, DoubleField, BooleanField, \
     DateTimeField, fn, DeleteQuery, FloatField, SQL, TextField, JOIN
@@ -18,6 +18,7 @@ from playhouse.flask_utils import FlaskDB
 from playhouse.pool import PooledMySQLDatabase
 from playhouse.shortcuts import RetryOperationalError
 from playhouse.migrate import migrate, MySQLMigrator, SqliteMigrator
+from playhouse.sqlite_ext import SqliteExtDatabase
 from datetime import datetime, timedelta
 from base64 import b64encode
 from cachetools import TTLCache
@@ -58,8 +59,8 @@ def init_database(app):
             max_connections=connections,
             stale_timeout=300)
     else:
-        log.info('Connecting to local SQLite database...')
-        db = SqliteDatabase(args.db)
+        log.info('Connecting to local SQLite database')
+        db = SqliteExtDatabase(args.db, journal_mode='WAL')
 
     app.config['DATABASE'] = db
     flaskDb.init_app(app)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -60,7 +60,7 @@ def init_database(app):
             stale_timeout=300)
     else:
         log.info('Connecting to local SQLite database')
-        db = SqliteExtDatabase(args.db, journal_mode='WAL')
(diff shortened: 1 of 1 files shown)


Stereolab42:
Outstanding. This actually makes the scanner usable for me at large scale, I was about to give up.

Dunkr23:
Great! Thank you for this fix!

lastrolo:
I tried this commit and it definitely helped. My stuff ran for longer, without question. I did still find accounts in cool down though. I think besides just improving the concurrency of SQL lite, actually stopping a DB lock issue from putting an account in cool down is probably helpful too??

Stereolab42:
One issue I've found with this is that a presumably temporary file called "pogom.db-wal" grows enormous over time... I've seen it up to 3.6GB. The next time I run the scanner it seems to hang for a long time processing this file. I setup my script to delete this file prior to running the scanner without issue, but I'm not sure if that's a good thing.

Diblo:
@Stereolab42 Try again - I'm not sure this will solve your problem, but this fix reduce the size of pogom.db-wal.

Stereolab42:
Why was this closed?

Diblo:
Sry. I had to many windows open and closed the wrong one :(

aldarund:
>Outstanding. This actually makes the scanner usable for me at large scale, I was about to give up.

Large scale with sqllite? really? Its like make zero sense.

Stereolab42:
Sorry to disappoint, I know absolutely zero about databases. Are you saying with MariaDB (per https://downloads.mariadb.org/) there will be no crippling locking problems at scale (10+ accounts), obviating the need for this patch?

aldarund:
@Stereolab42 yes. sqllite not supposed to be used by multiple writers. its more a testing database rather than a real one. And yes you wont have lock problem with maria\mysql.  There occasionally can be deadlocks, but it faaaaaaaar less frequent and will happen only with really large number of accounts like 200+.

Diblo:
@Stereolab42 what @aldarund say is that you should setup/ host a MySQL server or rent one. However, I have troubles with to many connections to my rent MySQL server. So the solution is to host your own. I'm sure on that @aldarund have the time to explain how you it if you don't have the time or skills. I do not have the time to help you.

simonbuehler:
fix works great for me, sqlite is actually faster for me spawnpoint scanning 2200 points with 30 accounts than a local mysql server ...?

jccwright:
@Sterolab42 ...at scale with 10 accounts 😏... I run with anywhere between
75 - 1000 at a time... now THAT's scale...

simonbuehler:
please merge this, map works good only with this PR

simonbuehler:
@FrostTheFox please merge

Artifice:
@Diblo Can you update this to resolve the conflicts? I'd like to push to get this merged!

Diblo:
Threading module can run up to 886 threads simultaneously (_Easy 600 accounts_). When running Pokemon Go Map with 600 accounts (plus various threads) it is much more than SQLite can handle without WAL. Pokemon Go Map is therefor unable to use its full potential with SQLite .

**Eg.** We are 7 friends who are using Pokemon Go Map. We are covering Odense city (3rd largest city in Denmark) and the surrounding area. The cover area is equivalent to 100 steps (461.81 kilometers^2). We are running 3 servers on 1 computer. A spawn scanner with 500 accounts, a hex scanner with 350 accounts and a web server. So @smc87 it's easy to imagine running Pokemon Go map with just SQLite. - We got into problems with just 250 accounts when we tested the the possibilities of the program.

**This patch is only 3 lines and is about getting the full potential out of Pokemon Go Map.  As well making the program more user-friendly.**

jde1126:
@Diblo Can you explain how you have the map set up? Do you have encounter Pokemon on? I want to have a City-wide map for Charlotte, but if I have encounter on, it keeps IP banning....

sebastienvercammen:
We've had a long discussion about this between devs on Discord and we've decided to merge this PR, but not without some necessary comments.

1. sqlite should never be used with RocketMap in _any_ form of "production environment", at most it should be used for (small scale) testing purposes only.
2. RocketMap is built on multithreading (and running multiple instances), i.e. multiple writers: sqlite does **not** support multiple writers.
3. Going forward, we won't change code specifically to keep supporting sqlite. Performance and best practices are more important than trying to support a storage format that isn't even intended for this kind of project.
4. With the current codebase, support for sqlite is already non-existant: any regular-sized setup will break compatibility with sqlite as we bundle SQL requests to reduce the query amount, which sqlite doesn't support because we're already above the maximum supported host parameter count (sqlite's 999 vs MySQL's 65535).
5. And lastly, considering the above points, we won't go out of our way to help people run the map (aka not for a test) with sqlite. That includes our Discord: we're not customer support and we're intentionally telling you now not to use sqlite for the map (besides testing).

Sqlite says it best:

> SQLite is often used as a surrogate for an enterprise RDBMS for demonstration purposes or for testing.

Issue #1129, opened 2016-09-04
[Feature Request] Search based on country


nicktan147:
I have been sharing my tools for a month now, since some of are not my  are out of my country, they used to scan their area for pokemon, and ended up my accounts running the map some are being banned. Can we create a function to limit the search function to limit the search result?

I searched through google, and found this,
http://stackoverflow.com/questions/8282026/how-to-limit-google-autocomplete-results-to-city-and-country-only

I tried to implement it doesn't work. Not sure if I put on the wrong js file, could anyone enlighten me?

jaquer:
I don't know how to make it user-configurable, but here's how to make the change yourself:

Open `static/js/map.js` in an editor and search for `searchBox`. You should find this line:

`var searchBox = new google.maps.places.SearchBox(document.getElementById('next-location'))`

Right after it, add a line like this:

`var autocomplete = new google.maps.places.Autocomplete(document.getElementById('next-location'), {componentRestrictions: {country: 'us'}})`

Replace `us` with your country's [ISO 3166-1 alpha 2 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)

sebastienvercammen:
Map is not built to be a platform to scan every visitor's location. Closed.

PR #1270, opened 2016-09-11
Change searchbox into Autocomplete object.

jaquer:
## Description

Changed the "Change Search Location" into Autocomplete object instead of Searchbox.
## Motivation and Context

The Autocomplete object has several benefits: the results are limited to actual places, not including category suggestions, the results can be limited to certain countries, coordinate boundaries, place types, etc.

Autocomplete fields are more flexible than Searchbox fields: [Google documentation](http://goo.gl/pxqxK7)

This is the groundwork to meet feature request #1129, and add any other kind of boundary restriction.
## How Has This Been Tested?

I've tested this searching for exact addresses and business names. The functionality does change slightly, I believe for the better. See screenshot summary below.

**Before:**
Searching for broad terms/names lists results with a magnifying icon. These are Google places suggestions.

Choosing the first result changes the location to the first business that matches that description, according to Google. Notice this business name was not shown to the user beforehand.

**After**
Searching for the same broad term now limits the results to actual places:

## Types of changes

- [x] All of the above, depending on your point of view
## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -276,6 +276,9 @@ function initSidebar () {
 
   searchBox.addListener('place_changed', function () {
     var place = searchBox.getPlace()
+
+    if (!place.geometry) return
+    
     var loc = place.geometry.location
     changeLocation(loc.lat(), loc.lng())
   })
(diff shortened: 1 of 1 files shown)


static/js/map.js
```diff
@@ -268,20 +268,15 @@ function initSidebar () {
...
-    var places = searchBox.getPlaces()
-
-    if (places.length === 0) {
-      return
-    }
-
-    var loc = places[0].geometry.location
+  searchBox.addListener('place_changed', function () {
+    var place = searchBox.getPlace()
+    var loc = place.geometry.location
```
beccasafan: needs a check that `place.geometry` exists. ie, type gibberish into the box and hit enter.

FrostTheFox:
Tested, OK

Issue #1142, opened 2016-09-04
Scanner keeps searching


KyleBoyer:
The map scan status fetching occurs every 5 seconds, yet even when nothing is fetching the map, it keeps scanned. Essentially, the map scan status fetching is a heartbeat. This means that after a period of time, with no heartbeat, the scanner should auto-stop the search to preserve the accounts from being banned. 

## Expected Behavior
1. (Start scanner python script["runserver.py"])
2. Open map web UI, scanner begins scanning.
3. Close map web UI, scanner stops scanning.
Repeat 2-3

## Current Behavior
The map continues scanning regardless of the heartbeat.

## Possible Solution
Call `post_search_control("off")` after a heartbeat has been absent for X time(60 seconds)

## Steps to Reproduce (for bugs)
1. Start scanner python script("runserver.py")
2. Open Web UI
3. Close Web UI

## Context
I have a shared scanner running, and only a few people use it. Thus, the scanner application is running 24/7. The accounts used for this scanner are banned quickly, after one night of leaving the search on.

## Your Environment
* Version used: Latest as of Sunday, September 4th, 2016 at 6PM CST
* Environment name and version (e.g. Python 2.7): Python 2.7.12
* Operating System and version (desktop or mobile): Ubuntu 16.04 - Desktop

justsomedudeonthenet:
This is not desired behaviour for many of us.  I want it tracking every single pokemon that spawns in my city so I have it for statistical analysis.

LordNeoZ:
You're not getting the accounts banned due to running 24/7 but for being teleported by 2 or more people moving the marker at the same/adjacent time. Most of us run the scanner 24/7 to get statistics/heatmaps without getting banned. You should consider to split your workers and configure 1 scanner per person or add workers and extend the area (but with fixed location)

KyleBoyer:
Prior to having it be a shared map, I still had this issue. However, it appears to be a new type of ban - accounts can still login, and fetch pokestop and gym locations, but no Pokémon. I switch accounts used, and it actually picks up on the Pokémon. I had this happen when running on my local machine(not shared), overnight. I understand some people would not like this behavior, but all that needs to be done is add a --heartbeat-timeout(or -hb) argument(0 being no timeout).

LordNeoZ:
That could be a soft ip ban, try to use a proxy because your IP was probably already flagged.

DiscoTim:
[...]

Issue #1132, opened 2016-09-04
[Feature Request] Scan only when client connects


cbogithub:
http://feathub.com/PokemonGoMap/PokemonGo-Map/+30

It would be good to host a scanner that is on standby 24/7 without being ip banned. If the scanner only scans when a client connects this may be possible.

DiscoTim:
I've been thinking about doing a PR for this, and it would be very useful for me and a lot of others running small maps for few a people. 

I can't envision all use cases. but for this simple scenario, i was thinking that it basically toggles the scan switch when a user connects and keeps it on for 30 minutes or hour or so past the last request.

of course the drawback is that when a user connects they'd need to wait for results, but that is a given. it just forgoes the need to have to start and stop the server when its not being used.

KyleBoyer:
> I've been thinking about doing a PR for this, and it would be very useful for me and a lot of others running small maps for few a people.
> 
> I can't envision all use cases. but for this simple scenario, i was thinking that it basically toggles the scan switch when a user connects and keeps it on for 30 minutes or hour or so past the last request.
> 
> of course the drawback is that when a user connects they'd need to wait for results, but that is a given. it just forgoes the need to have to start and stop the server when its not being used.
> 
> obviously it would be triggered by a command option and not be the default behavior.

Do you have a status on this?

KyleBoyer:
This issue can be closed, as it was added with PR #1271

FrostTheFox:
Feature Requests are no longer are taken via GitHub. Please submit FRs using http://vote.devkat.org

PR #1271, opened 2016-09-11
Fully Implement On Demand Mode(Conflicts resolved)

KyleBoyer:
## Description
I made changes to a few files, to track when the last web request was(heartbeat). This then has a timeout that will stop the bot from scanning if there is no activity in a certain period of time, configured by "-odt" or "--on-demand_timeout". Thus, "--on-demand_timeout 900" will activate this mode, and set the timeout to 15 minutes. One thing to note, is that the timeout is in seconds. Finally, the scanner will also startup in standby mode(waiting to start) until the first client connects.

## Motivation and Context
Solves https://github.com/PokemonGoMap/PokemonGo-Map/issues/1142 and https://github.com/PokemonGoMap/PokemonGo-Map/issues/1132

## How Has This Been Tested?
Tested locally and on my server, by connecting and disconnecting, turning off and on on demand

N/A
## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -15,7 +15,7 @@ from collections import OrderedDict
 
 from . import config
 from .models import Pokemon, Gym, Pokestop, ScannedLocation, MainWorker, WorkerStatus
-
+from .utils import now
 log = logging.getLogger(__name__)
 compress = Compress()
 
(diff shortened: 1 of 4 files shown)


mew905:
great for amazon ECS servers, reducing the load while not being used is handy to earn back those CPU credits. Would this work with a remote -os instance and local workers though?

KyleBoyer:
Where is the time in service check occur? I'm looking now, by if you know off hand it may help.

k4n30:
@KyleBoyer this doesn't work on -ns workers (those without a server). Any chance of updating to include those which don't have a web server?

PR #1283, opened 2016-09-12
auth_service for google acct should be 'google' not 'gmail'

taihun-at-somewhere:
Recent commit have made choose between 'ptc' and 'gmail' for account csv, but the auth service really should be 'google' according to the way pgoapi is handled.


FrostTheFox:
oh wow lol

PR #1287, opened 2016-09-13
Keep map updating while background on mobile

DiscoTim:
This PR ensures map updates will still occur when a browser is minimized/background on mobile, by implementing a webworker which calls some of the updating functions.

Generally interval timers are suspended/slowed when a browser is background on mobile, which prevents the map from updating and also prevents notifications from occurring.

## How Has This Been Tested?

Tested on Win ff/chrome/edge. Android Chrome/FF. Should work on IOS Safari, as it supports the necessary functions, but needs testing.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -32,9 +32,10 @@ var searchMarker
 var storeZoom = true
 var scanPath
 
-
 var selectedStyle = 'light'
 
+var updateWorker
+var lastUpdateTime
 
 var gymTypes = ['Uncontested', 'Mystic', 'Valor', 'Instinct']
 var audio = new Audio('static/sounds/ding.mp3')
@@ -1017,6 +1018,7 @@ function updateMap () {
     if ($('#stats').hasClass('visible')) {
       countMarkers()
     }
+    lastUpdateTime = Date.now()
   })
 }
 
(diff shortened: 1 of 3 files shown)


slideup-benni:
I like your idea, but not everybody has enough mobile bandwidth. This would make the map for them unusable. Is it possible, that you could make it optional in the options-bar?

DiscoTim:
@slideup-benni your right, it should be optional, and i think also the option only appear when on mobile.

bluelatios:
How about this
When you choose to receive notification for certain pokemon, it will send you an email with google maps link in it
Then you just click the link, and maps app will show you the mark
Is it even possible?
I think it works great when you use cloud vps, and scanning certain area non stop with -ss command

Darkknightz:
@bluelatios That iswhat https://github.com/kvangent/PokeAlarm and webhooks do.

bluelatios:
@Darkknightz  wow, feels like i've been living under a rock. Thanks! :D

DiscoTim:
@bluelatios that can certainly be done, as @Darkknightz pointed out but its out of scope if this PR.

This pr is simply to keep the app updating when its minimized on mobile.

bluelatios:
@DiscoTim sorry just one noob question, how do you run all this script on mobile? Android to be precise

DiscoTim:
@slideup-benni I've put some thought in to this, and i am wondering if it would be useful to use a combo box with selections for updates as automatic/background/none?

I can tell you from testing for example, that the app on its own, before this script takes over, will continue to update on mobile for 5 to 10 min after its background.

Also i am considering where this option would go, whether it be just be a simple toggle or combo with the options above. ideally it would go under a 'General settings' or 'Mobile Settings' tab but i am not going to introduce that for one option.

So maybe under 'Location and Search Settings' as 'Background update'?

DiscoTim:
@bluelatios might be a little tricky if you are not familiar with git, but basically in a nutshell you would fork the main project and merge this pr in to it.

LordNeoZ:
do you still need an SSL cert on the site to make notifications work right?

DiscoTim:
@LordNeoZ for chrome yes, its a requirement of chrome and the same with geolocation, that won't change. but SSL is free from letsencrypt, its just a matter of setting it up.

LordNeoZ:
@DiscoTim, yup, i know, was just wondering if setting a background webworker would change anything on that.
I really hope this gets merged. I used an apk to keep the webpage alive, but this is a far better aproach.

jchristi:
@DiscoTim unrelated to this PR, but curious if you if websockets also updates in the background on mobile. I feel like websockets would be the most optimal way of using the least amount of data, but if it doesnt update in the background then maybe service worker is the best compromise.

DiscoTim:
@jchristi this uses a webworker as opposed to a serviceworker, and you can even use a websocket in a webworker. the polling method the app uses now is terribly inefficient, so pretty much anything would be an improvement. afaik websocket would be best, with the client only requesting on a change, and the server only pushing changes, but would require a significant amount of front and backend work, and also not sure where things like webhooks stand in that.

jchristi:
@DiscoTim thanks for clarifying its a webworker, I misread initial description. I only recently learned about serviceworkers, so probably starry-eyed. Anyways, I like the idea of this having the potential to be combined with websocket :+1:

DiscoTim:
@jchristi yes they do get confusing :) serviceworkers are mostly useful for notifications, so they are still needed, as webworkers don't do that.

this is the 'type' of thing that needs to be combined with websockets https://github.com/PokemonGoMap/PokemonGo-Map/pull/1303 (once its finished, providing it ends up doing gyms/stops/pokemon). however even that just using the current polling method would drastically reduce data transferred.

static/js/map.js
```diff
@@ -1220,6 +1222,62 @@ function i8ln (word) {
...
+          Store.set('followMyLocationPosition', { lat: lat, lng: lng })
+        }
+      }
+    })
+  }
+}
+
+function createUpdateWorker () {
+  try {
+    if (isMobileDevice() && (window.Worker)) {
```
jchristi: Why do this only for mobile devices? any reason not to do this for desktop browsers too?

DiscoTim: for just this use of webworker, which is keeping the map updated, it isn't absolutely necessary. i think most desktop browsers don't suspend interval timers when a window isn't in the foreground. On desktop i think it would be extra overhead, for no reason.

PR #997, opened 2016-08-26
Add random delay, basic speed limit, randomize order of accounts

AngryRabbit1:
## Description

Added optional random delay (--random-delay) to command line so user can specify number of random seconds between 0 and argument to be added to scan delay each time the search worker thread sleeps in an attempt to avoid account bans.

Added optional rudimentary speed limit (--speed-limit) to command line options. Sleeps longer if next scan would go above the specified speed limit in a straight line.

Randomized the order of accounts at startup so they don't always log in in the same sequence.

Updated version of [https://github.com/PokemonGoMap/PokemonGo-Map/pull/870](url) with a newer version of the main develop branch and added additional features not present in 870.

[Edit: New Additions]
Added --max-speed-limit-sleep to prevent sleeping too long due to speed limit.
Added --max-accounts and --account-use-duration to facilitate swapping out of accounts every so often.

[Edit 2]
Rebased and removed -max-accounts and --account-use-duration because it overlaps functionality with PR 996.
## Motivation and Context

Accounts have been getting banned the past few weeks creating a need to constantly make more accounts. I have tried a few things (e.g. longer scan delays, lower number of workers, etc) to no avail, this is the first solution I have found that seems to remedy the situation. This PR also implements a few other measures to avoid bans.
## How Has This Been Tested?

On my own instance. @TackyBrad has tested the random delay as well.
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1301, opened 2016-09-14
Add speed limit

averylarry:
## Description

Shamelessly modified code from #1259, #997, #787, #737, maybe others.
Unlike other PRs, I am adding a speed limit and nothing else -- hoping maybe it will have a better chance at getting merged.

Updated config.ini.example to list the 2 new arguments.

Updated utils.py to add the 2 new arguments and verify their compliance
-sl  --speed-limit   Speed limit in kilometers per hour (would be trivial to change to meters per second if preferred).  Default = 0 (effectively disabled).  Could be -sp instead to confuse less with -st ??

-msld  --max-speed-limit-delay  Specifies in seconds the maximum delay allowed due to speed limit. Default = 0 (effectively disabled).  I'm open to naming suggestions -- it seems long

Updated search.py
Added check_speed_limit to calculate the necessary delay, if any, to keep under the speed limit.
Combined existing too late? logic with new too fast? (speed limit) logic since it doesn't make sense to delay for a speed limit only to find that you "arrived" too late to bother (for spawnpoint scanning).
## Motivation and Context

Teleporting too fast leads to account bans.  Using a speed limit minimizes too-fast teleportation and should reduce account bans.

## How Has This Been Tested?

Temporarily added logging of all the various calculations and manually verified proper behavior.  Ran extensively in -ss and beehive modes.  Moved the marker around.  Changed to ridiculous values to test edge cases.
Using somewhat normal settings (maybe a little slow) of 5 accounts 250 spawnpoints -ss -st 12 -sd 0 -sl 65 -msld 120 -- I would say under 10% of the time the speed limit is used, and it usually delays less than 30s.
win7 machine.  SQLite.

beehive scanning using -st 2 -sd 0 -sl 10 -msld 180 for illustration.  Includes a manual location change.

-ss scanning using -st 12 -sd 0 -sl 40 -msld 120

-ss scanning using -st 12 -sd 0 -sl 10 -msld 180 -msl 780 (ridiculous settings) to somewhat illustrate how the too-slow and too-fast work together

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change) I think?

## Checklist:

- [X] My code follows the code style of this project.  From what I can tell.
- [X] My change requires a change to the documentation.  Well, if config.ini.example counts.
- [X] I have updated the documentation accordingly.  Well, if config.ini.example counts.

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -472,14 +472,32 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
                         continue
 
                 # too late?
-                if leaves and now() > (leaves - args.min_seconds_left):
+                extra_delay = check_speed_limit(args, status['location'], step_location, status['last_scan_time'])
+                continue_time = time.time() + extra_delay
+
+                if leaves and continue_time > (leaves - args.min_seconds_left):
                     search_items_queue.task_done()
                     status['skip'] += 1
                     # it is slightly silly to put this in status['message'] since it'll be overwritten very shortly after. Oh well.
-                    status['message'] = 'Too late for location {:6f},{:6f}; skipping'.format(step_location[0], step_location[1])
+                    if time.time() > (leaves - args.min_seconds_left):
+                        status['message'] = 'Too late for location {:6f},{:6f}; skipping'.format(step_location[0], step_location[1])
+                    else:
+                        status['message'] = 'Skipping {:6f},{:6f}; outside time and speed constraints'.format(step_location[0], step_location[1])
                     log.info(status['message'])
                     # No sleep here; we've not done anything worth sleeping for. Plus we clearly need to catch up!
                     continue
+
+                # too fast?
+                if extra_delay:
+                    status['message'] = 'Too fast for {:6f},{:6f}; waiting {}s...'.format(step_location[0], step_location[1], extra_delay)
+                    log.info(status['message'])
+                    continue_time = time.time() + extra_delay
+
+                    while time.time() < continue_time:
+                        time.sleep(1)
+                        remain = int(continue_time - time.time())
+                        if remain:
+                            status['message'] = 'Too fast for {:6f},{:6f}; waiting {}s...'.format(step_location[0], step_location[1], remain)
 
                 # Let the api know where we intend to be for this loop
                 # doing this before check_login so it does not also have to be done there
@@ -496,7 +514,7 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
                 # Make the actual request (finally!)
                 response_dict = map_request(api, step_location, args.jitter)
 
-                # G'damnit, nothing back. Mark it up, sleep, carry on
+                # No response. Mark it up, sleep, carry on
                 if not response_dict:
                     status['fail'] += 1
                     consecutive_fails += 1
@@ -670,6 +688,26 @@ def calc_distance(pos1, pos2):
     d = R * c
 
     return d
+
+
+def check_speed_limit(args, previous_location, next_location, last_scan_time):
+    if args.speed_limit > 0 and last_scan_time > 0:
+        move_distance = calc_distance(previous_location, next_location)
+        time_elapsed = time.time() - last_scan_time
+
+        if time_elapsed <= 0:
+            time_elapsed = 0.001
+        projected_speed = 3600.0 * move_distance / time_elapsed
+
+        if projected_speed > args.speed_limit:
+            extra_delay = int(move_distance / args.speed_limit * 3600.0 - time_elapsed) + 1
+
+            if args.max_speed_limit_delay and extra_delay > args.max_speed_limit_delay:
+                return args.max_speed_limit_delay
+            else:
+                return extra_delay
+
+    return 0
 
 
 # Delay each thread start time so that logins only occur ~1s
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -22,6 +22,8 @@
 #gym-info:              # enables detailed gym info collection (default false)
 #min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
 #status-name:           # enables writing status updates to the database - if you use multiple processes, each needs a unique value
+#speed-limit:           # set speed limit in kilometers per hour (default 0 = disabled)
+#max-speed-limit-delay: # set maximum delay time for speed limit after which the scan will proceed despite exceeding the speed limit (default 0 = disabled)
 
 # Misc
 #gmaps-key:             # your Google Maps API key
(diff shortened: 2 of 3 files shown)


pogom/search.py
```diff
@@ -666,6 +684,27 @@ def calc_distance(pos1, pos2):
...
 
+def check_speed_limit(args, previous_location, next_location, last_scan_time):
+    if args.speed_limit > 0 and last_scan_time > 0:
+        move_distance = calc_distance(previous_location, next_location)
+        time_elapsed = time.time() - last_scan_time
+
+        if time_elapsed > 0:
+            projected_speed = 3600.0 * move_distance / time_elapsed
+        else:
+            projected_speed = 9999.9
```
Mr-Groch: After some thinking, IMHO better is to assume const time_elapsed than speed.

```python
time_elapsed = 0.001 if end_time - start_time == 0 else end_time - start_time

averylarry: Fine.  I'll probably change it.  I did the math -- there are only 2 ways this can make the difference between 0 seconds and 1 second speed delay:
1)  Someone actually uses a speed limit of 10,000 kilometers per hour or greater.
2)  The next location is extremely close.  This is dependent on the speed limit.  A speed limit of 108 kilometers per hour equals 3 centimeters per millisecond.  So if you use 108 for the speed limit the next location would have to be within 3 cm to make the difference between a 0 second delay and a 1 second delay.

darkelement1987:
Awesome, just what i was looking for

beccasafan:
Pulled this down and am running it on an -st 10 -sd 10 with 556 spawn points. To test, I'll be running normally, with -sl200 -msld 60, and with -sl300 -msld 60. I'm using 2 accounts, and the instance generally runs < 30s behind with an extremely rare failure and about a 1% empty response.

Right off the bat I was a bit confused by the arguments and msld being required. If msld isn't specified, the description says the default is 0 which is "effectively disabled". I interpreted this as that no max was enforced and the speed limit check would force waiting as long as necessary to meet the speed limit.

What actually happens is that it checks the extra delay against the default of 0 for the max, 0 is always the smaller, so it uses that number. Thus, no extra delay is ever used because 0 always comes back from check_speed_limit. Not specifying -msld disables -sl as well.

As far as the math goes, instead of check_distance, could geopy.distance be used? .distance or .great_circle? (I don't necessarily understand the math of the difference between them, but I know the python library has those functions...)

wobblybobz:
Just a thought... For most of us we have more accounts than workers running. would it not be a good idea maybe to assign a new account to a worker from the list if max speed limit delay is hit so there are no spawn points missed and i guess it would also help with cycling accounts so their usage does not flag 24/7 usage?

averylarry:
@beccasafan You caught me.  Your understanding of msld was my intention but I missed it.  I'm changing/testing it now and I'll update the code pretty soon so it will wait indefinitely if msld is 0 or unspecified.

averylarry:
@wobblybobz I see people like your idea.  At the moment that is beyond my understanding/coding skills.  I also like keeping this PR simple.

But to address your idea philosophically, I think it would be backwards to manage accounts from a speed limit.  Only considering spawnpoint scanning -- what I think would need to be done is a very drastic change in the workers/scheduler.  Right now the scheduler is fairly dumb.  By that I mean it simply takes a list of locations, sorts them by spawntime, and hands them out in order to whatever worker is ready for a new location.  Instead, the scheduler should be made intelligent.  This starts hinting into the traveling salesman problem.  Or really more like a taxi cab company.  The scheduler should look at all the spawnpoints happening in the next <max-speed-limit-delay> seconds and try to figure out amongst all of the workers the best way to cover all of them within the specified speed limit.  In other words, the speed limit function would morph into a sophisticated scheduler.  To do that, I think both the workers and the accounts need to be added into the database.  The scheduler would need to know all of the workers current status and expected availability to make an intelligent decision.

Long winded.  I guess what I'm saying is -- until someone tackles the really complicated sophisticated scheduler (and the workers can be dumb); we're left with adding small things to the workers to help them help themselves.

averylarry:
OK -- tested a small code tweak.  I ran with -sl 1 and no msld.  You can see it skipped a lot of locations because it wouldn't be able to reach them under the given speed limit before it despawned.  The locations that were finally found are waiting quite a long time:

capture

I'll let it run a little longer and then I'll commit the code change.

Spiritvn:
It would be work very well with https://github.com/PokemonGoMap/PokemonGo-Map/pull/1271 then the BAN rate will be very limited, i will try to merge and use these tonight, let's see

averylarry:
I've been running this for 5 days with no problems.

4RK4N:
also been running this for a week now. no issues.

averylarry:
Now I have to fix for #1097

averylarry:
Fixed (hopefully) conflicts with #1271 and #1097

averylarry:
I put the same code into the new utils.py and search.py from recent merges.  Working great so far!  All of 10 minutes.

4RK4N:
testing out the new code with the latest commits to develop branch

FrostTheFox:
@averylarry Can you PM me on discord? Thunderfox#5680

Stereolab42:
This patch did not turn out to be useful for me. Previously I was spawn-scanning an area of around -st 55 and using -sd 30 to prevent teleport bans, which worked pretty well. I was hoping this patch would let me reduce the number of workers I needed by reducing -sd, but it doesn't act like I expect it would. What I expected was that if a particular account was about to violate the speed limit, a different account would be chosen. Instead it seems to just put the worker to sleep? Thus, the overall practical effect for me was exactly the same as increasing -sd since all my workers would just immediately sleep after completing a search item and waiting out the existing -sd. Maybe this is useful for a much smaller search area, but not in my case.

averylarry:
@Stereolab42 Indeed.  The main benefit of this PR is that it accomplishes a speed limit VERY simply.  It doesn't involve scheduling, rather it involves the individual workers independently slowing themselves down.  It generally won't make anything more efficient -- it's function is to delay a worker when necessary.  But when you care about account bans from teleporting too far too fast . . .  (It can even just be used to show you that you need more workers.)

PRs #787 and #1386 accomplish a speed limit by modifying the scheduler.  That is a far more complex code change and they introduce their own issues.

Mr-Groch:
Why closed? It was our only protection against teleport bans...

averylarry:
This was a simple code change.  If it was going to be merged it would have been by now.  The general consensus seems to be a desire to wait for a better implementation involving the scheduler instead of the worker like #787 and #1386.
As a simple code change, I could re-create it easily if the winds of desire change to favor this approach.  Until then, there's no reason to clutter up the list of open PRs.

PR #1303, opened 2016-09-14
Only send updated and newly visible Stops/Gyms/Pokémon and not-'hidden' Pokémon in the JSON from server to client.

Darkknightz:
## Description

At the moment when you hide pidgeys or rattata's the json you pull from the server still includes these Pokémon.

Also only sends pokestops/gyms/pokémon that have changed since last request and those uncovered while zooming/moving.

## Motivation and Context

Not everyone has unlimited mobile bandwidth and this should also help with bandwidth usage on rented private servers.

This should greatly reduce the json pull every 5 seconds, and especially useful when implementing background updates like: PR #1287

## How Has This Been Tested?

I have tested this locally (single user), i could not see a difference in performance.

Maybe someone with more users can see if this has huge database/cpu impact.

_Edit: 
My concern came from me using Pokemon.get_active_by_id. By now this concern is outdated. And should in fact lower load on the server significantly saving time on SQL/JSONIFY/GZIP._

Before/After Picture
beforeafter

Not hiding any Pokemon.

Above screenshots show difference this will make after 1st request. Only sending modified stops/Pokémon
## Types of changes

## Checklist:

!For the merger: This PR changes the db scheme adding last modified to Pokémon.


DiscoTim:
this is interesting, i am wondering too about only pulling notified Pokemon. The polling method is terribly inefficient, with the server returning the same pokemon over and over. I understand why it is the way it is, because things get complicated. ideal would be something like the server pushes only new, and the client verifies that it received.

Darkknightz:
Ideally the server would only send new Pokémon yea, maybe we can do a 'init' request and a new Pokémon 'since latest' request. That would require more coding, filtering ID's was already in app.py so I just abused that. This is a quick-win, as long as requesting all active Pokémon and requesting Pokémon by 'id' is not heavier on SQL and other calculations.

ottowayne:
TBH I was thinking of going in the exact opposite direction and letting the client do all the filtering.
This would make it possible to allow e.g. Cloudflare to cache for a very short time when there are multiple users requesting data over and over again. More complex queries are more load for the database then if it'd be changed to bulk dumps.
Of course it'd have to be optional in case there are some people covering half their country.

DiscoTim:
I like what @Darkknightz was touching on, with only send me new pokemon since time_stamp, the timestamp could be backed up a bit to ensure nothing was missed (also have repeats) but definitely reduce overall bandwidth because you would not be getting 15 minutes of the same pokemon over and over.  would querying with a timestamp increase DB overhead much?

Darkknightz:
it would need to be an alteration of this in models.py

```
def get_active_by_id(
--snip--
query = (Pokemon
                     .select()
                     .where((Pokemon.pokemon_id << ids) &
                            (Pokemon.disappear_time > datetime.utcnow()) &
                            (Pokemon.latitude >= swLat) &
                            (Pokemon.longitude >= swLng) &
                            (Pokemon.latitude <= neLat) &
                            (Pokemon.longitude <= neLng))
                     .dicts())
```
 it already queries timestamp.

Something where  Pokemon.disappear_time > (datetime.utcnow()   + (15 min - 10 seconds.))

This might get the Pokemon that have spawn for longer than 15 min again but again this is another quick-win.

So first request should be get_active

ottowayne:
@Darkknightz That's not a good solution. I already thought about stuff like this for another patch i worked on. You can't be sure that every new Pokemon found has a later disappear time than the previous one due to parallel workers, failing workers and so on.
To solve this the way you are thinking of, you'd have to add an autoincrementing id or "found timestamp" to the database to be sure that you are retrieving all newly found Pokemon from the database.

Darkknightz:
You are right, finding Pokémon late would not work with what I mentioned. A found time-stamp to query on would work.


Again, this PR is just a quick-win without much alteration to code. Any further optimization would deserve its own PR.

DiscoTim:
@ottowayne i am not familiar with the this database, but does it not include a timestamp for when the record was created, and if so could that not be used in conjunction with what @Darkknightz  is suggesting?

Darkknightz:
I'm not comfortable enough with the SQL / scheme versions in this project to touch it just yet but something like this: 
http://stackoverflow.com/questions/4870536/mysql-automatically-store-record-creation-timestamp

In combination with a query that queries this timestamp, on now()-10 seconds (or what ever the background update max delay would be)

Just have to make sure that if the map has been in active too long that a new 'init' request is done. Or a cookie that stores last_request and use that instead of now() - 10 seconds.

DiscoTim:
I am looking at the database, some have last_modified but not the pokemon table. Also, the client could send what timestamp it wants, and the client can track what timestamp it last sent/requested.

Darkknightz:
Or add datetime.utcnow() to the json and use this value in the next request. If this value is unknown then request all Pokemon.disappear_time > datetime.utcnow().

Darkknightz:
The same can be done for pokestops and gyms, if thats not already done.

DiscoTim:
gym and pokestop do have the last_modified

Darkknightz:
if you add send time to the json, you can filter pokestops/gyms from the next json.

Would only send bytes in followup requests, gziping it would almost be a waste of CPU :p

I'll see if i can get this to work, for pokestops/gyms atleast. (if someone doesn't beat me too it)

DiscoTim:
awesome! :) like you, i am not sure what is involved in getting the last_modified into the pokemon table.

Darkknightz:
having the server send you the time in the JSON also eliminates time difference(/drift) and transfer time between user/server.

Not sure if this creates conflicts with other projects like iPokeGo and the android client, having extra data in the JSON I guess aslong as i stay out of the Pokemon/Gym/Pokestop collection it should be fine.

DiscoTim:
extra values in the json should just be ignored, and as you said if the value doesn't exist it just sends it all.

Darkknightz:
Need to adjust code so it requests new pokestops when zooming out or moving around

DiscoTim:
unless I'm missing something, between the pokemon/gyms/stops this is going to be a massive percent drop in data usage... and a lot less processing on the client.. pretty sweet.

quixfz:
just a comment on this. i did this on my server in another way. i changed the pokemons table the following way:
- encounter_id is a unique index. (but not primary anymore).
- added column eid (uint32) as an auto_increment primary id.
this way, i'm not sending the encounter_id at all in the json, also no spawn id, and less precise coordinates. i'm sending the eid instead (which is muuch smaller), for this i had to modify the upsert to do a "insert ignore" on mysql ('insert ... update on duplicate' would be better). a "replace" would increase the eid and pokemons would get double notifications in the browser.
the json-poller saves the max-eid and only requests the newer ones. and also sends the ignore list, so those dont get sent either. and the whole data is compressed http too.
i did most of this on a php-mysql "web-frontend" that gets synced every 4 seconds. this way the scanners dont get hammered with requests. http://pg.festzeit.ch/

Darkknightz:
@quixfz When IV's get implemented the Pokémon will change after its created and updated with moves/stats. I am afraid your work around will not work then.


Can anyone tell me if you can select pokestops with a NOT IN select query, at the moment I select twice and exclude one from the other.

Darkknightz:
adding Gyms tomorrow!

DiscoTim:
nice :) also, it doesn't look very tricky to get a a last_modified field in to Pokemon table.

last_modified looks the way to go, even if iv's don't make it in, other things may touch it down the road.

are you on discord? hit me up there if you are.

DiscoTim:
I am wondering if a simpler way instead of tracking the coords, is that if the client pans or zooms the map, the client simply resets/zero's the timestamp it sends back?

Darkknightz:
I guess you mean pokestops

DiscoTim:
yes i meant pokestops, i was just thinking ahead :)

jchristi:
Here's an idea: download and store spawn points on the client, have client request only data for currently active but spawn points that haven't been requested already.

tsuncp:
i got 
AttributeError: 'list' object has no attribute 'update'
on
 File "D:\PKMMAP\PokemonGo-Map\pogom\app.py", line 135, in raw_data
    d['pokemons'].update(Pokemon.get_get_active(swLat, swLng, neLat, neLng, oSwLat=oSwLat, oSwLng=oSwLng, oNeLat=oNeLat, oNeLng=oNeLng))

can you help me to solve this :(

Darkknightz:
Pull again, it should fix it.

tsuncp:
have u faced 
2016-09-18 02:42:50,032 [    db-updater-8][        models][ WARNING] (1364, u"Field 'last_modified' doesn't have a default value")... Retrying

tsuncp:
u"Field 'last_modified' doesn't have a default value")... Retrying

after recreate the schema and running the latest code
sql error

Darkknightz:
Sorry First time adding column this should fix it

Darkknightz:
Had tested my self on empty db, which apparently didn't need default value

tsuncp:
ok sorry. jsut my own problem.

Darkknightz:
If you change your db version back to 7 it should now successfully create the new column with default value null

tsuncp:
errr .. actually no problem with the default value , 
i just miss this.......

'last_modified': datetime.utcnow()

pogom/models.py
```diff
@@ -578,11 +659,18 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue):
...
     gyms = {}
+    skipped = 0
 
     cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells']
     for cell in cells:
         if config['parse_pokemon']:
             for p in cell.get('wild_pokemons', []):
+
+                # Don't parse pokemon we've already encountered. Avoids last_modified getting updated on rescanning.
+                if Pokemon.get_encountered_pokemon(p['encounter_id']):
```
jchristi: See @DiscoTim 's https://github.com/PokemonGoMap/PokemonGo-Map/pull/1272#pullrequestreview-721031. Current code will do a SQL SELECT for every pokemon in the list. Suggested optimization will only run a query once for each cell.

jchristi: Here is my suggestion for exact code: https://gist.github.com/jchristi/e694774c3e924dbf8c9b5cc89ab44a38

Darkknightz:
Thanks, I'll add it tomorrow

pogom/models.py
```diff
@@ -83,17 +83,55 @@ class Pokemon(BaseModel):
...
+        query = (Pokemon
+                 .select()
+                 .where((Pokemon.encounter_id == b64encode(str(encounter_id))) &
+                        (Pokemon.disappear_time > datetime.utcnow()))
+                 .dicts()
+                 )
+        pokemon = []
+        for a in query:
+            pokemon.append(a)
+        return pokemon
```
jchristi: just use a simple list comprehension: `return [ pokemon for pokemon in query]`

pogom/models.py
```diff
@@ -83,17 +83,55 @@ class Pokemon(BaseModel):
...
+
+    @staticmethod
+    def get_active(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None, oSwLng=None, oNeLat=None, oNeLng=None):
+        if not (swLat or swLng or neLat or neLng):
```
jchristi: This is a bug. The previous code was `if swLat is None or swLng is None or neLat is None or neLng is None:`. When you apply `not`, you also have to reverse the other logical operators (so `OR` becomes `AND`). So your code should be: `if not (swLat and swLng and neLat and neLng):` which is equivalent to `if not swLat or not swLng or not neLat or not neLng` which is (more or less) equivalent to the original. While the former might be less code, I think the later is more readable.

Darkknightz:
Yea get encountered will not be needed with the new code to exclude already found pokemon

pogom/models.py
```diff
@@ -124,7 +162,7 @@ def get_active(swLat, swLng, neLat, neLng):
 
     @staticmethod
     def get_active_by_id(ids, swLat, swLng, neLat, neLng):
-        if swLat is None or swLng is None or neLat is None or neLng is None:
+        if not (swLat or swLng or neLat or neLng):
```
jchristi: same issue here

pogom/models.py
jchristi: Method can be removed

Darkknightz:
I think this is ready for review/testing.

Unless anyone has more quick fixes? 

I think scannedlocations/spawnpoints should be optimized in separate PR.

Darkknightz:
Updated screenshot to reflect latest changes.

pogom/models.py
```diff
@@ -578,11 +646,26 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue):
...
+                encounter_ids = [b64encode(str(p['encounter_id'])) for p in wild_pokemon]
+
+                query = (Pokemon
+                         .select()
+                         .where((Pokemon.disappear_time > datetime.utcnow()) & (Pokemon.encounter_id << encounter_ids))
+                         .dicts())
+
+                encountered_pokemon = [(p['encounter_id'], p['spawnpoint_id']) for p in query]
+            for p in wild_pokemon:
+                if (b64encode(str(p['encounter_id'])), p['spawn_point_id']) in encountered_pokemon:
```
jchristi: if `wild_pokemon` is empty, then this should cause: `NameError: name 'encountered_pokemon' is not defined`. You need to instantiate `encountered_pokemon` to an empty list outside of the `if` statement.

Darkknightz:
If wild_pokemon is empty it will not go into the for loop therefor not give an error?

jchristi:
Oops? So you aren't having any errors?

pogom/models.py
```diff
@@ -83,17 +83,42 @@ class Pokemon(BaseModel):
...
+                     .where(((Pokemon.disappear_time > datetime.utcnow()) &
+                            (((Pokemon.latitude >= swLat) &
+                              (Pokemon.longitude >= swLng) &
+                              (Pokemon.latitude <= neLat) &
+                              (Pokemon.longitude <= neLng))) &
+                            ~((Pokemon.disappear_time > datetime.utcnow()) &
+                              (Pokemon.latitude >= oSwLat) &
+                              (Pokemon.longitude >= oSwLng) &
+                              (Pokemon.latitude <= oNeLat) &
+                              (Pokemon.longitude <= oNeLng))))
```
jchristi: Am I reading this right?
```python
.where(((Pokemon.disappear_time > datetime.utcnow()) &
                            (((Pokemon.latitude >= swLat) &
                              (Pokemon.longitude >= swLng) &
                              (Pokemon.latitude <= neLat) &
                              (Pokemon.longitude <= neLng))) &
                            ~((Pokemon.disappear_time > datetime.utcnow()) &
                              (Pokemon.latitude >= oSwLat) &
                              (Pokemon.longitude >= oSwLng) &
                              (Pokemon.latitude <= oNeLat) &
                              (Pokemon.longitude <= oNeLng))))
```
is equal to this:
```
where 
    disappear_time > now() and latitude >= swLat and longitude >= swLng and latitude <= neLat and longitude <= neLng \
    and (disappear_time <= now() or latitude < oSwLat or longitude < oSwLng or latitude > oNeLat or longitude > oNeLng)
```
I really don't like the single NOT operator, it makes it much more difficult to read.

Darkknightz:
If your code excludes all the active Pokémon in the old coords from the current coords then that's the same. Although I find yours harder to read :/ I cant even wrap my head around it, it could be that it's 0:10 AM here..

jchristi:
I'm still testing so we'll see

primaxius:
@Darkknightz you mentioned in the description that this breaks notify and hidden? Is that the case? Does this break any current functionality in any way shape or form?

Darkknightz:
@primaxius If you hide a Pokemon thats on your notification list, the Pokemon does not get send so it doesn't get processed. And you wont get a notification. With all the other changes it should be possible to send hidden pokemon or only Pokemon that are on both lists if this a required feature. 

Also looking decided to look further into spawnpoints/scannedlocations.

@DiscoTim has helped me with an updated map.js. Almost ready to push, so it will only send updated scanned/spawnpoints.

jchristi:
@Darkknightz why would anyone hide a pokemon they want to be notified about? Aren't they mutually exclusive?

Darkknightz:
Users do the darndest things ;)

jchristi:
Yeah, but if its shown to be a bug rather than a feature, then this PR becomes a non-breaking change.

DiscoTim:
i could see that, you may not want to see all the pidgey's , weedel's, and ratatta's, but might want to get notified about perfect iv ones. i remember seeing someone mention notifications for hidden on featurehub.

DiscoTim:
@primaxius @jchristi @Darkknightz doesn't break current functionality, could never get notifications for hidden.

Darkknightz:
Zooming in was the last thing I wanted to fix, unless there are bugs that are uncovered while testing. I think this is it :)

nunopitbull:
Can this awsome PR be merged?:)

nunopitbull:
It's working fine and one moment to the other...:
http://prntscr.com/clj73z

Darkknightz:
I think you might have some other PR's interfering with my code.

Darkknightz:
Running current develop next to this PR saw after like 15 min you could see this map drying out, worse on chrome than on Firefox. To make sure we need to create some overlap in the requests, this should make sure the client gets it the changed item.

FabianSperrle:
This seems to prevent expired lured pokestops from switching their icon back to normal, unlured state

Darkknightz:
I realized that yesterday, last_modified on stops is not always within map updates, will need a last_scanned.

beccasafan:
When running the migration, how long does it take / how many rows does your pokemon table have?

beccasafan:
Ok, pulled this down and compared side-by-side without the PR and with the PR, using map.setZoom and map.setCenter to make sure things matched up to then make sure the number of visible markers of each type also lined up.

1. The initial migration took 17 minutes to run (13.5m rows) - that's...***ouch***.
2. When changing the map, this PR results in 1 large request before bringing the size back down. This large request is larger than the requests coming from the map without the pr. 
 - For instance, when zooming out:
   - Without the PR, the map is steady at 4kb, zoom out, and steady at 12.5kb.
   - With the PR, the map is steady at 650b, zoom out, there's a 51kb request, then a 4.6kb, then 5kb, then 2.5kb, then 5kb, then 3kb. 
 - That large request stands out...I guess I didn't expect anything to go over what the map without the PR returned? But then it bouncing around between 2kb and 5kb is a bit odd (though much more minor) given that the data was only changing by 1-2 pokemon/scans as I wasn't changing the map.
3. After zooming out, the number of pokemon/scans were off by 1-2 for a bit. Granted, the 2 maps weren't issuing the request at the exact same time so a variance is possible, it just persisted after zooming out more than zooming in or panning did.

Things I did not explicitly verify get updated on the map correctly:
 - A pokestop becomes lured
 - A lured pokestop expiring
 - A gym changing teams
 - Gyms being updated from anything that's part of `--gym-info` scanning

Overall, this is very close. Running side-by-side the numbers matched up and I definitely saw the significant reduction in response sizes. The length of the migration makes me ***very uncomfortable***, and making sure the state changes from pokestops and gyms get picked up correctly should be done after that code fix.

Darkknightz:
Thanks, i will look into it.

Don't really get what goes wrong when you zoom, tho. 51kb vs 12.5kb original should be impossible. I havent seen anything like it in my testing. 

I'll see if I can spot the error!

Pokeme5005:
First of all, thank you so much Darknightz, and everyone else helping to work on this code. This would really help with me and my friend's bandwidth problems on our phones.

This may be outside the scope of the project, but I'm wondering if it possible to add a "loading" sign while the map is populating? Something like this map is doing? pokejoin.com

Thank you again!

DiscoTim:
@Pokeme5005 you should do a feature request for this. i think the way they are doing it is maybe too jarring? I had thought though a connected/loading/updating indicator was a good idea, and simple enough to do.

Darkknightz:
@beccasafan I think we found why the request is bigger then develop, we are sending the new fields "last_updated" per pokemon. I have changed the query to not return this field, as it is not needed (for now).    (altho 12.5kb vs 51kb is alot..)   (Will push changes soon)

Also fixed pokestops on the backend.
Gyms should be send every time it's been scanned (in order to update "Last Scanned" in the gym popup) so should get new data as it has been scanned.

> The length of the migration makes me _**very uncomfortable**_
I guess you haven't tried 1272 yet, that one adds 5 columns to the pokemon table. I wonder how long that migration would take for you.

beccasafan:
1272 is going in first, but this is good to go pending the rebase that'll be needed.

(sorry! hehe)

Darkknightz:
As said no problem, will get cracking on that rebase

primaxius:
@Darkknightz You rock!

Darkknightz:
@primaxius 
@beccasafan 

So, apparently I had no idea what I was doing. But thanks to @DiscoTim I made it work! 

Here you go, merge conflicts begone!

Barloklol:
There seems to be an issue with -gi enabled, always getting "Exception message: local variable 'stop_ids' referenced before assignment". It's working fine when I disable it, running newest develop build.

Thank you so much for your work btw Darkknightz, this pull is amazing!

Darkknightz:
Whoopsie, here you go

hkuiyo:
How about using websockets to achieve this?

jchristi:
@hkuiyo web sockets is only the transport mechanism so first things first is to make the server send less total data. Further improvements can definitely be made via web sockets but open source projects generally prefer small, iterative improvements rather than giant changes. Web sockets would be a very large change, but is definitely being considered by many devs as a future change.

Issue #1296, opened 2016-09-13
Snorlax is classified as "Rare"


DrAtari:
Snorlax is classified as "Rare", thus it doesn't show a notification when PokemonGo-Map is setup to notify about "Very Rare" and "Ultra Rare" Pokemon. Is it supposed to be like that, or is it a bug?

skaforey:
I found this very strange as well, not sure how the rarities were decided.

filippounitn:
I agree, it should be very rare istead of rare.

DiscoTim:
rarity i think is a large part subjective, and depends on your location.  i'd like to see a system where it calculates the rarity based on your data of previous sightings or a combination of both.

FrostTheFox:
I'm not 100% sure what rarity was originally determined by, but it really should be taken with a grain of salt (except when it comes to things like pidgey/weedle.)

SoonDead:
Although Rarity is location dependant, there is more or less a general consensus about really valuable pokémon.

Snorlax is REALLY valuable.

Since notification should notify about what you really want to collect, you generally want to know if there is a snorlax around.

Since most people don't want to spend 15 minutes every time they set this up on a new mobile, they just check the 2 top rarities. This is helpful and is generally ok. Except one would expect *the* snorlax to be there.

CatsFaninLex:
Which file do I need to edit to change this?

DiscoTim:
@CatsFaninLex \static\data\pokemon.json then npm install

crckdns:
the Spawnrate was changed... snorlax is pretty common now, same with other "rare" spawns from days earlier

FrostTheFox:
I've bumped it up to very rare, although in reality these cant ever be close to accurate without location based checking. #1305

3Trace:
And Dratini is "common" or "uncommon" one or the other.

PR #1305, opened 2016-09-14
Bumped up snorlax

FrostTheFox:
Relating to #1296, personally wouldn't prefer to be changing pokemon rarity left and right but Snorlax probably is more Very Rare than Rare.

### DIFF
--- a/static/data/pokemon.json
+++ b/static/data/pokemon.json
@@ -1811,7 +1811,7 @@
   },
   "143": {
     "name": "Snorlax",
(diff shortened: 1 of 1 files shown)


FrostTheFox:
waiting on travis

toshioson:
Help needed. Changing the Rarity in pokemon.json reflect in the notification of the map in the browser (nor when you click on the pokemon). Something more need to be done? Thanks.

PR #1235, opened 2016-09-09
Advance working with proxies (proxies from file, reloading, rotation)

Badiboy:
Update 1. Allows defining proxyfile with the list of proxies (cli or config). Disable proxy check result output if more the 10 proxies configured. Move all proxy checking from 'runserver.py' to 'proxy.py'.

Update 2. Added possibility to define the periodical proxy reloading (-pxr/--proxy-refresh) from file and rechecking their live status. File can be independently updated from the outside of PoGoMap. Reloading is done in parallel with main search loop and do not stall it at all.

Update 3. Added posibility to enable proxy rotation (before one proxy was binded to one search thread forever). Modes are fixed (as before), round-robin (cyclic rotation) and random (random assignment). Also --proxyfile changed to --proxy-file (as if pull request is still pending) for style compatibility with other proxy options.

Update 4. Additional option -me/-max-empty to switch account when nothing found for some consecutive tries. Usually it means that IP is soft banned. Off by default. Reasonable to use with proxies, otherwise no reason to switch anything: the IP will remain unchanged.

PR #1309, opened 2016-09-15
Fix proxy queue blocking

Braini01:
## Description

Added a task_done also for the case of banned proxies.
## Motivation and Context

If there is one or more banned proxies in the list of given proxies the .join command would be blocking the whole process because there was no task_done method call in this case.
## How Has This Been Tested?

Tested it on a linux server and the observed problem was solved.
## Types of changes

## Checklist:


Badiboy:
Allready implemented in #1235, but let's there be another pull request. It is small and maybe will be commited faster then #1235. :)

primaxius:
I'd rather see 1235 :)

onilton:
@Badiboy This is already accomplished now that 1235 got merged. Can you close your pull request :smile:  ?

PR #873, opened 2016-08-20
Prevent header from cutting off bottom of accordion

Studio-XP:
## Description
As it stands, if the contents of #nav are taller than the window (ie mobile user), the accordion doesn't scroll all the way to the bottom. It stops short 3.5em (the height of #header).

## Motivation and Context
Mobile users are unable to access options that are located at the bottom of the accordion

## How Has This Been Tested?
Reproduced by making the browser small enough to force the accordion to have to scroll. It would only scroll to a certain point.

Fresh clone, modified _nav.scss and ran build commands

Desktop (Win10): Chrome 52, Firefox 48, Internet Explorer 11
Mobile (Android 6.0): Chrome 52, Firefox 48

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1343, opened 2016-09-20
Prevent header from cutting off bottom

Studio-XP:
Same as #873, only this applies to the stats bar instead of the menu bar

## Types of changes

## Checklist:


Becold:
+1 for Chrome and Firefox.
But this need to be tested on some browsers. (like [Android Browser](http://caniuse.com/#search=calc))

primaxius:
+1 follows same code as previous fix for menu.

DiscoTim:
was wondering why position: fixed; top: 3.5em; wouldn't work? that just curiosity, if this works then great!

PR #1365, opened 2016-09-22
Remove unused js stuff

LordNeoZ:
classie and modernizr was used to load the spawnpoints with svg-overlay and make them "bounce", now both aren't being used (svg-overlay and snap.svg-min not even load at all) so there is no point in keeping them

## Description
Removed classie
Removed svg-overlay
Removed modernizr
Removed snap.svg-min
Modified map.html template to not load classie and modernizr
Modified statistics.html template to not load modernizr

## Motivation and Context
Just to improve loading speed / less requests

## How Has This Been Tested?
On my own server, multiple instances for about a day. Presed all the buttons and on multiple page loads.

## Types of changes

## Checklist:


primaxius:
These are files that are part of the framework for the site. As 2 files aren't loaded they have no impact, As for modernizr and classie they are actually used for backwards compatibility for older browsers as I was told. Closing this PR as it doesn't speed anything up and removes functionality.

LordNeoZ:
The only tests that modernizr is calling are csstransitions and js.
csstransitions is only used in the svg-overlay.js, wich isn't even loaded.
without js, there is no way the site will run at all, and also, there is no error handling in case it doesn't support js (like a "get a better browser" message).
Classie is only used inside of svg-overlay.js, wich again, is never loaded, so even if it could be used for backwards compatibility, it wouldn't work without being loaded in the first place.
Also, as i undestand, svg-overlays was replaces by images because of older browsers not being able to handle that much svg redraws.

So far i've tested it in:
Windows: Chrome, Firefox, Microsoft Edge, Internet Explorer 11
Android: Chrome, Firefox, Native browser, Google Now browser (not sure if it's the same as native browser but it looks different).
Linux: Chrome, Firefox
iPad mini: Safari, Chrome

Despite it does break the loading in Internet Explorer 8, it doesn't even matter, because map doesn't load even with the modernizr and classie (not even supported by the gmaps api).
BTW, by checking the gmaps api, only newer browsers are supported, so i don't think we should focus on backwards compatibility.

PR #1370, opened 2016-09-23
Removing old style spawnpoint scanning

primaxius:
## Description

This would remove the non TBTERRA style spawnpoint scanning.  Not sure if anyone still uses this but I don't believe so and should have been removed when the other style was introduced.

This method is inefficient as it will continuously scan ALL spawnpoints, with no scheduling sending the same amount of requests as normal scanning but far more teleporting and far few locations.  
## Motivation and Context

Spawnpoint scanning using this method is inefficient and cause confusion with people trying to run the map with proper spawnpoint scanning.  So many times telling people they are using the wrong spawnpoint scanning.
## How Has This Been Tested?
Both regular hex and spawnpoint scanning tested with no issues on Linux (Ubuntu 16.06)

## Types of changes

## Checklist:


primaxius:
Closing as more people than I thought use this right now however I may submit a change to alleviate confusion between the 2

PR #1383, opened 2016-09-25
Change cool-off time display to minutes and only print it once

springjools:
## Description

Change cool-off notification in account recycler to minutes instead of seconds, and only print it once when the cool off happens and not every minute while someone is cooled off
## Motivation and Context

Less spam
## How Has This Been Tested?

Tested on local machine

09-25 07:20 account-recycler  pogom.search   INFO     Account xxx needs to cool off for 119.0 minutes due to exception
## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -215,8 +215,8 @@ def account_recycler(accounts_queue, account_failures, args):
                 account_failures.remove(a)
                 accounts_queue.put(a['account'])
             else:
-                if not 'notified' in a:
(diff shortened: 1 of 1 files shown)


NGPriest:
Wouldn't -ps be easier?

springjools:
-ps is print status, how is that relevant?

This is a very minor fix to the already existing output of the account recycler thread.

springjools:
I tested it once and it worked, but I also do not really know how to induce cool-offs. I tried with letting the program run and plug off the ethernet cable etc, but it didn't work. But that text in screenshots is copied from the one time I could make a cool-down error.

FrostTheFox:
I tested this before, forgot to accept changes or anything

PR #1307, opened 2016-09-15
add scan here function

Samael27:
## Description
Adds a "scan here" Floating Action Button at bottom center of the map. When clicked/touched, will set the current map center as new search location.

## Motivation and Context
When outside hunting Pokemon you often want to check a few nearby places before deciding which way to go. This is way more convenient when you can just move the map one or two kilometers/miles and tap "scan here" instead of having to type in the name of some location.

## How Has This Been Tested?
Tested using docker.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

This is the updated #16 PR and #1231 PR including the last changes.

PR #1231, opened 2016-09-09
Add scan-here button

Samael27:
## Description
Adds a "scan here" Floating Action Button at bottom center of the map. When clicked/touched, will set the current map center as new search location.

## Motivation and Context
When outside hunting Pokemon you often want to check a few nearby places before deciding which way to go. This is way more convenient when you can just move the map one or two kilometers/miles and tap "scan here" instead of having to type in the name of some location.

## How Has This Been Tested?
Tested using docker.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

THis is the updated #16 PR including the last changes.

PR #1398, opened 2016-09-27
Add scan here function

Samael27:
## Description
Adds a "scan here" Floating Action Button at bottom center of the map. When clicked/touched, will set the current map center as new search location.

## Motivation and Context
When outside hunting Pokemon you often want to check a few nearby places before deciding which way to go. This is way more convenient when you can just move the map one or two kilometers/miles and tap "scan here" instead of having to type in the name of some location.

## How Has This Been Tested?
Tested using docker.

## Types of changes

## Checklist:

This is the updated #16 PR, #1231 PR and #1307 including the last changes.

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -73,7 +73,7 @@ class Pogom(Flask):
             self.search_control.clear()
         fixed_display = "none" if args.fixed_location else "inline"
         search_display = "inline" if args.search_control and args.on_demand_timeout <= 0 else "none"
-		scan_display = "none" if (args.only_server or args.fixed_location or args.spawnpoint_scanning) else "inline"
+        scan_display = "none" if (args.only_server or args.fixed_location or args.spawnpoint_scanning) else "inline"
 
         return render_template('map.html',
                                lat=self.current_location[0],
@@ -82,7 +82,7 @@ class Pogom(Flask):
                                lang=config['LOCALE'],
                                is_fixed=fixed_display,
                                search_control=search_display,
-							   show_scan=scan_display
+                               show_scan=scan_display
                                )
 
     def raw_data(self):
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -73,6 +73,7 @@ class Pogom(Flask):
             self.search_control.clear()
         fixed_display = "none" if args.fixed_location else "inline"
         search_display = "inline" if args.search_control and args.on_demand_timeout <= 0 else "none"
+		scan_display = "none" if (args.only_server or args.fixed_location or args.spawnpoint_scanning) else "inline"
 
         return render_template('map.html',
                                lat=self.current_location[0],
@@ -80,7 +81,8 @@ class Pogom(Flask):
                                gmaps_key=config['GMAPS_KEY'],
                                lang=config['LOCALE'],
                                is_fixed=fixed_display,
-                               search_control=search_display
+                               search_control=search_display,
+							   show_scan=scan_display
                                )
 
     def raw_data(self):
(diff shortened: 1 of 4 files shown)


Slivo-fr:
Same as #16 and #1307 reforked and rebased to get a clean PR

CalcMan:
How do you intend to handle teleporting workers?

Slivo-fr:
Either you didn't read or you didn't understand the changes. Out of topic comment

CalcMan:
If users start roving on the map clicking scan here, the search worker is going to start bouncing around a lot.

Slivo-fr:
Exactly as if user type a new location, or activate "Search follow me".
If you don't want your map to move, you have to use fl or ss.
What is your point ?

beccasafan:
Without knowing anything, is the purpose clear and distinctive from the "center map" button from google maps?

Slivo-fr:
Yeah, it change the scanning location. Mostly usefull for nomad hexa scan

beccasafan:
I know that...I'm asking if it's clear enough for someone who's looking at it for the first time.

And I'd really like a screenshot added to the description. :smile:

Slivo-fr:
See https://github.com/PokemonGoMap/PokemonGo-Map/pull/16#issuecomment-238073030 for screenshots. 
Seems clear enough for someone that just setup the map imo

Slivo-fr:
Actual screenshot
image

DiscoTim:
i think this might be too easy to teleport workers, maybe its not clear what it does, and might be too easy to touch accidentally. is this button disabled and hidden with -nsc and lock search marker? i think a scan here is useful but not sure this is the best implementation at this point. if there was a confirmation dialog 'Scan current location? yes/no' after you touch the button that might solve a lot of issues.

Slivo-fr:
I'm using it since #16 and never accidentally used it. 
This button isn't disabled and hidden with nsc or lock marker. 
It's not an issue on NSC as it cannot disable it. It's not really an issue with marker lock which is meant to avoid accidental dragging of marker (you can still change it's position typing another location)
The confirmation dialog isn't required imho

DiscoTim:
if lock search marker is meant to prevent accidental dragging of the marker, then a button could certainly accidentally be pressed as well i would think. what does -fl do with the button? is there anyway to hide the button or disable it?

cbogithub:
@DiscoTim , disable and enable in side bar control

```
 function updateLocationMarker (style) {
@@ -251,10 +261,12 @@ function createSearchMarker () {
 var searchControlURI = 'search_control'
 function searchControl (action) {
   $.post(searchControlURI + '?action=' + encodeURIComponent(action))
+  $('#scan-here').toggleClass('disabled', action === 'off')
 }
```

Samael27:
You can disable the button with the -ss and -fl options ;)
But I can assure you that I use it a lot on my phone and I never miss clic it. Apply my PR and try it to make you an opinion with it ;)

DiscoTim:
I could be mistaken, that is just my concern with this pr. i think the time when the button would be most likely miss-clicked would be during panning and zooming..etc.. and it seems just a little too easy to teleport workers half way around the world. also new users are obviously going to click this button just to see what it does.

Slivo-fr:
Try it yourself, we're using it for month without any fail.
Plus user is most of the time looking the map where his worker already scan, it should reduce first discovery teleport issue

Slivo-fr:
After many discussions, some of you were  concerned about mistake click or the button not being obvious enough.
We added a switch to display the button, off by default, that will display an alert to make the user aware of potentials risks.

User will now know what it's about, and must activate this button. Is it safe enough now ?

PR #1432, opened 2016-10-01
update debian instillation

Wtcheesecakes:
updated for easier instillation for noobs like me :)

### DIFF
--- a/docs/basic-install/linux.rst
+++ b/docs/basic-install/linux.rst
@@ -53,7 +53,7 @@ If your output looks as above, you can proceed with installation:
 
 .. code-block:: bash
         
-	cd ..
+	cd ~/
 	sudo apt-get install git
 	git clone https://github.com/PokemonGoMap/PokemonGo-Map.git
 	cd PokemonGo-Map
--- a/docs/basic-install/linux.rst
+++ b/docs/basic-install/linux.rst
@@ -52,7 +52,11 @@ After install, check that you have the correct versions in your environment vari
 If your output looks as above, you can proceed with installation:
 
 .. code-block:: bash
-
+        
+	cd ..
+	sudo apt-get install git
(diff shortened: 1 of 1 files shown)


docs/basic-install/linux.rst
```diff
@@ -52,7 +52,11 @@ After install, check that you have the correct versions in your environment vari
 If your output looks as above, you can proceed with installation:
 
 .. code-block:: bash
-
+        
+	cd ..
```
Ashex: use `cd ~/` instead, this way the working directory is known for the rest of the steps.

docs/basic-install/linux.rst
```diff
@@ -52,7 +52,11 @@ After install, check that you have the correct versions in your environment vari
...
+	sudo apt-get install git
+	git clone https://github.com/PokemonGoMap/PokemonGo-Map.git
+	cd PokemonGo-Map
 	sudo -H pip install -r requirements.txt
 	sudo npm install
```
Ashex: sudo should not be required for the first npm install, only when the `-g` flag is specified since it requires system access to modify stuff. If you can remove sudo there and for grunt build then it should be fine. The npm/grunt stuff isn't in your PR but it is good for keeping documentation in a sane healthy state :)

docs/basic-install/linux.rst
Wtcheesecakes: yep very true good addition

docs/basic-install/linux.rst
Wtcheesecakes: yeah but might as well leave it it does no harm

docs/basic-install/linux.rst
Wtcheesecakes: and done

PR #1439, opened 2016-10-02
Add option to exclude Pokemon by rarity

Jeyms:
## Description

Add option to exclude Pokemon by rarity
## Motivation and Context

Currently, we have to manually enter the name of the Pokemon on the list one by one, or by selecting exclude, once the Pokemon shows up on the map. As different areas have their own set of common Pokemon, the values can be updated by editing the static/data/pokemon.json file.
## How Has This Been Tested?

Have been using this change for two weeks based on the previous develop branch. Have also tested this on current develop branch with no issues.

image
## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -680,6 +680,10 @@ var StoreOptions = {
     default: [],
     type: StoreTypes.JSON
   },
+  'remember_select_rarity_excluded': {
+    default: [],
+    type: StoreTypes.JSON
+  },
   'remember_select_notify': {
     default: [],
     type: StoreTypes.JSON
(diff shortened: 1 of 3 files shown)


aguita:
I love the idea of this, but I'd like this to be synced with https://github.com/PokemonGoMap/PokemonGo-Map/pull/686.   The rarity of the 'mons in the .json file for my area is way off.  Example:  Dratini is labeled as uncommon, but in my area is Very Rare and the Dragonite is ultra rare.

DiscoTim:
to be inline with #1303, and prevent unnecessary data transfer, this should either combine the excluded ids for rarity and specific pokes and send that in the request eids, or account for excluding rarities in the backend.

sebastienvercammen:

PR #1386, opened 2016-09-26
Speed scheduler merge

Artifice:
Create new scheduler Speed scan, option -speed
Enhance ScannedLocation table to allow for minimal scanning to identify spawn points
Add gym and pokestop checking to 0/0/0 scans to allow rescanning after bad scans
Add last action time and location to worker status DB to use to prevent teleporting on program restarts
Add performance statistic reporting for -speed scan option
Add SpawnSighting db table to capture spawn sightings with times and time_til_hidden stamps for later use in identifying spawn types

PR #1476, opened 2016-10-06
simple speed limit

averylarry:
## Description

Recreate #1301 
## Motivation and Context

simplistic speed limit.  Just thought I'd resurrect it yet again.
## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -891,15 +891,48 @@ def search_worker_thread(args, account_queue, account_failures,
                         continue
 
                 # Too late?
-                if leaves and now() > (leaves - args.min_seconds_left):
+                extra_delay = 0
+                # Speed scan does it's own speed calculation as part of the
+                # scheduler so we don't want the worker to do it again.
+                if not args.speed_scan:
+                    extra_delay = check_speed_limit(args.kph,
+                                                    [status['latitude'],
+                                                     status['longitude']],
+                                                    step_location,
+                                                    status['last_scan_date'])
+                continue_time = time.time() + extra_delay
+                if leaves and continue_time > (leaves - args.min_seconds_left):
                     scheduler.task_done(status)
                     status['skip'] += 1
-                    status['message'] = messages['late']
-                    log.info(status['message'])
+                    if time.time() > (leaves - args.min_seconds_left):
+                        status['message'] = messages['late']
+                    else:
+                        status['message'] = (
+                            'Skipping {:6f},{:6f}; outside time and speed ' +
+                            'constraints.'
+                            .format(step_location[0], step_location[1]))
+                    log.warning(status['message'])
                     # No sleep here; we've not done anything worth sleeping
                     # for. Plus we clearly need to catch up!
                     continue
 
+                # Too fast?
+                if extra_delay > 0:
+                    status['message'] = (
+                        'Too fast for {:6f},{:6f}; waiting {}s...'
+                        .format(step_location[0], step_location[1],
+                                extra_delay))
+                    log.info(status['message'])
+                    continue_time = time.time() + extra_delay
+
+                    while time.time() < continue_time:
+                        time.sleep(1)
+                        remain = int(continue_time - time.time())
+                        if remain > 0:
+                            status['message'] = (
+                                'Too fast for {:6f},{:6f}; waiting {}s...'
+                                .format(step_location[0], step_location[1],
+                                        remain))
                 status['message'] = messages['search']
                 log.debug(status['message'])
 
@@ -1206,6 +1239,24 @@ def calc_distance(pos1, pos2):
     d = R * c
 
     return d
+
+
+def check_speed_limit(kph, previous_location, next_location, last_scan_date):
+    if kph > 0:
+        move_distance = calc_distance(previous_location, next_location)
+        time_elapsed = (datetime.utcnow() - last_scan_date).total_seconds()
+
+        if time_elapsed <= 0:
+            time_elapsed = 0.001
+
+        projected_speed = 3600.0 * move_distance / time_elapsed
+        log.debug('Move distance: %s k; time elapsed: %s s; Projected speed: '
+                  '%s kph', move_distance, time_elapsed, projected_speed)
+        if projected_speed > kph:
+            extra_delay = int(move_distance / kph * 3600.0 - time_elapsed) + 1
+            return extra_delay
+
+    return 0
 
 
 # Delay each thread start time so that logins occur after delay.


iluretar:
Will adding more accounts reduce the delay time?

Janglinator:
Why won't admins pull this?

averylarry:
Probably won't ever be merged because coding a better scheduler would render this code obsolete and then it would have to be pulled out/deprecated.  I resurrected it just because I thought people might like to use it for trying to figure out the parameters Niantic uses to captcha-block accounts.

FlarumHostingCo:
What is a good number to use for speed-limit: in the config file?

averylarry:
Adding more accounts will _on average_ reduce the delay time.
Good value for speed limit is unknown.  Cars on the highway can travel 120+ kph.  People fast walk maybe 10kph for short distances.  _In the past_ 100kph was pretty safe, 60kph is vehicle-in-town kind of speed.  Don't know right now what triggers Niantic's bans so . . .

averylarry:
Hopefully it's obvious that a slower/safer speed means larger delays which need more workers/accounts to keep from getting behind on your scans.  This DOES NOT make workers more efficient, it DOES NOT reduce account management.

iluretar:
Im getting a lot of skipping, outside time and speed constraints.
60 accounts -st 15 2k spawnpoints

averylarry:
Unless something else is going on, you'll need to increase the speed limit, max-speed-limit-delay, or workers/accounts.
"outside time and speed constraints" means the worker cannot reach the spawnpoint under the speed limit before it's too late to bother scanning it.

You can test this by increasing your speed limit (maybe double to test) and seeing if it skips less.  **Presuming** that I don't have an unknown bug in the code -- the message basically means your workers are "speeding" in order to cover the area you have specified with the number of workers you are using.

Arengo:
First of all, nice implementation. Works well! I haven't figured out a effective speed limit yet, used to run super safe 15 km/h with huge amount of workers but I think there is still room for being backup-safe and being faster. Any experiences from you? 
Additionally, the next step would be a better scheduling - aiming for the spawn point which is the closest or most efficient regarding other workers and their scheduling.

averylarry:
PRs #787 and #1386 accomplish a speed limit by modifying the scheduler.  They haven't been updated, though.

danielcha:
I guess this is going to become more important with the recent update where sightings don't show up if you are going to fast: https://www.reddit.com/r/pokemongodev/comments/57pco4/another_preventive_serverside_measure_aimed_at/

averylarry:
Resurrected again.  Still don't expect it to be merged in favor of a better scheduler (eventually).

omgms:
well i'm using this at the mean time

winhhh:
How do you pull this properly? I've started from scratch on a fresh Windows install, following the basic install guide on readthedocs.io and the scanner works, then when I try:
`git fetch origin pull/1476/head:SpeedLimiter
git checkout SpeedLimiter`

I get `warning: unable to rmdir pokecrypt-pgoapi: Directory not empty
Switched to branch 'SpeedLimiter'`

kbtbc:
Speed limit works great! 
Max sleep paramater is giving error:
runserver.py: error: unrecognized arguments:  --max-speed-limit-sleep

averylarry:
I've used both -sleep and -delay in the past.  I may have inadvertently switched when I rebased.

neutron666:
How to use with which values ? Always getting "too fast for ..." and waiting delay time.

averylarry:
That's what it should do.  The worker will wait until it can reach the new scan location under the configured speed limit.

neutron666:
So whats the fastest safely working?

Am 18.11.2016 6:28 nachm. schrieb "averylarry" <notifications@github.com>:

averylarry:
I haven't heard of anyone having problems with speed limit 35.  I have had success all the way up to about 42 BUT the code will round up to the nearest second so I'm not certain of the exact speed where Niantic stops returning sightings.

jposton1987:
Downloaded the PR where can I change the speed limit

averylarry:
It's a command line option.  runserver.py -sl 35 (plus whatever other arguments you want).  Also can be set in config.ini

coreyman2000:
image

conflicts with latest merge

m12lrpv:
This needs to be pulled as a priority. It represents a safety check on the entire application which will always be applicable regardless of any future schduler changes.

Imoq:
Would love to use this one, if those conflicts can be solved from dev branch. Pretty, please? :)

sclo012:
The conflicts are very easy to resolve, not sure why they aren't done yet. Mostly it's related to comments and nothing important.

simonbuehler:
@averylarry pls rebase

simonbuehler:
why closing this, this is a must have and works great

averylarry:
rebase

averylarry:
rebase
uses the new -kph argument
does not have --max-speed-limit anymore
isn't used when using the new --speed_scan method

FrostTheFox:
No idea what changes were pending, tagging for review, will probably look at this monday

Imoq:
Seems to be working well! Perhaps worth to integrate it to develop now that it doesn't interfere with --speed as some people are still -ss (for clustering scanning from json).

averylarry:
rebase

averylarry:
Persistent now.  Removed some extra statuses and used the existing ones from the db instead.

averylarry:
Accidental rebase.

pogom/search.py
```diff
@@ -859,6 +877,23 @@ def calc_distance(pos1, pos2):
     return d
 
 
+def check_speed_limit(args, previous_lat, previous_lon, next_location, last_scan_date):
```
sebastienvercammen: For consistency (`next_location`), change `previous_lat, previous_lon` to `previous_location` and pass as a tuple. And replace `args` with `kph`; there's no need to keep passing around the entire args object.

pogom/search.py
```diff
@@ -859,6 +877,23 @@ def calc_distance(pos1, pos2):
...
+    if args.kph > 0:
+        previous_location = [previous_lat, previous_lon]
+        move_distance = calc_distance(previous_location, next_location)
+        time_elapsed = (datetime.utcnow() - last_scan_date).total_seconds()
+
+        if time_elapsed <= 0:
+            time_elapsed = 0.001
+        projected_speed = 3600.0 * move_distance / time_elapsed
```
sebastienvercammen: Could you add an extra whiteline above 888? Just to make it clearer, especially for people not used to reading python.

pogom/search.py
```diff
@@ -576,15 +576,33 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
                         continue
 
                 # Too late?
-                if leaves and now() > (leaves - args.min_seconds_left):
+                if args.speed_scan:
```
sebastienvercammen: This part is unnecessary. You can set `extra_delay = 0` by default and only adjust if it matches any specific conditions (e.g. `if not arg.speed_scan`). It makes code clearer.

pogom/search.py
```diff
@@ -576,15 +576,33 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
...
+                    if time.time() > (leaves - args.min_seconds_left):
+                        status['message'] = messages['late']
+                    else:
+                        status['message'] = 'Skipping {:6f},{:6f}; outside time and speed constraints.'.format(step_location[0], step_location[1])
+                    log.warning(status['message'])
                     # No sleep here; we've not done anything worth sleeping for. Plus we clearly need to catch up!
                     continue
 
+                # Too fast?
+                if extra_delay:
```
sebastienvercammen: `extra_delay` is an integer, so comparison with an integer is preferred (`if extra_delay > 0`). Makes code more readable; people will know what `extra_delay` is, without possibility for confusion. _(In this case it might seem silly, but it's a good habit.)_

pogom/search.py
```diff
@@ -576,15 +576,33 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
...
+                    status['message'] = 'Too fast for {:6f},{:6f}; waiting {}s...'.format(step_location[0], step_location[1], extra_delay)
+                    log.info(status['message'])
+                    continue_time = time.time() + extra_delay
+
+                    while time.time() < continue_time:
+                        time.sleep(1)
+                        remain = int(continue_time - time.time())
+                        if remain:
```
sebastienvercammen: Same comment here as above: `if remain > 0`.

pogom/search.py
averylarry: My argument for the way I had it is that it executes less code.  It makes an evaluation on the if statement every time (if args.speed_scan).  Then it sets the extra_delay one time depending on the outcome of the if statement.
They way you suggest will set extra_delay to 0 every time, evaluate the if statement every time, and then also sometimes set the extra_delay again.

pogom/search.py
averylarry: I'm not sure if I changed this quite the way you'd like.

pogom/search.py
sebastienvercammen: Python defines variables at the top of the local scope, but it only assigns the value when it reaches an assignment.

For example, you could do:
```
def foo():
  print x
  x = 5
```
and it would work, even though `x` _"should"_ not exist yet at that point. You'll get an `UnboundLocalError`, but it knows `x`. :)

You're right that the assignment wouldn't happen each time, but assigning the integer `0` is a trivial instruction.

averylarry:
rebase

pogom/search.py
sebastienvercammen: @averylarry Perfect! Much better.

averylarry:
@smc87 Wanna test and give an opinion on the current version?

averylarry:
rebase

PR #1496, opened 2016-10-13
Suggest Node.js v6 in basic instructions

chowderchow:
Node.js v4 gave errors when compiling with npm build, most notably on ARM platforms (RPis etc) but I've seen the same errors on Windows platforms too.

### DIFF
--- a/docs/basic-install/index.rst
+++ b/docs/basic-install/index.rst
@@ -90,6 +90,8 @@ In order to run from a git clone, you must compile the front-end assets with nod
  * `Windows <https://nodejs.org/dist/v4.5.0/node-v4.5.0-x64.msi>`_
  * `OSX <https://nodejs.org/dist/v4.4.7/node-v4.4.7.pkg>`_
  * Linux -- refer to the `package installation <https://nodejs.org/en/download/package-manager/>`_ for your flavor of OS
+ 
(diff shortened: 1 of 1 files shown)


FrostTheFox:
What kind of errors? v4 is LTS, so it's sort of the "stable" one.

FrostTheFox:
Maybe, I would say if you receive errors to upgrade to v6.

docs/basic-install/index.rst
```diff
@@ -90,6 +90,8 @@ In order to run from a git clone, you must compile the front-end assets with nod
  * `Windows <https://nodejs.org/dist/v4.5.0/node-v4.5.0-x64.msi>`_
  * `OSX <https://nodejs.org/dist/v4.4.7/node-v4.4.7.pkg>`_
  * Linux -- refer to the `package installation <https://nodejs.org/en/download/package-manager/>`_ for your flavor of OS
+ 
+(Please use Node.js v6 to avoid build errors)
```
jsherin: considering the links above point to v4.5 and v4.4.7, I don't think it is right to add this without correcting them

FrostTheFox:
5d5af62f4d17a533d259920a9e58db77bdf9695c

PR #1519, opened 2016-11-07
use only resent SpawnPoints

evenly-epic-mule:
instead of using all ever gained SpawnPoints only use the resent ones

## Description
I'm currently getting tonnes of SpawnPoints from the last 4 (or more) changes
which finally makes SpawnPoints Scanning inefficient even compared to HexScanning 

## Motivation and Context
filter get_spawnpoints_in_hex by disappear_time

## How Has This Been Tested?
yes, I had 2.6k spawnpoint, after the update 400
compared with a newly setup database this gives about the same value

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -346,13 +346,15 @@ class Pokemon(BaseModel):
         query = (query.where((Pokemon.latitude <= n) &
                              (Pokemon.latitude >= s) &
                              (Pokemon.longitude >= w) &
-                             (Pokemon.longitude <= e)
+                             (Pokemon.longitude <= e) &
+                             (Pokemon.disappear_time > (datetime.now() - timedelta(days=1)))
                              ))
         # Sqlite doesn't support distinct on columns
-        if args.db_type == 'mysql':
-            query = query.distinct(Pokemon.spawnpoint_id)
-        else:
-            query = query.group_by(Pokemon.spawnpoint_id)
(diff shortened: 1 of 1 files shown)


jsherin:
did you consider just using the purge_data flag to purge the old ones?

evenly-epic-mule:
yep, but what if you want to keep the data for analysis?

averylarry:
Couple different fixes for this are being worked on.  It's not easy.

3Trace:
I don't understand, you resent spawn points? What would you have instead?

danielcha:
In case anyone was as confused as I was. It's a typo. He means "recent" spawnpoints. I have been struggling with the same issue. A lot of spawnpoints I have in my database are no longer active. However I think this shouldn't be a default option. Imagine the case where you map a huge area, and only 24/7 scan a small part. After a week you decide to scan a larger area, but those spawnpoints are no longer "recent". This should be behind a flag, but is useful!

MyAdler:
You could have a strike system with spawnpoints. If scanning when a pokemon is expected finds 0 pokemon the spawnpoint gets a strike. Three strikes and we no longer use that spawnpoint for scanning unless its found again. Finding a pokemon at a spawnpoint clears the strikes.

Arengo:
I do like MyAdler's approach.

sebastienvercammen:

PR #1494, opened 2016-10-12
Move spawnpoints to seperate table for easier access

Darkknightz:
## Description

```
Move spawnpoints to seperate table for easier/faster access
```
## Motivation and Context

Generating spawnpoints from pokemon table everytime is very CPU consuming

Also fixes an issue that was introduced in #1303, sending only new spawnpoints would miss the ability to determine if the spawnpoint is "special" (multiple spawn times) or not.
## How Has This Been Tested?

Tested locally with -mock
## Screenshots (if appropriate):

The following is using a relatively small db. 498997 Pokes, 3727 spawnpoints.

On CPU I now barely see MySQL use anything when using map to generate spawnpoints vs full usage on old map.

before:
capture2

after:
capture

First startup generates the spawnpoints (and is updated when pokes are upserted)
capture
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1527, opened 2016-11-11
Valid TTH check & proper spawn times. --Fast-scan updates

mew905:
## Description
Detects long spawns (but not hidden time), checks TTH value, if it's a good value (you catch it within 90 seconds of despawn) it flags it in the database, if not, it checks the spawn_id against the database to see if it's been scanned with a valid timer in the past, grabs the most recent one, calculates the difference between then and now, and provides you with a future spawn time. Also happens to detect long spawns.

Also includes a database pruner activated with -di flag (--delete-invalid) that will prune all pokemon with invalid despawn time AFTER the map's assumed 15 minute despawn.

## Motivation and Context
Prepping for spawn scan, still useless without a proper speed limiter, but I dont like being told something's not possible.... plus user complaints of pokemon being on the map, but not on the game, were annoying.

## How Has This Been Tested?
On my own map, see screenshot below

image
image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -717,7 +717,7 @@ def hex_bounds(center, steps):
     return (n, e, s, w)
 
 
-def construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid):
+def construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid, nptime):
     pokemons[p['encounter_id']] = {
         'encounter_id': b64encode(str(p['encounter_id'])),
         'spawnpoint_id': p['spawn_point_id'],
@@ -725,6 +725,7 @@ def construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid):
         'latitude': p['latitude'],
         'longitude': p['longitude'],
         'disappear_time': d_t,
+        'last_modified': nptime,
     }
 
     if valid > 0:
@@ -814,26 +815,25 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
                 d_t = datetime.utcfromtimestamp(
                     (p['last_modified_timestamp_ms'] +
                      p['time_till_hidden_ms']) / 1000.0)
+                nptime = datetime.utcfromtimestamp(p['last_modified_timestamp_ms'] / 1000.0)
                 valid = 1
             else:
                 # lets grab the times from our database to compare to
-                query = (Pokemon
-                         .select(Pokemon.disappear_time, Pokemon.last_modified)  # we only need the times
-                         .where((Pokemon.spawnpoint_id == p['spawn_point_id']) & (Pokemon.disappear_time < datetime.utcnow()) & (Pokemon.valid == 1))  # grab matching spawn point ID with a valid timer
-                         .order_by(Pokemon.disappear_time.desc())  # sort by most recent
-                         .limit(1)  # we only want one
-                         .dicts())
-
-                # make this a datetime object now
                 nptime = datetime.utcfromtimestamp(p['last_modified_timestamp_ms'] / 1000.0)
-                if query.count > 0:
+                try:
+                    query = (Pokemon
+                             .select(Pokemon.disappear_time, Pokemon.last_modified)  # we only need the times
+                             .where((Pokemon.spawnpoint_id == p['spawn_point_id']) & (Pokemon.disappear_time < datetime.utcnow()) & (Pokemon.valid == 1))  # grab matching spawn point ID with a valid timer
+                             .order_by(Pokemon.disappear_time.desc())  # sort by most recent #we only want one
+                             .limit(1)
+                             .dicts())
                     db = query[0]
-                    diff = (nptime - db['last_modified']).total_seconds()  # compare the new discovery time to the db spawn
-                    diff = int(diff / 3600) + 1  # get how many full hours have elapsed since the old time
+                    diff = (nptime - db['last_modified']).total_seconds()  # compare the new pokemon discovery time to the database pokemon
+                    diff = int(diff / 3600) + 1  # get how many full hours have elapsed since the old pokemon
                     d_t = db['disappear_time'] + timedelta(hours=diff)  # add an hour to the old time because we still need a future time
-                    valid = 1  # validate the timer
-                else:
-                    d_t = nptime + timedelta(minutes=15)  # assume 15 minutes
+                    valid = 1  # validate the timer 
+                except:
+                    d_t = nptime + timedelta(minutes=15)
                     valid = 0  # invalidate the timer
 
             printPokemon(p['pokemon_data']['pokemon_id'], p['latitude'],
@@ -848,7 +848,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
                                                  spawn_point_id=p['spawn_point_id'],
                                                  player_latitude=step_location[0],
                                                  player_longitude=step_location[1])
-            construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid)
+            construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid, nptime)
             if args.webhooks:
                 wh_update_queue.put(('pokemon', {
                     'encounter_id': b64encode(str(p['encounter_id'])),
@@ -1173,6 +1173,11 @@ def clean_db_loop(args):
                          .delete()
                          .where((Pokemon.disappear_time < datetime.utcnow()) &
                                 (Pokemon.valid < 1)))
+                query.execute()
+
+                query = (Pokemon
+                         .delete()
+                         .where(Pokemon.disappear_time > Pokemon.last_modified))
                 query.execute()
 
             log.info('Regular database cleaning complete')
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -726,16 +726,8 @@ def construct_pokemon_dict(pokemons, p, encounter_result, d_t, valid, nptime):
         'longitude': p['longitude'],
         'disappear_time': d_t,
         'last_modified': nptime,
+        'valid': valid,
     }
-
-    if valid > 0:
-        pokemons[p['encounter_id']].update({
-            'valid': 1,
-        })
-    else:
-        pokemons[p['encounter_id']].update({
-            'valid': 0,
-        })
 
     if encounter_result is not None and 'wild_pokemon' in encounter_result['responses']['ENCOUNTER']:
         pokemon_info = encounter_result['responses']['ENCOUNTER']['wild_pokemon']['pokemon_data']
@@ -797,7 +789,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
         # For all the wild pokemon we found check if an active pokemon is in the database
         query = (Pokemon
                  .select(Pokemon.encounter_id, Pokemon.spawnpoint_id)
-                 .where((Pokemon.disappear_time > datetime.utcnow()) & (Pokemon.encounter_id << encounter_ids) & (Pokemon.valid == 1))
+                 .where((Pokemon.disappear_time > datetime.utcnow()) & (Pokemon.encounter_id << encounter_ids))
                  .dicts())
 
         # Store all encounter_ids and spawnpoint_id for the pokemon in query (all thats needed to make sure its unique)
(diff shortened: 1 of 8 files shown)


mew905:
throws a list out of range exception, Trying to find the cause but not being good at python makes me even worse at debugging

pogom/models.py
```diff
@@ -802,9 +814,27 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
...
+                query = (Pokemon
+                         .select(Pokemon.disappear_time, Pokemon.last_modified)  # we only need the times
+                         .where((Pokemon.spawnpoint_id == p['spawn_point_id']) & (Pokemon.disappear_time < datetime.utcnow()) & (Pokemon.valid == 1))  # grab matching spawn point ID with a valid timer
+                         .order_by(Pokemon.disappear_time.desc())  # sort by most recent
+                         .limit(1)  # we only want one
+                         .dicts())
+
+                # make this a datetime object now
+                nptime = datetime.utcfromtimestamp(p['last_modified_timestamp_ms'] / 1000.0)
+                if query.count > 0:
```
mew905: possible source of list out of range exception?

mew905: well this is awkward. Apparently I cant Git

mew905:
"Disappear time unknown" doesn't work now. Thanks, Travis >.> I wish I knew how to js

omgms:
It broke again hahaha

pogom/utils.py
```diff
@@ -154,6 +154,9 @@ def get_args():
     parser.add_argument('-pd', '--purge-data',
                         help='Clear pokemon from database this many hours after they disappear \
                         (0 to disable)', type=int, default=0)
+    parser.add_argument('-di', '--delete_invalid',
```
sLoPPydrive: Better call it "delete-invalid" - use dash instead of underscore.

sLoPPydrive:
I'm having the issue that after I refresh my browser no Pokemon will be shown. This might be a UI only problem as Pokemon are being found and webhooks being triggered.

Maybe it is a problem that I use --delete_invalid for both the map process (no scanner) and the scanner processes (no web server)?

sLoPPydrive:
Ah well, and no Pokemon details are shown. I have encounters enabled (for most Pokemon types).
image

Extrememist:
Have you tried it without -di ?

midadami:
It seems it works fine without -di, I am starting to get some 30min spawns. I know the timers might be wrong and actually checking its dificult.

sLoPPydrive:
Giving it another try without `-di`. So far the Pokemon won't disappear when I refresh the browser page, yay! But clicking on a Pokemon still doesn't show any details. After checking out the pull request I did a `pip install -r requirements.txt --upgrade` (not needed for this PR I think) and `npm install`. What could be the cause? :-/

Barloklol:
It's some error in the map.js file, trying to figure it out

Extrememist:
all i know is i didn't apply the rest of the update so i cannot help much further. :D

Barloklol:
The code cleanup fucked it, use the map.js code from this commit https://github.com/PokemonGoMap/PokemonGo-Map/pull/1527/commits/a3bcec961ac3db0d96050135993e808cd0174c0f

KoertJanssens:
Shouldn't we delete 
`if (b64encode(str(p['encounter_id'])), p['spawn_point_id']) in encountered_pokemon:
                # If pokemon has been encountered before dont process it.
                skipped += 1
                continue`
from _models.py_ in order to update the valid time if we scan a pokemon who is already in the DB?
Or better: skip all the valid Pokemon?

sLoPPydrive:
I just looked at `map.js` and learned about the backticks in JS (template literals). What I still don't get is the conditional concatenation of template literals like this:
```
var contentstring = `
	<div>...<div>`
	if (valid === 1) {
	  `....`
	} else {
	  `Disappear time unknown`
	}`</div>`
```
Is this even proper JS syntax? Couldn't find an example anywhere...

midadami:
Im getting some 1h30 min spawn haha. Not sure if thats right.

quixfz:
great idea, looking forward to a working version.

but, this one doesnt work. problems i see:
- i installed it and getting 1h30 spawns
- see KoertJanssens's comment. currently only spawns scanned (for the first time) in the last 90 seconds (before hidden) get a valid=1 flag.

- _last_modified returned by api is different then last_modified from db (my understanding at least). api time is when it appeared. db time is when it was last modified (currently == appear time), but when updating disappear_time, last_modified should be updated and that should lead to it being sent to the web gui. it should be possible the generate the new disappear_time with the old one. (without using last modified) PS: ignore all that, doenst matter when point above gets fixed._

my understanding of this: run hex-scan (at least an hour) with faster then 90seconds scan time to get a valid time for each spawnpoint. after that you can slow down the scan.

i'm currently running dozens of st3 sd11 scanner, with 3 accounts each. this should take under 90 seconds.

Barloklol:
I removed the line as suggested by KoertJanssen  and currently testing in all my cities. I noticed 1h+ timers as well, not sure if the others are correct. Told my users to report back to me how the "safe" timers are working out, so we'll see.

mew905:
yeah, not sure if map.js works anymore, works on my map as you can see from the screenshots, but travis had me change the if valid === 0 logic to stop the check errors and I dont know how to fix the new logic, home assistance in this area would be well appreciated. Only changes are under pokemonLabel logic.

1h+ timers are definitely an error, but not from my 'fix'. last_modified goes into the database as default utctimenow() when it is upserted, if it exceeds the calculated disappear_time when it goes into the database (could happen because of database locks or slow connections) then the spawn_check logic produces a 1+hour spawn. I did an update in the -di logic that cleans up those times, as they're invalid anyway (how did we find a despawned pokemon? we didn't), so I changed it so it drops the UTC equivalent last_modified_ms into the database instead of utctimenow(). This should add it to the database as "we found it at this time" instead of "we inserted it at this time".

slideup-benni:
I think to delete pokemon from the database to get correct spawn points is not a good idea. You will also remove them from the stats-history when you delete them. A better way is to move spawnpoints to a separate table like here 

https://github.com/PokemonGoMap/PokemonGo-Map/pull/1494

Anyway. You've did a good job! thank you

static/js/map.js
```diff
@@ -362,35 +362,56 @@ function pokemonLabel (name, rarity, types, disappearTime, id, latitude, longitu
...
-	</div>
-	${details}
-	<div>
-	<a href='javascript:excludePokemon(${id})'>Exclude</a>&nbsp;&nbsp
-	<a href='javascript:notifyAboutPokemon(${id})'>Notify</a>&nbsp;&nbsp
-	<a href='javascript:void(0);' onclick='javascript:openMapDirections(${latitude},${longitude});' title='View in Maps'>Get directions</a>
-	</div>`
-  return contentstring
+  if (valid > 0) {
+	var contentstring = `
```
nullpixel1: You should indent with 4 spaces here

static/js/map.js
```diff
@@ -362,35 +362,56 @@ function pokemonLabel (name, rarity, types, disappearTime, id, latitude, longitu
...
+		<div>
+		Location: ${latitude.toFixed(6)}, ${longitude.toFixed(7)}
+		</div>
+		${details}
+		<div>
+		<a href='javascript:excludePokemon(${id})'>Exclude</a>&nbsp;&nbsp
+		<a href='javascript:notifyAboutPokemon(${id})'>Notify</a>&nbsp;&nbsp
+		<a href='javascript:void(0);' onclick='javascript:openMapDirections(${latitude},${longitude});' title='View in Maps'>Get directions</a>
+		</div>`
+  }  else {
```
nullpixel1: Multiple spaces before else

static/js/map.js
```diff
@@ -362,35 +362,56 @@ function pokemonLabel (name, rarity, types, disappearTime, id, latitude, longitu
...
+		var contentstring = `
```
nullpixel1: Travis expects 4 spaces, but there are 2 tabs instead

nullpixel1: `contentstring` is already defined, you redefined it

static/js/map.js
```diff
@@ -362,35 +362,56 @@ function pokemonLabel (name, rarity, types, disappearTime, id, latitude, longitu
...
+		<div>
+		Location: ${latitude.toFixed(6)}, ${longitude.toFixed(7)}
+		</div>
+		${details}
+		<div>
+		<a href='javascript:excludePokemon(${id})'>Exclude</a>&nbsp;&nbsp
+		<a href='javascript:notifyAboutPokemon(${id})'>Notify</a>&nbsp;&nbsp
+		<a href='javascript:void(0);' onclick='javascript:openMapDirections(${latitude},${longitude});' title='View in Maps'>Get directions</a>
+		</div>`
+	}  return contentstring
```
nullpixel1: Travis wants two spaces rather than a tab

nullpixel1: Multiple spaces before return

mew905:
No longer deletes active, invalid pokemon. Didn't know upsert meant "update if insert has matching unique ID". Now updates existing invalid pokemon with newly scanned pokemon for those who happen to scan the same pokemon multiple times before despawn.

-di is only a database cleanup tool that erased despawned invalid pokemon, as well as pokemon whose insertion time was after despawn time, and is no longer necessary with the changes in models.py.

static/js/map.js
```diff
@@ -362,12 +362,11 @@ function pokemonLabel (name, rarity, types, disappearTime, id, latitude, longitu
...
       `
   }
-  var dstime = ''
 
   if (valid > 0) {
-    dstime = `<div>Disappears at ${pad(disappearDate.getHours())}:${pad(disappearDate.getMinutes())}:${pad(disappearDate.getSeconds())}<span class='label-countdown' disappears-at='${disappearTime}'>(00m00s)</span></div>`
+    var dstime = `<div>Disappears at ${pad(disappearDate.getHours())}:${pad(disappearDate.getMinutes())}:${pad(disappearDate.getSeconds())}<span class='label-countdown' disappears-at='${disappearTime}'>(00m00s)</span></div>`
   } else {
-    dstime = `<div>Disappear time unknown</div>`
+    var dstime = `<div>Disappear time unknown</div>`
```
nullpixel1: Change to ```dstime = '<div>Disappear time unknown</div>' ```

nullpixel1: Travis is asking for single quotes, which are `''` these, so mayve

Arengo:
PR#1530 and this should collaborate.

mew905:
@arengo we actually discussed the logic before this came about. He was working on his but because he was busy and couldnt test until later i decided to make an attempt of my own. I had no idea the approaches would be so different. Because im not familiar with what python code uses what CPU cycles i have no idea whose implementation is more efficient but as far as i can tell, both work equally well.

tokar86a:
That sad. But which going to be implemented in the map? Should be nice to see a fix to the spawn problem.

houkop:
+1 to tokar86a, May of us can't decide whith one is better and would be in implemented.

Skorfius:
has anyone else noticed this pr breaks Pokemon auto updating on the map, without manually refreshing the page?

jackey1510:
Yes, sometimes it will not show some of the Pokémon spawned lately. Need to manually refresh the browser.

sLoPPydrive:
Yep, same here. Some Pokémon are missing and are only shown after manual refresh.

mew905:
Yes, its an issue i brought up in #pr as my map doesnt do it, and barlok said his did, but the latest version fixes it for him. Clearly thats not the case for everyone.

Barloklol:
It's still a problem for some of my users, but for me it seems fixed. At least it refreshes now, but apparently it's missing some pokemon here and there until you fully refresh. Not really sure what's causing this because it seems fine when I check out the packets in Chromes network details.

sLoPPydrive:
I bet it's the DB query to fetch only the new Pokemon for the given viewport (if such a query exists). Its conditions will most probably leave some Pokemon out.

AdrienNguyenWorldline:
UPDATE pokevislyon.pokemon SET valid = 1 WHERE TIMESTAMPDIFF(SECOND,last_modified,disappear_time) < 90; 

Should be executed when updating database

Obihoernchen:
I think this can be closed. We have -speed now and there are a lot of merge conflicts with last update ~3 month ago.

sebastienvercammen:
Closing old PR, we've had SpeedScan for a while now.

PR #1550, opened 2016-11-16
Manual Captcha Solving

paraselene:
## Description
Manual Captcha Solving without 2captcha

## Motivation and Context
Someone may not have 2captcha, they need to solve the captcha themselves, it provides a easy way for them to solve the captcha

## How Has This Been Tested?

1. Replace __domain__ with your PokemonGoMap domain and add a bookmarklet, provide your PokemonGoMap domain to runserver.py using -mcsd or --manual-captcha-solving-domain

javascript:(function (){document.getElementsByTagName('head')[0].appendChild(document.createElement('script')).src='__domain__/inject.js?'+Math.random();}());

2. Wait for an account to encounter the captcha

3. Go to https://pgorelease.nianticlabs.com/

4. Use the bookmarklet above

screen shot 2016-11-16 at 9 50 54 pm

5. Solve the recaptcha

6. Recaptcha token will be saved to Database in Token table

7. token_request will get a token which created after the request made and return to the search worker and verify the captcha challenge

p.s.: if token_request cannot get a token which created after the request made within manual-captcha-solving-allowance-time(default: 60s), token_request will return 'ERROR' and append the account to account_failures

## Types of changes

## Checklist:


fu012343210:
Looks great! 

So when an account get's captcha'd, how is the user supposed to know? If they only have 60 seconds to solve it, they have to act pretty quickly.

paraselene:
You will see Account xxx is encountering a captcha, starting manual captcha solving

fu012343210:
Hm, so then do you know how long we can leave them for and still have the captcha remain valid?

Arengo:
Nice approach, but not too friendly to implement yet. Did I understand it correctly, that the Webserver destination wants to run the capture solving, not the scanning destination? This would for example be an issue for me - running my Webserver and Database on another server than my actual scanning processes.

iluretar:
i think that opennig a WebDriver like PoGo-Captcha when no 2captcha key is given would be a better solution to this problem [example](https://github.com/marbink/PoGo-Captcha/blob/master/pogo-captcha.py#L133-L170)

fu012343210:
Well @iluretar there is appeal in having an approach like the one proposed here, as it can be done on any computer, by multiple users, with hardly any setup besides the initial setup.

mdnpascual:
maybe add a new html template so you can solve it on the webserver and the possibility of people accessing the maps to solve them?

Is this possible ?

paraselene:
@Arengo : Yes, the map web server and scanning destination must be using the same machine.

@iluretar : Anyone with the bookmarklet can help to solve the recaptcha. Eg: You can solve it inside mobile chrome.

@mdnpascual : It's not possible to have a html template inside the map. The recaptcha must be done under https://pgorelease.nianticlabs.com/ domain

iluretar:
I see..but then users will be clicking the bookmarklet blindly hoping a captcha is solved right?

neskk:
@paraselene I like what you're trying to do, thanks for sharing your work!

One major draw-back right now:
- When running a large number of instances, the way you lock the file where the token is written to and read from only allows for a single captcha to be solved at any given time, resulting in others being ignored and consequentially adding it to failed accounts.

paraselene:
Thanks to @seikur0, the basic idea comes from PGO-mapscan-opt
https://github.com/seikur0/PGO-mapscan-opt/wiki/Captcha-Solving---Current-State
I just modify the inject.js and return the token to server using img tag and create token_captcha.txt automatically.

paraselene:
@neskk, it's just a version 0.1 prove of concept. The main idea is that anyone could blindly help solving the captcha. I could do better by creating another queue to solve captcha instead of keep waiting for the token_captcha.txt in the search thread, creating a database for storing token, creating a notification system for user to know is that any unresolved challenge...etc

neskk:
@paraselene I didn't mean to criticize your effort, it's an awesome skeleton and I'll try to improve it to have at least a different file for each instance based on status-name. If I get this working then I'll try to setting up a database for captchas to enable having an ID attached to each request so that javascript can pick up captcha'ed workers from a centralized pool.

Please tell me if you would do anything differently or how you would do it since I just started thinking on this.
Once again thanks for your great work!

animefandk:
Please yes! Add this!

Arengo:
I, as well as a lot of the other linux scan network users, would appreciate to counter the "one-machine" issue. Overall, this is the right approach and very promising but when this will be merged, there has to be a real guide to it since the booklet is not handled with intuition.

ic4roswings:
@paraselene I have been working on this solution, for a couple days now, mine tho uses chromedriver.exe with selenium which I think is a lot cleaner than using files, are you instered in merging them ?

weizh888:
Thanks for implementing this feature. When running the runserver.py, do I still include the command -cs to enable the captcha solving? Thanks

weizh888:
I saw Account xxx is encountering a captcha, starting manual captcha solving in console, but where do I go to solve it? I clicked https://pgorelease.nianticlabs.com/, but it is 404 error

paraselene:
@ic4roswings, someone may run the PokemonGoMap without GUI, they can't use chromedriver. That's why I use bookmarklet to send back the token to the server.

@weizh888, After solving one captcha, the token should be sending back to server. Please check if it solved the captcha challenge in the server.

shockler:
Can I make it so that user who connect to my map solve captcha for me?

thatsweater:
Just to be clear,
 - Does not work with -os/-ns setups such as beehives, setups using proxies, and setups using multiple computers

As well i was having trouble getting it to trigger at all so i could even test the first captcha, it would just get 000's all day, how long does it take for it to trigger?

paraselene:
@neskk, I make some modification. Token are stored in Database now. So we can have multiple Scanners running and solve recaptcha At the same time.

neskk:
@paraselene you rock! I meant to do this yesterday but I'm studying for an exam on Saturday and since I figured this would require a little more work than I initially thought I postponed it :) At some point I even thought of putting the accounts on database too but this way is simpler. Great job!

Edawg478:
When a captcha isnt solved in the 60 seconds will a new account takes its place? Then once the captcha is solved will that account go back into rotation? Im just thinking that it would be nice to solve a handful at a time without causing dead spots. Then when you come back to the pc after catching some mon you can solve them.

thatsweater:
using -cs -mcsd 'domain' with a -ns, captchas are now triggered from 0/0/0, however it does not appear that the bookmarklet works either with the dns (domain), or the IP address directly of the server.

thatsweater:
Any way to make it....not chinese?

paraselene:
@Edaw478, a new account will take its place. Actually, you can solve the captcha using bookmarklet in your phone when you are catching pokemon, haha

lluisd:
I don't understand well the 60 seconds that you have to resolve the captcha. As I understand you have 60 seconds to resolve the captcha when the "Account xxx is encountering a captcha, starting manual captcha solving" message is shown. But what happens if you want to resolve it 10-30 minutes or 2hours later? I am not watching the console log during 24h day.

paraselene:
@lluisd, You account will be back after account-rest-interval which is 7200s after and you can solve the captcha again. You can set account-rest-interval to a lower value

lluisd:
@paraselene I still think that is to difficult to know when the 60 seconds  will start since you dont have any external notifier. To make it work you are forced to watch the console log and wait for that message.Perhaps will be better to dont  rest the account and restart the captcha again each 60 seconds so then Ive time to resolve it.

paraselene:
@llusid, it will send webhook and tell you how many token is needed now.

neskk:
@paraselene It's almost a perfect solution, you're the man ;)
To achieve perfection, we need a way to signal the front-end, not just webhook (which is a great feature nevertheless), that there are workers receiving Captcha so that map users would actually help out right away and don't let the account go to sleep.
I would also consider making a cycle of X attempts to manually solve the Captcha, i.e. when a worker gets captcha'ed it will notify the database, wait 60sec for solving and if manual solve fails it would retry the manual process until it fails X times. I know this isn't a perfect solution since it will stall scanning... Another possibility would be having a separate account queue/list for captcha'ed accounts. 
The perfect solution really would be a separate page managed my pogomap that would activate the queue of captcha'ed accounts once a user logged in and maybe display an IFrame with the captcha right there on the map site. Once that captcha is solved it would jump to the next account that needs captcha solving.

Just putting my thought out there, not trying to diminish your work, keep it up ;)

lluisd:
Yeah I am also thinking like @neskk . It will be great to have a queue of accountd that need captcha and start resolving them whem you want (by accessing some new page). The most important is also to not spend more time by resolving a captcha of an account that is acting as the current worker since you are wasting the worker hex time. 

Perhaps is difficult to adapt some new complex solution to the current code, your work is great and i also vote for a solution that works without a desktop server machine (with GUI).

suppero:
@paraselene why do you close it?

PR #1530, opened 2016-11-12
Valid times on map (SpawnPoint scanning OK, read inside) [Last update: 10dec]

marbink:
## Description
Added algorithm to get monsters disappear time from previous scans results.
On the frontend inside the popup you will see clearly when the time has been set by prediction, when it has been set by scanner at 15min and when it has been set by server response.

**Spawnpoint Scanning Warning:** Pay attention, This PR adjusts time BUT DON'T limits speed. If you don't want to merge a PR that limits speed, use SS on small areas or using small beehives. Not sure of this, but i think that using SS is easier to teleport accounts. 

**[Update x2 - 10dec2016]** For those who are/were using this PR that want to migrate valid TTHs found on new db, follow these steps:
- Make a backup of the whole db
- Drop `ScannedLocation` and `WorkerStatus` tables
- Git clone PGM master 
- Run runserver.py for some seconds, just to make it to create new tables.
- Run this query to migrate data: `INSERT INTO spawnpoint
SELECT
spawnpoint_id as id,
latitude,
longitude,
max(last_modified) as last_scanned,
"hhhs" as kind,
"????" as links,
0 as missed_count,
UNIX_TIMESTAMP(max(last_modified)) % 3600 as latest_seen,
UNIX_TIMESTAMP(max(last_modified)) % 3600 as earliest_unseen
FROM pokemon
where time_detail = 1 and spawnpoint_id NOT IN (SELECT id as spawnpoint_id from spawnpoint)
group by spawnpoint_id
`
- If you don't get any error in the previous step, you can proceed deleting the `time_detail` column (not needed anymore): `ALTER TABLE pokemon DROP COLUMN time_detail`

**Note:** This query will migrate valid disappear_time on new tables, BUT it won't set the appropriate duration. It will be set during new scans.

**[Update - 09dec2016]** Talked with 1501 author. Maybe we find a way to migrate this PR TTH data to the new db format and compatible with SpeedScan. Let's see if I can write a migration script this weekend. 

**[Update - 30nov2016]** Added index on time_detail column. For those who were already using this PR before this update, you should execute this query to adjust your DB: `ALTER TABLE pokemon ADD INDEX pokemon_time_detail (time_detail ASC);`

**[Update - 27nov2016]** Added one more CLI param: `-unv` or `--only-unvalid`
[...]

PR #1367, opened 2016-09-23
Beehive Multiple Account Support

Janglinator:
This fork adds an option for a new pattern in multiple account searching.
## Description

These changes allow users to specify that they wish to use a beehive pattern, rather than a single hexagon path with multiple accounts playing leap-frog and getting banned.  Coupled with PR #1301, multiple accounts should be able to run indefinitely without a ban.

The example below is using 19 individual accounts, each with a step of 5.  `-bh -st 5 -w 19`
Each hexagon uses a _single_ worker, eliminating all jumps from every worker.

How it works:  I use multiple schedulers, on per worker, to perform searches.  Any scheduler modes (`-ss`, `--spawnpoints-only`) still work the same, but each scheduler is given its own point of origin, calculated in a circular pattern around the origin ('-l').  Rather than putting all workers on ONE path, it calculates an individual path for EACH individual worker, using the same algorithm as before.  Check out the images below.

What it DOES do:
- Allows for multiple workers to be used, while specifying the exact step-size of each individual worker, to maximize total search area.
- Calculates multiple separate paths, one per worker, and allows workers to scan their paths individually.
- Guarantees no workers will ever jump more than the diameter of a single search query (unless account is placed on hold).
- When one worker fails, another worker replaces it (assuming there are extra available).
- Causes the `-st` property to be interpreted as the step limit for _individual_ workers, rather than the step size for _all_ workers.  Some consider this a negative; I consider it a positive because I now know _exactly_ the area each individual worker is responsible for covering, allowing me to maximize the total search area.
- Significantly decreases the unnecessary ram and cpu overhead of running multiple terminal threads to ensure accounts don't play hopscotch. 🎉 
- Eliminates the need to calculate individual paths for multiple accounts; that's basically all this PR does  😆

What it does NOT do:
- Change the closed circuit search path, or any path calculations implemented. (See screenshots below)
- Implement [schristmann's suggestion](https://github.com/PokemonGoMap/PokemonGo-Map/issues/809#issuecomment-241069686) of using multiple workers around a single scanner path.
- Conflict with [schristmann's suggestion](https://github.com/PokemonGoMap/PokemonGo-Map/issues/809#issuecomment-241069686).  We could simply implement a way to specify workers per behive path when it IS implemented, utilizing both techniques at once.
- Cause ANY properties to be ignored.  The `-st` property is re-interpreted as being the steps for each individual worker, rather than step size for ALL workers.

[...]

PR #1551, opened 2016-11-16
Solve Manual Captcha with ChromeDriver on Demand

ic4roswings:
Provides a way to solve captchas without 2captcha payed service 
## Description
After 2 consecutive 0/0/0 Returns checks the account in order to obtain the URL of the captcha if it exist opens a Chrome instance to solve the captcha It waits 30 sec or Until the captcha was solved to retrive the information

EDIT: CHROMEDRIVER IS REQUIRED, IN THE PYTHON SCRIPTS FOLDER

## Motivation and Context

Some people like myself can't make a payment in the 2captcha website
It solves https://github.com/PokemonGoMap/PokemonGo-Map/issues/1470

## How Has This Been Tested?

Running an instance waiting for the captcha to raise the script itself was tested with at least 50 captcha accounts that I Have 

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/Community-Tools.md
+++ b/docs/extras/Community-Tools.md
@@ -24,3 +24,8 @@ _Example: I have an ST of 7, a delay of 10, and need my scan to be around 100 se
 Constantly in development - but allows the generation of a multi-worker .bat file with ease, along with custom flags. Can read account and location data right off a .csv file (no need to edit the files either - just set the right columns) and allows you to customize everything from the starting parameters to the naming of each instance through an easily configurable config.ini. 
 _Example: I need to search a few small areas very very quickly with accounts I've generated through the_ [PTC Acc Gen](https://github.com/skvvv/pikapy) 
 _Example: I have a list of locations and a list of accounts, but I don't want to go through the tedium of having to create the .bat myself_
+
+## [HoneyFill - Multi-account-multi-location generator in python](https://github.com/joostsijm/HoneyFill)
+### Easily parsing your accounts and coordinates.
+Generate .sh file to start PokemonGo-Map with given coordinates. Use your .csv files to fill the users and coordinates. Providing Command line arguments to configure the output. Allows setting the step size of multiple coordinates, and amount of users it's using. Useful when you need to regenerate a scanner starter with new accounts.
+_Example: python2 generate.py -a EXAMPLEaccounts.csv -c EXAMPLElocations.csv_


iluretar:
i think you should be using captcha_token as variable name instead of linea

Arengo:
We all love Travis. . .. ...

totoberg123:
I guess this is only for Windows?

ic4roswings:
@iluretar yeah maybe once I get it to compile xD

iluretar:
Maybe make it wait 60s instead of 30? pikaptcha do 60s and it doesn't seem to be too long

ic4roswings:
@iluretar I only made it 30 cause if you are not on your desktop when the captcha prompts i'll pause your scanning for as long as the timeout goes, but maybe it can be modified to be an argument

iluretar:
But it only pauses scanning for that worker which shouldn't be a problem, one cool thing would be a flag to first try manual and if that fails go for 2captcha, i'm at my desk most of the day and have pgm running in a side monitor so i can be resolving captchas without issue, but when i go out it would be awesome if it used 2captcha.

ic4roswings:
@iluretar As far as I know all the process are paused waiting for the captcha response with 2captcha or manual

iluretar:
Been running it for about 30min and solved around 10 captchas, all workers kept going while i solved them.

ic4roswings:
are you sure ?? that sounds like multitasking to me

iluretar:
captcha
Maybe i'm interpreting this wrong but it seems like the other workers keep going

ic4roswings:
Yeah, you're right I haven't see that, is kind of wierd I don't understand how is it working that way, 1 Minute it is then

iluretar:
i think that time should be the time that recaptcha url is valid to maximize efficiency, it that time known?

gneissone:
@totoberg123 No, it's working fine for me on OS X. Maybe it's obvious, but to make this work you need to download [chromedriver](https://sites.google.com/a/chromium.org/chromedriver/) and put it somewhere in your PATH. Then make sure you're including the -cs or --captcha-solving flag when you start runserver.py.

Bart274:
How far is this from being merged?

springjools:
The problem with checking just consecutive 0 0 0 is that the order of the accounts are always the same. If you hex-scan an area which has, say water or a mountains, that area will likely have a valid 0 0 0 and since the accounts are always in the same order, the same account will always hit that area.

Bart274:
@springjools that's not true, if you have for example step limit of 5, you have 61 areas. so each account will always scan the same area if your number of accounts is a multiple of 61, ohterwise they switch locations...

springjools:
Ah yeah, that's true of course, how stupid of me. But if you get a bunch of errors and break before round 2, then this is still true.

Bart274:
@ic4roswings I just tried your PR, I use it on a map that uses 2 accounts and both keep hitting 0/0/0 but I never get a captcha. Anything wrong?
http://pastebin.com/raw/5qujnkUz

chowderchow:
@Bart274 Probably means that they're legitimate empty scans, not captcha'd accounts.

Bart274:
No, I forgot to enable captcha solving in my config :p

midadami:
What version of chromedriver should we use? With default one I get chromedriver not found. With last one chromedriver crashes.

ic4roswings:
The latest one 2.25

mdnpascual:
I cannot make this work with chromedriver 2.25 with PR #1367, #1476, and #1530 

I'm using [marbink's pogo captcha solver](https://github.com/marbink/PoGo-Captcha) and the chrome driver is fine. But when I use this one, chrome will open for a second and just closes.

springjools:
I tried this and it works nicely. However, if you have many captchas, the older ones get hidden in the background. Would it be possible to randomise the position of the new window that appears so they don't all stack?

mdnpascual:
Do you know what part of the code I need to edit to reduce timeout time of account if I fail to answer the captcha below the timeout? 2 hours seems to be excessive timeout if you fail to answer them atm

simisbeat:
I'm new here, can someone tell me how to add this manual captcha solving to my po go map? thank you :)

GubblGubbl:
Unfortunately the chromedriver is not available for the arm architectures like on Raspberry Pis.

wobblybobz:
Working great, just wish there was a way that we could force the prompt for captcha request in case we miss the 60 second window from when it auto pops up. So we don't have to wait for the cool down timer to lapse or to force a restart of the server.

maybe a button on the status page that sends an ajax request to the server that starts the chromedriver stuff for that specific account?

Bart274:
@ic4roswings I also like the suggestion @wobblybobz made: would it be possible to add a button on the status page so you can manually launch the captcha sequence again if you happen to miss the previous one so you can uncaptcha your accounts a lot sooner without having to wait for the cool off?

PokeSearcher:
I'd suggest verifing chrome driver on startup rather than only when a captcha is requested. Its currently diffucult to test/diagnose issues (in my case an older unsupported chromedriver.exe).

Given that the test opens a window i suggest not doing it on search-only process, I changed my runserver.py locally to do:

`    if args.captcha_key is None:
        if not args.no_server:
            val_chrome = chrome_verifier()
            if not val_chrome:
                log.error('Using Chromedriver however its not installed.')
                sys.exit()
        log.info('Using Chromedriver captcha solving.')
    else:
        log.info('Using 2captcha captcha solving.')
`

iluretar:
@ic4roswings can you rebase please?

PR #1556, opened 2016-11-18
Reduced and improved the requests to emulate an actual API call (as an app would)

chowderchow:
## Description
***IMPORTANT
You will need to enable the elevation API here:***
https://console.developers.google.com/apis/api/elevation_backend

1. Instead of calling checkChallenge whenever 2 consecutive empty calls have been made, it's now scripted to check with every getMapObjects call. 

2. Modified map_request to mimic an actual request envelope as from an app 

3. Modified step_location to contain actual altitude

PSEUDO-CODE FLOW:
```
req = api.create_request()
req.check_challenge()
req.get_map_objects(...)
...
response = req.call()
return response
```
This will make only one API call, so no extra calls due to captcha checking as in the current implementation.

All the captcha flags are the same as before:
```
-cs --captcha-solving
-ck --captcha-key
-cds --captcha-datasitekey (OPTIONAL)
```

## Motivation and Context
Some people were 'experiencing' a higher observed captcha rate while using the current map. I think it may be due to the extra check_challenge request every 2 consecutive empty scans and improper API calls.

image

## Types of changes

## Checklist:


paraselene:
Seems Good

crckdns:
that's an "all-in-one-fix" :D purrrfect

smc87:
if you have pulled this before Chowder has had a chance to add to requirements.txt then:
`sudo -H pip install geocoder`

k4n30:
Can the captcha function be updated to automatically report captchas that fail. I've had it happen to me 3 times in the last hour and unless you watch it closely you are paying for bad captcha returns

https://2captcha.com/setting - under additional API requests

willydk:
@smc87 damn... I got 124 captchas in 4 hours...
@k4n30 what would the log say if I get bad captcha?

k4n30:
2016-11-19 16:44:48,221 [ search-worker-1][        search][    INFO] Account xxxxx failed verifyChallenge, putting away account for now

That's what I get now.  If you're asking me what it should say is the same thing plus something like captcha reported (using the api command I linked)

willydk:
i see those in the logs now, thanks

shockler:
I test this PR with 100 new acc and got no captcha for 6 hours
After that i use -ari 900 and -asi 900 and i keep hitting captcha

iluretar:
Google documentation says elevation api is capped at 2500 requests per day on the free tier, are the workers hitting the api each scan?

omgms:
Do I have any problem if I get this error when I start "unable to retrieve altitude from google api, setting to 0"

shockler:
@omgms Chowder said that he will fix it this weekend

Saprophator:
Still getting captchad alot with this pr. settings -sd 40 -st 10 -ld 30

tomballgithub:
How can you tell if you are getting captcha'ed a lot besides looking at your balance on 2captcha?  My balance is going down $2 per day but I would really like to be able to look at the logs or data to see what is really going on per instance.  I also would like to be able to verify that changes like this one or settings tweaks make a difference or not.

willydk:
look at the logs for 2captcha

willydk:
nohup python runserver.py >> nohup_1.out &

tomballgithub:
Many thanks...

simonbuehler:
the Elevation requests should be cached and refreshed when new search location is more than [insert reasonable default distance here] Meters away from the cached value. Add some jitter to it and this should be good. 
The 2500 requests/day limit would be hit otherwise too fast and make this optimization useless

SkOODaT:
works great but doesnt work lol....google api is 2500 dailylimit with 91 accounts its filled within 30 mins

FrostTheFox:
Closed by request for new PR.

PR #1566, opened 2016-11-20
Emulate the actual API calls made as if it was from an app

chowderchow:
get_map_objects, encounter and fort_details/get_gym_details were modified to mimic an actual request envelope made by an Android app

## Description
With these changes, the captcha detection has been moved to be checked with every GMO request, since the response now contains the CHECK_CHALLENGE field. Thus removing the need for the captcha_request function. 

## Motivation and Context
Hopefully helps to avoid ban/softban/captcha detection

## How Has This Been Tested?
Previously ran with these changes in another PR, changes were unaltered and carried forward.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -823,9 +823,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
                 encounter_result = req.download_settings()
                 encounter_result = req.get_buddy_walked()
                 encounter_result = req.encounter(encounter_id=p['encounter_id'],
-                                       spawn_point_id=p['spawn_point_id'],
-                                       player_latitude=step_location[0],
-                                       player_longitude=step_location[1])
+                                                 spawn_point_id=p['spawn_point_id'],
+                                                 player_latitude=step_location[0],
+                                                 player_longitude=step_location[1])
                 encounter_result = req.call()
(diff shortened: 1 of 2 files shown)


therealssj:
I dont think you need to do 
` x = req.check_challenge()` just doing `req.check_challenge()` should queue the request and in the end you do `x = req.call()` like you have done

simonbuehler:
Shouldn't correct elevation values be part of this ? Like getting from Google API, saving in DB, using the value for all locations within x Meters distance. If location is too far away, look up DB for data and if not already Cached call API to get New value,save to DB and done

therealssj:
I can say that the elevation doesn't matter much. Google API has rate limits so its better to leave it out.

chowderchow:
@therealssj it's for verbose/very verbose logs 

@simonbueler elevation implementations will be in a separate PR.

PokeSearcher:
I've got roughly 80 accounts running on individual positioned processes with -st 4 -sd 15 and been trying this pull request for the last couple of hours. Completely subjectively (I didn't record numbers) it didn't seem to make much difference to captchas, I still had maybe 25-30% of the accounts captchad.

FrostTheFox:
Double testing this now, should be good.

k4n30:
@chowderchow assuming you forgot to include reporting bad captchas when they are verified against the API?

PR #1570, opened 2016-11-20
Partially account for invalid TTH values

averylarry:
## Description
Set disappear time = 1/1/1900 when TTH is unknown and no previous valid TTH values
have been recorded for the spawnpoint.
Add argument -dt, --default-spawn-timespan to indicate what timespan should be used
for unidentified spawntypes. Currently uses a default of 30 minutes since 30 minute spawns
seem to be prevalent now where 15 minute spawns used to be heavily most common.
Changes the front end to provide more information regarding the status of spawns
and spawnpoints.
Spawnpoint color changing hardcoded to 30 minute spawns.
Does not identify spawntypes.
Does not change the DB schema.

image

image

image

image

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -100,7 +100,7 @@ class Pokemon(BaseModel):
     def get_active(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None, oSwLng=None, oNeLat=None, oNeLng=None):
         rightnow = datetime.utcnow()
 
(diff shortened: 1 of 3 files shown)


smint86:
The screens above look great. Haven't stumbled upon this yet. Would it be possible to rebase this to make it work with the current development status? Having the times on the map like formatted above would be great, especially together with using the -speed scheduler.

averylarry:
Too much work to rebase with so little chance of it getting merged.

PR #1581, opened 2016-11-24
retry failed search item

paraselene:
## Description

If the search worker failed to scan, it will retry for the failed search item next round

## Motivation and Context

Currently, If it failed in one search item, it won't retry for the same location again

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -482,6 +482,8 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
 
     log.debug('Search worker thread starting')
 
+    retry_failed_search_item = None
+
     # The outer forever loop restarts only when the inner one is intentionally exited - which should only be done when the worker is failing too often, and probably banned.
     # This reinitializes the API and grabs a new account from the queue.
     while True:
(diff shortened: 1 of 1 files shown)


ic4roswings:
@paraselene  Does it works with 0/0/0 or if a captcha is encountered ?

neskk:
Nicely done ;)

sclo012:
Is there any way to rebase it?

pogom/search.py
```diff
@@ -561,12 +566,14 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
...
+                        if retry_failed_search_item is None:
+                            search_items_queue.task_done()
                         continue
 
                 # Too late?
                 if leaves and now() > (leaves - args.min_seconds_left):
-                    search_items_queue.task_done()
+                    if retry_failed_search_item is None:
+                        search_items_queue.task_done()
                     status['skip'] += 1
```
neskk: @kbtbc I think I had this problem before too, it really doesn't make sense to retry a "Too late" scan...
We need to ensure that a "Too late" item is not scheduled for a retry and if for some reason we're retrying an item that is "Too late" then we need to set `retry_failed_search_time =None`.

sLoPPydrive:
Should be closed as #1585 is the cleaner solution that has additional checks whether a scan is really empty (-> no gyms, no nearby pokemon, no pokestops) or just does not have any spawns.

PR #1583, opened 2016-11-26
Manual step_distance control

smc87:
## Description
Add console control for step_distance in schedulers.py

## Motivation and Context
Simple captcha/teleportation avoidance idea, mimic the app better by setting -sd to around 10 seconds (as per normal app behavior, set this value yourself) then vary the step distance so that you move / contact the server more like a regular player. I.E you are checking in with the server every 10 seconds (as per app), but not moving further than you would normally in a 10 second window.

added flag: -sdi --scan-distance 0.070

**EDIT: just a note that this will cause scan overlap depending on your settings and might not be useful for scanning, but more just as a proof of concept into captcha avoidance.** 

EDIT 2: You still need to set -sd yourself.

## How Has This Been Tested?
only a couple of line change,tested on my *nix deployment

## Types of changes

## Checklist:
- [x] I have updated the documentation accordingly. - well the --help page


smc87:
Appreciate any feedback as I haven't been running constantly for a few days ($$$) so don't have great stats to compare with but it looked anecdotally better from watching the -ps screen

eXPeri3nc3:
I normally define -sd as 20 in my normal runs, if I pull this pr to test do I still need to set -sd manually or it does it automatically?

smc87:
You will need to set manually somewhere around 10 ideally (i guess?).
This PR only adds the ability to manually configure the distance between scans, it doesn't impose any other changes at all.

SkOODaT:
this works perfect for captcha avoidance i had a single scanner running for 30 min -st 1 and it gathered 3 captchas with this running havent even seen 1 so far 20 mins in

omgms:
@SkOODaT any update on captcha rate?

shockler:
i set -sdi 0.05 and got 218 captcha/h with 95 workers ...

SkOODaT:
i ended up getting 1 in 40 mins on the single account so i do think this does help

smc87:
I have started using a random value for the step_distance and also a random value you every scan delay (within limits). Currently running a test.

smc87:
Neither method seemed to make much difference. Closing.

PR #1586, opened 2016-11-26
On the fly spawn clustering

mew905:
## Description
OTF Spawn clustering allows you to run multiple locations or beehive without the need to dump a .json file, cluster it, then use it permanently. This is far more flexible, as if it even finds new spawns within the hex bounds by chance it can expand to catch them. Simply apply -ss and -clss to cluster your spawn scanner as it goes. It's literally just the cluster tool in the map files now. Built for compatibility with 1530

## Motivation and Context
Flexibility. Didn't want to have to dump spawn points.

## How Has This Been Tested?
Personal map.

## Types of changes

## Checklist:

The clustering tool we all know and love is added into the map.

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -158,6 +158,8 @@ def get_args():
                         action='store_true', default=False)
     parser.add_argument('-ss', '--spawnpoint-scanning',
                         help='Use spawnpoint scanning (instead of hex grid). Scans in a circle based on step_limit when on DB.', nargs='?', const='nofile', default=False)
+    parser.add_argument('-clss', '--sscluster',
+                        help='Cluster spawnpoints before use (with -ss and no .json)', action='store_true', default=False)
     parser.add_argument('--dump-spawnpoints', help='Dump the spawnpoints from the db to json (only for use with -ss).',
                         action='store_true', default=False)
     parser.add_argument('-pd', '--purge-data',
@@ -309,6 +311,10 @@ def get_args():
             errors.append('Missing `username` either as -u/--username, csv file using -ac, or in config.')
         else:
             num_usernames = len(args.username)
+
+        if args.sscluster:
+            if args.spawnpoint_scanning is False:
+                errors.append('You enabled spawnpoint clustering, but without using spawnpoints!')
 
         if args.location is None:
             errors.append('Missing `location` either as -l/--location or in config.')


tokar86a:
I dont really understand how this work. How should i be scanning?

mew905:
instead of using a spawn .json file to run spawn scan and running that through the clustering tool to get an optimized spawn point file, this just moves the clustering tool into the map, and clusters straight from the database as you scan.

mew905:
think I fixed the issue with this creating more spawn points

simonbuehler:
great PR! it should reduce calls/captchas even more and having it inbuild is great!

mew905:
Have you had a chance to use it yet? I have scanned locations disabled to prevent database locks so I cant see, but a user pointed out the issue and why it increased spawn points.

SkOODaT:
i had issues with the most recent change to cluster.py workers threw errors... had to revert to before the change

simonbuehler:
didn't test yet as i stopped scanning after the forced update - can this be tested with some mocked api provider?

forcewill:
Im really interested in this approach, how does this deals with new spawns discovered during ss scanning?

mew905:
It refreshes every time it gets a new queue. If it discovers more spawns, itll cluster those new spawns in according to whatever PRs you may be running

tokar86a:
Have any one test this yet to see if it reduce the captchas?

mew905:
how about this switch from hex to spawn scan :)
image

urinella:
Using this with PR 1530 causes a problem using -st 10. 
It's going more steps than the "10". Anyone else with this issue?

simonbuehler:
@urinella have screenshot of the scanned area? 1530 shouldn't influence that

mew905:
Hmm... How far are we talkin? Outer edge could pick up a spawnand assign it
to the worker but it should generally stick to the hex bounds according to
models.py

urinella:
@mew905 Sorry I forgot this.
It's not that far: -st 10 is like -st 12/13 in my case.

tokar86a:
@mew905 this sure looks nice. So if i understand this right when you add the -clss it also searching for new spawns that is close bay? I only trying to get a understanding on how this work. I feel a little confused right now. I am sorry about this. 

So its like updating it self whit the new spawn points that it found?

simonbuehler:
i think when scanning during -ss a spawnpoint at the outermost ring, the scan with the spawnpoint centered may reveal a nearby spawnpoint that was out of reach during hex scan. This might cause the list to grow, not sure if that means that spawnpoint scanning will slowly lead to a ever increasing area?

mew905:
Yes and no, it could pick up a previously unseen spawn point and expand,
but it clusters what get spawnpoints in hex finds in the database, so no
matter what they *should* always remain in their designated area

simonbuehler:
im running this with -bh and get many duplicate scans for the same point like this
screenshot 2016-12-01 12 27 34

i think the calculation should filter duplicates between the workers out so this won't happen anymore

sclo012:
How exactly does -ss work at the same time with -bh? Anyways, is there any point of clustering or computing computing spawnpoints whenever get_spawnpoints_in_hex is called? It is much more efficient to just compute it once when the scanner starts and not over and over again whenever the queue is empty. Especially if the db is huge. In normal operation, it is not usual to gain new spawn points in the scanned area except for one case (see above), and even if so, it's better to only compute them when scan is restarted.

simonbuehler:
@sclo012 after a quick look at the code it seems that for each area that's assigned to a worker the points are collected and then clustered. I think the problem is that the hex-scan borders are not kept but expanded a bit so points get assigned to two overlapping worker areas

mew905:
Perhaps give rise to a new table? Theyre not 100% static either according
to some users. Migrating spawn points, changes in spawn time would all
render a spawn .json useless. The advantage they have is they are static.
But you're correct, a semi-permanent database or file consisting of said
spawns would be best. Perhaps when empty scans rises above a threshold we
do a temporary hex scan to determine if spawn times changed? A sort of
"consecutive empties" on the spawn point itself?

On Dec 2, 2016 2:55 AM, "simonbuehler" <notifications@github.com> wrote:

@sclo012 <https://github.com/sclo012> after a quick look at the code it
seems that for each area that's assigned to a worker the points are
collected and then clustered. I think the problem is that the hex-scan
borders are not kept but expanded a bit so points get assigned to two
overlapping worker areas

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/PokemonGoMap/PokemonGo-Map/pull/1586#issuecomment-264404660>,
or mute the thread

simonbuehler:
i wouldn't go that route and doing implicit hybrid scanning, its much easier to understand that there is hex scanning and spawnpoint scanning and the user who wants run both can still do that with additional scanner processes but thats advanced stuff.

If i read the code correct the computing is only done when a new location is selected, so no overhead here?
Just make the Clustering stay within limits, apart from that everything works as it should imo.

bbdoc:
This PR doesn't seem to be compatible with PR 1530... Tried to merge and getting following error :

```
2016-12-02 10:55:35,246 [      MainThread][     runserver][    INFO] Parsing of Pokestops disabled.
2016-12-02 10:55:35,246 [      MainThread][     runserver][    INFO] Encountering pokemon enabled.
2016-12-02 10:55:35,253 [      MainThread][        models][    INFO] Connecting to MySQL database on db:3306
2016-12-02 10:55:35,261 [      MainThread][        models][   ERROR] Your database version (10) appears to be newer than the code supports (9).
2016-12-02 10:55:35,261 [      MainThread][        models][   ERROR] Please upgrade your code base or drop all tables in your database.
```

PR 1530 updates DB to v10 while this PR seems to be awaiting v9...

sclo012:
@bbdoc you have to merge this PR into 1530, or merge 1530 into your own before this one.

bbdoc:
@sclo012. Don't get what you mean. I'm already using PR1530 for quite some time so it already upgraded the DB to v10 (Does it on the first launch). I now would like to add this PR for clustering so I merged it into my existing latest+1530 implementation, but getting this error message.

sclo012:
That shouldn't be any problems merging, this code doesn't modify the db itself, and I merged it fine earlier today. All it does is add one line to call the clustering.

mew905:
yeah, not sure why your db schema version would revert to 9, as sclo mentioned it only modifies the one line, all the rest are in "new" files which wouldn't impact your db at all.

foltincsaba:
I would add time_treshold as an argument. In denser areas I needed 420 secs to avoid skipping. In other places 300 secs was enough.

fantasy4u:
I have use the -ss and -clss command in step 5, it shows the total spawnpoint is less than just use -ss, is it normal. For example, -ss and -clss >> the schedulers show me there are total 100 spawnpoint, if  I just use the -ss he schedulers show me there are total 153 spawnpoint.

simonbuehler:
@fantasy4u that's the whole point of clustering, reducing and unifying nearby points (in space / time)

@mew905 could you check how to make sure the clustering stays within limits so no double scanning in beehive mode will occur ?

llamamastafu:
Could you rebase this?

Arengo:
I have huge DB performance issues on startup.

pogom/cluster.py
```diff
@@ -0,0 +1,149 @@
+import random
+
+import clsmath
+
+
+class Spawnpoint(object):
```
sebastienvercammen: Class name conflicts with currently existing `SpawnPoint`.

pogom/cluster.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.spawnpoint_id is not None:
+            obj['spawnpoint_id'] = self.spawnpoint_id
+        obj['lat'] = self.position[0]
+        obj['lng'] = self.position[1]
+        obj['time'] = self.time
+
+        return obj
+
+
+class Spawncluster(object):
```
sebastienvercammen: Rename to `SpawnCluster` for naming consistency.

pogom/cluster.py
```diff
@@ -0,0 +1,149 @@
...
+        # update centroid
+        f = len(self._spawnpoints) / (len(self._spawnpoints) + 1.0)
+        self.centroid = clsmath.intermediate_point(spawnpoint.position, self.centroid, f)
+
+        self._spawnpoints.append(spawnpoint)
+
+        if spawnpoint.time < self.min_time:
+            self.min_time = spawnpoint.time
+
+        if spawnpoint.time > self.max_time:
```
sebastienvercammen: Should be `elif` rather than a separate `if`, otherwise it'll unnecessarily check for `>` even if it already was `<` before.

pogom/cluster.py
```diff
@@ -0,0 +1,149 @@
...
+
+def test(cluster, radius, time_threshold):
+    assert cluster.max_time - cluster.min_time <= time_threshold
+
+    for p in cluster:
+        assert clsmath.distance(p.position, cluster.centroid) <= radius
+        assert cluster.min_time <= p.time <= cluster.max_time
+
+
+def main(raw):
```
sebastienvercammen: Function should be renamed to something that makes more sense outside of the scope of the file (e.g. `filtered = cluster.main(filtered)` in models.py).

pogom/utils.py
```diff
@@ -310,6 +312,10 @@ def get_args():
         else:
             num_usernames = len(args.username)
 
+        if args.sscluster:
+            if args.spawnpoint_scanning is False:
+                errors.append('You enabled spawnpoint clustering, but without using spawnpoints!')
```
sebastienvercammen: Rephrase:
```
... but without spawnpoint scanning!
```
because other schedulers do still have spawnpoints, they just don't have the SS scheduler.

pogom/clsmath.py
```diff
@@ -0,0 +1,54 @@
+from math import acos, atan2, cos, degrees, radians, sin, sqrt
+
+R = 6378137.0
```
sebastienvercammen: With `R` only being used in one spot, this should be a local constant rather than a global.

sebastienvercammen:

Issue #1554, opened 2016-11-17
-j not working as expected


danwie:
## Expected Behavior
I expected the location change a little everytime time searching.

## Current Behavior
It seems jitter is not working. Worker always searching at same location as before.

```
$ python runserver.py -ns -l '39.464769, 27.912343' -j -st 1 -np -fl -nsc -k "AAAA..AAAA" -u USER -p 'PASS'
2016-11-17 10:13:53,606 [      MainThread][     runserver][    INFO] Parsed location is: 39.4648/27.9123/22.7332 (lat/lng/alt)
2016-11-17 10:13:53,606 [      MainThread][     runserver][    INFO] Parsing of Pokemon disabled
2016-11-17 10:13:53,609 [      MainThread][        models][    INFO] Connecting to local SQLite database
2016-11-17 10:13:53,613 [ search-overseer][        search][    INFO] Search overseer starting
2016-11-17 10:13:53,614 [ search-overseer][        search][    INFO] Starting account recycler thread
2016-11-17 10:13:53,615 [ search-overseer][        search][    INFO] Starting search worker threads
2016-11-17 10:13:53,616 [ search-worker-0][        search][    INFO] Waiting to get new account from the queue
2016-11-17 10:13:53,616 [ search-overseer][        search][    INFO] New location caught, moving search grid
2016-11-17 10:13:53,616 [ search-worker-0][        search][    INFO] Switching to account Xapimo17
2016-11-17 10:13:53,616 [      db-cleaner][        models][    INFO] Regular database cleaning complete
2016-11-17 10:13:53,617 [ search-worker-0][      auth_ptc][    INFO] PTC User Login for: Xapimo17
2016-11-17 10:13:54,437 [ search-worker-0][      auth_ptc][    INFO] PTC User Login successful.
2016-11-17 10:13:54,437 [ search-worker-0][      auth_ptc][    INFO] Request PTC Access Token...
2016-11-17 10:13:54,675 [ search-worker-0][      auth_ptc][    INFO] PTC Access Token successfully retrieved.
2016-11-17 10:14:14,678 [ search-worker-0][        search][    INFO] Searching at 39.464769,27.912343
2016-11-17 10:14:15,737 [ search-worker-0][        models][    INFO] Parsing found 0 pokemons, 18 pokestops, and 7 gyms.
2016-11-17 10:14:53,620 [account-recycler][        search][    INFO] Account recycler running. Checking status of 0 accounts
[...]

PR #1591, opened 2016-11-27
Adding altitude

gonjiak:
Before anything else, last time I developed something were about 6 years ago, I am totally out and also never worked with python (But don't worry, code is working). 

## Description
Using Google Elevation API to receive an actual altitude for a scanned area. If successful, every consecutive call for the entire map instance will use the given altitude and jitter the value. If unsuccessful, the default value will be used.

https://console.developers.google.com/apis/api/elevation_backend is required
Following commands can be used:

--altitude 
you can set the default altitude in meter, preset is 13m and a floating 0-1 meter will be added.
--altitude_range
range in meter can be added to the normal altitude,  e.g. with 13m default and 1 meter range a random altitude between 12m and 14m gets generated.

## Motivation and Context
Receiving altitude from google API is very limited, except you want to pay a lot money. This solution just do API calls until one successful returns, which mostly is done with the first attempt. The Idea is indeed to create a more realistic PoGo API request to reduce captchas. So far the testings were quite satisfactorily, reducing captchas by 70 - 80% with proper settings, but further testings have to be made.

## How Has This Been Tested?
Tested on my own map, works.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project. (no idea)


smc87:
What about a single (or a few) calls to google API to get a rough altitude and then randomize with that as your base?

gonjiak:
@smc87 that's a good idea, my problem is that I have like no time to actually get into a new (for me) matter. I will look into it, if it's not too complex I can probably implement it.

sirWest:
@smc87 this altitude api was once tried in one of more known pogo bots and the api calls peaked very fast, google has quite a strict limit on this per day so it doesn't make sense to use it.

smc87:
@sirWest - The limit is 2.5k calls daily for the free plan. The idea here would be to only make one or a few calls on initial startup (schedulers.py) to get a rough altitude, then randomise from that value.
We should no make a call for every location lookup.

The idea being that no matter if you are scanning in mexico vs netherlands at least the map should provide a roughly correct altitude for your area.

sirWest:
@smc87 yea that's a good idea, ofc 90% of that random range should be above the last api value, since mostly you can be higher in buildings etc but rarely underground... also, remembering spawnpoint's altitude would benefit a lot for spawnpoint scanning and have all of their altitudes saved into database after some time of random scanning. Many good ideas :)

simonbuehler:
nice idea, in #1566 there was an attempt with elevation api but not properly implemented:

It should be something like this: getting from Google API for current location, save to DB, using the value with some random jitter for all locations within x kilometers/meters distance (using haversine function). If location is too far away, look up DB for cached elevation data  and if not already Cached call API to get New value ,save to DB and done
if api call limits are reached, log the error and use old values, better than nothing

Holzberg:
Being that right now we are not specifying an altitude, does that mean we are passing 0?  And does that mean that if I am in Denver, CO that I am a mile underground?  Seems like that would be an easy way for Niantic to tell if someone is using this map.

gonjiak:
Okay, added Elevation request. Each scan a elevation request is made, until there is a successful return (should be first scan attempt in general, so one API call in total). This altitude will be used for every scan, still jittering the values. If there is no successful elevation request, default altitude from config will be used.

@simonbuehler it's a good idea, but out of my capabilities 

@Holzberg yes, it either send 0 for common or 40.32 for spawnpoint scanning.

simonbuehler:
the last changes do look good to me and my ideas were maybe a bit over-engineered; only thing is that there are now two jitter settings, one "--jitter" for lat, long and the one you introduced, maybe this could unified somehow as  jitter should probably be used on all three values by default

gonjiak:
@simonbuehler as you already mentioned at the #1554 issue, with -jitter as default it would be really easy to unify both features. Other way i can think of is to just merge the altitude_range into -jitter. If -jitter is not used, the altitude just change by 0-1 meter (float) and if used -jitter the lat, lng and alt get jittered.

gonjiak:
I just runned my map with this PR for 10 hours straight. Either Niantic changed their captcha policy, or the changes to altitude got a massive impact on captchas. I got 15! captchas in total, using spawn point scanning on 1250 points and 200 worker. Thats 11273 Area scans, which makes it 1 captcha for 752 scans. That's 1/4 - 1/5 from what I got before

Can someone confirm this?

Spurlex:
Using Google for Elevation is not good. Rate limits will happen fast and then it won't work. IMO better solution is needed.

gonjiak:
@Spurlex This is actually the solution to not hit the Elevation API rate limit and still get an approximated altitude for the scanned area. After first successful API call no further calls are made and the altitude given is used for every consecutive scan.

Spurlex:
@gonjiak Yes its fine if you scan one location. What happens when you scan thousands.

gonjiak:
@Spurlex It's for the entire map instance, not just a single area. I am scanning 1250 spawn points and the altitude used for every scan is based on one API call made right at the beginning.

SkOODaT:
@Spurlex this is better then having all workers set to 0

eXPeri3nc3:
I ran this PR for a few hours now, the captcha triggered is much lower compared to previous runs. Will monitor longer to see how it goes

gonjiak:
11 hours of scanning of 1250 spawnpoints and I just got 7 captchas, works pretty well for me

jorgitoh:
But 200 workers for 1250 spawns is a huge overkill. Using a big -sd like 60 sec u only need 20 accounts to cover that ammount of spawns.

shockler:
@gonjiak -alt 13 -altr 1, am I using the right command?

gonjiak:
@shockler the default -alt is 13 already, so you don't have to add that. The default altitude value shouldn't be used for scans anyways if you have google's elevation activated, which it should be by default if i remember correctly.

simonbuehler:
@gonjiak please fix the conflicts and @FrostTheFox please merge then. really needed

gonjiak:
@simonbuehler done after some problems with upstreaming to current develop branch.. (github idiot)

smint86:
So this works with -speed now?

gonjiak:
@smint86 no, I have to revamp this PR because the current solution is quite floppy. Actually, this PR was just to fill the gap that there is no altitude used, until someone more talented makes a suitable solution. I will see what I can do here, but it will take some time.

danwie:
@smint86 I'm currently testing -speed (including PR #1617) and this PR. I have much fewer captchas with this setup than before (only -speed). Note: I haven't checked the source code or output - but it seems to be working even though @gonjiak says the solution is "sloppy". I have used this setup for about 3 hours only - will let you know if I run into anything noticeable.

gonjiak:
@danwie I was talking about the coding, not the functionality itself

k4n30:
@FrostTheFox any chance of a merge?

sebastienvercammen:
As mentioned on Discord, the quality of the code in this PR is not good enough to be merged.

The following needs to be changed:

1. There is no good reason for the code to be in-line. It should be separated from the searchers and put in its own elevation function. This will also fix the duplicate code you've got going on.
2. Results of queries to the elevation API should be stored. Countries don't suddenly change elevation, so only a single query to the API should be made per location. I suggest creating a DB Model specifically for these entries (no null values, unique per location). Don't mess with the current models, they're not appropriate for storing this data in an efficient manner.
3. To add to #2, moving a few meters in real life doesn't suddenly change elevation enough for it to be relevant, so I suggest adding a radius search in the DB caching code: if it finds a stored elevation within x meters, it should just use that one and store it.

onilton:
@gonjiak If you are not interested in this anymore (or want help) I can help with the changes. I Just need write access to the branch :)

gonjiak:
not working on this pr anymore because i stopped playing due of lack of time. Feel free to use this code as base and make a new PR.

onilton:
Ok. So can you close  this one so I can open another one?

Em ter, 20 de dez de 2016 3:18 PM, gonjiak <notifications@github.com>
escreveu:

sebastienvercammen:
Done @onilton 

Remember to keep the previously mentioned feedback in mind while posting your own :)

PR #1601, opened 2016-12-02
Simple Location Links

kevdotnet:
- added the option to move to a location with URL arguments

## Description
Automatically moves the map to the coordinations specified in the URL.

## Motivation and Context
Makes it possible to send links with location information for use with PokeAlarm for example.

## How Has This Been Tested?
On a running map with PokeAlarm and many users.

## Types of changes
- add URL arguments
- make map jump to the coords specified

## Checklist:

- [x] My code follows the code style of this project. (pretty sure, just a little change)
- [ ] My change requires a change to the documentation. (not really, might be nice though)

Automatically moves the map to the location specified in a link like this: 
http://yourdomain.com/?lat=49.903014&lon=8.5882217

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -75,9 +75,17 @@ class Pogom(Flask):
         search_display = "inline" if args.search_control and args.on_demand_timeout <= 0 else "none"
         scan_display = "none" if (args.only_server or args.fixed_location or args.spawnpoint_scanning) else "inline"
 
+        map_lat = self.current_location[0]
+        map_lng = self.current_location[1]
+        if request.args:
+            if request.args.get('lat') is not None:
+                map_lat = request.args.get('lat')
+            if request.args.get('lon') is not None:
+                map_lng = request.args.get('lon')
+
         return render_template('map.html',
-                               lat=self.current_location[0],
-                               lng=self.current_location[1],
+                               lat=map_lat,
+                               lng=map_lng,
                                gmaps_key=config['GMAPS_KEY'],
                                lang=config['LOCALE'],
                                is_fixed=fixed_display,
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -75,13 +75,9 @@ class Pogom(Flask):
         search_display = "inline" if args.search_control and args.on_demand_timeout <= 0 else "none"
         scan_display = "none" if (args.only_server or args.fixed_location or args.spawnpoint_scanning) else "inline"
 
-        map_lat = self.current_location[0]
-        map_lng = self.current_location[1]
-        if request.args:
-            if request.args.get('lat') is not None:
-                map_lat = request.args.get('lat')
-            if request.args.get('lon') is not None:
-                map_lng = request.args.get('lon')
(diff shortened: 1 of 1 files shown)


FrostTheFox:
Just to confirm, this just moves the view right?

h0us3cat:
Would be nice if we can right click on map and copy url/coordinates from there.

EdenGottlieb:
@FrostTheFox It moves the map scan coordinations, not just the view.
This doesn't implement a solution to fixed location, even if fixed location is turned on it will change the map scan locations (which it shouldn't).
Also, setting map_lat and map_lng can be much cleaner. Like this:
`
if request.args and not args.fixed_location:
    map_lat = self.current_location[0] or request.args.get('lat')
    map_lng = self.current_location[1] or request.args.get('lon')
`

kevdotcom:
Although the console output said "New Location caught, moving search grid", it remained scanning at the same location.

FrostTheFox:
had to revert this, messed up, recommitting

Imoq:
@kevdotnet @kevdotcom I'm running my server with these parameters:

`python runserver.py -l "location" -os -nsc -fl --db-max_connections 312 -H 127.0.0.1 -P 80`

But the links to the map aren't working, it always centers to default -l no matter what lat/lng you pass as an argument. I assume it's because -fl but I am not sure. Do you have an idea what the issue could be?

Thanks!

kevdotcom:
I was always using -nsc and -fl... Did you do "npm run build" after pulling?

Imoq:
Yes, of course I did. Tried from mobile and from web browser. I'm stuck, not sure what else can I try.

Imoq:
nvm, my own stupidity... the argument I was passing was "lon" instead of "lng" as it was a fix from before it was implemented. Fixing now. Thanks and sorry!

Imoq:
btw, it turns out that there was a further modification that rendered this unusable when using -fl (can see the modification here: https://github.com/PokemonGoMap/PokemonGo-Map/commit/289fa2f879270178c594cd571891559e366c407c ).

They way to fix it (thanks to @kevdotcom ) is to revert manually app.py like this:

```
root@ip-172-31-28-74:~/PokemonGo-Map/pogom# diff app.py app.py.orig
81,85c81,83
<         if request.args:
<             if request.args.get('lat') is not None:
<                 map_lat = request.args.get('lat')
<             if request.args.get('lon') is not None:
<                 map_lng = request.args.get('lon')
---
>         if request.args and not args.fixed_location:
>             map_lat = request.args.get('lat') or self.current_location[0]
>             map_lng = request.args.get('lon') or self.current_location[1]
root@ip-172-31-28-74:~/PokemonGo-Map/pogom# 
```

EdenGottlieb:
@Imoq The idea is that this should not work when using -fl. The -fl flag means there is no way to move the scan coordinates of the map.

Imoq:
@EdenGottlieb the SCAN coordinates do NOT move, only the CENTER of the map where the coordinates are passed as argument.

Issue #1618, opened 2016-12-09
Uncaught exception when running runserver.py


HenrySia:
## Current Behavior

> ./runserver.py 
> 2016-12-08 18:07:24,658 [      MainThread][     runserver][   ERROR] Uncaught exception
> Traceback (most recent call last):
>   File "./runserver.py", line 305, in <module>
>     main()
>   File "./runserver.py", line 168, in main
>     str(position[0]), str(position[1]))
> TypeError: 'NoneType' object has no attribute '__getitem__'

## Steps to Reproduce (for bugs)

1. git pull --recursive https://github.com/PokemonGoMap/PokemonGo-Map
2. Check you're on commit c8cb6468ed037c6b0fd8d265a745ce3f72863b95
3. npm install && npm run build
4. Modify config.ini, run ./runserver.py

## Your Environment

* Version used: c8cb6468ed037c6b0fd8d265a745ce3f72863b95
* Environment name and version (e.g. Python 2.7): Python 2.7
* Operating System and version (desktop or mobile): Ubuntu 16.04.1 LTS

springjools:
Step 4 is incomplete: Modify config.ini, run ./runserver.py

Needs more info about what arguments are fed into the command line/config

HenrySia:
Hi @springjools , here's my config.ini. I've been using the same one since last month.
Note: I removed my usernames, password, and gmaps-key.

```
# DO NOT USE NOTEPAD TO EDIT CONFIG FILES!! USE NOTEPAD ++ INSTEAD.
# Authentication settings
#auth-service:          # ptc (default) or google
username: [bla1, bla2, bla3]
password: secretpassword

# Database settings
#db-type: sqlite        # sqlite (default) or mysql
#db-host:               # required for mysql
#db-name:               # required for mysql
#db-user:               # required for mysql
#db-pass:               # required for mysql
#db-port:               # default 3306

# Search settings
location: Cal Anderson Park, Capitol Hill, WA
#no-gyms:               # disables gym scanning (default false)
#no-pokemon:            # disables pokemon scanning (default false)
#no-pokestops:          # disables pokestop scanning (default false)
#scan-delay:            # default 10
#step-limit:            # default 12
#gym-info:              # enables detailed gym info collection (default false)
#min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
#status-name:           # enables writing status updates to the database - if you use multiple processes, each needs a unique value

[...]

PR #1620, opened 2016-12-10
Fix the location check so that it works

springjools:
Fix the location check so that it works

## Description
Check that location is valid before requesting alt

## Motivation and Context
Fix issue #1618 

## How Has This Been Tested?
Tested on my computer for different locations

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -161,7 +161,7 @@ def main():
     else:
         log.debug('Looking up coordinates in API')
         position = util.get_pos_by_name(args.location)
-        
+
     if position is None or not any(position):
         log.error("Location not found: '{}'".format(args.location))
         sys.exit()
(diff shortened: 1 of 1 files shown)


springjools:
image

springjools:
Well, I happened to choose this way, I'm sure it could also be done another way. But the try block checks for that a valid altitude is returned for the position, but it doesn't except for errors associated with the position in the first place, and also in that exception the user is told that an altitude couldn't be found, whereas the error is that the location cannot be parsed. So, they are different checks.

runserver.py
```diff
@@ -162,6 +162,9 @@ def main():
         log.debug('Looking up coordinates in API')
         position = util.get_pos_by_name(args.location)
 
+    if position is None or not any(position):
+        log.error("Location not found: '{}'".format(args.location))
```
sebastienvercammen: Replace double quoted `"string"` with `'single quoted'`, and the error message with a slightly more descriptive one (e.g. `'Could not find location: LOCATION, are you sure you've entered a valid one?'` and it's good for me.

PR #1640, opened 2016-12-19
Make -ps respect log level

smc87:
## Description
switch_status_printer in search.py now respects verbose and very verbose arguments

## Motivation and Context
Good to do.

## How Has This Been Tested?
Local
To test: run -ps switch in combination with -v or -vv switch. Once running switch to log view, and you should see the correct log level.

## Types of changes
Couple of line fix to preserve log level

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -62,7 +62,7 @@ def jitterLocation(location=None, maxMeters=10):
 
 
 # Thread to handle user input.
-def switch_status_printer(display_type, current_page):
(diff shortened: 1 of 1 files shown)


mew905:
Thank you! this will make it much easier to figure out why accounts are failing. With a log that zooms by its extremely difficult to check if an account is even dead, and you have a limited time to find out why. With this, you can immediately see dead accounts with -ps and check the logs to identify the problem :) 👍

PR #1641, opened 2016-12-19
Throw exception if attempting -speed with no_gym and no_pokestop

smc87:
- Early in the code. Could move back to schedulers or elsewhere that anyone thinks more appropriate.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -186,7 +186,7 @@ def main():
     if args.no_gyms:
         log.info('Parsing of Gyms disabled.')
         if (args.scheduler == "SpeedScan") and args.no_pokestops:
-            log.critical("Please remove -nk (no pokestops) or -ng (no gyms) to enable -speed scanning.")
+            log.critical("Please enable pokestop or gym parsing to enable -speed scanning.")
             sys.exit(1)
     if args.encounter:
         log.info('Encountering pokemon enabled.')
--- a/runserver.py
+++ b/runserver.py
@@ -186,7 +186,8 @@ def main():
     if args.no_gyms:
         log.info('Parsing of Gyms disabled.')
         if (args.scheduler == "SpeedScan") and args.no_pokestops:
-            raise Exception("Please remove -nk (no pokestops) or -ng (no gyms) to enable -speed scanning.")
+            log.critical("Please remove -nk (no pokestops) or -ng (no gyms) to enable -speed scanning.")
+            sys.exit(1)
     if args.encounter:
         log.info('Encountering pokemon enabled.')
 
--- a/runserver.py
+++ b/runserver.py
@@ -185,8 +185,8 @@ def main():
         log.info('Parsing of Pokestops disabled.')
     if args.no_gyms:
         log.info('Parsing of Gyms disabled.')
-    if (args.scheduler == "SpeedScan") and args.no_gyms and args.no_pokestops:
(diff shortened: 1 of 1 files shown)


forcewill:
To use speedscan we must scan also gyms or pokestops?

smc87:
Yes, its in the docs, otherwise you get 0/0/0 scans.

WalterMagnum:
Not that this is the place,  but if no gyms / pokestops are found it thinks you are over the speed limit.

MysticalOS:
That explains a few things. where I live there are no pokestops or gyms.

averylarry:
I'm not sure this is actually a requirement.  I'd prefer to leave it as a warning instead of a fatal error and exit.

kevdotcom:
Not everybody does it with arguments... Many people probably turned it off in the config.

You should change the Exception accordingly.

smc87:
Updated the message to be more generic.

averylarry:
@smc87 But there's no reason other than excessive warnings to have gyms or pokestops with -speed.  I think that I *should* be able to run -speed without gyms and pokestops if I want to (which I do, because I don't ever care about gyms or pokestops and I don't want them in my database).  That's why I advocate for a warning instead of a full exception/exit the program.  I'd even suggest a warning with some type of "press any key to acknowledge that -speed works better with gyms or pokestops" prompt.

I think a warning is good.  I think the rest of it should be fixed by modifying the -speed logging/warning messages.

sebastienvercammen:
@averylarry I was told on Discord by participating members in @Artifice's original speed PR that it requires either gyms or pokéstops because it bases its definition of a "good scan" on the result of those gym/pokéstop scans.

From Artifice's PR:

```
Speed Scheduler uses visible Gyms and Pokestops to determine if a scan is valid. Try adding gyms and stops back into your scan.
```

If you want to review speed scanning and how gyms/pokéstops relate to them (to see if everything works properly without them), you're welcome to. If it does work perfectly, include @Artifice and me in your mentions, because it means I've gotten wrong information from the author and something needs to be changed.

TheNewDarkOne:
Ah that explains why it was pretty useless when I ran it. It was literally dropping VALID spawn points because it couldn't figure out what was going on. I am rural and there are LITERALLY no pokestops or gyms near me. If it's using them to determine if a scan is valid then it will certainly have no idea what's going on where I live. I honestly just need it to scan everything every 15 minutes until it finds all spawn points, period. it should just do a good ole beehive to get spawn points, THEN start rescanning spawn points to get TTH. all in all i feel -speed was probably the worst of the 3 PTRs they could have merged for valid TTH. spawn points.

averylarry:
models.py lines 1457 - 1467 should be changed to include nearby_pokemon.  It should not speculate regarding -ng and -nk since it can actually lookup the arguments.  Lines 1469 - 1471 should be merged into 1457 - 1467 to provide a single clear warning message regarding 0/0/0 scans.

averylarry:
PR #1662 alleviates the misleading/confusing logging and reduces (eliminates?) speed scanning's reliance on gyms/pokestops.

smc87:
Closed in favour of: https://github.com/PokemonGoMap/PokemonGo-Map/pull/1662
Which  actually fixes the problem, rather than preventing running with the desired options.

PR #1642, opened 2016-12-20
Fix the number of db connections

onilton:
## Description
It was creating a LOT of connections when multiplying by the
number of accounts, we should multiply by the number of workers
(and that is actually the number of connections needed)

## Motivation and Context
Avoid a exploding number of db connections

## How Has This Been Tested?
In my own map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -46,8 +46,8 @@ def init_database(app):
     if args.db_type == 'mysql':
         log.info('Connecting to MySQL database on %s:%i', args.db_host, args.db_port)
         connections = args.db_max_connections
-        if hasattr(args, 'accounts'):
-            connections *= len(args.accounts)
+        if hasattr(args, 'workers') and args.workers > 0:
+            connections *= args.workers
         db = MyRetryDB(
             args.db_name,
             user=args.db_user,


Arengo:
Good find, could be modified a bit to adapt to users' configurations like
````
if hasattr(args, 'accounts'):
connections *= len(args.accounts)
elif: hasattr(args, 'workers') and args.workers > 0:
        connections *= args.workers
else: connections = args.db_max_connections
```

onilton:
@Arengo this is already done:

      connections = args.db_max_connections
      if hasattr(args, 'workers') and args.workers > 0:
         connections *= args.workers

Remember that --db-max_connections is actually  "Max connections (per thread) for the database."

Arengo:
@onilton you took out the `args.account` but that will result in db_max_connections for users which don't call '-w' in their process!? That's why I re-added this if-statement

onilton:
No, because  in pogom/utils.py:

       # Make max workers equal number of accounts if unspecified, and disable account switching.                                                                       
            if args.workers is None:
                args.workers = len(args.accounts)
                args.account_search_interval = None

pogom/models.py
```diff
@@ -46,8 +46,8 @@ def init_database(app):
     if args.db_type == 'mysql':
         log.info('Connecting to MySQL database on %s:%i', args.db_host, args.db_port)
         connections = args.db_max_connections
-        if hasattr(args, 'accounts'):
-            connections *= len(args.accounts)
+        if hasattr(args, 'workers') and args.workers > 0:
+            connections *= args.workers
```
smc87: I'm abit confused at the logic here. Its not actually counting the number of workers, you instead are counting tuples inside the workers object and that list seems to be repeated x5 and then those connections are not always utilized as far as I can see.

onilton: I am not sure if I understood write you wrote :stuck_out_tongue:  Sorry. :disappointed: 

But I will try to explain it anyway.... Forgive me if sound stupid, if I say anything obvious or anything else... I was confused reading this code at first too. As you pointed out, I kept the previous logic, just fixed what was wrong.

First, '--db-max_connections'  is 'Max connections (per thread) for the database.' (default=5). If you think you will have a thread for each worker, than it is also the max db connections by worker, that's why it is multiplied. 

`args.workers` is an int parameter with the number of workers, I don't need to count them, it is already the integer I need.

The final question: Why I would need more connections per worker/thread? Despite most of the writes operations being done through que db_queue, the read operations like the selects for spawnpoints or pokemon info are done through those db connections.

This final part I can't explain very well because I don't know pewee very much and the rest of the code, but anyway I discovered it through experimentation: 

These read operations sometimes are intensive and although in theory I would only need one single connection per thread/worker, this setup failed for me.


The logic here is for each worker

shockler:
I got error while using -w 1

smc87:
What error?

pogom/models.py
onilton: As I said in the end, I don't get this also :smile: But I know from my tests that if I put one connection per worker PGM will run out of connections from the pool almost instantly.

onilton:
@shockler and what is the value of --db-max_connections ? The default one?

pogom/models.py
smc87: Ah ok, sorry i wasn't sure if you understood what i was saying.
I would imagine the reason the pool is exhausted immediately if connections = worker number because you need a connection for the db-updater and wh-updater (maybe others too?).

Imoq:
Even if I'm not a developer, I believe I'll pull this one tonight to test, as the "increase threads via --db-threads" error message is driving me nuts, have gone up to --db-threads 50 (starting from 2, 5, 10, etc.) until it went away. I believe current -speed implementation has serious issues with DB connections.

pogom/models.py
onilton: Yeah I thought about that and I created something like connections = num_workers + CONSTANT (1,2,3,4). But I still got no connection available errors with this :disappointed: . Seems like the workers right now for some reason demand more than one connection...

shockler:
@smc87 @onilton "ValueError: Exceeded maximum connections." after that, my python crashs.
If I use more than 1 worker, it's ok.

onilton:
@shockler And... what about --db-threads ? 

With 12 db-max_connections this shouldn't be an issue...   =/

Also, are you using the main develop or any other pulls/branchs? 

If db-threads is not the issue, I think we can't just multiply by workers, I think we need a minimum number of connections before multiplying...

pogom/models.py
jaake77: Curious if instead of workers x 5 you need workers x2 + 3 or w/e. Sorry I'm learning a lot.

smc87:
@sebastienvercammen Try logging out the value of `connections` and see if it is the same as the int arg.workers - I expected this for args.accounts but not args.workers

drallieiv:
I am looking forward to this

Alderon86:
This would be pretty neat if its correct implemented! looking forward to use!

onilton:
@kevdotcom  No, its not simply  `connections *= args.db_max_connections* args.workers * args.workers_per_hive`

I think you got wrong what wph means. The total workers are not `-w` * `-wph`. Actually wph is how you want to divide workers per hive. The total workers are still the number you provide in `-w`

Anyway nince that you mentioned it, since for each hive we spin a new scheduler and the scheduler needs an db connection we need to account the number of schedulers in our connections calc

onilton:
@shockler Can you provide me your complete `./runserver.py` args?

onilton:
@sebastienvercammen Ok. Something that I know:

We need to align some of the terms you used because it was confusing in my mind, in PGM utils we have:


    parser.add_argument('--db-max_connections', help='Max connections (per thread) for the database.',                                                                   
                        type=int, default=5)
    parser.add_argument('--db-threads', help='Number of db threads; increase if the db queue falls behind.',
                        type=int, default=1)

That's why I didn't get your `(-w * --db-threads)`. Anyway, with this is mind, a possible solution (or a point of start) could be something like:

     -w * --db-max_connections + --db-threads  + 1 (for clean_db_loop) +1 (for overseer/scheduler)

I would rather not go for (the safe) `+modifier` since I think that's what db-max_connections default of 5 is already about.
    
I always saw `--db-max_connections` (per-thread) as directly related to workers, but if we want we could also make it the "per-thread" literally for all threads not just worker ones:

    (-w + db-threads) * db-max-connections + 1 (for clean_db_loop) +1 (for overseer/scheduler)

I am still investigating what else needs a dedicated connection, if you guys remember anything please comment here.

sebastienvercammen:
@onilton Your right, it can definitely get confusion, especially because some config items aren't very clear in their definitions (some of them come from a time before we ran multiple instances).

To clarify/reiterate:

- An _instance_ is a python process.
- `-w` is the number of search worker threads for an instance.
- `--db-threads` is the number of database threads **per search worker thread**.
- `--db-max_connections`  is **not** per search worker (or per database) thread, it is _per instance_. The documentation in utils.py is wrong. As you will see in [models.py](https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/models.py#L45), max connections is used in `init_database`, which is called only by `runserver.py` (main function).

The confusion with `--db-max_connections` arises from the original calculation of total database connections to use:
```
        connections = args.db_max_connections
        if hasattr(args, 'accounts'):
            connections *= len(args.accounts)
```
in this context, the wording _"Max connections (per thread)"_ refers to the fact that they multiply by the number of accounts (which in this implementation happens to be the number of search worker threads, they used one thread for every account). But the actual number of _maximum connections_ is set _per instance_.

That's why, to be clearer, we should use _max connections_ only in relation to _instances_ rather than _threads_. The wording in utils.py should also be changed.

This is also why I said the amount of maximum DB connections should be at least `(-w * --db-threads)` to support the _"ideal situation"_ mentioned in my comment above.

sebastienvercammen:

PR #1645, opened 2016-12-21
Altitude

onilton:
# Description
This is a continuation of #1591. I picked where he left, rebased with current develop, fixed some issues, and I will try to improve the solution to make it more likely to be merged

From the previous description:

> Using Google Elevation API to receive an actual altitude for a scanned area. If successful, every consecutive call for the entire map instance will use the given altitude and jitter the value. If unsuccessful, the default value will be used.
>

https://console.developers.google.com/apis/api/elevation_backend/overview **is required**
Following commands can be used:

--altitude
you can set the default altitude in meter, preset is 13m and a floating 0-1 meter will be added.
--altitude-variance
range in meter can be added to the normal altitude, e.g. with 13m default and 1 meter range a random altitude between 12m and 14m gets generated.
--no-altitude-cache
Do not cache fetched altitudes in the database. This implies fetching the altitude only once for the running instance.

What I am planning to do:
* Make Altitude work with Speed Scheduler now (DONE)
* Isolate the previous duplicated code in utils.py (DONE)
* Store previous requested altitudes to avoid more api calls (DONE)
* Rethink the parameters/arguments? (DONE)

## How Has This Been Tested?
In my local map.

## Types of changes

## Checklist:

- [X ] I have updated the documentation accordingly.

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -59,8 +59,9 @@ from operator import itemgetter
 from datetime import datetime, timedelta
 from .transform import get_new_coords
 from .models import (hex_bounds, Pokemon, SpawnPoint, ScannedLocation,
-                     ScanSpawnPoint, LocationAltitude)
+                     ScanSpawnPoint)
 from .utils import now, cur_sec, cellid, date_secs, equi_rect_distance
+from .altitude import get_altitude
 
 log = logging.getLogger(__name__)
 
@@ -275,7 +276,7 @@ class HexSearch(BaseScheduler):
         # Add the required appear and disappear times.
         locationsZeroed = []
         for step, location in enumerate(results, 1):
-            altitude = LocationAltitude.get_altitude_by_loc(location)
+            altitude = get_altitude(self.args, location)
             locationsZeroed.append(
                 (step, (location[0], location[1], altitude), 0, 0))
         return locationsZeroed
@@ -434,8 +435,8 @@ class SpawnScan(BaseScheduler):
         # locations = [((lat, lng, alt), ts_appears, ts_leaves),...]
         retset = []
         for step, location in enumerate(self.locations, 1):
-            altitude = LocationAltitude.get_altitude_by_loc([location['lat'],
-                                                             location['lng']])
+            altitude = get_altitude(self.args, [location['lat'],
+                                    location['lng']])
             retset.append((step, (location['lat'], location['lng'], altitude),
                            location['appears'], location['leaves']))
 
@@ -602,7 +603,7 @@ class SpeedScan(HexSearch):
 
         generated_locations = []
         for step, location in enumerate(results):
-            altitude = LocationAltitude.get_altitude_by_loc(location)
+            altitude = get_altitude(self.args, location)
             generated_locations.append(
                 (step, (location[0], location[1], altitude), 0, 0))
         return generated_locations
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -54,6 +54,8 @@ import sys
 from copy import deepcopy
 import traceback
 from collections import Counter
+import random
+import requests
 from queue import Empty
 from operator import itemgetter
 from datetime import datetime, timedelta
@@ -64,10 +66,11 @@ from .utils import now, cur_sec, cellid, date_secs, equi_rect_distance
 
 log = logging.getLogger(__name__)
 
-
 # Simple base class that all other schedulers inherit from.
 # Most of these functions should be overridden in the actual scheduler classes.
 # Not all scheduler methods will need to use all of the functions.
+
+
 class BaseScheduler(object):
 
     def __init__(self, queues, status, args):
@@ -130,8 +133,8 @@ class BaseScheduler(object):
                 step_location[0], step_location[1], remain),
             'late': 'Too late for location {:6f},{:6f}; skipping.'.format(
                 step_location[0], step_location[1]),
-            'search': 'Searching at {:6f},{:6f}.'.format(
-                step_location[0], step_location[1]),
+            'search': 'Searching at {:6f},{:6f},{:6f}.'.format(
+                step_location[0], step_location[1], step_location[2]),
             'invalid': ('Invalid response at {:6f},{:6f}, ' +
                         'abandoning location.').format(step_location[0],
                                                        step_location[1])
@@ -157,6 +160,8 @@ class BaseScheduler(object):
 # Hex Search is the classic search method, with the pokepath modification,
 # searching in a hex grid around the center location.
 class HexSearch(BaseScheduler):
+    elevation = False
+    altitude = 0
 
     # Call base initialization, set step_distance.
     def __init__(self, queues, status, args):
@@ -170,7 +175,9 @@ class HexSearch(BaseScheduler):
             self.step_distance = 0.070
 
         self.step_limit = args.step_limit
-
+        self.gmaps = args.gmaps_key
+        self.altitude_range = args.altitude_range
+        self.altitude_default = args.altitude
         # This will hold the list of locations to scan so it can be reused,
         # instead of recalculating on each loop.
         self.locations = False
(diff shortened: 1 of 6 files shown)


kevdotcom:
We could also make gmaps api calls for all steps at the start of scanning and save them for later use in the database.

mew905:
do you have a before and after 2captcha graph to back up a 70-80% drop?? I hard-coded my altitude to 550m (an approximation of my area's altitude) but I havent really noticed much. Empirical data would be great to back up such a large claim

onilton:
@kevdotcom That's the idea, it is already planned, you can see this in the description.
@hoyeeleo Oh, didn't noticed that, you are right, I will try to fix it.

onilton:
@mew905 Sorry mew905, this was from the old pull request, I just copied it. I don't have data to backup this claim. I will remove that sentence.

onilton:
@hoyeeleo I checked what you said and got to the conclusion that it isn't needed.  `schedule()` uses `self.scan` location info that is filled in `location_changed()` that itself always calls `self._generate_locations()`, the method that I changed. :slightly_smiling_face:  Makes sense?

hoyeeleo:
@onilton I tested it and you are right! However `SpeedScan.altitude = altitude` and `altitude = get_altitude(...)` should have their order switched. Otherwise the altitude is not stored. Same for `HexSearch` and `SpawnScan`.

onilton:
@hoyeeleo I think the current code is right, take a second look. Anyone don't pay attention in this too much since it will be removed in my next commits.... (spoiler) :smile:

quixfz:
i just got my own elevation api running. currently only data for switzerland is avaible, but i'm working on the whole world. seeing that db-caching is working now, i intend to open it up for pogo-map users.

here's some info (i'm using the SRTMGL1 dataset):
http://wiki.openstreetmap.org/wiki/SRTM
http://gis.stackexchange.com/questions/43743/how-to-extract-elevation-from-hgt-file

todo: error-checking, bilinear-interpolation, add data from whole world

here's the api: http://pg.festzeit.ch/alt.php?lat=46.2&lng=7.9

easily changed in the pogomap code. the other possible way would be to add a hgt parser to this code and so everyone only needs to download the needed hrt file.

mew905:
I appreciate what youre doing quizfx, if we only need approximations for altitudes then we could use yours. But niantic could also be looking for exact altitude values, since they use gmaps for that, and i have seen variance between services, we might be best off to use the gmaps api, as limited as it is

quixfz:
vertical altitude output of gps is at best precise to a few meters (much worse the horizontal position). no point in having 100% exact values. google maps will also often use srtm data.

themastrdudeh:
Trying to run this in centos but i just keep getting "Unable to retrieve altitude from Google APIs; setting to 0", it works from my windows test machine.

Any clues?

quixfz:
you're probably hitting the google api limit (2500 requests per day)...

themastrdudeh:
@quixfz 
I don't think i am since i have no problems running it on my windows machine.

Win:
```
12-23 18:28:38       MainThread  root           DEBUG    Using coordinates from CLI directly
12-23 18:28:38       MainThread  requests.packages.urllib3.connectionpool INFO     Starting new HTTPS connection (1): maps.googleapis.com
12-23 18:28:38       MainThread  requests.packages.urllib3.connectionpool DEBUG    "GET /maps/api/elevation/json?locations=12.3456,34.5678 HTTP/1.1" 200 161
12-23 18:28:38       MainThread  root           DEBUG    Local altitude is: 12.1234567891m
12-23 18:28:38       MainThread  root           INFO     Parsed location is: 12.3456/34.5678/12.1234 (lat/lng/alt)
12-23 18:28:38       MainThread  root           INFO     Encountering pokemon enabled.

```
Lin:
```
12-23 18:27:45       MainThread  root           DEBUG    Using coordinates from CLI directly
12-23 18:27:45       MainThread  requests.packages.urllib3.connectionpool INFO     Starting new HTTPS connection (1): maps.googleapis.com
12-23 18:27:45       MainThread  root           ERROR    Unable to retrieve altitude from Google APIs; setting to 0
12-23 18:27:45       MainThread  root           INFO     Parsed location is: 12.3456/34.5678/0.0000 (lat/lng/alt)
12-23 18:27:45       MainThread  root           INFO     Encountering pokemon enabled.
```

Obihoernchen:
@themastrdudeh I bet you miss the proper CAs. Try to install python-certifi.

forcewill:
@mew905 I believe that even if niantic is looking for altitude they will have to have a delta variation of error since low end GPS  aren't that accurate from what I read http://stackoverflow.com/questions/9361870/android-how-to-get-accurate-altitude
I'm really anxious to test this PR will do when I come back from vacations

neskk:
I liked the previous version where you didn't  store anything in the database at all and just did one request per hex.
Current implementation has some problems, specially when search location is not fixed:

1. Initialization time can be long if step-size is big, since it has to request Elevation for each step-location.
A single -st 10 instance will require 54 API calls.

2. Extra overhead for the database: each time the schedulers regenerate the search item queue it will query the database for all those locations again. Plus, LocationAltitude data adds redundancy to db because locations are also being stored in ScannedLocation.

3. Whenever you change the coordinates you want to scan it will trigger the whole elevation API request cycle, since most likely you'll never position the marker twice in the same exact spot, causing all queries to fetch previous altitude to fail, then new requests to the API, then more queries to insert the results.

I vote for reversal of last commit.
Commit: 6643255b9c480cbed4e20be39d677043c20f4e28 has worked perfectly for me, provides a decent aproximation to the "real" altitude for the zone, it's not expensive for the database, it doesn't make more that one request to Gmaps API for each hive.

Spurlex:
http://api.pokehuntr.com/api/elevation?coords=34.009376287758926,-118.49720478057863

Anyone can use this API if you would like :)

sebastienvercammen:
Removed the link competition in here: this is a PR for PokémonGo-Map, not an advertising website for everyone's "custom elevation API".

onilton:
@neskk I won't revert to that commit, sorry. The last commit is not only about storing altitude, I refactored a lot of things that were ugly/repeated in the previous version

People already discussed this on Discord, it is a known issue, I already known what I've got do, but didn't worked on it yet.

Anyway I also plan to allow people to just not use gmaps altitude, so they can just pass the --altr and --altr-range by theirselves

onilton:
TO-DO (just a reminder)

* replace altitude-range  with variance (Seb's suggestion)
* avoid too many db calls
* use nearby altitudes

neskk:
@onilton the commit I referenced was after most the refactoring you've done and just before the database changes. How do you propose to solve all those things?

onilton:
Trying to do some of the requests in batch,
Also using nearby altitudes would help with both location changes and db redundancy

neskk:
Maybe if `get_alt_from_loc(location)` did some kind of math to lookup altitude from an area, let's say, a square that is centered around `location` it would solve the "API requests" problem, because small adjustments to `search_location` wouldn't trigger another round of API requests. But the query is probably gonna be more expensive and can lead to slow-downs when refreshing the search_item_queue.
I'm glad to help and discuss ideas for this but I don't see a "easy way out"  for some of the problems I've referred, meaning implementation will be complex.

pogom/schedulers.py
```diff
@@ -276,7 +275,9 @@ def _generate_locations(self):
         # Add the required appear and disappear times.
         locationsZeroed = []
         for step, location in enumerate(results, 1):
-            locationsZeroed.append((step, (location[0], location[1], 0), 0, 0))
+            altitude = LocationAltitude.get_altitude_by_loc(location)
```
sebastienvercammen: Missing check if calls to DB are disabled.

pogom/schedulers.py
```diff
@@ -599,8 +600,12 @@ def _generate_locations(self):
                 loc = get_new_coords(loc, xdist, WEST)
                 results.append((loc[0], loc[1], 0))
 
-        return [(step, (location[0], location[1], 0), 0, 0)
-                for step, location in enumerate(results)]
+        generated_locations = []
+        for step, location in enumerate(results):
+            altitude = LocationAltitude.get_altitude_by_loc(location)
```
sebastienvercammen: Missing check if calls to DB are disabled.

pogom/utils.py
```diff
@@ -94,6 +95,17 @@ def get_args():
                         type=int, default=1)
     parser.add_argument('-l', '--location', type=parse_unicode,
                         help='Location, can be an address or coordinates.')
+    parser.add_argument('-alt', '--altitude',
+                        help='Default altitude in meters.',
+                        type=int, default=13)
+    parser.add_argument('-altv', '--altitude-variance',
+                        help='Variance for --altitude in meters',
+                        type=int, default=1)
+    parser.add_argument('-naltdb', '--no-altitude-db-cache',
```
onilton: But we are still caching in memory, right?

That's why I added the db thing. 

But, I can remove it, no problem.

onilton:
@jaake77 fixed in the last commit

PR #1654, opened 2016-12-30
add default notify and excluded moons for web

catnet23:
add support to set exclude and notify pokemons in static/js/maps.js
just add your exclude or notfy pokedex id to the variable.

as example, change:
static/js/map.js line 27: var excludedPokemon = []
to:
var excludedPokemon = [16, 19]

same with:
static/js/map.js line 28: var notifiedPokemon = []
to
var notifiedPokemon = [131, 143, 149]

pr_map
pr_map1
pr_map2

don't forgott after change

$ npm run buld
and restart the web process

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -66,23 +66,23 @@ var audio = new Audio('static/sounds/ding.mp3')
 // Functions
 //
 
-function set_default_value(exclude, notify) {
-  if (exclude) { 
-    Store.set('remember_select_exclude', get_localstore_value('remember_select_exclude')) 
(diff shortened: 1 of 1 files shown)


kevdotcom:
You can already change all the defaults in map.common.js

`  'remember_select_exclude': {
    default: [13,16,19,21,96],
    type: StoreTypes.JSON
  },
  'remember_select_notify': {
    default: [143,149],
    type: StoreTypes.JSON
  },`

catnet23:
thanks @kevdotcom didn't saw that, a user aks me for that.

PR #1656, opened 2016-12-31
Modify bad scan

averylarry:
## Description
Added sightings/nearby Pokémon to the logic for identifying a bad scan.
Changed the way a speed violation is logged and processed.

## Motivation and Context
We can tell more about 0/0/0 scans (bad scan, speed violation, nearby Pokémon).  Let's report on them properly.  Works nicely with the new nearby Pokémon count added to the logging.

## How Has This Been Tested?
Running on my small Windows deployment.

## Types of changes

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1480,7 +1480,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
             # a speed violation.
             log.warning('No nearby or wild Pokémon but there are visible gyms or pokestops. Probable speed violation.')
             #
-            # If we're not parsing pokestops or gyms and we want a probably speed violation to be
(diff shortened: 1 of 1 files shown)


averylarry:
1) Need someone who knows -speed really well to double check that the "bad scan" logic is still sound.
2) Need to decide if a probable speed violation should be treated as a "bad scan" for -speed.

mew905:
I thought speed scan already did this? It's where I got my idea for using the nearby list in 1588 speed limiter before 1501 was merged. Or did it not work properly?

averylarry:
Speed scan does not use nearby_pokemon when determining a bad scan.  It only uses wild and forts.  I added nearby to the logic.
Rebased to PR  1662.

PR #1660, opened 2017-01-01
Fix for the annoying infobox when swiping on touch/mobile device

ultrafunkamsterdam:
On mobile devices, especially on larger/denser maps, zooming or swiping often results in multiple info-modals. I think this is annoying. 
If you want information about a pokemon,gym or pokestop you should be able to open info modal by just tapping it.  

## Description
Modified map.js
Wrapped mouseover event listener in if-statement :  if ( ! isMobileDevice && ! isTouchDevice )
Could easily be reversed by removing the if statement in case you absolutely need the popups on swipe so hard.

## Motivation and Context

It's not required, but solves a lot of irritation and phone trhow-aways.

## How Has This Been Tested?
'implemented' it in my map and tested it on iPhone6 (chrome,safari), Galaxy S7 (chrome, samsung stock browser). All working fine. Also on desktop it works fine.

## Types of changes

- [  ] New feature (non-breaking change which adds functionality)
- [ x ] Breaking change (fix or feature that would cause existing functionality to change)
- [ x ] Fix aggression issues ;)

## Checklist:


ultrafunkamsterdam:
mistake

PR #1662, opened 2017-01-01
Modify bad scan and logging to include nearby_pokemon

averylarry:
## Description
Added sightings/nearby Pokémon to the logic for identifying a bad scan.
Changed the way a speed violation is logged and processed.
## Motivation and Context
We can tell more about 0/0/0 scans (bad scan, speed violation, nearby Pokémon). Let's report on them properly. Works nicely with the new nearby Pokémon count added to the logging.
## How Has This Been Tested?
Running on my small Windows deployment.
## Types of changes

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -24,9 +24,9 @@ from cachetools import TTLCache
 from cachetools import cached
 
 from . import config
-from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, get_args, \
-    cellid, in_radius, date_secs, clock_between, secs_between, get_move_name, get_move_damage, \
-    get_move_energy, get_move_type
+from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, \
+    get_args, cellid, in_radius, date_secs, clock_between, secs_between, \
+    get_move_name, get_move_damage, get_move_energy, get_move_type
 from .transform import transform_from_wgs_to_gcj, get_new_coords
 from .customLog import printPokemon
 log = logging.getLogger(__name__)
@@ -44,7 +44,7 @@ class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
 
 def init_database(app):
     if args.db_type == 'mysql':
-        log.info('Connecting to MySQL database on %s:%i',
+        log.info('Connecting to MySQL database on %s:%i...',
                  args.db_host, args.db_port)
         connections = args.db_max_connections
         if hasattr(args, 'accounts'):
@@ -58,7 +58,7 @@ def init_database(app):
             max_connections=connections,
             stale_timeout=300)
     else:
-        log.info('Connecting to local SQLite database')
+        log.info('Connecting to local SQLite database...')
         db = SqliteDatabase(args.db)
 
     app.config['DATABASE'] = db
@@ -81,7 +81,7 @@ class BaseModel(flaskDb.Model):
 
 
 class Pokemon(BaseModel):
-    # We are base64 encoding the ids delivered by the api,
+    # We are base64 encoding the ids delivered by the api
     # because they are too big for sqlite to handle.
     encounter_id = CharField(primary_key=True, max_length=50)
     spawnpoint_id = CharField(index=True)
@@ -101,18 +101,19 @@ class Pokemon(BaseModel):
         indexes = ((('latitude', 'longitude'), False),)
 
     @staticmethod
-    def get_active(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None, oSwLng=None, oNeLat=None, oNeLng=None):
+    def get_active(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None,
+                   oSwLng=None, oNeLat=None, oNeLng=None):
         now_date = datetime.utcnow()
-        # now_secs = date_secs(now_date)
         query = Pokemon.select()
         if not (swLat and swLng and neLat and neLng):
             query = (query
                      .where(Pokemon.disappear_time > now_date)
                      .dicts())
         elif timestamp > 0:
-            # If timestamp is known only load modified pokemon.
+            # If timestamp is known only load modified Pokemon.
             query = (query
-                     .where(((Pokemon.last_modified > datetime.utcfromtimestamp(timestamp / 1000)) &
+                     .where(((Pokemon.last_modified >
+                              datetime.utcfromtimestamp(timestamp / 1000)) &
                              (Pokemon.disappear_time > now_date)) &
                             ((Pokemon.latitude >= swLat) &
                              (Pokemon.longitude >= swLng) &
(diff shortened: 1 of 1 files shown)


averylarry:
1. Need someone who knows -speed really well to double check that the "bad scan" logic is still sound.
2. Need to decide if a probable speed violation should be treated as a "bad scan" for -speed. It wasn't previously and I didn't make it that way but I think it should be. No nearby and no wild but there are forts = bad scan (it's not necessarily a speed violation, but it's very likely). I have commented out code in 2 places that would accomplish this.

TheNewDarkOne:
Where I live, it's often possible for there to be 0 pokestops 0 gyms 1 pokemon with 0 nearby. to put in perspective. I had a 30 step beehive that ran for days and found less than 60 spawn points. i'm a rural as rural can get. which is why playing game is near impossible without scanners cause you'd wander around for hours trying to find stuff. On the plus, where I live I don't have to see that god awful pokestop tracker. On the neg, my old tracker is usually empty or has 1 pokemon at most in it and figuring out where it is without scanner was impossible until mapping out spawn points. After that it's trial an error.

TLDR, it's so hard to tell  violations from there really just being nothing there where I am and that sucks :( I miss 15 min TTH

vlalo001:
This fixed my problem for many of my instances that were getting "0/0/0 Bad Scan". Thank you SO MUCH!!!!

averylarry:
@TheNewDarkOne Have you thought about using spawnpoint scanning?  You can dump the spawnpoints to a file, then it's really important to compress the file (using the clustering utility in Tools).  Finally, use #1476 to use spawnpoint scanning with a speed limit.  That's how I run my map.  Speed scanning just has a few limitations that I don't like.

averylarry:
> Also needed to readd ScannedLocation.visible_forts(step_location) to the logic on #1662. This is necessary for the scan to find forts within the 1000m radius and mark the scan as good. It should be added back in IMO @averylarry

In what specific way?  Finding visible forts would only be useful in a full 0/0/0/0 situation, which I left as a "bad scan".  Adding visible_forts would leave you with a bad scan if True and then ... what do we do with False?  That's a larger question for the overall -speed scanning logic I did not want to tackle with this PR.  Therefore I left that scenario as a "bad scan", which is not different from the way it was.

In other words -- I removed a set of scans that were definitely NOT bad from the "bad scan" status AND I made the logging more accurate (and I removed a DB call).  I did NOT change any other existing -speed scanning logic (like what to do with a legitimate 0/0/0/0 scan or what to do with a 0/0/N/N probable speed violation).

cydine:
You're right. I hadn't considered what would happen if visible_forts was false. My bad!

This is where max retries would be beneficial:
0/0/0/1 - Bad scan. Keep retrying
0/0/0/0, 0/0/1/0, 0/0/1/1 - Max x retries before scan is marked as good with no spawns.

Mr-Groch:
def visible_forts(step_location) in pogom/models.py need to be updated according to API scan radius change - 500m (distance = 0.5). IMHO this should be first step needed for fixing bad scans problem. If no pokemon nearby and all forts are outside scan range - this visibility fix causes problems when distance not synced with API. Referring to closed #1685 (I think that it was closed too fast).

Imoq:
I started getting 0/0/0 on latest update, and I wasn't getting any before. I was using latest develop branch but hadn't updated pgoapi in a while (only used to do git pull updates). Very strange that I only get the 0/0/0 on 3 specific locations (-st 5) with *specific accounts*.

Anyway, was trying to test if this solves the issue, but it looks like it isn't compatible with latest develop to use 0.51 API --hash-key, I'll test (and hoping this solves them) whenever this PR gets updated.

Thanks!

averylarry:
rebase
pep8
grammar

averylarry:
I renamed pokemons to pokemon in the code except for lines 1549, 1554.
I changed
if len(something):
    do something
to
if something:
    do something
about 10 times.
Lines 1549 - 1583 and 1901 - 1916 is the bulk of the actual new code regarding bad scan logic.
Pretty much everything else is reducing line length to 79 and grammar modification for log lines and comments.

averylarry:
Just seeing if Travis will run now.

nullpixel1:
Yeah, works good for me. +1 👍

neskk:
Can't even see clearly what actual code logic changed with all the cosmetic changes mixed up... *sigh*

sebastienvercammen:
@neskk The "cosmetic changes" are to apply PEP8. It was an intentional decision to apply it over time so we could start guaranteeing code quality and readability.

I agree a separate commit would have helped, but we work with what we have.

You might not be used to PEP8 or working with other people, but you have PR rank now on our Discord (given to you by Fox). I expect better from you.

Some respect for other people's work is a minimum.

pogom/models.py
```diff
@@ -1473,67 +1525,82 @@ def hex_bounds(center, steps=None, radius=None):
...
+        # we'd like to know whether or not *any* were found to help determine
+        # if a scan was actually bad.
+        wild_pokemon += cell.get('wild_pokemons', [])
+
+        forts += cell.get('forts', [])
+
+    # If there are no wild or nearby Pokemon . . .
+    if not wild_pokemon and not nearby_pokemon:
+        # . . . and there are no gyms/pokestops then it's unusable/bad.
+        if not forts:
```
sebastienvercammen: Why was `and ScannedLocation.visible_forts(step_location)` removed here and not replaced by anything?

averylarry: Because the scan is unreliable and should be rescanned whether or not there should have been forts.
A 0/0/0/0 scan with no expected visible forts -- we don't know if it's valid or speed limit or bad.  So we rescan it 3 times (I think from the other PR).
A 0/0/0/0 scan WITH expected visible forts -- we know it's bad.  So we rescan it 3 times (I think from the other PR).

So we do the same thing either way.  The only reason to bring visible_forts back into the equation is if you pass 3 states back to the speed scan logic in schedulers.py.  Known good, known bad, and unknown.  Right now the speed scan logic only uses 2 states, good and bad.

averylarry: Because the scan is unreliable and should be rescanned whether or not there should have been forts.
A 0/0/0/0 scan with no expected visible forts -- we don't know if it's valid or speed limit or bad.  So we rescan it 3 times (I think from the other PR).
A 0/0/0/0 scan WITH expected visible forts -- we know it's bad.  So we rescan it 3 times (I think from the other PR).

So we do the same thing either way.  The only reason to bring visible_forts back into the equation is if you pass 3 states back to the speed scan logic in schedulers.py.  Known good, known bad, and unknown.  Right now the speed scan logic only uses 2 states, good and bad.

averylarry: Because the scan is unreliable and should be rescanned whether or not there should have been forts.
A 0/0/0/0 scan with no expected visible forts -- we don't know if it's valid or speed limit or bad.  So we rescan it 3 times (I think from the other PR).
A 0/0/0/0 scan WITH expected visible forts -- we know it's bad.  So we rescan it 3 times (I think from the other PR).

So we do the same thing either way.  The only reason to bring visible_forts back into the equation is if you pass 3 states back to the speed scan logic in schedulers.py.  Known good, known bad, and unknown.  Right now the speed scan logic only uses 2 states, good and bad.

averylarry:
Because the scan is unreliable and should be rescanned whether or not there should have been forts.
A 0/0/0/0 scan with no expected visible forts -- we don't know if it's valid or speed limit or bad.  So we rescan it 3 times (I think from the other PR).
A 0/0/0/0 scan WITH expected visible forts -- we know it's bad.  So we rescan it 3 times (I think from the other PR).

So we do the same thing either way.  The only reason to bring visible_forts back into the equation is if you pass 3 states back to the speed scan logic in schedulers.py.  Known good, known bad, and unknown.  Right now the speed scan logic only uses 2 states, good and bad.

PR #1670, opened 2017-01-02
Ignore bad scans

birdstream:
## Description

Added an option to disable checking for bad scans
## Motivation and Context

If scanning a rural area or any place where there may be a scarcity of pokestops and gyms, some scan locations may never produce a good scan (no pokestops or gyms close enough, and no spawns). This is problematic for speedscan where you may end up having a situation where workers gets trapped trying to rescan the same locations over and over and the map just dies. This is just a dirty fix and should probably never be merged. The tradeoff is naturally some spawns may get missed. But if you are confident you're not getting softbans it should work pretty well. I suppose there could be some better logic to this, like just trying to rescan for a set amount of times and then give up. I'm not good with coding so if anyone would like to help out, I'd appreciate it :)

## How Has This Been Tested?

On my own map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1456,7 +1456,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
 
     # Check for a 0/0/0 bad scan
     # If we saw nothing and there should be visible forts, it's bad
-    if not len(wild_pokemon) and not len(forts) and ScannedLocation.visible_forts(step_location):
+    if not len(wild_pokemon) and not len(forts) and ScannedLocation.visible_forts(step_location) and not args.ignore-bad-scans:
         log.warning('Bad scan. Parsing found 0/0/0 Pokemon/pokestops/gyms')
         log.info('Common causes: captchas, IP bans, or using -ng and -nk arguments')
         return {
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1456,7 +1456,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
 
     # Check for a 0/0/0 bad scan
     # If we saw nothing and there should be visible forts, it's bad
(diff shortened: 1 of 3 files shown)


averylarry:
I think #1662 will fix your problem, or at least help it.

birdstream:
@averylarry I actually tried that PR but it didn't fix it for me :/ The map still came to a halt after a while. What i really can't understand is why, though. Because by the little knowledge I have, looking at the code this shouldn't be happening. Scans that doesn't find any nearby stops or gyms at a given location should never result in a bad scan if no stops or gyms should be seen there? Funny thing is i did not have this problem before.. This started happening around christmas. Cleared the database, fresh install did not help. Makes me wonder if Niantic changed the detection radius for stops and/or gyms?

Skorfius:
pretty sure it has to do with the cell id radius being dropped to 500

TheNewDarkOne:
Yeah I don't think this is solution either. I thin avery's commit should help as long as there is one pokemon on nearby, of course i'm rural enough that 30 minutes out of every hour there isn't even anything on nearby. that said, it's progress. Ultimately mapping took such a big hit in rural in terms of trying to map spawn TTH. I just use classic beehive and call it a day because it's only thing that works. My only wish is if someone would make it so beehive stores TTH when it's found since I'm not going to run -speedscan because as you have noticed it doesn't work in rural, it gets stuck in rescan loops because it expects pokemon when a lot of the time there just aren't any.

If beehive stores TTH for speed scan to use later then eventually fi I just run enough workers and just flat out beehive the area for a while eventually spawn point scanning might be viable.

birdstream:
@TheNewDarkOne ignoring bad scans is the only way my map survives atm. But yeah it's a really bad solution. It should not be merged! Im looking into schedulers.py right now to figure out how to do exactly what you suggested, like letting it retry a set amount of time and then carry on..

birdstream:
Submitted a new PR #1672  to more gracefully handle bad scans

PR #1671, opened 2017-01-03
Added cell id radius to fix most bad scans

Skorfius:
## Description
Speed uses pokestops and gyms to determine bad scans from good scans.
The master branch of pogodev api changed this to 500, so I changed it back to 1000.

This fix wont work for everyone if there was no gyms or pokestops around to begin with.
I was getting too many empty results and this fixed my map and allowed it to go from less than 10% on the initial scan to over 90% in the first hour of running it.

## Motivation and Context
wasted $30 on captchas from pure bad scans in 3 days, it had to be fixed.

## How Has This Been Tested?
My map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -775,7 +775,7 @@ def map_request(api, position, jitter=False):
         scan_location = position
 
     try:
-        cell_ids = util.get_cell_ids(scan_location[0], scan_location[1])
+        cell_ids = util.get_cell_ids(scan_location[0], scan_location[1], radius=1000)
         timestamps = [0, ] * len(cell_ids)
         req = api.create_request()
         response = req.get_map_objects(latitude=f2i(scan_location[0]),


sebastienvercammen:
Unfortunately this isn't a fix, it's a quick hack to work for some very specific users. Bad scans (which is also not a right term: most cases are good scans but they have nothing to compare to) should still be looked into properly.

Obihoernchen:
But I think we have to set cell id radius to >= 900 for gym scanning (-np).
https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/schedulers.py#L143

bbdoc:
@sebastienvercammen It could still help on rural zones as speedscan will find pokestops and gyms within 1000m that it didn't find within 500m. I do agree this is no real fix as if the first stop is 1001m away you still end up with a bad scan, but is there any good reason not to adapt to 1000m already ? I had no issues on my map at the time PR1501 was initially released and now my Initial scanned is freezing around 10%, I guess this change from 1000m to 500m on pogodev api is the reason behind it...

Obihoernchen:
@bbdoc the client is requesting 500m, so requesting 1000m all the time could be an alert for niantic.

bbdoc:
would it be possible to make this value cofigurable through command line ? I have an instance where I do need it, even if it means more captchas, but another one where it's not needed, so better stick to 500m. Thanks

bbdoc:
I've been testing that since yesterday. Pretty happy with it. It helped a lot on my gym only scan, got better results on my rural scan, and really didn't notice any increase in my captcha rate, so I guess Niantic isn't monitoring this.

Anyway in my opinion, better solution would be to keep 500m the default but make it a configuration option. Ideally this option should adapt all references to 500m in the code.

Obihoernchen:
I think 1000 for no_pokemon + configurable (for rural areas), otherwise default 500 would be the best solution.

MikeShiner:
I went from 90% bad scans to 0 with this patch. Definitely better than disabling max-failures but not sure if this a necessary fix for the majority of players.

sebastienvercammen:
Closing this PR in favor of continuing discussion of a proper fix in one of the current open PRs (in order of currently preferred approaches): #1662, #1672.

PR #1672, opened 2017-01-03
Give up on bad scans

birdstream:
## Description

This is a better solution to #1670 
Rather then disabling bad scan check, instead give up after failing for more than 5 times (can be set with -bsr)

## Motivation and Context

Some steps just won't get a succesful scan and get stuffed back in the search queue forever.

## How Has This Been Tested?

On my own map
Console output

## Types of changes

## Checklist:

- [ ] I have updated the documentation locationsaccordingly.

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -822,7 +822,7 @@ class SpeedScan(HexSearch):
                     item['done'] = None
                     log.info('Putting back step %d in queue', item['step'])
                 else:
-                    log.info('Step %d failed scan for more then %d times! Giving up...', item['step'], self.args.bad_scan_retry)
+                    log.info('Step %d failed scan for more than %d times! Giving up...', item['step'], self.args.bad_scan_retry)
             else:
                 # Scan returned data
                 self.scans_done += 1
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -818,8 +818,11 @@ class SpeedScan(HexSearch):
             # If we had a 0/0/0 scan, then unmark as done so we can retry, and save for Statistics
             elif parsed['bad_scan']:
                 self.scans_missed_list.append(cellid(item['loc']))
-                item['done'] = None
-                log.info('Putting back step %d in queue', item['step'])
+                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry:
+                    item['done'] = None
+                    log.info('Putting back step %d in queue', item['step'])
+                else:
+                    log.info('Step %d failed scan for more then %d times! Giving up...', item['step'], self.args.bad_scan_retry)
             else:
                 # Scan returned data
                 self.scans_done += 1
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -819,7 +819,7 @@ class SpeedScan(HexSearch):
             elif parsed['bad_scan']:
                 self.scans_missed_list.append(cellid(item['loc']))
                 # Only try for a set amount of times (BAD_SCAN_RETRY)
-                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry or self.args.bad_scan_delay == 0:
+                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry or self.args.bad_scan_retry == 0:
                     item['done'] = None
                     log.info('Putting back step %d in queue', item['step'])
                 # If scan failed too many times, mark as scanned and carry on
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -818,11 +818,13 @@ class SpeedScan(HexSearch):
             # If we had a 0/0/0 scan, then unmark as done so we can retry, and save for Statistics
             elif parsed['bad_scan']:
                 self.scans_missed_list.append(cellid(item['loc']))
-                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry:
+                # Only try for a set amount of times (BAD_SCAN_RETRY)
+                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry or self.args.bad_scan_delay == 0:
                     item['done'] = None
                     log.info('Putting back step %d in queue', item['step'])
+                # If scan failed too many times, mark as scanned and carry on
                 else:
-                    log.info('Step %d failed scan for more than %d times! Giving up...', item['step'], self.args.bad_scan_retry)
+                    log.info('Step %d failed scan for %d times! Giving up...', item['step'], self.args.bad_scan_retry)
             else:
                 # Scan returned data
                 self.scans_done += 1
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -903,13 +903,14 @@ class SpeedScan(HexSearch):
             elif parsed['bad_scan']:
                 self.scans_missed_list.append(cellid(item['loc']))
                 # Only try for a set amount of times (BAD_SCAN_RETRY)
-                if self.scans_missed_list.count(cellid(item['loc'])) < self.args.bad_scan_retry or self.args.bad_scan_retry == 0:
+                if self.args.bad_scan_retry > 0 and (
+                        self.scans_missed_list.count(cellid(item['loc'])) >
+                        self.args.bad_scan_retry):
+                    log.info('Step %d failed scan for %d times! Giving up...', item[
+                             'step'], self.args.bad_scan_retry + 1)
+                else:
                     item['done'] = None
                     log.info('Putting back step %d in queue', item['step'])
-                # If scan failed too many times, mark as scanned and carry on
-                else:
-                    log.info('Step %d failed scan for %d times! Giving up...', item[
-                             'step'], self.args.bad_scan_retry)
             else:
                 # Scan returned data
                 self.scans_done += 1
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -58,6 +58,7 @@ log = logging.getLogger(__name__)
 # Most of these functions should be overridden in the actual scheduler classes.
 # Not all scheduler methods will need to use all of the functions.
 class BaseScheduler(object):
+
     def __init__(self, queues, status, args):
         self.queues = queues
         self.status = status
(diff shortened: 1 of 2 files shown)


sclo012:
Has anyone figured out why some steps will always return 0/0/0 bad scan? My map isn't even rural and every step should see at least pokestops or gyms, so 0/0/0 is definitely an error. The accounts are not captcha and definitely not banned since they scan other steps perfectly fine.

TheNewDarkOne:
Sclo012 probably are bad scans then. The very definition of a bad scan is a scan that returns nothing that should NOT return nothing. The issue here is telling bad scans from scans that are SUPPOSED to return nothing. Ultimately at end of day, because of using an old blocked API and among other things not being sent correctly (altitude?, etc), there are a lot of actual bad scans. Hopefully when on a valid api among other things there will be less of them. Meanwhile still helps to try and find a way to deal with scans in rural areas that don't have gyms/stops/nearby

birdstream:
Is there an easy way to see lat & long for a given step? Damnit im gonna go there myself and launch the game to see what happens! :P

averylarry:
@sclo012 Try using 1662 and see what the logs show.  It'll provide at least a little more information.

bbdoc:
Hi,

Trying to understand how you are solving this. What will happen after the configured number of bad scan ? Will it mark the band as done and move on so Initial scan goes towards 100% ? What will happen if I stop the script and relaunch, will it go over all those empty areas again for the configured number of attempts or are they writen in db as abandonned ?

Thanks

bbdoc:
btw, when giving up, could you please add step number you are giving up ? Would make the log easier to follow up ;-)

bbdoc:
How is it that I see it giving up several times on the same step ? If it's giving up it shouldn't scan it anymore, looks like somthing is going wrong : 

```
2017-01-04 10:59:41,053 [search-worker-17][    schedulers][    INFO] Step 4612 failed scan for 5 times! Giving up...
2017-01-04 11:01:44,503 [search-worker-17][    schedulers][    INFO] Step 4612 failed scan for 5 times! Giving up...
```

birdstream:
@bbdoc The step is never completely given up on, it's not marked as being "bad" in the database or anything. It will only give up on it for that queue. When the scheduler queues new items it will be included again. However it seems odd that you get that message with like only 2 mins apart.. should be around 5 mins.. How did you implement my PR? git pull? Running some other PR's?

@kevdotcom You're right. But i figured since not all users need this, I wanted to have explicit control over this, and the option to disable it (setting -bsr to 0)

bbdoc:
OK thanks @birdstream I do understand it now... Still annoying has I have a few steps that will then be scanned every 5 minutes when I know there is Nothing there. Do you have any idea if I could in some way flag those in my DB so they are considered as done and not scanned anymore ? And how do I make the link between those step numbers and locations in the DB ?

birdstream:
@bbdoc I think they should be done after a while, but i'm not sure. I havent been able to test this long enough to see what happens :) I've run the map for a couple of hours and initial has creeped up to 99.96% from about 99.7 (running -st 30 beehive with 60 workers) I'm not entirely confident in how all of the code works since I'm not a programmer. In fact, all python knowledge i posess today comes from tinkering with this project. So it's been serving more than one purpose for me. Programming is fun :)

I'll look into the flag steps in the database thing and see if i can figure that out. I'm also looking for a way to link steps to locations...

averylarry:
I tend to think that changing how a bad_scan is determined is better than messing with what to do with a bad_scan.
Or maybe leaving the bad_scan status as-is but adding another "uncertain_scan" status from models.py (when no Pokemon and no forts and no forts expected from the database).  Then schedulers.py can take that information and after having the same uncertain_scan 2 times it will use 0/0/0 as the valid scan.

Does that make sense?

sclo012:
@averylarry The bad scans that I have are actual 0/0/0 where it is unable to scan anything, but there actually are pokemons/stops/gyms there. If left alone, the map will always queue something to scan there since the current logic for bad scan in -speed is to unmark them and requeue them.

It is also possible to get that to happen if there is in fact nothing there anyway

averylarry:
@sclo012 1662 accounts better for nearby Pokemon and pokestops/gyms (even with -nk and -ng).  That's why I wanted you to try it.  This PR (in my opinion) is related to the exclusive scenario when there are no Pokemon AND there are no gyms or pokestops close enough to be returned in a scan.

sclo012:
@averylarry I tried your 1662 and it actually told me there are actually nothing for those 0/0/0 but in fact there are, since I used another scanner and can scan those point.

averylarry:
Do you remember what the log said?  If it said:
'Bad scan. Parsing found absolutely nothing.'
and you're confident there were pokestops or gyms nearby, then I tend to think it was an actual bad scan.  Which is different than "there are not gyms/pokestops nearby to help determine a bad scan".
Maybe combining #1662 and #1671 will get it (almost) right for you.

sclo012:
It said 'Bad scan. Parsing found absolutely nothing.', so pretty sure it's an actual bad scan, but it will queue for it infinitely because for some reason it is never going to be able to scan correctly on that step.

birdstream:
And as far as i can understand by looking at the code, 0/0/0 scans should not be regarded as bad scans if nothing is expected to be there. It calculates a hex bound and pulls forts from database that are within this hex bound..

Hey.. maybe thats the problem?? Isn't scan radius circular? If so.. i suppose there is a chance some steps are in the outer corners of this hex bound and may be considered to have visible forts but in reality doesn't?? hmmm

birdstream:
i meant of course the fort may be in the outer corner.. sorry

averylarry:
@birdstream #1662 adds nearby Pokemon to the equation.  So it needs a full 0/0/0/0 scan (0 wild Pokemon, 0 nearby Pokemon, 0 gyms, 0 pokestops) before it is treated as bad.  Additionally, it checks for gyms and pokestops even with -nk and -ng (and it doesn't check the DB to see if there were supposed to be gyms/pokestops).  Those 2 things alone make much of the issue go away.

cydine:
Combining #1662 and #1671 gets it spot on for me. Thanks to all involved!

birdstream:
@averylarry i tried that PR but it didnt help me out :/ still got bad scans

Anyway.. with this PR running while i was at work, initial scan has indeed reached 100% now and TTH is at 99 %

averylarry:
It's true -- 1662 will still treat a legitimate 0/0/0/0 scan as a bad scan.  I didn't try to mess with the overall logic of -speed scanning; instead I was only reducing the really poor log comments and eliminating some things that clearly shouldn't be bad scans that were being treated as such.  I didn't try to figure out what to do with unknown/unreliable scans (where we don't know if 0/0/0/0 is true or bad).  I also didn't change speed limited scans to be bad scans though I think they should be (not that we know for certain that a 0/0/N/N scan is a speed violation).
If any of that makes sense.

cydine:
I would say that without applying a 1000m scan radius, speed scan would be useless for those in rural or semi rural areas. PGM has a long way to go before it is even close to mimicking  the game client. I don't think setting scan radius to 1000 will have any effect on account bans - it's going to happen anyway.
A scan radius up to 1500m will be accepted by the server according to the API.

I live in a semi rural area - 9 pokestops and 2 gyms, I changed speed limited scans to be bad scans as I have enough forts in a visible radius. Those who are full rural would be better leaving this as default I think.

averylarry:
@cydine see comment on 1662 PR.

bbdoc:
I've set -bsr 0 on one of my instance but I'm getting : 

`Got empty response. Retried 1 times. Giving up`

Shouldn't 0 disable it ? Looks like it doesn't...

birdstream:
@bbdoc No, this PR is not of any value to regular hex scanning, i think. I't cannot fix them, it's just there to avoid steps from being put back in the queue over and over again if it cannot make a successful scan.
Is that exactly how the message is printed in the log? because if so it's something different and not from this PR

kevdotcom:
I observed the bad scans a little further and I noticed that it not only keeps repeating the same step, but it is also always the same account that tries to scan this step and whenever I just restart the scan, it suddenly keeps repeating a different step... but with the same account!

I think it's just some kind of softban, related to the accounts. After a couple hours of rest time, they work like before.

sclo012:
@kevdotcom I find that my bad scan is tied to the step itself, different accounts will try to scan the same step, all will return 0/0/0/0. Whenever they scan another step, it returns valid results. I also add that I only start to see this problem after using the .51 api, it is not as obvious when I used the .45

birdstream:
@sclo012 That is the same in my case, my bad scans are tied to the same steps. And is why i created this PR in the first place.

@kevdotcom What you are suggesting can pretty much be achieved with --max-empty already. Tried that, but it would just eat through the spare accounts and then die...

kevdotcom:
@sclo012 @birdstream How do you know that other accounts don't scan the step aswell? The scheduler doesn't send a different account to the step that got a bad scan. It's the same account repeating the step over and over again until the queue refreshes and (in my case) another step keeps repeating itself... 

Do you have some log output?

birdstream:
@kevdotcom Because like i said, i tried with --max-empty in the beginning. Then if an account have had to many empty scans it would be switched out. But that didn't help, it would just eat through all of my spare accounts and then come to a halt, waiting to get new accounts. I have like 5-6 steps on my map that behaves like this. And it is always the same steps it gets stuck on. The speedscan scheduler grabs the closest available worker to scan a step, and may grab another one if it's close enough while the first one that just scanned and failed is on delay... This eventually results in more and more workers gathering around the failing steps and the map just commits suicide...

I wish there was a way to visualize the bad scans on the map, perhaps by marking them grey instead of green on the map (with "scanned locations" toggled) maybe i can add that in this PR

Mr-Groch:
def visible_forts(step_location) in pogom/models.py need to be updated according to API scan radius change - 500m (distance = 0.5). IMHO this should be first step needed for fixing bad scans problem. If no pokemon nearby and all forts are outside scan range - visibility check causes problems when distance not synced with API. Referring to closed #1685 (I think that it was closed too fast).

bbdoc:
I also don't understand why it was closed so quickly. I tried both this one and nearby pokemons and changing radius to 1000 did a far better job for me than those 2 together...

sebastienvercammen:
@Mr-Groch @bbdoc To clarify: #1685 was closed because we already knew 0/0/0 scans were related to #1662 + #1672 + jitter + location + step distance. And we already had finished private branches for most of those.

sebastienvercammen:
Changes made:

- Logging changed to `+1`, to include the original scan attempt (and log total tries, not only retries).
- Default bsr decreased to 2 (which makes 3 scans total), there is no reason to have more. In fact, setting bsr to 1 would be good enough for most circumstances.
- Conditions changed to read as a positive rather than a negative, this improves readability. Settings the `args` condition first also means the second condition doesn't need to be evaluated when `bsr == 0`.

PR #1674, opened 2017-01-04
Ensure that jittered location is passed to all api calls

pgmaker:
## Description
The jittered location was only being passed to the api in the map request. The set position and other calls had a different location.
This fixes that scenario by ensuring that the jittered location is passed to all of the relevant api calls.
That way the account appears to be in 1 location only and not 2 locations at once.

## How Has This Been Tested?
I tested it extensively in a more complex environment then ported it to a clean copy and ran a short functional test of these changes in isolation.

## Types of changes

## Checklist:

- [ X] My code follows the code style of this project. 
Possibly not. I'm a c++ dev and new to python but the change is small and probably ok.
- [ X] My change requires a change to the documentation.
N/A
- [ X] I have updated the documentation accordingly. 
N/A

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -72,6 +72,7 @@ def jitterLocation(location=None, jitter=False, maxMeters=10):
 
     return scan_location
 
+
 # Thread to handle user input.
 def switch_status_printer(display_type, current_page, mainlog, loglevel):
     # Disable logging of the first handler - the stream handler, and disable it's output.
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -60,7 +60,7 @@ def jitterLocation(location=None, jitter=False, maxMeters=10):
         # jitter it, just a little bit.
         origin = geopy.Point(location[0], location[1])
         b = random.randint(0, 360)
-        d = ((math.sqrt(random.random()) * float(maxMeters)) + 5) / 1000
+        d = ((math.sqrt(random.random()) * float(maxMeters))) / 1000
         tmp_scan_location = geopy.distance.distance(kilometers=d).destination(origin, b)
 
         scan_location = [tmp_scan_location.latitude, tmp_scan_location.longitude, location[2]]
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -53,12 +53,33 @@ TIMESTAMP = '\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\00
 
 
 # Apply a location jitter.
-def jitterLocation(location=None, maxMeters=10):
-    origin = geopy.Point(location[0], location[1])
-    b = random.randint(0, 360)
-    d = math.sqrt(random.random()) * (float(maxMeters) / 1000)
-    destination = geopy.distance.distance(kilometers=d).destination(origin, b)
-    return (destination.latitude, destination.longitude, location[2])
+#def jitterLocation(location=None, maxMeters=10):
+#    origin = geopy.Point(location[0], location[1])
+#    b = random.randint(0, 360)
+#    d = math.sqrt(random.random()) * (float(maxMeters) / 1000)
+#    destination = geopy.distance.distance(kilometers=d).destination(origin, b)
+#    return (destination.latitude, destination.longitude, location[2])
+
+
+def jitterLocation(location=None, jitter=False, maxMeters=5 ):
+    
+        # create scan_location to send to the api based off of position, because tuples aren't mutable
+    if jitter:
+        # jitter it, just a little bit.
+        origin = geopy.Point(location[0], location[1])
+        b = random.randint(0, 360)
+        d = ((math.sqrt(random.random()) * float(maxMeters)) + 5) / 1000
+        tmp_scan_location = geopy.distance.distance(kilometers=d).destination(origin, b)
+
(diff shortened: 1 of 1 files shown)


averylarry:
I don't think this is necessary with the current code.  There is no request made to the API in check_login when the current auth session is still valid.  When a re-login *is* done -- there's a 20 second sleep built into check_login after a successful login (current line 764).  At most you'd be moving 0.5m/s, or 1.8kph from the login location (step_location) to the jittered scan location.

pogom/search.py
```diff
@@ -52,13 +52,25 @@
...
-    return (destination.latitude, destination.longitude, location[2])
+# Handle Jittering if it is configured.
+def jitterLocation(location=None, jitter=False, maxMeters=10):
+
+    # create scan_location to send to the api based off of position, because tuples aren't mutable
+    if jitter:
+        # jitter it, just a little bit.
+        origin = geopy.Point(location[0], location[1])
+        b = random.randint(0, 360)
+        d = ((math.sqrt(random.random()) * float(maxMeters)) + 5) / 1000
```
averylarry: Why would you add 5 meters here?  You're making a 5 meter to 15 meter donut around the scan location.

averylarry:
Looking even more -- I don't see where the scan_location is used at all in check_login.

pogom/search.py
pgmaker: oops. Yeah in my own codebase I reduced it to 5m forced plus maxmeters of 5 so the jitter couldn't be less than 5m and couldn't be more than 10 but I left the max meters as 10m.

pgmaker:
the api.set_position loads the coordinates into the api object. How they're used from there is difficult to trace because of the nature of protobuf api's and python (at least from the perspective of a c++ dev)
So it's better to be safe than sorry until more information on the api and how it works comes to hand.

smc87:
@averylarry I understand why it might not seem needed, but I think its generally a good idea to emulate the app as close as possible. EG the 20 second wait...maybe we shouldn't actually be doing that? 

@pgmaker I agree with the "better safe than sorry" approach. 

as mentioned, although it doesn't make the call itself, it is the param that gets applied to your next call from "api.*" -- I've seen errors thrown about player location not set when i have tried setting device_info after previously using setlocation.

Will test shortly.

smc87:
@pgmaker Do you think we should update gym_request to use scan_location too?

averylarry:
parse_map uses explicit lat/lon -- and it's my understanding that the explicit location takes precedence over the previous set_position location.
I vaguely remember there was a reason for the 20 second delay, but I don't have details.  It's probably related to IP bans.
If you want to emulate real life, my phone's GPS is sketchy at best.  As is my cell phone internet.  I don't think this will help -- I mean the way it is now is pretty much MORE jitter.

Just to be clear, this might be nice to do anyway just for consistency, but I do not think it actually is a bug or something that *needs* to be changed.

Now -- if you want to change it anyway, then I would suggest something more like this:

```
if args.jitter:
    scan_location = jitterLocation(step_location)  #don't bother with sending args.jitter to the function
else:
    scan_location = step_location
api.set_position(*scan_location)
```

No wait -- seb prefers this:
```
scan_location = step_location
if args.jitter:
    scan_location = jitterLocation(step_location)  #don't bother with sending args.jitter to the function
```
And then actually remove the explicit setting of location in map_request.  Neither check_login nor map_request nor gym_request etc. would then need to have location passed to it.

Or maybe get rid of set_position entirely and only use explicit location in check_login, map_request, and gym_request.  This actually feels safer to me.

pgmaker:
I don't know enough about the api to remove the call to set_position but the other suggestions are no issue to do.

smc87:
@averylarry line 220 in https://github.com/pogodevorg/pgoapi/blob/master/pgoapi/pgoapi.py 
seems to suggest you are right and that the API's position will be overridden by requests with specific position set.

I think we still need to set_position before first login though, but this could be moved into the check_login routine, as we don't need to do it every time.

smc87:
@averylarry I just tried getting rid of the set_position() call for subsequent calls and relying only on the and it messes up the scanning. 
I guess we do infact need to manually update the location of PGoApi instance via set.position(), as it doesnt seem to be updated by the PGoApiRequest instances that are called from within it.

@pgmaker - sorry I feel I have derailed your PR comments somehow.

pgmaker:
@smc87 Dont sweat it. You've done the test that confirms the need to call set position.
I have another commit to do that i need to test first.

cydine:
The jitter function can be tidied up a bit. I would suggest:

```python
def jitterLocation(location=None, maxMeters=10):`

    bearing = random.randint(0, 360)
    distance = ((math.sqrt(random.random()) * float(maxMeters))) / 1000

    destination = get_new_coords(location, distance, bearing)
    log.info('Jittered to: %f, %f', destination[0], destination[1])

    return (destination[0], destination[1], location[2])
```

and then you can remove the geopy imports as well.

pgmaker:
I agree. Many of the coordinate calculations are unnecessarily over the top for what this is. Earth curvature corrections are generally pointless under 15km and only if you're after high precision which we're not.
Reducing the complexity of the calculations should theoretically improve performance of the application.
I do think it's a task for a separate PR though.

sebastienvercammen:

PR #1684, opened 2017-01-08
Improve refresh queue performance

trew:
## Description
Decrease the amount of sql queries when refreshing the queue.

## Motivation and Context
There is an insane amount of sql queries going on when refreshing the queue. Depending on the setup, there could be thousands of different queries to the database when in fact only a few is needed.

## How Has This Been Tested?
Tested on a windows machine with -st 16. ( around 1700 spawnpoints)
Time taken to perform queue refresh was reduced from 5-10 second to about half a second.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -744,7 +744,6 @@ class ScannedLocation(BaseModel):
     def _q_init(scan, start, end, kind, sp_id=None):
         return {'loc': scan['loc'], 'kind': kind, 'start': start, 'end': end, 'step': scan['step'], 'sp': sp_id}
 
-
     @classmethod
     def get_by_locs(cls, locs):
         lats = [loc[0] for loc in locs]
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -744,6 +744,29 @@ class ScannedLocation(BaseModel):
     def _q_init(scan, start, end, kind, sp_id=None):
         return {'loc': scan['loc'], 'kind': kind, 'start': start, 'end': end, 'step': scan['step'], 'sp': sp_id}
 
+
+    @classmethod
+    def get_by_locs(cls, locs):
+        lats = [loc[0] for loc in locs]
+        lons = [loc[1] for loc in locs]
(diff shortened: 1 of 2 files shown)


Zyrix:
Works very well for me as it decreased my query refresh time from around 5-7 minutes to 45-55 seconds.

Barloklol:
This PR puts a lot of strain on your DB if you scan lots of instances, so I recommend adjusting these values if you use MySQL with InnoDB:

wait_timeout
If not by default at 28800sec (8h), I recommend setting it back to that since I got tons of aborted connections with my 1h value, which messed up my DB performance

thread_cache_size
Check your max_used_connections and set it bigger than that

table_open_cache
Check if your Opens increase quickly, if not no need to adjust. Should be around the number of tables in your DB

table_definition_cache
Use this formula: 400 + (table_open_cache / 2)

query_cache_size
I recommend disabling it with a 0 since our tables are constantly changing. Also set query_cache_type to 0 to disable it completely.

innodb_open_files
Use the same value as table_open_cache

innodb_read_io_threads
Since we're doing lots of read operations I suggest increasing it by a lot. I maxed it out at 64, might be too much for slower systems though

I'm sure someone with more MySQL experience can correct me on some of that, I just tweaked until it worked. I have refresh queue times ranging from 10-15secs with 100 instances running on a single DB now without problems. Very good PR, improved refresh speed by up to 95% for me

eskimo9:
Works very well, went from 90-100 % cpu usage on python to 30% and mysql is loaded heavier. Good job.

onilton:
@trew I fixed the merge conflicts, ping me if you want to save sometime and get them.

KoertJanssens:
Your latest commit gives me 'Too many values to pack' errors
Before that it was running fine and improving performance!

trew:
Thanks for spotting that, @KoertJanssens . Turns out locs contains 3 values, which invalidates the use if zip.

PR #1689, opened 2017-01-09
Update schedulers.py (self.step_distance = 0.900 to self.step_distance = 0.500)

h0us3cat:
## Description

if self.args.no_pokemon:
            self.step_distance = 0.900

to

if self.args.no_pokemon:
            self.step_distance = 0.500
## Types of changes

## Checklist:

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -98,7 +98,7 @@ class HexSearch(BaseScheduler):
 
         # If we are only scanning for pokestops/gyms, the scan radius can be 900m.  Otherwise 70m.
         if self.args.no_pokemon:
-            self.step_distance = 0.900
+            self.step_distance = 0.500
         else:
             self.step_distance = 0.070
 


cydine:
There are a few other instances in the code where this change is needed too. Another in schedulers.py, in models.py and in scan.py.

bbdoc:
Agree with @cydine... Changing only this value can results in lots of issues. I tried it and the reverted as my beehive was not calculated correctly anymore, had lots of holes between the different hives. All references need to be changed.

Obihoernchen:
That's bad and by far not complete. Why not adding radius to https://github.com/PokemonGoMap/PokemonGo-Map/blob/develop/pogom/search.py#L789 ?

sebastienvercammen:
This PR missed other old references to the distances. We have a more complete PR ready on a private fork but can't publicize/merge yet because the changes are tied to other upcoming PRs that change behavior.

PR #1692, opened 2017-01-09
Fix scan radius in no pokemon mode

Obihoernchen:
## Description
pogoapi changed the default cell_id radius from 1000m to 500m (see: https://github.com/pogodevorg/pgoapi/pull/156).
This breaks the `--no-pokemon` scanning mode because this mode assumes a radius of => 900m.
Change the scan radius to 1000m in no pokemon mode to restore previous functionality.

## Motivation and Context
- gym scanning misses a lot of gyms without this change

## How Has This Been Tested?
- locally

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -776,7 +776,7 @@ def check_login(args, account, api, position, proxy_url):
     time.sleep(20)
 
 
-def map_request(api, position, jitter=False, scan_radius):
+def map_request(api, position, jitter=False, scan_radius=500):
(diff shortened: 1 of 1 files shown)


Arengo:
Tried it out locally, works for me.

sebastienvercammen:
PR closed: The distance change in the API was intentional and for good reason, the app only uses a 500m range. Our distances should be reworked to use 500m, we shouldn't be forcing the API to old behavior.

Discussion can continue in relevant PRs: #1662

PR #1695, opened 2017-01-09
Display dialog on map when requests to /raw_data error or timeout

hokiepokedad:
## Description
For larger map implementations, there are cases where a user's marker settings (criteria) is wide open by selecting pokemon, gyms, scanned locations, spawnpoints, etc. Also, they may have zoomed out to include many hive locations for their large viewport.  Doing so can potentially time out a request.  In addition to wide open criteria, their connection could also be a factor in timing out (WiFi or mobile).

While an error is returned for the timeout, it is not being handled on the client side and the user sees nothing.  This will now display a dialog with the following text:

Title: Reduce marker settings
Body: Hmm... we're having problems getting data for your criteria. Try reducing what you're showing and zooming in to limit what's returned.

## Motivation and Context
I've had frequent questions from a bunch of my users saying "no pokemon" appear on the map when it was due to the wide open criteria or slow connections.

## How Has This Been Tested?
My local maps.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1029,6 +1029,19 @@ function loadRawData() {
                 rawDataIsLoading = true
             }
         },
+        error: function() {
+            if (!$timeoutDialog) {
+                var opts = {
+                    title: 'Reduce marker settings'
+                };
+
+                $timeoutDialog = $('<div>Hmm... we\'re having problems getting data for your criteria. Try reducing what you\'re showing and zooming in to limit what\'s returned.</div>').dialog(opts);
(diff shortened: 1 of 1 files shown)


Arengo:
Would you mind to first rebase you're PR cleanly that we see only the changes related to the PR or is it just HASH 0b1628a... ?

hokiepokedad:
@Arengo - my apologies, git noob here.  all checks passed and things should look okay now.

Arengo:
@hokiepokedad no problem, I am sure I am the bigger git noob of us two.

PR #1708, opened 2017-01-12
Prevent "Speed violation?" message when using --no-pokemon

snooter:
## Description
Using the --no-pokemon flag will always result in no nearby_pokemons or wild_pokemon.

## Motivation and Context
Prevents log spam.

## How Has This Been Tested?
Run in a remote area without --no-pokemon to see the message, run in a busy area with --no-pokemon to suppress the message.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1466,7 +1466,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue, a
             'bad_scan': True
         }
 
-    if not len(nearby_pokemons) and not len(wild_pokemon):
+    if not len(nearby_pokemons) and not len(wild_pokemon) and not args.no_pokemon:
         log.warning('No nearby or wild Pokemon. Speed violation?')
         log.info("Common causes: not using -speed, deleting or dropping the WorkerStatus table without waiting before restarting, or there really aren't any Pokemon in 200m")
 


sebastienvercammen:
You didn't check the code properly, `nearby_pokemons` is still used as usual because the Niantic servers will return this data (they are unaware of any "no Pokémon" flag), even with Pokémon scanning disabled.

It's `wild_pokemon` that is used for Pokémon parsing later on which will be empty.

This PR does not fix anything.

Obihoernchen:
@sebastienvercammen I disagree.
This PR is totally fine because the speed violation message is wrong for no pokemon mode. **There is no speed limit for gym/pokestop scanning**. `nearby_pokemons` and `wild_pokemon` can be empty, it doesn't matter. Gyms/stops are scanned regardless.

snooter:
I'm still wrapping my head around all of the inner workings here, but I agree with @Obihoernchen .  I'm probably exceeding the speed limit, which is resulting in zero nearby_pokemons and wild_pokemon, but that doesn't matter because the -np flag explicitly says I don't care about that, and I'm still getting the gym information I requested. 

All I know is that with this PR, when using -np for gym scanning, my logs are spam-free, and nothing is changed when not using the -np flag.

snooter:
To simplify: this PR only suppresses a "no pokemon found" message when using --no-pokemon, which seems pretty innocuous.

PR #1714, opened 2017-01-13
Detect captchas also when catcha solving is not enabled to avoid bad scans and not needed retries.

lluisd:
## Description
Avoid register a captcha as a bad scan when we get a captcha inside the api response with the _captcha-solving (-cs)_ option not enabled.

Avoid trying to rescan again with that account until we reach the max-empty attempts which by default is disabled (it can be an infinte loop). So, put the account to cool down with a message related to captcha encountered as son as we detect it.

## Motivation and Context
Some of us we don't use any external service to solve captchas. So, when a captcha is encountered we receive bad scan 0/0/0 messages in our logs which can be confusing and the account tries to rescan again and again unless you specify a max-empty number of attempts (waste of time). 

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -694,10 +694,9 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
                 # todo's to db/wh queues.
                 try:
                     # Captcha check.
-                    if args.captcha_solving:
-                        captcha_url = response_dict['responses'][
-                            'CHECK_CHALLENGE']['challenge_url']
-                        if len(captcha_url) > 1:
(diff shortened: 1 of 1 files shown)


averylarry:
I wonder if this would be better in the parse_map function in models.py.

lluisd:
Closed. 

Closed. New one opened https://github.com/PokemonGoMap/PokemonGo-Map/pull/1727 to create the right commits (pep8 and changes) with a new own branch to be more easy to be updated/rebased in the future.

PR #1717, opened 2017-01-15
2captcha balance checking

birdstream:
## Description

Pause scanning in case 2captcha balance reaches under limit

## Motivation and Context

Last morning i woke up to realize that my 2captcha balance was zero and so all of my accounts had been been captcha'd. Not nice. This PR aims to take care of that, and automagically suspend scannnig to avoid this scenario. Thought this should be in there...

## How Has This Been Tested?

On my own map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -39,7 +39,7 @@ from pgoapi.exceptions import AuthException
 
 from .models import parse_map, GymDetails, parse_gyms, MainWorker, WorkerStatus
 from .fakePogoApi import FakePogoApi
-from .utils import now, get_tutorial_state, complete_tutorial
+from .utils import now, get_tutorial_state, complete_tutorial, captcha_balance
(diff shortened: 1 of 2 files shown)


trew:
Do you really need to check every 30 seconds? Imo that should be configurable.

m3g4mInD:
@trew  change for the value you whant  "time.sleep(1500)"

birdstream:
Added 2Captcha wallet balance to ps output. I know i should not be using globals but i just couldn't figure out how to avoid that. If anyone would like to give directions on how to do that, it'd be appreceated! I still have a lot to learn. However, using globals in this case should not be harmful since only one thread modifies balance...

birdstream:
I'll make a fix when i get back from work.

birdstream:
Rebased #1726

PR #1721, opened 2017-01-16
add time to popup notification

catnet23:
## Description
This PR improve the popup-notification with IV, disappear-time and time until disappear.

## Motivation and Context
I've saw a lot this popup notification but had always to click on it to get information about this pokemon.
to improve this behaivoir i added the useful information directly to the popup-notification

## How Has This Been Tested?
This PR i've tested on my own Map in different cities. Running since day one of my own Map without any issue.

```
4.4.26-gentoo
Python 2.7.12
pip 7.1.2 from /usr/lib64/python2.7/site-packages (python 2.7)
```

popupnoty
## Types of changes
- [ `x`] New feature (non-breaking change which adds functionality)

## Checklist:
- [ `x`] My code follows the code style of this project.

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -866,27 +866,6 @@ var StoreOptions = {
         default: '',
         type: StoreTypes.Number
     },
-    /*
-        text place holders:
-        <pkm> - pokemon name
-        <prc> - iv in percent without percent symbol
-        <atk> - attack as number
-        <def> - defense as number
-        <sta> - stamnia as number
-    */
-    'notify_title': {
-        default: '<pkm> <prc>% (<atk>/<def>/<sta>)',
-        type: StoreTypes.String
-    },
-    /*
-        text place holders:
-        <dist>  - disappear time
-        <udist> - time until disappear
-    */
-    'notify_text': {
-        default: 'disappear at <dist> (<udist>)',
-        type: StoreTypes.String
-    },
     'showGyms': {
         default: false,
         type: StoreTypes.Boolean
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -866,6 +866,27 @@ var StoreOptions = {
         default: '',
         type: StoreTypes.Number
     },
+    /*
+        text place holders:
+        <pkm> - pokemon name
+        <prc> - iv in percent without percent symbol
+        <atk> - attack as number
+        <def> - defense as number
+        <sta> - stamnia as number
+    */
+    'notify_title': {
+        default: '<pkm> <prc>% (<atk>/<def>/<sta>)',
+        type: StoreTypes.String
+    },
+    /*
+        text place holders:
+        <dist>  - disappear time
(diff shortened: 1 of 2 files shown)


static/js/map.js
```diff
@@ -605,6 +605,50 @@ function isRangeActive (map) {
...
+    'hour': hour,
+    'min': min,
+    'sec': sec
+  }
+}
+
+function notifyText (item) {
+  var perfection = getIv(item)
+  var ivtext = perfection.toFixed(1) + '% (' + item['individual_attack'] + '/' + item['individual_defense'] + '/' + item['individual_stamina'] + ')'
+  var dtime = new Date(item['disappear_time'])
```
onilton: disappear_time text could have its own function? maybe `disappearText(disappearTime)`

I think the same goes for untiltext

catnet23:
@onilton maybe it's better to add it in map.common.js? with some replacement like placeholder?

static/js/map.js
catnet23: @onilton maybe it's better to add it in map.common.js? with some replacement like placeholder?

catnet23: or i can do it in notifyText it self als return value.
because do we need it in other section? if wee don't need it, then i will put in notifyText anayway.

onilton: It could be, but I don't have a strong opinion on this, to be honest.

static/js/map.common.js
```diff
@@ -886,6 +886,28 @@ var StoreOptions = {
...
+
+  /*
+    text place holders:
+    <pkm> - pokemon name
+    <prc> - iv in percent without percent symbol
+    <atk> - attack as number
+    <def> - defense as number
+    <sta> - stamnia as number
+  */
+  'notify_title': {
```
lluisd: notify_title and notify_text are not parameters for the map. StoreOptions object is used to save default values for options that can be set and get from the map interface, not for saving string templates.

static/js/map.js
```diff
@@ -605,6 +605,67 @@ function isRangeActive (map) {
...
+function getIv (item) {
+  if (item['individual_attack'] !== null) {
+    return 100.0 * (item['individual_attack'] + item['individual_defense'] + item['individual_stamina']) / 45
+  }
+
+  return 0
+}
+
+function getTimeUntil (time) {
+  var now = +new Date()
```
lluisd: Why not extract the part that calculates and returns a timestring with the time left in that existing function: "updateLabelDiffTime" to avoid duplicate code.

static/js/map.js
```diff
@@ -605,6 +605,67 @@ function isRangeActive (map) {
   return Store.get('showRanges')
 }
 
+function getIv (item) {
+  if (item['individual_attack'] !== null) {
```
lluisd: There is a lot of places where we calculate the IV % with the harcoded calculation duplicated three times. I will suggest to extract to a function like that and use it for the three usages + your usage.

Issue #1638, opened 2016-12-18
[Feature Request] Make use of nearby_pokemon at forts information


saintmagician:
Edit: This is NOT about lured pokemon. This is about "nearby" pokemon for areas where the new tracker is available. See this screenshot: http://imgur.com/gallery/Mj4Fl
The new tracker displays "sightings" (pokemon within 200m of you, that is not near a pokestop) and "nearbys" (pokemon located at a visible pokestop). The game client only ever displays up to 9 nearbys, however the API call to get map info gives you all possible nearbys. This feature request is about displaying these nearbys on the map next to their respective pokestop. Ideally, we'd even be able to match them up to known spawn points. 

When you do the API call to get back map info, you get the following information: capturable_pokemon, wild_pokemon, nearby_pokemon

As far as I can see, capturable pokemon list is the same as the wild pokemon list. PokemonGo-Map currently parses all wild_pokemon and inserts them into the db. The nearby_pokemon list is used only to generate the 'Nothing on nearby_pokemons or wild. Speed violation?' message. 

The nearby_pokemon list contains two kinds of nearby pokemon: pokemon within 200m (appears in game as 'sightings'), and pokemon with attached fort information (appears in game as 'nearby'). The latter category could be parsed and displayed in PokemonGo-Map. You could display them on the pokestop that they are attached to. Although these pokemon won't appear exactly on the pokestop, they are generally within 70m of the pokestop. This would be good enough for most people. The upside is that you can see a LOT of nearby_pokemon - up to 1km away I believe. For example, a single scan near where I live gives about 60 nearby pokemon attached to various pokestops. 

example of nearby pokemon (within about 1km, always attached to a fort that is currently visible)
{'fort_image_url': u'http://lh4.ggpht.com/bkaTb8_dO93oGayYan4HZuWJZgClnI-By1QkGLIU6bkT-rMl3-pC-mNgRf5aH-tYm19D4Ae5_yHJKw6sUlo', 'fort_id': u'39120c0642e34915927f7ffbaf9de9d3.16', 'distance_in_meters': 705.1959838867188, 'encounter_id': 6974047907822660857L, 'pokemon_id': 98}

example of sightings pokemon (within 200m, not attached to a fort)
{'encounter_id': 17934254007354986880L, 'pokemon_id': 16}

example of wild pokemon:
[...]

PR #1724, opened 2017-01-17
Refactored webhooks.

sebastienvercammen:
## Description
Two major changes to webhooks:

1. Removed thread locking in favor of capturing the exception when several threads get into a race condition. This immensely improves webhook throughput, and all webhooks still get sent: on exception, we requeue the failed webhook.
2. Included the type of search scheduler that's being used (e.g. SpeedScan, HexSearch) and the TTH percentage found (in case of SpeedScan, `-1` when not applicable). This is required for our public webhook server.

## Motivation and Context
- Severe reduction in webhook thread requirements, while maintaining delivery rate.
- TTH found is useful for data verification and spawn timer verification.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1548,6 +1548,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     scheduler_name = scheduler.__class__.__name__
     tth_found = getattr(scheduler, 'tth_found', -1)
 
+    if tth_found > -1:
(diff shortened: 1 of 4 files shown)


jorgitoh:
Hi @sebastienvercammen. There is a way to filter which pokemons the map send to the webhook?. Im using the issue #1638 to find nearby pokemons and the amount of pokemons is huge. Would be good if we can filter the pokemons before send that to the wh, because we ll reduce the amount of threads too. Regards!

PR #1723, opened 2017-01-16
Adds total summaries at top of status web page across all map instances 

tomballgithub:
## Description
Adds total summaries at top of status web page across all map instances whether in group-by-mainworker mode or not.  This was waiting on release of PR #1604 and basically adds up the stats for each main_worker and generates stats across all workers.  

These total stats include:  active workers, successes, failures, empties, skips, and captchas.

I had to muddle through learning quite a bit about javascript to make this work.  I am open to suggestions on how to clean things up.

The per hr and per month stats use an average of the elapsed time of each instance, assuming that they are launched at approximately the same time.  Any other method would be extravagant.  If instances are launched random times from each other the top-level per hr and per month numbers are not trustworthy.

## Motivation and Context
If you have multiple instances, there is not a way to see the totals across all instances.  PR #1604 added per-main-worker support and this builds upon that

## How Has This Been Tested?
Tested on my Windows server with 10 instances of 5-25 workers each
Verified totals and calculations are correct
Verified behavior works in group-by-mainworker mode and other mode

## Screenshots (if appropriate):
image
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1726, opened 2017-01-17
2captcha balance (rebased)

birdstream:
## Description

Add captcha balance checking to pause scanning if balance reaches under limit.

## Motivation and Context

If captchta wallet balance reaches zero, the scanner doesn't stop and will eventually result in all accounts being captcha'd. This can be avoided by periodically check the balance and stop when it reaches a desired limit.

## How Has This Been Tested?

On my own map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -267,7 +267,7 @@ def captcha(args, pause_bit):
         if args.captcha_key and captcha_balance(args.captcha_key) <= args.captcha_balance_limit:
             pause_bit.set()
             log.info('Searching paused due to 2captcha balance is under limit!')
-        time.sleep(30)
+        time.sleep(args.captcha_balance_interval)
 
 
 def worker_status_db_thread(threads_status, name, db_updates_queue):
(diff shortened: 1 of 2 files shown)


Alderon86:
if you could add that to the already exisiting captcha data on the status page, that would be super awesone!

tomballgithub:
Per @alderon86's comment, this would be great to align with PR #1723 and show that data at top of web status page

birdstream:
I actually had that in the original PR but it kept conflicting with the latest days commits so i just left it for now until things settle.

BrownSlaughter:
Instead of pausing it might be a nice idea to somehow send a notification or webhook if the balance goes below the limit to notify the user

catnet23:
@BrownSlaughter why not using the options on 2captcha it self?

M4d40:
> if you could add that to the already exisiting captcha data on the status page, that would be super awesone!

> I actually had that in the original PR but it kept conflicting with the latest days commits so i just left it for now until things settle.

@birdstream i think the captcha stuff is mostly done for now, could you please re-check if you could implement it?

PR #1727, opened 2017-01-17
Detect captchas also when catcha solving is not enabled to avoid bad scans and not needed retries.

lluisd:
## Description
Avoid register a captcha as a bad scan when we get a captcha inside the api response with the captcha-solving (-cs) option not enabled.

Avoid trying to rescan again with that account until we reach the max-empty attempts which by default is disabled (it can be an infinte loop). So, put the account to cool down with a message related to captcha encountered as son as we detect it.

## Motivation and Context
Some of us we don't use any external service to solve captchas. So, when a captcha is encountered we receive bad scan 0/0/0 messages in our logs which can be confusing and the account tries to rescan again and again unless you specify a max-empty number of attempts (waste of time).

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -200,8 +200,10 @@ def status_printer(threadStatus, search_items_queue_array, db_updates_queue, wh_
 
                     status_text.append(
                         status.format(item, time.strftime('%H:%M', time.localtime(threadStatus[item]['starttime'])),
-                                      threadStatus[item]['username'], threadStatus[item]['proxy_display'],
-                                      threadStatus[item]['success'], threadStatus[item]['fail'], threadStatus[item]['noitems'],
+                                      threadStatus[item]['username'], threadStatus[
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -807,11 +807,11 @@ def search_worker_thread(args, account_queue, account_failures, search_items_que
...
                 try:
                     # Captcha check.
-                    if args.captcha_solving:
-                        captcha_url = response_dict['responses'][
-                            'CHECK_CHALLENGE']['challenge_url']
-                        if len(captcha_url) > 1:
-                            status['captcha'] += 1
+                    captcha_url = response_dict['responses'][
+                        'CHECK_CHALLENGE']['challenge_url']
+                    if len(captcha_url) > 1:
```
sebastienvercammen: Python-esque = `if captcha_url:`

lluisd:
Done.

PR #1739, opened 2017-01-22
Improve webhooks

neskk:
Remove `scheduler_name` and `tth_found` from the standard webhook message.
## Description

We don't need to send `scheduler_name` and `tth_found` bundled with *every* webhook message.
Since other schedulers don't even have a use for this, I changed back the structure of webhook message to its previous form:
```
data = {
        'type': message_type,
        'message': message
}
```
To send SpeedScanner scheduler updates, I added in search.py - search_overseer_thread() a webhook update message that is sent whenever the percentage of TTH found changes. This message will only be sent if SpeedScanner is enabled.

**UPDATED:** new SpeedScanner scheduler update message:
```
data = {
        'type': 'scheduler',
        'message': { 'name': <scheduler_name>,
                     'instance': <instance_status_name>,
                     'tth_found': 100.0,
                     'spawns_found': 54
        }
}
```

Also did some 'minor text fixes' adding and correcting some punctuation.

**UPDATED:** Added a config parameter `-whsu`, `--webhook-scheduler-updates` to enable/disable sending webhook updates with scheduler status.

## Motivation and Context

Helps to reduce traffic sent from webhooks.
Improves compatibility with traditional/older webhook consumers.
Reduces "context leakage": keeps scheduler out of `parse_map()` and `parse_gym()` functions.
Can be refactored later to include more information relative to schedulers.

## How Has This Been Tested?

Tested in my local machine, win7 x64, python 2.7.12, using PokeAlarm to check if webhooks were arriving.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -399,18 +399,10 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb, db_updat
 
     stats_timer = 0
 
+    scheduler_tth_found = 0
+
     # The real work starts here but will halt on pause_bit.set().
     while True:
-        # TODO: send regular updates when tth_found changes
-        '''
-        scheduler_name = scheduler.__class__.__name__
-        tth_found = getattr(scheduler, 'tth_found', -1)
-
-        if tth_found > -1:
-        # Avoid division by zero. Keep 0.0 default for consistency.
-        active_sp = max(getattr(scheduler, 'active_sp', 0.0), 1.0)
-        tth_found = tth_found * 100.0 / active_sp
-        '''
 
         if args.on_demand_timeout > 0 and (now() - args.on_demand_timeout) > heartb[0]:
             pause_bit.set()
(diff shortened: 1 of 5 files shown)


pogom/search.py
```diff
@@ -462,6 +464,20 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb, db_updat
...
+        if args.speed_scan:
+            tth_found = getattr(scheduler_array[0], 'tth_found', -1)
+
+            if tth_found > -1:
+                # Avoid division by zero. Keep 0.0 default for consistency.
+                active_sp = max(getattr(scheduler_array[0], 'active_sp',
+                                        0.0), 1.0)
+                tth_found = tth_found * 100.0 / float(active_sp)
+
+            if scheduler_tth_found < tth_found:
```
sebastienvercammen: We should move this block of code out of the overseer's main block of code, so the difference between the overseer's job and "extras" (in this case, webhooks) becomes clearer.

We should also consider replacing
```
if scheduler_tth_found < tth_found:
```
with something similar to
```
if (tth_found - scheduler_tth_found) > 0.1:
```
to only update once it passes a certain interval. Float precision would otherwise trigger it as an update every single time, which probably isn't what we want.

neskk: I can make the changes right now.
Previously every webhook message sent `tth_found`, so I figured that it would be nice to send a message _every time_ the percentage of `tth_found` changed.
I could put this code in a function, but since it's only used once and still has few functionalities, I would let that refactoring happen when/if we add more features.

sebastienvercammen: As discussed, we do want to rework it so all scheduler types have some form of separate webhook handling, but since that's outside of the scope of this PR, moving it into its separate function (in a proper scope) will already help keep things clean.

A second point was to rethink the fields that were being sent. For now it was only the name and `tth_found` (in percentage), but these are some suggested fields to add:

- `status_name`: can be used by the receiving server to identify source instances.
- `spawns_found`: total spawns found.
- `tth_found` as a total number of spawnpoints of which we have the TTH (rather than a percentage), in combination with `spawns_found`.

The next step is to think about sending regular account status data (non-identifying data: total accounts, total captchas, captcha rate), and search thread data (good scans, bad scans).

PR #1747, opened 2017-01-24
Modify 'Bad scan' logic so that initial scans complete

rocketrobot99:
## Description
Change the code to treat 'bad scans' in a similar fashion to 'Empty scans'

## Motivation and Context
Currently 'Bad scan' locations are retried until '--bad_scan_retries' is reached, then the step is put back onto the schedule. This means the schedule keeps attempting to scan these locations and never completes. It is assumed that 'Bad scans' only occur because they are in rural areas far from Gyms and Pokestops. In the past, bad scans could occur in dense populations, test have shown that this is no longer the case.

## How Has This Been Tested?
I have run a large map, the initial scan now completes, whereas previously it would not.

The following test was run, just to be sure that 'Bad scans' only occur in rural areas. 'Bad scan' locations were recorded and shown on the map next to spawn points. this provides a sense check. If any of these locations are near a spawn point then there is something wrong.

image

To me, this clearly shows the 'bad scan' locations (green circles) are not near any spawns, gyms or Pokestops, so should be treated as 'empty'.

## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1607,7 +1607,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     if not wild_pokemon and not nearby_pokemon:
         # . . . and there are no gyms/pokestops then it's unusable/bad.
         abandon_loc = False
-        
+
         if not forts:
             log.warning('Bad scan. Parsing found absolutely nothing.')
             log.info('Common causes: captchas or IP bans.')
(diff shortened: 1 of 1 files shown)


WalterMagnum:
This fixed my bad scan issues.

PR #1567, opened 2016-11-20
manual recaptcha solving

paraselene:
## Description

Manual Captcha Solving without 2captcha

## Motivation and Context

Someone may not have 2captcha, they need to solve the captcha themselves, it provides a easy way for them to solve the captcha

## How Has This Been Tested?

1. Go to /bookmarklet under your map.
eg: If your maps is http://xyz.com, go to http://xyz.com/bookmarklet

2. Drag the Solve Recaptcha Bookmarklet to your bookmark bar

3. Wait for an account to encounter the captcha

4. Click the bookmarklet one time, it will redirect you to http://pgorelease.nianticlabs.com/

5. Click the bookmarklet one more time, it will show the Recaptcha

screen shot 2016-11-16 at 9 50 54 pm

5. Solve the recaptcha

6. Recaptcha token will be saved to Database in Token table

7. token_request will get a token which created after the request made and return to the search worker and verify the captcha challenge

p.s.: if token_request cannot get a token which created after the request made within manual-captcha-solving-allowance-time(default: 60s), token_request will return 'ERROR' and append the account to account_failures
Please set manual-captcha-solving-domain to your map domain eg: http://xyz.com

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1754, opened 2017-01-25
Captcha handling rework

neskk:
Captchas have been around for a long time and pretty much everyone is using 2captcha to get rid of them. We wanted an alternative method, one that would work without much trouble to map users.

Code refactoring and restructuring.
Support for manual captcha solving using the browser and some tricks with javascript :)

## Description

Provides a way to manually solve captchas using the browser and a bit of javascript trickery.
This PR also moves some code around so we get a cleaner `search.py` and did some refactoring in captcha handling.

Most of this work is based on #1567 which did a great job on this since it showed us the light at the end of the tunnel 👍
The main difference between this PR and 1567 is that we don't try to solve the captcha while the account is scanning. If an account encounters a captcha and `--captcha-key` is not set, the account is put on a special queue for captcha'd accounts, where it will wait for a token.

@kevdotcom gets the credit for suggesting this new approach a couple of days ago and my first reaction was that this method wouldn't work. But after some tests, we found that it actually worked pretty well and decided to implement it.

The queue is managed by a separate thread that checks every 15 sec the database for token submissions.
For each valid token found in database, the captcha-overseer will launch a new thread that logins into that account and tries to validate the token using verifyChallenge. If successful the account is put back in the account_queue to be reused. If the process fails, the account remains in the captcha_queue waiting for another token.

Also added webhook update messages when captchas are found/solved/failed.

Captcha handling page now displays the number of remaining captcha'd accounts once the token is submitted.

- Moved check_login() and tutorial functions to account.py
- Moved captcha check and handling to captcha.py
- jitterLocation() was moved to utils.py and renamed to jitter_location()

## Motivation and Context

Refactored / moved code around, because search.py was getting to confusing and too much code on a single file is never a good thing. Captchas seemed like a good place to start refactoring since we were adding functionality aswell.

Proper method to solve captchas manually, without having to listen to webhooks or monitoring the status page.
Fixed token "double spending": since we're using a transaction to query for tokens we no longer require mutex locks and we make sure that two instances shall never retrieve the same tokens.

## How Has This Been Tested?

Tested on local machine (win7 x64) and on remote server (ubuntu 16.04). Both running python 2.7.12.
Tested captcha handling on Chrome and Firefox.

image

## Types of changes

## Checklist:


neskk:
I want suggestions for configurable parameters, captcha solving page (inject.js) and webhook update messages ;) Thank you

Obihoernchen:
Add captchas left to inject.js 👍

neskk:
@Obihoernchen It does that already but only after token submission.
But you're right, the number of captchas remaining should be displayed at all times in inject.js 👍

Obihoernchen:
oh yeah I just saw that you added this message :D

neskk:
Bug fixed: JSON encode wouldn't accept Decimal types, so added explicit cast to `int`.

Separated stats retrieval from token submission. Stats are now displayed independently of token submission.
base-captchas

paraselene:
Nice work, I have closed #1567 since I don't have time to maintain it. Thanks @neskk.

sLoPPydrive:
First of all: Awesome! :D People always wanted to have this exact functionality, not being limited to the few seconds in which they have to solve the captcha or else it times out.

Personally I found captchas were increasing when swapping out accounts, either on a regular basis or because the captcha failed, so I implemented so-called "captcha strategies" in my personal fork (branch "MIX_10" if anyone likes to check it out), most noticeable the "manual-first" strategy (besides "manual-only" or "automatic") which would wait for a token in the database like #1567 did and after the timeout it would solve the Captcha via 2Captcha, all without logging out the account. I also added a command for print-status ('c') to rotate through the captcha strategies, so that when I am at my computer I set it to manual-first because I have the time to solve the captchas and when I'm on the run I set it to automatic, all without restarting the scanner(s). It was a perfect solution for me, still running against 0.45 API, so I might try to add my captcha strategies to this one as well. :-)

Great work! Thank you, @neskk !

pogom/search.py
```diff
@@ -904,72 +911,43 @@ def search_worker_thread(args, account_queue, account_failures,
...
+
+                            if args.webhooks:
+                                wh_message = {'status_name': args.status_name,
+                                              'status': 'encounter',
+                                              'account': status['username'],
+                                              'captcha': status['captcha'],
+                                              'time': 0}
+                                whq.put(('captcha', wh_message))
+
+                            account_queue.task_done()
```
sLoPPydrive: Does this make sure that the currently scanned location is being retried by the next account? I assume it does because only `scheduler.task_done(status, parsed)` marks a location as "done", am I right?

pogom/models.py
```diff
@@ -1043,7 +1063,7 @@ class WorkerStatus(BaseModel):
     fail = IntegerField()
     no_items = IntegerField()
     skip = IntegerField()
-    captcha = IntegerField(default=0)
+    captcha = IntegerField()
```
sLoPPydrive: Why remove the default value? Having 0 as default seems to be reasonable.

bbdoc:
Hi @neskk . Wonderful PR... I have been using 1567 for quite some time and this looks like a very good way forward.

May I suggest a few things : 
- Would be very nice to be able to still set a captcha-key, but together with a some threshold. E.g. I want my accounts to be put on hold waiting for manual captcha, but go to 2captcha if more than xxx accounts are awaiting a captcha.
- Is there a way to not show "working accounts" ? Don't really want my users to know how many accounts are scanning.
- Will it be possible to localize "Working Accounts", "Remaining Captcha" and "Failed accounts" ? Only 3 lines of text but would be better if we could get it in local language. Or is it simply a template to update ?
- Maybe asking too much here but the last missing piece would be to be able to show remaining captchas directly on the map (e.g. on top of it) ;-)

2 questions as well : 
- What are "Failed accounts" ?
- Are the counts shown on the same webpage as captcha solving ? Does it mean we still need to go through the whole bookmarlet stuff or not (create favorite, replace with code and clic twice) ?

Thanks

walaoaaa1234:
This PR is really awesome.
If using mobile app to solve captchas more wonderful.

pogom/search.py
```diff
@@ -904,72 +911,43 @@ def search_worker_thread(args, account_queue, account_failures,
...
+                                                   captcha_url, account,
+                                                   account_failures, whq):
+                                # Make another request for the same location
+                                # since the previous one was captcha'd.
+                                scan_date = datetime.utcnow()
+                                response_dict = map_request(api, step_location,
+                                                            args.no_jitter)
+                        else:
+                            status['message'] = ('Account {} has encountered' +
+                                                 'a captcha.').format(
```
blueberry0101: This sentence is missing a space, I guess because of the + you didn't notice it.

pogom/search.py
neskk: Yes, you're right.
Right now it's the schedulers job to detect bad scans and to retry the search item if it failed.
I also wanted to have a look at this, maybe another PR, because, "conventional" schedulers right now don't have bad scan detection/handling.
This PR tries to change as little as possible from the default behavior we had previously.

pogom/models.py
neskk: None of the other fields have default values. This is handled well by the code, so my OCD made me remove that ;)

pogom/search.py
neskk: Thank you, adding a fix to next commit 👍

neskk:
@bbdoc I've worked on a hybrid captcha mode previously just as @sLoPPydrive did. Right now I just wanted to offer a simple alternative to 2captcha.

- I like your idea of having some kind of threshold that once reached, enables 2captcha solving. Maybe for a future PR?
- It's pretty simple to hide "working accounts", all you need to do is edit "templates/inject.js" and remove a couple of lines.
- I don't really know how localization is being handled in this project so I can't help you with that, but I can tell you that it also comes down to just a few edits in "templates/inject.js".
- I thought about showing captcha stats in main page but I got lazy and just made the stats for captcha page. It's fairly simple to add this because everything you need is already being sent by the server via /get_stats.
- Failed accounts, are accounts that were added to account_failure list, usually this means banned, but can be other errors as well.

Right now stats are only shown in captcha page, it's fairly simple to change this, but right now I don't have much time left. If anyone helps me with the markup/style to put "remaining captchas" on map.html I'll do it today :)

averylarry:
"Waiting 15 seconds before next token query."
Do we really need to see that in the logs at log.info level every 15 seconds?  Feels like a verbose/debug level log to me.

neskk:
@averylarry you're absolutely right, forgot that, I was using it for debug.

pogom/captcha.py
```diff
@@ -0,0 +1,232 @@
...
+    if args.proxy:
+        # Try to fetch a new proxy
+        proxy_num, proxy_url = get_new_proxy(args)
+
+        if proxy_url:
+            log.debug('Using proxy %s', proxy_url)
+            api.set_proxy({'http': proxy_url, 'https': proxy_url})
+
+    # Jitter location up to 100 meters
+    location = jitter_location(location, 100)
```
averylarry: Why jitter 100m?  You'll have to wait 10s after doing this to allow for moving 100m back to where you were.

neskk: I thought it would be strange if the attempt to uncaptcha the account was made in the same spot as the original map_request(). Maybe 100m is too much, I can adjust later, but the account has plenty of time between uncaptcha and the next scan because the login process will be repeated.

averylarry: I don't see any reason why that would be strange.

The login process doesn't take any time if the account is still authorized.

evilmumi:
Having the problem that the captcha Website Show 0 to solve 0 working etc. 
And Status Website dont work

spammer23:
Same here, and Log Shows:
2017-01-27 12:27:35,235 [       Thread-91][           app][   ERROR] Exception on /get_stats [GET]
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/pi/PokeTest/RocketMap/pogom/app.py", line 67, in get_account_stats
    stats = MainWorker.get_account_stats()
  File "/home/pi/PokeTest/RocketMap/pogom/models.py", line 1053, in get_account_stats
    dict = {'working': int(account_stats[0]),
TypeError: int() argument must be a string or a number, not 'NoneType'

bbdoc:
Little stupid thing but could you make the box a big bigger so it doesn't wrap text when you have more than 100 workers (at least on Android) ?

averylarry:
@evilmumi @spammer23 Make sure you have -mcd set correctly . . ?

evilmumi:
I add -mcd http://ghetthhfgifft.no-ip.info/pgo but on the bookmarklet Page the Link is... No-ip. Info: nö pgo and a :   so i changed it manually but no Chance

spammer23:
@averylarry it is correct, i can resolve captchas - but this message is spamming my log until i close the website

neskk:
@evilmumi @spammer23 This is a mistake I made. The database is not being updated as it should be, causing lots of errors because we try to access columns on MainWorker table that aren't there.
Try manually dropping the table "MainWorker", don't worry this hasn't got important information, it should work for most issues. I didn't want to increase the version of the db_schema but I think I need to do it otherwise people will keep getting these errors.

@bbdoc I will adjust the width, thanks for the input.

evilmumi:
Can i than use mcd with slash?  Domain/pgo?

neskk:
@evilmumi yes, you should set -mcd to whatever URL your front-end instance runs. This is necessary in order for captcha tokens to be delivered to the database. You can even set up a "special" instance that only listens to captcha tokens and runs on a specific URL you set using -mcd so that even if front-end is down, captcha tokens can still be delivered.

templates/bookmarklet.html
```diff
@@ -3,11 +3,16 @@
...
 </head>
-<body>
-<script>
-
-document.write("<a href=\"javascript:(function(){if(window.location.href!='http://pgorelease.nianticlabs.com/'){window.location.href='http://pgorelease.nianticlabs.com/'}else{document.getElementsByTagName('head')[0].appendChild(document.createElement('script')).src='"+window.location.protocol+"//"+window.location.hostname+":"+window.location.port+"/inject.js?'+Math.random();}}());\">Solve Recaptcha</a>")
-</script>
-</body>
+  <body style="width:660px; margin:40px auto;">
+    <h3>Don't save THIS page in bookmarks!</h3>
+    <p>The bookmarklet to solve captchas is the link just bellow:</p>
```
sLoPPydrive: It's "below" :-)

sLoPPydrive:
I have 2 things that bug me:

1. Webhooks: My previous implementation was sending webhooks when solving captchas manually (encounter, solved, failed) and was quiet when captchas were solved automatically via 2Captcha. Your implementation is more or less the other way round: When I provide a 2Captcha key I get all the webhook messages and when I solve manually I only get the captcha encounter. I know this depends heavily on your use case but for my personal map it was nice to get direct feedback via webhook when I solve a captcha manually and I don't want to get ANY notification if captcha solving was done automatically.

2. The captcha solving page is refreshing on a regular basis. This is good if I am not currently solving a captcha but it is really bad if you do. Refreshing the captcha means you can still click on the images but a click on "Verify" simply does nothing and you have to start over again. I know you cannot just let the captcha solving page as it is because after some time it somehow times out and captchas won't be solved any more. But you could also remove the <div> element where you actually solve the captcha when you do the refresh. And of course refresh only after the maximum time possible.

neskk:
@sLoPPydrive you're right about those things.
Webhook messages need improvement when manual captcha is active, will do this in a couple of hours.

I also want to develop a strategy to use both manual and 2captcha approaches at the same time.
I'm thinking about how to threshold the activation of 2captcha method. It could be done based on a ratio between workers and captcha'd accounts, i.e. if 50% of workers are captcha'd, it would automatically start sending captchas to 2captcha instead of adding more accounts to captcha_queue.
Another approach would be time-based, e.g. after waiting 2 hours for captcha token it would remove the account from captcha_queue and process it using 2captcha.
I really don't know what is the best method, maybe a conjugation of both?

Captcha refresh page is set to a reasonable value, I know it can reset the answer to the current captcha you're solving. 30secs refresh seemed a good guarantee that tokens will almost always arrive valid for the account to uncaptcha. I wanted to fiddle with this limit, obviously pushing it as high as possible while ensuring token's validity, but I need empiracal data, feedback. Have you tried this PR? All tokens validate? Can you set the refresh timer to an higher value and keep checking if tokens are still valid?

bbdoc:
@neskk Imo, it would be good to have 2 switchs like 
--max-manual-captcha-workers 
--max-manual-captcha-time

First one would send to 2captcha if you reach a certain number of worker (I think it's better than %, we all know how many accounts and workers we have when launching our scripts). Second one would send it of a certain time.

It's important if using those that to take care that : 
- older captcha is always solved first.
- each instance has its own count if running multiple instances, as those switch would be set at the instance level. Don't know if your PR is currently able to differentiate between instances.

evilmumi:
I tried now with a fixed Port without /map or so. Because it wont work for me. Now i good still 3zeros and Spam at log and no captcha solve will come to database. How can i help to fix?
`[2017-01-28 16:31:47,117] ERROR in app: Exception on /get_stats [GET]
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/src/1754/RocketMap/pogom/app.py", line 87, in get_account_stats
    stats = MainWorker.get_account_stats()
  File "/usr/src/1754/RocketMap/pogom/models.py", line 1053, in get_account_stats
    dict = {'working': int(account_stats[0]),
TypeError: int() argument must be a string or a number, not 'NoneType'
2017-01-28 16:31:47,117 [      Thread-141][           app][   ERROR] Exception on /get_stats [GET]
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/python/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/src/1754/RocketMap/pogom/app.py", line 87, in get_account_stats
    stats = MainWorker.get_account_stats()
  File "/usr/src/1754/RocketMap/pogom/models.py", line 1053, in get_account_stats
    dict = {'working': int(account_stats[0]),
TypeError: int() argument must be a string or a number, not 'NoneType'`

DragonHeart69:
captcha
The only page i load @ this moment is the captcha page. Tis eat my memory... & the permanent reload is very irritating. Can this made that only the stats refresh, without reload all?

neskk:
@evilmumi as I've said before, you need to drop database table MainWorker because it wasn't modified as it should. Drop the table (delete it) MainWorker, then restart the server - the table will be recreated and I think those errors will stop.

@DragonHeart69 Don't know what you mean, I didn't noticed any major RAM consumption change, also Yandex is using 1.3gb and that is not from this PR. 157Mb of RAM for python sounds about right, I don't see anything wrong so please formulate better your comment.

kevdotcom:
@bbdoc 

"older captcha is always solved first."

This is already the case as we are using Python's queue module for the captcha'd accounts.

DragonHeart69:
@neskk The only open tablat in yandex is the page for captcha's. After 15min, i have 1.2gb ram need. @ this moment the page is so slowly, I have to close the page and re-open

neskk:
@DragonHeart69 maybe use a decent browser? Try Firefox or Chrome... there's no reason for captcha page to behave so badly.
Page reload is annoying but prevents invalid tokens from being sent. I must add a timer on page so people know that a refresh is inbound.
There's lots of things not necessarily connected to this PR that can be causing that issue...

evilmumi:
@neskk use pokemongomapdb;
Drop table mainworker;
Table g0ne Problem still there

neskk:
@evilmumi I can almost guarantee you that this is being cause by a problem in database. It tries to fetch information from MainWorker table, and fails at line 1102 of models.py because the value retrieved is None. Can you check the MainWorker table and verify that it has three fields starting with `accounts_` ? If it doesn't, that's what's causing the problem. Otherwise, I don't have a clue why it isn't working for you :/ I've tested this on Ubuntu and Windows, and it worked perfectly.

evilmumi:
@neskk i think Here is the problem. Table is empty. 

mysql> use pokemongomapdb
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> show tables
    -> ;
+--------------------------+
| Tables_in_pokemongomapdb |
+--------------------------+
| gym                      |
| gymdetails               |
| gymmember                |
| gympokemon               |
| mainworker               |
| pokemon                  |
| pokestop                 |
| scannedlocation          |
| scanspawnpoint           |
| spawnpoint               |
| spawnpointdetectiondata  |
| token                    |
| trainer                  |
| versions                 |
| workerstatus             |
+--------------------------+
15 rows in set (0.00 sec)

mysql> select * from mainworker
    -> ;
Empty set (0.00 sec)

neskk:
@evilmumi I think what's causing this problem is that you probably haven't set a `-sn`/`-status-name` for that instance, which disables database status updates. I'm adding some fixes to prevent this error, but ultimately if you want this to work you need to set `--status-name` so that statistics are updated in database.

evilmumi:
@neskk Thank you. Now no error any more.  Status Show me all worker need captcha. 
Captcha site Show all workers as active AND failed no missing captcha. 
I can solve captcha but worker still unactive.  Please help one more. We are near my friend.

neskk:
@evilmumi you need to enable captcha solving ;) `-cs`/`-captcha-solving`

pogom/captcha.py
neskk: Removed this... now it check if `args.no_jitter is True` and uses default jitter distance: 10m

evilmumi:
@neskk works like a charm. Thank you

pogom/models.py
sebastienvercammen: The fields defined here are used to created the tables when you're using a new database, so I wouldn't just remove a default value. We should recheck, and perhaps it's best to give all of them explicit default values so there's no confusion.

evilmumi:
@neskk now i have the same problem as @sLoPPydrive with refresh while solving

stdu00:
Hi, nice work, it is working great.
But would it be possible to switch to 2captcha if the captcha is not solved manually during a specific amount of time (manual_capatcha_timeout)
Thx

PokemonCatcher:
I got this working today. Very nice feature… definitely at the top of my wish list!

Some comments:
- Updated/consolidated documentation is required. The hints in this thread and in #1567 are unclear and/or obsolete. 
- If the capture image iframe is visible, form reload must be suppressed. Otherwise if you get a tricky multi-stage captcha (like the street numbers), the page reloads and you cannot submit it.
- I am still unclear how this is supposed to be configured with one -os thread and several -ns threads (needs to be in the documentation).
- The status page needs to be updated when a worker has been un-captchad. Something like PR#1778.

Great work, thank you for developing this feature. It is SO much better than logging in to each account on a phone!

evilmumi:
@neskk i edited this two lines in inject.js
```
//setTimeout(refreshStats, 1000);

//setInterval(refreshCaptcha, 30000);
```
looks better now with(out) refresh

walaoaaa1234:
1st time load
after solving some captchas
(chrome) memory keep increasing

neskk:
I am aware of that memory leak, fixing it and adding several other features.

bbdoc:
@neskk I see you just implemented a --manual-captcha-timeout. Do you still have plans to enable a mode in which it would go to 2captcha based on number of workers awaiting captcha ? Typically I use -w40 with 60 accounts, so don't want more than 20 accounts awaiting captcha. Time is not really relevant...

neskk:
@bbdoc time is relevant, if you can average how many captchas you have per hour you can set a comprehensive timeout that suits you. Still ironing last bugs, will think about it when I'm finished.

bbdoc:
Of course I can start working on some kind of average, but that still means sometimes it will use 2captcha when it's not really needed, and some time I won't have enough accounts anymore because it didn't get to 2captcha quickly enough... I've seen my number of captcha multiplied by 3 some days compared to the previous day because Niantic changes Something, which will simply make any kind of average based on past result totally irrelevant.

Would be some much more logical to base my settings on the number of workers I need to keep busy on my map so I still can reach 100% spawns, and imo, does correspond to what most people are doing with their map (using -w with some spare accounts).

sebastienvercammen:
@bbdoc You can always add more accounts as "spare" for when your captcha rate goes up. This makes more sense than having to predict timeout based on captcha rate, which would inevitably approach 0s timeout for manual captcha solving when captcha rate goes up past a certain point.

templates/inject.js
```diff
@@ -0,0 +1,76 @@
...
+    document.getElementsByTagName('head')[0].appendChild(script);
+
+    script = document.createElement('script');
+    script.src = 'https://www.google.com/recaptcha/api.js';
+    script.type = 'text/javascript';
+    document.getElementById('recaptcha').appendChild(script);
+    timer = {{timer}};
+}
+
+function refreshCaptcha() {
```
walaoaaa1234: unused function

templates/inject.js
```diff
@@ -0,0 +1,76 @@
...
+            last_res = res;
+            $.post('{{domain}}/submit_token', {token: res}, function(data) {
+                if(data == "ok"){
+                    $('#status').text('Captcha token submitted.');
+                } else {
+                    $('#status').text('Failed to submit captcha token.');
+                }
+                $('#status').fadeIn(200);
+            });
+            setTimeout(initCaptchaPage, 1500);
```
walaoaaa1234: setTimeout(refreshStats, 1500);

pogom/search.py
```diff
@@ -1153,42 +1147,7 @@ def gym_request(api, position, gym):
...
-    if args.captcha_key is None:
-        token_needed += 1
-        if args.webhooks:
-            whq.put(('token_needed', {"num": token_needed}))
-        while request_time + timedelta(seconds=args.manual_captcha_solving_allowance_time) > datetime.utcnow():
-            tokenLock.acquire()
-            if args.no_server:
-                # multiple instances, use get_token in map
-                s = requests.Session()
-                url = "{}/get_token?request_time={}&password={}".format(args.manual_captcha_solving_domain, request_time, args.manual_captcha_solving_password)
```
j16sdiz: I notice you removed this "get_token" function to retrieve token from remote db..
We have a few map on different server with different database that want to share the same capthcha solving page... 

To make the code simpler, maybe we can provide a 2captcha-like api and allow the 2captcha code use alternative api url?

drallieiv:
how do you use bookmarklets on mobile ?

Is anyone working on a desktop or mobile app that just calls the API and runs it's own chromedriver locally ?

neskk:
@drallieiv it's possible to use it on mobile phones, I've only tested with Chrome (mobile) and you have to manually add a bookmark with the javascript code, for instance, Name: `Solve Recaptcha`, URL: `javascript:...`. After that, just start typing `Solve Recaptcha` on the URL field of the browser, it should auto complete and once you click it click the effect should be the same. Remember you have to repeat the process, only on second click on the bookmarklet it will show up the captcha page.

bbdoc:
@neskk I'm coming back on this one as there seems to be Something going wrong with page refresh. When I click on my bookmark twice, I do get to the captcha solving page and see correct stats (accounts/captcha/fails) Under it. Anyway, if I solve a captcha, it refreshes the page and stats are changed to 0/0/0. If I want the correct count again, I have to either wait until the end of the countdown so it refreshes again, or close the page and click twice on the bookmarklet favorite again. Anything than can be done so that it refreshes with the correct stats after solving a captcha ?

neskk:
Statistics are refreshed after the refresh timeout (every 30sec) or whenever you submit a captcha token. This means that what you're asking is already being done. What can happen is that the request for statistics wasn't answered and so the stats will show 0/0/0. This is a webserver problem and the more refreshes you do, more stress is put in the webserver instance.

bbdoc:
Statistics are always refreshed correctly on first visit using the favorite, as well as after the refresh timeout, but are NEVER refreshed after I submit a captcha. I don't think this can be a webserver problem as I was testing that being alone on the server with only a few captcha every hour, no real stress on the web interface. At least this is what is happening on my android phone, tested both on the default browser and chrome.

neskk:
@bbdoc you're probably not using the latest version - git pull. I say this because a similar issue has been addressed recently.

bbdoc:
Indeed still on the commit from Feb 3... I don't update systematically if I don't see anything of interrest for me in latest commit... didn't see anything related to that in the latest commits but I'll give it a try. Or did you change something in the script itself ? I'm still using a favorite I had created before the PR was commited...

PR #1759, opened 2017-01-26
Add spawnpoint type to pokemon webhook data.

sebastienvercammen:
## Description
Moved some duplicate code out to its own function, and included spawnpoint type in the pokémon data being sent.

## Motivation and Context
Sending tth_found only in `scheduler` webhook messages isn't enough: timing doesn't always match, and we can't guarantee we can match pokémon spawn data with `scheduler` messages, so we can't retrieve the exact type (spawnpoint status).

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1574,7 +1574,7 @@ def hex_bounds(center, steps=None, radius=None):
 
 # todo: this probably shouldn't _really_ be in "models" anymore, but w/e.
 def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
-              api, now_date):
+              api, now_date, tth_found_pct):
     pokemon = {}
     pokestops = {}
     gyms = {}
@@ -1789,7 +1789,8 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'disappear_time': calendar.timegm(
                         disappear_time.timetuple()),
                     'last_modified_time': p['last_modified_timestamp_ms'],
-                    'time_until_hidden_ms': p['time_till_hidden_ms']
+                    'time_until_hidden_ms': p['time_till_hidden_ms'],
+                    'tth_found': tth_found_pct
                 })
                 wh_update_queue.put(('pokemon', wh_poke))
 
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1574,7 +1574,7 @@ def hex_bounds(center, steps=None, radius=None):
 
 # todo: this probably shouldn't _really_ be in "models" anymore, but w/e.
 def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
-              api, now_date, tth_found_pct):
+              api, now_date):
     pokemon = {}
     pokestops = {}
     gyms = {}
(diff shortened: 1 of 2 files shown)


neskk:
I'm against this change and my previous webhook rework was to fix this.

- We'll be sending the same data twice, one thing is if you sent the mon's specific tth, that I could agree with, but you're sending the same tth_found percentage that is sent from wh_status_update().

- If you need higher update frequency on scheduler updates that once every second I feel something is not being done correctly.

- This breaks the standard Pokemon webhook message format, causing webhook implementations to start dumping warnings/errors about an extra field.

- My previous argument also was that after tth_found stabilizes you're sending redundant data to webhook.

Maybe generate an hash ID for each scheduler so you can better distinguish them. This is definitely a feature that my previous PR fell short with, but it shouldn't be too hard to do.

sebastienvercammen:
@neskk 


1. The pokémon's TTH is already sent, it's one of the standard fields always received from Niantic's API, but these changes are for our public webhook server which can use it for analysis/validation (even in the long term).
2. As explained in more detail on Discord, the implementation of "scheduler updates" falls very short in practice: in a controlled test with a minimal amount of maps sending data, we hit over 25000 incoming requests per minute (non-scheduler data) that couldn't reliably match with a scheduler update.
3. "Webhooks should not be smart" was the basis of their implementation. Webhooks are already duplicate data being sent where that is preferable over giving someone access to your database (or cloning it).
4. 2+3 = with a large amount of incoming requests, a minimal amount of extra bandwidth is _much_ more scalable than having to do extra database queries for every single request. Using identifying data (e.g. hash IDs) for each scheduler would not only increase the size of the data sent (although less frequently), it still wouldn't solve the scalability problem. Accessing request data is simply much more scalable than having to do custom matching.
5. Scheduler updates aren't just TTH and have no relation to the existing webhook messages. If someone wants to be able to disable scheduler updates and they still want to contribute to the public webhook server, there's no reason that shouldn't be possible.
6. Any implementation that breaks when a single extra field is added, is a wrong implementation.

Using a single extra field (whether it's global TTH percentage or TTH status for the current pokémon doesn't matter, either works) is a fair compromise.

neskk:
@sebastienvercammen
You're **not** sending pokemon's TTH data, you're sending the overall tth_found percentage which is schedulers data, it's not the same thing.
This also makes the scheduler update messages that we send completely useless and might just send that function to the garbage.
Scheduler update messages could be sent at schedulers discretion. I don't understand the need to change pokemon webhook message to include info not related to that pokemon. For instance, whenever the scheduler finds a new tth, it could send a message with poke_id, poke_tth and overall_tth. I feel this is a much cleaner way to achieve what you're trying to do, but this is just my opinion.

sebastienvercammen:
@neskk 

You should re-read what the code does:
```
'time_until_hidden_ms': p['time_till_hidden_ms'],
```
TTH stands for "time till hidden".

Scheduler updates have a different scope and can be adjusted to report many more scheduler fields than what it's doing now: it still serves its purpose.

Adjusting global TTH to TTH status for that specific pokémon is trivial, as was also discussed in Discord. That should solve your problem with the data not being related to only that Pokémon.

Webhooks also already filter out duplicate requests and only send the message when something significant has changed, this TTH status field would just become one of the significant fields - it still wouldn't send anything duplicate, as it wouldn't send it if nothing has changed.

neskk:
@sebastienvercammen you're right, I failed to see that sorry.
BUT, you're sending both TTH for pokemon (which I'm ok with) an schedulers tth_found_pct which will be duplicate data over wh_status_update() and will never be considered a duplicate since it's totally a different message.
Right now, the only thing I'm strongly against is sending tth_found_pct in a pokemon wh message.

sebastienvercammen:
@onilton Moved from tth_found_pct to spawnpoint status as reported by the model itself. Does that fix your concerns?

neskk:
Sorry for being picky, for me this makes much more sense. 👍

PR #1762, opened 2017-01-26
Create hashing.md

jagauthier:
I wrote this up to answer some questions about hashing I've been seeing in #help.  It would be nice to add a !hash macro that could reference this as well.

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/hashing.md
+++ b/docs/extras/hashing.md
@@ -1,0 +1,36 @@
+# Hash Keys
+
+## What are hash keys for?
+Up until API version 0.45 the pgoapi was able to directly access Niantic servers and make calls.
+Because of the complex cat and mouse game played, Niantic upped the complexity with API 0.51.  The only way to access that API was through a hash key.
+Legacy 0.45 could still be accessed.
+on January 19th, 2017 Niantic did a force update of the Pokemon Go Client. This forced the actual clients to use API version 0.53.  This also requires a hash key.
+At this writing legacy 0.45 is still available.
+
+However, with a forced update Niantic can easily see clients authenticatin using 0.45. These accounts can/will be banned.
+
+It is recommended to use hash keys to replicate authenticity of your client and reduce in-game captch costs.
+
+## Where do I get a hash key?
+[Check out this FAQ](https://talk.pogodev.org/d/55-api-hashing-service-f-a-q)
+
+## How many RPMs will I use?
+There is no perfect way to know. There are many variables that must be considered.  Including your step size, spawn spoints, encounters.
+
+Please don't ask *"What if my step size is _x_, and I have encounters for _y_ Pokemon"*
+
+We still don't know.  Get a key, turn on your map and see if it works.
+If you are getting rate limited then either get more keys, or lower your calls (encounters, step size, etc)
+
+
+## What does HashingQuotaExceededException('429: Request limited, error: ',) mean?
+Any variant of this means you've exceeded the Requests Per Minute that your hashing key allows
+
+## How about [ WARNING] Exception while downloading map: HashingOfflineException('502 Server Error',)
+Hash server is most likely offline.
+
+## And this one? BadHashRequestException('400: Bad request, error: Unauthorized',)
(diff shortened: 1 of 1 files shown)


docs/extras/hashing.md
```diff
@@ -0,0 +1,36 @@
+# Hash Keys
+
+## What are hash keys for?
+Up until API version 0.45 the pgoapi was able to directly access Niantic servers and make calls.
+Because of the complex cat and mouse game played, Niantic upped the complexity with API 0.51.  The only way to access that API was through a hash key.
+Legacy 0.45 could still be accessed.
+on January 19th, 2017 Niantic did a force update of the Pokemon Go Client. This forced the actual clients to use API version 0.53.  This also requires a hash key.
+At this writing legacy 0.45 is still available.
```
sebastienvercammen: Rewrite this block, it needs to be simplified.

These things are largely irrelevant:
```
Up until API version 0.45 the pgoapi was able to directly access Niantic servers and make calls.
Because of the complex cat and mouse game played, ...
Legacy 0.45 ... ("legacy 0.45" isn't a thing)
```
We're not trying to be a history book (if that _was_ your intention, perhaps a HISTORY.md instead?).

docs/extras/hashing.md
```diff
@@ -0,0 +1,36 @@
...
+
+However, with a forced update Niantic can easily see clients authenticatin using 0.45. These accounts can/will be banned.
```
sebastienvercammen: Typo in `authenticatin`, and speculation:
```
These accounts can/will be banned.
```
This is also incorrect, as it doesn't require an update, Niantic always knows what version you're using:
```
with a forced update Niantic can easily see clients authenticatin using 0.45
```

docs/extras/hashing.md
```diff
@@ -0,0 +1,36 @@
...
+
+It is recommended to use hash keys to replicate authenticity of your client and reduce in-game captch costs.
```
sebastienvercammen: Rewrite:
```
It is recommended to use hash keys
```
We don't recommend anything, because we still would like to avoid getting users to pay for hashing servers if we could. It's better to rephrase to explain the consequences (captcha and ban rate) instead.

docs/extras/hashing.md
```diff
@@ -0,0 +1,36 @@
...
+## Where do I get a hash key?
+[Check out this FAQ](https://talk.pogodev.org/d/55-api-hashing-service-f-a-q)
+
+## How many RPMs will I use?
+There is no perfect way to know. There are many variables that must be considered.  Including your step size, spawn spoints, encounters.
+
+Please don't ask *"What if my step size is _x_, and I have encounters for _y_ Pokemon"*
+
+We still don't know.  Get a key, turn on your map and see if it works.
+If you are getting rate limited then either get more keys, or lower your calls (encounters, step size, etc)
```
sebastienvercammen: Rewrite
```
or lower your calls (encounters, step size, etc)
```
to be more descriptive:
```
or lower your calls (disable encounters, disable gym details, decrease step size, etc)
```

docs/extras/hashing.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+## What does HashingQuotaExceededException('429: Request limited, error: ',) mean?
+Any variant of this means you've exceeded the Requests Per Minute that your hashing key allows
+
+## How about [ WARNING] Exception while downloading map: HashingOfflineException('502 Server Error',)
+Hash server is most likely offline.
```
sebastienvercammen: Rewrite
```
Hash server is most likely offline.
```
to:
```
Hashing server is temporarily unavailable (possibly offline).
```

jagauthier:
I'm fine with these changes.   But have you read this? 
https://pgm.readthedocs.io/en/develop/extras/faq.html
Not exactly technical, or user friendly.

sebastienvercammen:
@jagauthier 

1. It's not because someone else did something wrong, that you should as well.
2. That FAQ page is actually much more to the point than what you submitted.

But thanks for updating.

PR #1774, opened 2017-01-28
Fix scheduler trying to fetch items every second

neskk:
Quick fix for schedulers that run on hexes without spawnpoints.
## Description

This a **ugly** quick fix to prevent schedulers from hammering the database requesting new search items every second when there aren't any spawnpoints in that hex.
This can happen when using beehive mode with small step_size.

## Motivation and Context

Needs improvements, but this allows 3 attempts for schedulers to grab items, if they fail consecutively more than that, they won't be calling .schedule() anymore.

## How Has This Been Tested?

Testing on my server running an instance on a more remote location where some of the hexes in the beehive, unfortunately, don't pick up any spawnpoints. `-speed -bh`

base-schedulers
base-schedulers-status

## Types of changes

## Checklist:


neskk:
`xxxx	70	0	32	0	2	17:26:49 28 Jan 2017	Search aborting. Overseer refreshing queue.`

Started again with a clean database and still got the same result, I think this is really a bug and not just my setup going haywire.
Running this on all my instance and it didn't affect normal behavior.

neskk:
This breaks the initial scan because database is empty. Will need to figure other way to fix the problem when there's no spawnpoints in a scanned area.

PR #1780, opened 2017-01-28
Major webhook rework.

sebastienvercammen:
## Description
Changes how webhooks work (also made them thread safe), how requests are sent, and how we determine when the queue isn't being processed fast enough and the user should increase settings.

**From now on, it is highly recommended to leave `--wh-threads` at `1` (default).** If you have some sort of crazy setup and you infinitely get the error `try increasing --wh-concurrency or --wh-threads` (i.e. the queue never goes below 100), increase `--wh-concurrency`. If you need more than `--wh-concurrency 50`, you can consider adding an additional thread.

**Seeing the error once is fine.** Please don't increase anything if you see the error once and it instantly goes away. Only when the error never goes away, you should increase concurrency (as explained above).

**Changes:**

1. Optimized default settings (1s timeout, 3 retries instead of 5).
2. Made webhooks thread safe again. This fixes infinite queuing on LFUCache error.
3. Reworked the queue size error to use time: it will only log an error if the queue size has been over the limit for at least 5 seconds.
4. Moved synchronous `requests` to asynchronous `requests_futures`.
5. We now use a single Session object instead of creating one per request. This is not only more efficient in terms of re-using objects; requests to the same hosts will re-use the same underlying TCP connection instead of requiring a new one.

## Motivation and Context
Gives major performance increase, better throughput, and rather than requiring 25 threads, the test now only needed a single thread and default settings to keep the queue empty.

## How Has This Been Tested?
Tested extensively with Barlok on a large setup (st 26, w 100) and intentionally with two receiving webhook servers rather than only one.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -3,12 +3,21 @@
 
 import logging
 import requests
+from datetime import datetime
 from requests_futures.sessions import FuturesSession
+import threading
 from .utils import get_args
 from requests.packages.urllib3.util.retry import Retry
 from requests.adapters import HTTPAdapter
 
 log = logging.getLogger(__name__)
+
+# How low do we want the queue size to stay?
+wh_warning_threshold = 100
+# How long can it be over the threshold, in seconds?
+# Default: 5 seconds per 100 in threshold.
+wh_threshold_lifetime = int(5 * (wh_warning_threshold / 100.0))
+wh_lock = threading.Lock()
 
 
 def send_to_webhook(session, message_type, message):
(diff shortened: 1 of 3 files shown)


pogom/webhook.py
```diff
@@ -116,6 +123,41 @@ def wh_updater(args, queue, key_cache):
 
 # Helpers
 
+# Background handler for completed webhook requests.
+# Currently doesn't do anything.
+def __wh_completed():
```
trew: Why is this included if it doesn't do anything? Future use?

sebastienvercammen: Calling `.response` makes the process synchronous, because it would wait for the response to be complete. Using the background handler keeps it entirely async, and indeed allows for future use.

FrostTheFox:
+1 map didn't die

PR #1787, opened 2017-01-31
Niantic API force Watchdog

catnet23:
* added watchdog that check if niantic force a update
* if niantic force a update it will stop scanning

## Description

This Watchdog checks every given time, if niantic has forced a new API since you started the scanners.
## Motivation and Context

I did this PR because if i'm not at home and Niantic force a new API and i don't have any chance to stop the scanner this PR paused the scann.

## How Has This Been Tested?

Tested on my Map with a local Webserver to trigger the API Version changes.

$ echo -ne '\x0a\x060.53.0' > /tmp/index.html
$ php -S 0.0.0.0:1337 -t /tmp/

Replaced Niantic Version URL with my Local Server and start Scanner
Change local API Version
$ echo -ne '\x0a\x060.53.13' > /tmp/index.html

EDIT: perfect example today
```
2017-02-08 16:51:40,198 [ search-overseer][        search][    INFO] Enabling new API force Watchdog.
2017-02-08 19:49:04,504 [ search-overseer][        search][    INFO] Started with API: 0.53.0, Niantic forced to API: 0.55.0
2017-02-08 19:49:04,507 [ search-overseer][        search][    INFO] Scanner paused due Niantic API foce.
2017-02-08 19:49:04,514 [ search-overseer][        search][    INFO] Stop the scanner process until RocketMap has updated.
```

api_force
api_paused

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -354,7 +354,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
-    curr_api_version = '0.53.0'
+    curr_api_version = '0.55.0'
     api_check_time = 0
 
     '''
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -24,6 +24,11 @@ import sys
 import traceback
 import random
 import time
+import geopy
+import geopy.distance
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 import copy
 
 from datetime import datetime
@@ -350,6 +355,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
+    cur_plfe_version = None
 
     '''
     Create a queue of accounts for workers to pull from. When a worker has
@@ -420,6 +426,21 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
         t = Thread(target=worker_status_db_thread,
                    name='status_worker_db',
                    args=(threadStatus, args.status_name, db_updates_queue))
+        t.daemon = True
+        t.start()
+
+    # Create the key scheduler.
+    if args.hash_key:
+        log.info('Enabling hashing key scheduler...')
+        key_scheduler = schedulers.KeyScheduler(args.hash_key).scheduler()
+
+    # Create API version check thread
+    if args.plfe_version_check:
+        cur_plfe_version = get_plfe_version(args)
+        log.info('Enabling new API force Watchdog')
+        t = Thread(target=plfe_check_thread,
+                   name='niantic_api',
+                   args=(args, cur_plfe_version, pause_bit))
         t.daemon = True
         t.start()
 
@@ -1212,3 +1233,50 @@ def stagger_thread(args):
 # The delta from last stat to current stat
 def stat_delta(current_status, last_status, stat_name):
     return current_status.get(stat_name, 0) - last_status.get(stat_name, 0)
+
+
+class TooManyLoginAttempts(Exception):
+    pass
+
+
+def plfe_check_thread(args, cur_plfe_version, pause_bit):
+    while True:
+        # pause if niantic force a new version (to prevent get banned)
+        forced_api = get_plfe_version(args)
+        if (cur_plfe_version != forced_api and forced_api != 0):
+            pause_bit.set()
+            log.info('Started with API: %s, Niantic forced to API: %s' % (
+                cur_plfe_version, forced_api))
+            log.info('Scanner paused due Niantic API foce.')
+            log.info('Stop the scanner process until RocketMap has updated.')
+
+        time.sleep(args.plfe_version_check_interval)
+
+
+def get_plfe_version(args):
+    proxies = {}
+
+    if args.proxy:
+        num, proxy = get_new_proxy(args)
+        proxies = {
+            'http': proxy,
+            'https': proxy
+        }
+
+    try:
+        s = requests.Session()
+        s.mount('https://', HTTPAdapter(max_retries=Retry(total=3,
+                                        backoff_factor=0.1,
+                                        status_forcelist=[500, 502,
+                                                          503, 504])))
+        r = s.get(
+            'https://pgorelease.nianticlabs.com/plfe/version',
+            proxies=proxies,
+            verify=False)
+
+        return r.text[2:]
+
+    except Exception as e:
+        log.warning('error on API check: %s', repr(e))
+
+        return 0
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -24,12 +24,11 @@ import sys
 import traceback
 import random
 import time
-import geopy
-import geopy.distance
 import requests
+import copy
+
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
-import copy
 
 from datetime import datetime
 from threading import Thread, Lock
@@ -355,7 +354,8 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
-    cur_plfe_version = None
+    curr_api_version = '0.53.0'
+    api_check_time = 0
 
     '''
     Create a queue of accounts for workers to pull from. When a worker has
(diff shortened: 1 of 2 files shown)


tomballgithub:
nice, thanks, trying this

tokar86a:
this sure should be something awesome to have on the map :)

catnet23:
@tomballgithub did some changes if the website was not reachable because the SSL verify it brings in the sleep mode. Changed.

pogom/search.py
```diff
@@ -387,6 +390,16 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
         log.info('Enabling hashing key scheduler...')
         key_scheduler = schedulers.KeyScheduler(args.hash_key).scheduler()
 
+    # Create API version check thread
+    if args.plfe_version_check:
+        cur_plfe_version = get_plfe_version(args)
+        log.info('Enabling new API force Watchdog')
```
sebastienvercammen: For consistency:
```
log.info('Enabling new API Watchdog.')
```

pogom/search.py
```diff
@@ -329,6 +331,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
+    cur_plfe_version = None
```
sebastienvercammen: Rather than storing only the version retrieved at startup, the expected version could be passed as an optional config item (current version by default). This would avoid problems with unexpected instance restarts/crashes.

pogom/search.py
```diff
@@ -1286,3 +1299,46 @@ def stat_delta(current_status, last_status, stat_name):
...
+
+
+def get_plfe_version(args):
+    proxies = {}
+
+    if args.proxy:
+        num, proxy = get_new_proxy(args)
+        proxies = {
+            'http': proxy,
+            'https': proxy
```
sebastienvercammen: We shouldn't be using proxies to retrieve the last version, it doesn't add any value but slows down all requests and can cause issues with unstable proxies. The only instance where it can be useful is when your home/server IP is banned, but even then rotating proxies isn't the best option.

pogom/search.py
```diff
@@ -1286,3 +1299,46 @@ def stat_delta(current_status, last_status, stat_name):
...
+        }
+
+    try:
+        s = requests.Session()
+        s.mount('https://', HTTPAdapter(max_retries=Retry(total=3,
+                                        backoff_factor=0.1,
+                                        status_forcelist=[500, 502,
+                                                          503, 504])))
```
sebastienvercammen: Retries and backoff factor are probably overkill for just status checking. This code can be simplified.

pogom/search.py
```diff
@@ -387,6 +390,16 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
...
 
+    # Create API version check thread
+    if args.plfe_version_check:
+        cur_plfe_version = get_plfe_version(args)
+        log.info('Enabling new API force Watchdog')
+        t = Thread(target=plfe_check_thread,
+                   name='niantic_api',
+                   args=(args, cur_plfe_version, pause_bit))
+        t.daemon = True
+        t.start()
```
sebastienvercammen: This code shouldn't need yet another thread: it's the overseer thread's job to control scheduling and imo this includes these situations where an external factor/flag means the instances should be paused.

I would add a separate ready/status function to the overseer and move this.

pogom/search.py
Imoq: Many people run maps on banned IPs, and that's the main reason why we use proxies in first place. I believe if a proxy is configured on the map, it should be used to get the version as well.

Perhaps not rotating all of them (or perhaps yes? since we are using them to make the requests, anyway).

sebastienvercammen: @Imoq I understand the point of view of hosting on banned IPs. RocketMap was never about large scale and being hosted on paid servers (i.e. small scale, family/friends), and we should remind ourselves of that, but it's also possible for technical users to not want to host their own small maps on their own computers.

Or put differently: if someone appears to argue the need for it, in this case it's enough for me to keep it in. 👍

pogom/search.py
catnet23: So you mean the user should add the API version itself in the config? So he need to check before he start the scanner which version is now actuall. But we can do both like, if someone pass a version we don't need a check on startup. If not, we check on startup.
But i think the most of the user prefer the startup check.

pogom/search.py
catnet23: I need it on my Map, because my Hoster is banned. If no proxy is given no proxy will used.

pogom/search.py
catnet23: you're right, i added this just because my server was no reachable. we can remove it. but if you run the check just hourly, then you will loose one hour.

pogom/search.py
Imoq: I also don't have my home IP anything to do with maps, and many users do that, because they connect with their real accounts on home network and don't want the IPs to be related just in case there is a witch hunting.

In any case, thank you very much Seb, for being understanding!

pogom/search.py
sebastienvercammen: 1. Add current version as a clear, hardcoded constant in the scanner.
2. Add a `--no-version-check` config item. If the version check is not disabled (i.e. we want the version check), the map retrieves the current forced version and only starts scanning if it's the same version as the hardcoded value (on failure, don't start scanning either). Otherwise skip the check and just start scanning (but add a warning to logs that version check is being skipped). 
3. Since every new forced version requires a pgoapi update, updating the hardcoded value each time we update pgoapi is trivial (and most likely the best option).

pogom/search.py
```diff
@@ -523,6 +517,22 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
...
+                forced_api = get_api_version(args)
+                if (curr_api_version != forced_api and
+                   forced_api != 0):
+                    pause_bit.set()
+                    log.info(('Started with API: {}, ' +
+                             'Niantic forced to API: {}').format(
+                                 curr_api_version,
+                                 forced_api))
+                    log.info('Scanner paused due Niantic API foce.')
+                    log.info('Stop the scanner process until RocketMap ' +
```
sebastienvercammen: Move to separate helper function for readability.

pogom/search.py
```diff
@@ -442,6 +433,9 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
         t.daemon = True
         t.start()
 
+    if not args.no_version_check:
+        log.info('Enabling new API force Watchdog')
```
sebastienvercammen: End log sentences with a dot (`Enabling new API force Watchdog.`).

pogom/search.py
```diff
@@ -331,7 +331,8 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
-    cur_plfe_version = None
+    curr_api_version = '0.53.0'
+    api_time = 0
```
sebastienvercammen: Rename `api_time` to something more descriptive.

rndx539:
Since we support the 0.55 API we should wait until Niantic has forced to 0.55.
If we don't wait and Niantic force the new API the scanner will stop.

tobias86aa:
@catnet23 but can you dont add the next one in the code or something?

rndx539:
@tokar86a i can't because Niantic still returns 0.53.0
```
% curl 'https://pgorelease.nianticlabs.com/plfe/version' 

0.53.0
```
And if we change the API in the code to 0.55.0 is not equal to the response from Niantic itself.

pogom/search.py
```diff
@@ -1212,3 +1226,54 @@ def stagger_thread(args):
...
+        api_check_time = int(time.time()) + args.version_check_interval
+        forced_api = get_api_version(args)
+        if (curr_api_version != forced_api and
+           forced_api != 0):
+            pause_bit.set()
+            log.info(('Started with API: {}, ' +
+                     'Niantic forced to API: {}').format(
+                         curr_api_version,
+                         forced_api))
+            log.info('Scanner paused due Niantic API foce.')
```
sebastienvercammen: Typo: "force". Reword to:
```
Scanner paused: Niantic forced new API.
```

pogom/utils.py
```diff
@@ -384,6 +384,13 @@ def get_args():
                         help=("Complete ToS and tutorial steps on accounts " +
                               "if they haven't already."),
                         default=False)
+    parser.add_argument('-nvc', '--no-version-check', action='store_true',
+                        help='Disable API version check',
+                        default=False)
+    parser.add_argument('-vci', '--version-check-interval', type=int,
+                        help='Interval to check API version force in secs ' +
+                        '(default: 300)',
+                        default=300)
```
sebastienvercammen: Rather than querying every 300s, we should add some randomness ([-x, +x]), otherwise it would be too easy for them to monitor and flag IPs.

pogom/utils.py
```diff
@@ -384,6 +384,13 @@ def get_args():
                         help=("Complete ToS and tutorial steps on accounts " +
                               "if they haven't already."),
                         default=False)
+    parser.add_argument('-nvc', '--no-version-check', action='store_true',
```
sebastienvercammen: What do y'all think of using `-no-vc` as short format for disabling rather than `-nvc`, for clarity?

FrostTheFox: -novc

pogom/utils.py
FrostTheFox: mhh randomness

tomballgithub:
This worked for me today, everything stopped at the force update.  I had the older code that was coded to 0.53.0

pogom/utils.py
catnet23: What would you prefer? random.randint(60,300) or do we need a new commandline with the random between value?

sebastienvercammen: I'd rather not add another config item, but considering it'll be one that will be left to default in 99% of cases, I would say it's a necessary evil to maintain code quality (hardcoded values are bad!).

Between 1 to 5 minutes sounds good.

PR #1794, opened 2017-02-02
Display hash key expiration date on hash key status page (-PS)

hokiepokedad:
## Description

With the release 0.55 pgoapi, a new status was included to show the expiration of hash keys.  This will add a new column called "Expires" to the hash key status page on -PS.

## Motivation and Context

Show when keys expire in local time.

## How Has This Been Tested?

Local maps.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -1076,7 +1076,8 @@ class KeyScheduler(object):
             self.keys[key] = {
                 'remaining': 0,
                 'maximum': 0,
-                'peak': 0
+                'peak': 0,
+                'expires': 'N/A'
             }
 
         self.key_cycle = itertools.cycle(keys)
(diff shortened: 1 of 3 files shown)


bbdoc:
More than only displaying it, we need a mechanism to stop using an expired key. Just had the case of my key expiring. I had 2 other valid keys in my config, but PGM continued using the 3 of them with as a result that all accounts stopped scanning as my IP got banned from hashing Server. Any way to exclude an expired key from being used in order to avoid that ?

hokiepokedad:
@bbdoc agreed.  I actually ran into this myself last night.  Yes, I'll include that functionality. In fact I'm thinking of repurposing this PR to move hash keys into models so we can disable keys, store in db, and have a new table on the status web page similar to the hash keys status.  I'll try to get to that today.

hokiepokedad:
merging minds and working on @Alderon86 's PR

PR #1799, opened 2017-02-02
Fix no nearby pokemon warning if --no-pokemon is used

Obihoernchen:
## Description
Reopen of #1708 
- If you scan with --no-pokemon option you don't care for pokemon and there is no speed limit for gyms/stops
  - suppress the warning (`No nearby or wild Pokemon but there are visible gyms or pokestops. Possible speed violation.`) if scanning with --no-pokemon

## Motivation and Context
Annoying message which is incorrect in case of no_pokemon

## How Has This Been Tested?
my map gym/stop scanner

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1720,10 +1720,11 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
             log.info('Common causes: captchas or IP bans.')
             abandon_loc = True
         else:
-            # No wild or nearby Pokemon but there are forts.  It's probably
-            # a speed violation.
-            log.warning('No nearby or wild Pokemon but there are visible gyms '
-                        'or pokestops. Possible speed violation.')
+            if not args.no_pokemon:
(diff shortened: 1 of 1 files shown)


sebastienvercammen:
The error message is not incorrect, even when using `--no-pokemon`. The message is because the scanner didn't receive anything from Niantic's servers, which is an accurate error: `--no-pokemon` only disables the parsing of Pokémon, it doesn't disable receiving their data.

Obihoernchen:
@sebastienvercammen I know how this works but still 50% of this message is useless for scanning with no pokemon:
"No nearby or wild Pokemon" --> I specified no pokemon so I don't care for Pokemon at all
"but there are visible gyms or pokestops." --> This part is fine but I see it in the scan result anyways 
"Possible speed violation" --> wrong, there is no speed violation for gyms + stops. That's totally for Pokemon scanning though. 

I just don't see why it's a desired behaviour to get flooded with warning messages,  if you scan for gyms. There is no reason to scan gyms with <35 kph move speed.

PR #1806, opened 2017-02-04
Hashkey /status page report

Alderon86:
## Description
Displays currently used Hash Keys and their values on the status webpage.
It also gives you the ability to hide or show all the tables which are displayed.
Keys and values are written to db, expired Hash Keys are also removed via db-cleaner.
Displayed Keys on the Webpage will have their last 9 digits obfuscated because of reasons.
This PR also fixes issue #1929 

## How Has This Been Tested?
Several Maps

beaty

## Types of changes

## Checklist:

Thanks to all of #pr awesome Team! <3


Alderon86:
pls dont mind travis right now, this is just the base for new functions

pogom/search.py
```diff
@@ -938,14 +940,14 @@ def search_worker_thread(args, account_queue, account_failures,
...
-                response_dict = map_request(api, step_location, args.no_jitter)
+                response_dict = map_request(
+                    api, step_location, args.no_jitter)
                 status['last_scan_date'] = datetime.utcnow()
 
                 # Record the time and the place that the worker made the
                 # request.
                 status['latitude'] = step_location[0]
                 status['longitude'] = step_location[1]
-                dbq.put((WorkerStatus, {0: WorkerStatus.db_format(status)}))
```
myntath: Is there a reason for this?

Alderon86:
any suggestions? otherwise theres one review missing to get this merged

pogom/search.py
```diff
@@ -912,6 +915,19 @@ def search_worker_thread(args, account_queue, account_failures,
...
+                        'country': 'US', 'language': 'en',
+                        'timezone': 'America/Denver'})
+                    request.call()
+                except BadHashRequestException as e:
+                    pause_bit.set()
+                    log.warning(
+                        'Hash Key {} seems to be expired! {}'.format(key,
+                                                                     repr(e)))
+                    log.warning('Scanning Paused, please restart with' +
+                                'a valid Hash Key!')
```
neskk: You pause scanning because of expired Hash Key but maybe there are others available. This will lead to all areas to pause scanning even if just one hash key has expired because you'll keep rotating through all hash keys and whenever the "bad" one is used the scanning pauses for that worker.
You should move this check to HashKey scheduler or something but don't do this here and not this way.
check_login() could throw BadHashRequestException and if caught in search_worker_thread we would remove this key from scheduler after X fails and repeat the step with another hash key. If there aren't available hash keys in scheduler then you stop the scanning.

pogom/search.py
```diff
@@ -974,6 +991,39 @@ def search_worker_thread(args, account_queue, account_failures,
                         time.sleep(3)
                         break
 
+                    if args.hash_key:
+                        key_instance = (key_scheduler.keys[key])
```
neskk: captcha.py -> captcha_solver_thread() also uses hash keys, please add support there as well.

pogom/app.py
```diff
@@ -90,6 +92,25 @@ def get_account_stats(self):
         r.headers.add('Access-Control-Allow-Origin', '*')
         return r
 
+    def get_all(self):
```
neskk: Why have an extra page when everything can be shown on status? I thought you were including/using HashKeyManager template inside Status page, but I think hashkeys.html is not used by status page and if so, in my opinion, it could be removed. One less entry point/page to check and since all information is available on status I don't see a reason for this page to exist.

pogom/app.py
```diff
@@ -544,6 +565,7 @@ def post_status(self):
             d['login'] = 'ok'
             d['main_workers'] = MainWorker.get_all()
             d['workers'] = WorkerStatus.get_all()
+            d['HashKeys'] = HashKeys.get_all()
```
neskk: You don't need extra entry points at all. Hash key information is passed here and it's enough...

pogom/search.py
```diff
@@ -974,6 +991,39 @@ def search_worker_thread(args, account_queue, account_failures,
...
+                                from_zone = tz.tzutc()
+                                to_zone = tz.tzlocal()
+
+                                expires = expires.replace(tzinfo=from_zone)
+                                expires = expires.astimezone(to_zone)
+                                expires = expires.strftime(
+                                    '%Y-%m-%d %H:%M:%S')
+
+                        key_instance['expires'] = expires
+
```
Alderon86: maybe its better to just save the timestamp, convert it to local time in status.js so db entry can be used for what ever?

pogom/search.py
```diff
@@ -989,6 +1039,17 @@ def search_worker_thread(args, account_queue, account_failures,
                         step_location[0], step_location[1],
                         parsed['count'])
                     log.debug(status['message'])
+                    key_out = {}
+                    key_out[0] = key_instance
+                    key_out[0]['key'] = key
+                    HashKeys.upsert_keys(dbq, key_out)
+                    log.info(
```
Alderon86: i thought its useful for people who arnt using -ps or /status

pogom/models.py
```diff
@@ -2273,6 +2302,13 @@ def clean_db_loop(args):
                              (datetime.utcnow() - timedelta(minutes=2)))))
             query.execute()
 
+            # Remove expired HashKeys
+            query = (HashKeys
+                     .delete()
+                     .where(HashKeys.expires < datetime.utcfromtimestamp(
```
Alderon86: so what do you suggest when to remove them?

pogom/app.py
Alderon86: just thought of people who are using 20+ keys, the might not want to see this on workers page and check key stats on his own

pogom/search.py
Alderon86: i just did what @FrostTheFox mentioned on discord "an invalid hash-key should be a stop all imo".

pogom/app.py
Alderon86: if /hashkeys is removed this will be the only entry left. but not sure what to do, is the /hashkeys wished or not?

pogom/search.py
FrostTheFox: though i did say if there's more than one key it should remove the invalid key and continue

pogom/app.py
```diff
@@ -90,6 +92,25 @@ def get_account_stats(self):
         r.headers.add('Access-Control-Allow-Origin', '*')
         return r
 
+    def get_all(self):
+        args = get_args()
+        if args.status_page_password is None:
```
FrostTheFox: Unless hash keys are prone to be super similar, just cut like the last half of it off.

pogom/app.py
```diff
@@ -90,6 +92,25 @@ def get_account_stats(self):
...
+            abort(404)
+        return render_template('hashkeys.html')
+
+    def post_all(self):
```
FrostTheFox: Uh, I mean, I don't think I'm an expert or anything, but passing a password as a URL parameter sounds like something that shouldn't be done.

I know POST requests and GET requests are equally insecure over http (which most map users use), but this could also leave you open to issues with referrer stuff, server logs, browser history, etc.

pogom/app.py
Alderon86: im just thinking, if you cant protect your site well enough using reverse proxy/nginx and/or pw protection, isnt it your own fault? but if its wished this is gonna be changed

pogom/search.py
Alderon86: k got it

pogom/app.py
Alderon86: that would open even more issues, so no.

pogom/models.py
neskk: Only delete them after a couple of hours, or maybe a day have passed over expiration date.

pogom/app.py
neskk: @Alderon86 figuring out the IP behind reverse proxy isn't all that difficult. After that you can start a brute force attack almost effortlessly... this software is not made to be that secure because it didn't hold valuable stuff, now that you're putting hash keys on pages which security is questionable I think this conversation is pertinent.

pogom/app.py
neskk: I didn't suggest you to change it to GET request I said you could.
Only javascript uses that password to query status so the URL will be hidden everytime, no history, no referrer problems.

pogom/app.py
neskk: IMO I would remove it...
There's no additional features there, it's one more template to maintain and you added a nice toggle to status page that takes care of hiding info if you don't want to see it. I prefer only having to check one place to get all the stats. Don't know about other people this is only my views, I admit I can be a bit picky with details.

pogom/models.py
```diff
@@ -1668,6 +1668,35 @@ def get_valid(limit=15):
...
+            'maximum': 0,
+            'remaining': 0,
+            'peak': 0,
+            'expires': None,
+            'last_updated': None
+        }
+
+    @staticmethod
+    def upsert_keys(db_update_queue, keys):
+        db_update_queue.put((HashKeys, keys))
```
Alderon86: to get familiar with the db queue, removed it.

pogom/models.py
Alderon86: i'll set it to 1 day for now

pogom/search.py
Alderon86: moved to debug

pogom/search.py
Alderon86: ehm you mean the overseer thread?

pogom/search.py
Alderon86: k i will *** the last 4 or 5 digits then, should be done later this day

pogom/search.py
Alderon86: *outdated* since no changes were made to that part

pogom/models.py
```diff
@@ -2273,6 +2297,14 @@ def clean_db_loop(args):
                              (datetime.utcnow() - timedelta(minutes=2)))))
             query.execute()
 
+            # Remove expired HashKeys
+            query = (HashKeys
+                     .delete()
+                     .where(HashKeys.expires <
+                            (datetime.utcfromtimestamp(int(time.time())) +
```
neskk: `datetime.utcfromtimestamp(int(time.time())` = `datetime.utcnow()` ?

Alderon86: if i remove, the convertion to local tz, this should be still correct, since we only got timestamp left in the db!?

pogom/search.py
neskk: Yep no changes are required, my bad.

neskk: `key_instance = key_scheduler.keys[key]`

Alderon86: please try it yourself, travis doesnt like that

pogom/models.py
```diff
@@ -1668,6 +1669,15 @@ def get_valid(limit=15):
         return tokens
 
 
+class HashKeys(BaseModel):
+    key = CharField(primary_key=True, max_length=20)
+    maximum = IntegerField(default=0)
+    remaining = IntegerField(default=0)
+    peak = IntegerField(default=0)
+    expires = DateTimeField(null=True)
+    last_updated = DateTimeField(default=datetime.utcnow)
```
sebastienvercammen: Fields like `max`, `peak` (highest ever) and `average` are useful, but `remaining` is an attribute that's only relevant at runtime and most likely shouldn't be in the database.

pogom/search.py
```diff
@@ -987,6 +986,27 @@ def search_worker_thread(args, account_queue, account_failures,
...
                         break
 
+                    if args.hash_key:
+                        key_instance = key_scheduler.keys[key]
+                        key_instance['remaining'] = HashServer.status.get(
+                                'remaining', 0)
+
+                        if key_instance['maximum'] == 0:
+                            key_instance['maximum'] = (
+                                HashServer.status.get('maximum', 0))
```
sebastienvercammen: If your default is `0` when it doesn't have the field, your `if key_instance['maximum'] == 0:` condition can be removed: as long as it's `0` it will keep going into the branch.

pogom/search.py
```diff
@@ -1002,6 +1022,17 @@ def search_worker_thread(args, account_queue, account_failures,
                         step_location[0], step_location[1],
                         parsed['count'])
                     log.debug(status['message'])
+                    key_out = {}
+                    key_out[0] = key_instance
+                    key_out[0]['key'] = key
+                    dbq.put((HashKeys, key_out))
+                    log.debug(
+                            ('Hash Key {} has {}/{} RPM ' +
```
sebastienvercammen: `Hash key`

pogom/search.py
```diff
@@ -1002,6 +1022,17 @@ def search_worker_thread(args, account_queue, account_failures,
                         step_location[0], step_location[1],
                         parsed['count'])
                     log.debug(status['message'])
+                    key_out = {}
+                    key_out[0] = key_instance
+                    key_out[0]['key'] = key
```
sebastienvercammen: Why are you using a dict with index `0`? Did you mean to use an array, or a simple variable? And can you rename `key_out` to a clearer, less "temporary" variable name?

static/js/status.js
```diff
@@ -93,17 +108,102 @@ function processWorker(i, worker) {
...
+    mainKeyHash = 'global'
+    if ($('#hashtable_global').length === 0) {
+        addhash('global')
+    }
+
+    if ($('#hashrow_' + keyHash).length === 0) {
+        addhashtable(mainKeyHash, keyHash)
+    }
+
+    var lastModified = new Date(hashkey['last_updated'])
```
sebastienvercammen: Did you triple-check if `lastModified` is correctly being converted from a UTC timestamp into a local timestamp for the browser's user?

static/js/status.js
```diff
@@ -93,17 +108,102 @@ function processWorker(i, worker) {
...
+        RPM Left
+        </div>
+      <div class="status_cell">
+        Peak
+        </div>
+       <div class="status_cell">
+         Expires at
+       </div>
+       <div class="status_cell">
+         Last Modified
```
sebastienvercammen: Consistency: either use regular wording/capitalization (`Hash keys`, `Last modified`) or make all headers `First Upper Case` (`Expires At`). I vote for the former.

static/js/status.js
```diff
@@ -168,6 +266,13 @@ function compare(index) {
         return $.isNumeric(valA) && $.isNumeric(valB) ? valA - valB : valA.localeCompare(valB)
     }
 }
+function comparehashKeys(index) {
```
sebastienvercammen: Rename function: this function does not compare hash keys, it returns a comparing function.

pogom/models.py
Alderon86: If we dont save remaining in the database, what do you think would be the plast place for it? if i put it back to workerstatus remaining will jump with every call

static/js/status.js
Alderon86: quadruple checked now. its local date on the browser

pogom/search.py
Alderon86: it should be more like an array but this wont work if i need to remove remaining from db

static/js/status.js
```diff
@@ -93,17 +108,102 @@ function processWorker(i, worker) {
...
+    var lastModified = new Date(hashkey['last_updated'])
+    lastModified = lastModified.getHours() + ':' +
+        ('0' + lastModified.getMinutes()).slice(-2) + ':' +
+        ('0' + lastModified.getSeconds()).slice(-2) + ' ' +
+        lastModified.getDate() + ' ' +
+        monthArray[lastModified.getMonth()] + ' ' +
+        lastModified.getFullYear()
+
+    var expires = new Date(hashkey['expires'])
+    expires = expires.getHours() + ':' +
```
Alderon86: Please check the code again, it already is

onilton: No, it is still duplicated code. Maybe you didn't push your changes?

Alderon86: there are still couple of changes i still havent pushed yes

Alderon86:
App.py does not recieve remaining atm everything else is done i think

SiteEffect:
^awesome work ;-)

pogom/search.py
```diff
@@ -1031,8 +1031,13 @@ def search_worker_thread(args, account_queue, account_failures,
                         parsed['count'])
                     log.debug(status['message'])
                     hashkeys = {}
```
M4d40: Any reason for creating an empty dict and afterwards fill it with only 1 value?
Why not creating it with the value from the start?
`hashkeys = {
  0: {
      'key': key_instance['key'],
      ...
  }
}`
Or is it for better understanding?

Alderon86: this will move now to the scheduler, which already is holding all the data from beginning.
wait for an update^^

faith9:
What happens if multiple instances scan with the same key? 
It takes them all into account and adds up all the usages?

sebastienvercammen:
@faith9 This is not a question a developer would have asked. Github is not customer support, please wait until features are reviewed and released, or ask general questions like this on our Discord.

faith9:
@sebastienvercammen, ok, I'll try to switch to developer mode.

I've applied this PR to my setup and everything seems to work beautifully, except that the usage of the same hashkey in different instances seems to be ignored and only one instances is shown on the status page.

As a possible (quick and dirty) solution for multiple instances usage I suggest appending the status_name to the hashkey before storing in the database.

See this pull request: https://github.com/Alderon86/RocketMap/pull/2/files

image

sebastienvercammen:
@faith9 This is not necessary, RPM usage is stored as reported by BossLand servers for a hashing key, which means it's global use rather than per instance.

static/js/status.js
```diff
@@ -168,6 +266,13 @@ function compare(index) {
         return $.isNumeric(valA) && $.isNumeric(valB) ? valA - valB : valA.localeCompare(valB)
     }
 }
+function compareHashTable(index) {
```
Alderon86: sebastienvercammen 14 days ago Collaborator
Rename function: this function does not compare hash keys, it returns a comparing function.

now im confused

neskk: It can be thought as a design pattern used by multiple programming languages and it's commonly called a comparator (in Java you have Comparator and Comparable interfaces that enable you to sort/compare objects).
Javascript doesn't enforce these patterns as much as Java or C# but in truth what you have there is a comparator... well the inner function is, the outer function (`compareHashTable` returns a comparator) You can imagine that this can be useful if you had more than one way of comparing elements, `compareHashTable` could return a different comparator based on a parameter or something. Since this is not the case I think `compareHashKeys` it's an "ok" name or perhaps `hashkeyComparator` if you want to be more correct.

pogom/models.py
myntath: @sebastienvercammen although it's only relevant at runtime I think it is still necessary to have the remaining values in the database in order to make sure the information can be aggregated across different instances (with potentially different keys) and displayed on the web status page. This seems to be the approach for status messages which are also only relevant at runtime.

myntath:
I see that 'remaining' has been removed from the status page. I commented about this up higher but the reply seems to be buried because of changes already made. I think this is functionality we should keep in this pr. It is the most obvious statistic I expect to see when looking at hashkeys. Even though it is only valid at runtime I think it is best we write it to the database in order to be able to aggregate information about hashkeys across multiple instances and display them on the one webserver status page. This seems to be the same approach taken with status messages for workers.

pogom/models.py
sebastienvercammen: @myntath The RPM stats (remaining, peak, expires, ...) are all reported by BossLand. All instances receive this information when they query the BossLand servers with an API key. It doesn't make sense to store this information, which all instances are already receiving.

sebastienvercammen:
@myntath Remaining had to be removed from the database because it's not information that needs to be stored - it's only relevant at runtime and it's a derived attribute that's actually just `(maximum - used)`.

Whether it should have been removed from the status page is a different thing. I agree the status page should still show it, exactly because it's relevant at runtime, but it shouldn't be stored in the database. It most likely disappeared from the status page as a side effect of removing it from the database, and no one was paying attention to how it should actually be handled.

Alderon86:
@sebastienvercammen what would you consider the prefered way to actually handle it?

sebastienvercammen:
@Alderon86 Pass on status messages without being dependent on database updates only.

pogom/models.py
myntath: @sebastienvercammen An example of my particular use is I am running three instances. Two scanning instances each using different hashkeys and a third server only instance. We can only access the status page from the server only instance. However, we would still like to see details about the hashkeys from the two scanning instances. 

I understand how the data is synchronised between multiple instances with the same keys because of bossland but I think we do want to see status message for other instances (at least on the webserver /status page). This is the behaviour I get for workers, where I can monitor different workers from different instances of a server instance not using any workers at all. The only reason for storing the remaining, peak, average information is to efficiently make it available to the server only instances. Otherwise I agree there would be no need.

myntath:
@sebastienvercammen I have commented in detail in the burried comment chain above with my use case but I can't think of a more efficient way pass the status messages to server only instances from other instances compared to writing to the database. Passing variables between separate python processes will require writing to some external file or other data storage. We already have the database so why not use it.

sebastienvercammen:
@myntath I agree to your points, but `remaining = (max - used)`, or an alternative that's more accurate in the long-term: `remaining = (max - average)`, which are all fields still in the database. Remaining is a derived attribute.

pogom/schedulers.py
```diff
@@ -1119,18 +1119,28 @@ def get_scheduler(name, *args, **kwargs):
...
+                'expires': None
             }
 
         self.key_cycle = itertools.cycle(keys)
         self.curr_key = ''
 
+        hashkeys = self.keys.copy()
+        hashkeys_db = {}
+        for key in hashkeys:
+            hashkeys_db[key] = hashkeys[key].copy()
```
sebastienvercammen: Why do we copy the entire `self.keys` into `hashkeys`, to then have `hashkeys_db` which also copies individual elements with `hashkeys[key].copy()`, while `hashkeys` is never used afterwards?

pogom/models.py
```diff
@@ -1671,6 +1672,30 @@ def get_valid(limit=15):
...
+                 .where(HashKeys.key == key)
+                 .dicts())
+
+        return query[0] if query else {
+            'maximum': 0,
+            'average': 0,
+            'peak': 0,
+            'expires': None,
+            'last_updated': None
+            }
```
sebastienvercammen: Indentation is inconsistent.

pogom/app.py
```diff
@@ -550,6 +550,10 @@ def post_status(self):
             d['login'] = 'ok'
             d['main_workers'] = MainWorker.get_all()
             d['workers'] = WorkerStatus.get_all()
+            hashkeys = HashKeys.get_all()
+            for i, s in enumerate(hashkeys):
+                hashkeys[i]['key'] = s['key'][:-9] + '*'*9
```
sebastienvercammen: When you want to do some less readable stuff, comment it. Or do it in a readable way.

myntath:
@sebastienvercammen Yes I think lets pick remaining / used, that way we can still calculate average on the front end without saving to db. I don't really have a preference between remaining / used, used is more intuitive to me but I think it is actually remaining that is directly reported when querying the pgoapi HashServer and not used (instead we do max - remaining = used). I have commented the relevant call.

pogom/models.py
```diff
@@ -1671,6 +1672,30 @@ def get_valid(limit=15):
         return tokens
 
 
+class HashKeys(BaseModel):
+    key = CharField(primary_key=True, max_length=20)
+    maximum = SmallIntegerField(default=0)
+    average = SmallIntegerField(default=0)
```
sebastienvercammen: I can't seem to find where `average` is being set in your edited files. I did find a (pseudocode) `average = peak / 60 * elapsed`, which is not what an average statistic of usage is supposed to be. Average should be usage over time, not peak over time.

static/js/status.js
```diff
@@ -93,17 +108,102 @@ function processWorker(i, worker) {
...
+function processHashKeys(i, hashkey) {
+    var mainKeyHash = hashFnv32a(hashkey['key'], true)
+    var keyHash = hashFnv32a(hashkey['key'], true)
+    mainKeyHash = 'global'
+    if ($('#hashtable_global').length === 0) {
+        createHashTable('global')
+    }
+
+    if ($('#hashrow_' + keyHash).length === 0) {
+        addhashtable(mainKeyHash, keyHash)
```
sebastienvercammen: lowerCamelCase naming: `addHashtable`

static/js/status.js
```diff
@@ -272,4 +377,25 @@ $(document).ready(function () {
...
+        }
+    })
+    $('#showworker-switch').change(function () {
+        showWorkersTable = this.checked
+
+        $('#status_container .status_table').remove()
+        $('#status_container .worker').remove()
+
+        if (statusPagePassword) {
+            updateStatus()
```
sebastienvercammen: Each call you added to `updateStatus()` effectively schedules an additional timer. If you were to load the page and switch `#hashkey-switch` and/or `#showworker-switch` a few times, you'd be updating the status several times per second, up to a limit of an infinite amount of times (aka: you're breaking your interwebs, brother).

Alderon86:
@sebastienvercammen all my changes since @neskk reviewed again are locally and i havent pushed them yet, maybe you can wait for another review until i call you?

Alderon86:
@sebastienvercammen @neskk 
newstatus

is this what we are going for?

Alderon86:
@sebastienvercammen you can review again if you got some time.

walaoaaa1234:
my hash key expired time keep changing lol, don't know where's goes wrong

Alderon86:
@walaoaaa1234 nvm i found it

Alderon86:
@sebastienvercammen ok done, would be nice if you would review again.

pogom/schedulers.py
Alderon86: because i couldnt find any other way to keep -ps alive

static/js/status.js
```diff
@@ -93,17 +109,131 @@ function processWorker(i, worker) {
...
+            count: 0
+        }
+
+        hashkeys[key] = keyValues
+    }
+
+    // Calculate average value for Hash keys.
+    var writeIndex = hashkeys[key].count % 60
+    hashkeys[key].count += 1
+    hashkeys[key].samples[writeIndex] = hashkey['peak']
```
myntath: I have mentioned this to @Alderon86 but thought others might have ideas as well. I don't think we can call a value calculated by averaging the peak value an average. I would prefer to calculate with remaining, but an appropriate rename might also work

static/js/status.js
```diff
@@ -93,17 +109,131 @@ function processWorker(i, worker) {
...
+    var sumSamples = 0
+    for (var j = 0; j < numSamples; j++) {
+        sumSamples += hashkeys[key].samples[j]
+    }
+
+    if (numSamples > 0) {
+        hashkeys[key].average = sumSamples / numSamples
+    }
+
+    var remaining = hashkey['maximum'] - hashkey['peak']
```
myntath: This seems counter intuitive to me.  I expect remaining to be the requests remaining for this minute, not the least requests remaining ever.

static/js/status.js
Alderon86: i looked back at the remaining calculation thing u mentioned, that will average the Remaining Request you have left in total for this minute, so your thoughts might be wrong, 
Average use is used to middle your average use per minute, now that hashservers are fixed this will give a more accurate value of usage, then just averageing what u have left per minute (with using remaining in the calcuulation). So in the end we want to see how many rpms we use in average per minute, not the average of what we have left.

static/js/status.js
Alderon86: Pass me a method to keep the actual remaining visible and i'll do it ;)

myntath: See, my previous comments, the only reasonable way I can think of to do this (and not have cross instance problems) is to store the most recent remaining value in the database and then access it like any other related variable.

static/js/status.js
myntath: I don't know if you see different behaviour on the status page than me, but I see what I expect from the code. Every sample is the peak value, (see line 133 above) so the average is just slowly tracking towards the peak value.

I just sat and watched the ps output (which shows the true remaining) and the web page. All that the average did on the web page was slowly increase until it was equal to the peak. Then the web page remained static until the peak increased again. During this time the ps view showed that the remaining had decreased well below the peak and given this the average should also have been dropping.

The average does not and never can drop on the web status page it simply slowly increases to the current peak. To me this is pointless information. If my peak an hour ago was 100rpm but I have been using 10rpm ever since, I want to see that the average is 10 not 100.

To me the data we display in ps is great (maybe could add an average as well). The data on the web page is not, after a few minutes I only see peak.

static/js/status.js
Alderon86: @sebastienvercammen im lost, please provide help here

Arengo: I see no issue storing remaining in the database, too. Would solve a lot of strange workarounds which had to be done to explicitly not store it.

static/js/status.js
Arengo: True, assigning peak value for average calculation is no useful information since peak is not current peak on short time interval but maximum peak we had during the whole runtime.

static/js/status.js
Alderon86: Seb needs to answer this.
But yes, putting back remaining in would make it work like it was intendet to

Alderon86: beaty
this is how it could look with remaining in the db

jagauthier: I went over both versions of this in my local development copy. One with the remaining field, and one with out.

In order to achieve these statistics: Remaining, and a sensible average, we have to store the remaining field in the database.  It seems that it could be worked around with a single hash key, but once multiple hash keys are involved then more help is needed from the back end.

In my view, there are only two pieces of data here that are important:
1) Expiration - Convenience++ is all I can say.
    * Having this data could really be beneficial in the long run, by not using expired hash keys so this isn't so darn manual. (Another PR, later?). Of course, if we can read hashkeys from a file, like proxies this could be a very well rounded solution for longevity.  By a new hash key, put it the file. RM loads it.  Old keys expire, RM stops using it.

2) Remaining RPM - Knowing your remaining RPM is an absolute must if we merge this PR at all.  Can I eyeball a single hash key and figure it out? Yes. Can I do that with 10 hash keys?  No, and I don't want to.  If we choose not to include the remaining field, we are choosing to implement something that doesn't cater to what the users of RM want.

myntath:
After the last changes, I have tested again with fresh pull  and db giving my +1 to this now. So handy to have this data available, really helps avoid server downtime and troubleshoot hashing related problems.

Also +1 to jagauthier's idea of automatically stopping expired hashkeys. But I think that should be a future PR.

bbdoc:
I have been testing this PR for a few weeks already and it works as expected. I really think this need to get merged rapidly. We are now all forced to use an Hash Key and need visibility on it so we don't buy more than needed. We used to have all information from the -ps screen available on the status page as well which is not the case anymore. There are lots of people out there who have no access to the -ps screen because they are running under a docker, or are launching their maps with nohup or on crontab. All those people need this visibility on status screen !

FrostTheFox:
calm down peeps, i'm checking this today

FrostTheFox:
will check back a little later, if there's no other issues then merge 👍

Alderon86:
im just adding here, what i already wrote to fox. me and @myntath  tested very intense over the last month you can pm him for confirmation, but taking apart the key_instance, wont work. if remaining should be displayed on the web page, this will only work if the whole key_instance is upserted to db, otherwise we will get different values on each call because theres no way to say "this value appends to that key" it will result in jumping values between the keys and incorrect display. Also average would use "peak" for calculation then which also will result in a "none" real-time value. im lost about this, like you see it could be easily done, we are also saving values like captchas in the db, which are even not used after restarting the server, which in the end is also just relevant on runtime, because its reseted everytime, so why not save the remaining value from hash keys, which is, like so many folks said, one of the most wanted values to display besides peak.

static/js/status.js
```diff
@@ -92,17 +109,129 @@ function processWorker(i, worker) {
...
+        ('0' + (lastUpdated.getMonth() + 1)).slice(-2) + '-' +
+        ('0' + lastUpdated.getDate()).slice(-2) + ' ' +
+        ('0' + lastUpdated.getHours()).slice(-2) + ':' +
+        ('0' + lastUpdated.getMinutes()).slice(-2) + ':' +
+        ('0' + lastUpdated.getSeconds()).slice(-2)
+
+    var expires = new Date(hashkey['expires'])
+
+    // Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting
+    expires = expires.getFullYear() + '-' +
```
onilton: Please make this a method/function. You can even search in the javascript codebase, maybe the a method/function like this already exists.

static/js/status.js
```diff
@@ -92,17 +109,129 @@ function processWorker(i, worker) {
...
+    }
+
+    if (numSamples > 0) {
+        hashkeys[key].average = sumSamples / numSamples
+    }
+
+    var lastUpdated = new Date(hashkey['last_updated'])
+
+    // Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting
+    lastUpdated = lastUpdated.getFullYear() + '-' +
```
onilton: Please make this a method/function. Is the same code from below. You can even search in the javascript codebase, maybe the a method/function like this already exists.

pogom/app.py
```diff
@@ -554,6 +554,12 @@ def post_status(self):
             d['login'] = 'ok'
             d['main_workers'] = MainWorker.get_all()
             d['workers'] = WorkerStatus.get_all()
+            # Get Hash key values from database and obfuscate the last 9 digits
```
onilton: All this code can be moved to its own function/method. Just think like hashkey obfuscation should be post_status job. One option is to move the method to HashKey class, in a method like `get_all_obfuscated()` or `get_safe_all()`.

Another option is to just override get_all in haskey to return a field called 'obsfucated_key', 'safe_key' or 'display_key'

static/js/status.js
Alderon86: i just did what the existing code for worker last modified did, but ok if its wished that hard i will wrap around a function

static/js/status.js
Alderon86: see above

pogom/models.py
```diff
@@ -2506,11 +2506,14 @@ def bulk_upsert(cls, data, db):
...
     verify_database_schema(db)
-    db.create_tables([Pokemon, Pokestop, Gym, ScannedLocation, GymDetails,
-                      GymMember, GymPokemon, Trainer, MainWorker, WorkerStatus,
-                      SpawnPoint, ScanSpawnPoint, SpawnpointDetectionData,
-                      Token, LocationAltitude, HashKeys], safe=True)
-    db.close()
+    tables = [Pokemon, Pokestop, Gym, ScannedLocation, GymDetails,
+              GymMember, GymPokemon, Trainer, MainWorker, WorkerStatus,
+              SpawnPoint, ScanSpawnPoint, SpawnpointDetectionData,
+              Token, LocationAltitude]
```
DennyLoko: Why did you remove the `HashKeys` table?

jagauthier: Probably because this was merged into develop, https://github.com/RocketMap/RocketMap/pull/1964, he merged develop into his repo:
Merge branch 'develop' into rpm

He can just put it back, in the two places.

static/js/status.js
jagauthier: I think he should keep consistent with the existing code.  If a change is desired, then a different PR should go through and make those changes in totality.

onilton: @jagauthier It is good that the changes you introduce do not make the code worse (imho this is the case, you are duplicating code). 

And in the best scenario, the thing is to leave code in better state than you find.

Leave the current code as it can be acceptable, but inserting new bad code (without a really good reason) is not.

onilton: @Alderon86 it's important for you to understand that this is not just wished, this is **required** for us to maintain the code quality. 

Duplicated code in general is a bad thing, makes programs bigger than it should be and makes the program more error prone since fixing something in one of the replicas doesn't guarantee you fixed in the other replicas.

Also duplicate code also make it harder to refactor programs.

And there is a lot of other issues with duplicated code.

Forgetting about the duplicate code stuff, also, isolating your code in smaller functions usually make it easier to understand and follow the code, since it helps abstract low level implementation details into higher-level abstract ideas. In this case a function like `getFormattedDate(date)` would make easier to understard than just a sequence of `('0' + lastUpdated.blablabla()).slice(-2) + ' ' +`

Maybe @sebastienvercammen can help clarifying this.

Alderon86: This is way more clear then your last answer @onilton , thanks for pointing out why its bad!

pogom/models.py
```diff
@@ -2528,7 +2528,7 @@ def drop_tables(db):
               GymDetails, GymMember, GymPokemon, Trainer, MainWorker,
               WorkerStatus, SpawnPoint, ScanSpawnPoint,
               SpawnpointDetectionData, LocationAltitude,
-              Token]
+              Token, HashKeys]
```
DennyLoko: 👏 👍

pogom/app.py
```diff
@@ -566,6 +566,7 @@ def post_status(self):
             d['login'] = 'ok'
             d['main_workers'] = MainWorker.get_all()
             d['workers'] = WorkerStatus.get_all()
+            d['hashkeys'] = HashKeys.get_obfuscated_keys()
```
onilton: Wow, much better to read :slightly_smiling_face:  Congrats! :+1:

static/js/status.js
```diff
@@ -92,17 +113,114 @@ function processWorker(i, worker) {
...
+    for (var j = 0; j < numSamples; j++) {
+        sumSamples += hashkeys[key].samples[j]
+    }
+
+    if (numSamples > 0) {
+        hashkeys[key].average = sumSamples / numSamples
+    }
+
+    var lastUpdated = new Date(hashkey['last_updated'])
+    lastUpdated = getformattedDate(lastUpdated)
```
onilton: Much better to read, don't you think?  :slightly_smiling_face: 

I think it won't be bad here to put everything in one single line.

`var lastUpdated = getformattedDate(new Date(hashkey['last_updated']))`

Or... if you still want to keep the old format of two lines create two different temp variables (usually reusing a variable for differente purposes is not really a good thing). This is better:

```
var rawLastUpdated = new Date(hashkey['last_updated'])
var lastUpdated = getformattedDate(rawLastUpdated)
```

static/js/status.js
```diff
@@ -2,11 +2,24 @@
...
 var groupByWorker = true
+var showHashTable = true
+var showWorkersTable = true
+var hashkeys = {}
 
 // Raw data updating
 var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
 var lastRawUpdateTime = new Date()
 
+function getformattedDate(formattedDate) {
```
onilton: I have two comments here:

1 - `getformattedDate` should be camelCase, so `getFormattedDate` instead
2 - formattedDate as parameter gives the impression that the date is already formatted, use something like `rawDate` or `unformattedDate` as the parameter

static/js/status.js
```diff
@@ -2,11 +2,24 @@
...
+// Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting
+    formattedDate = formattedDate.getFullYear() + '-' +
```
onilton: There is no problem here in creating a new var to use with the rename I suggested above, like:

```
var formattedDate = unformattedDate.getFullYear() + '-' +
....
```

Really you don't need and shouldn't reuse vars specially if the semantic changes.

You can also just use return here instead of assigning to a var.

pogom/app.py
Alderon86: Thank you :)

static/js/status.js
```diff
@@ -92,17 +112,112 @@ function processWorker(i, worker) {
...
+            samples: [],
+            average: 0,
+            count: 0
+        }
+
+        hashkeys[key] = keyValues
+    }
+
+    // Calculate average value for Hash keys.
+    var writeIndex = hashkeys[key].count % 60
```
onilton: I think `count` here doesn't really reflect what this variable is meant to. When samples reach the limit of 60, count is not count anymore, right? Explicit is better. Maybe rename `count` to `nextSampleIndex` or `nextSampleWriteIndex` (or something like this)?

Alderon86:
Travis, please dont be ill :/

Alderon86:
Sorry for reopening, i hope travis has taken his pills!
And....he hasnt...

static/js/status.js
```diff
@@ -141,7 +140,7 @@ function processHashKeys(i, hashkey) {
     }
 
     if (numSamples > 0) {
-        hashkeys[key].average = sumSamples / numSamples
+        var average = sumSamples / numSamples
```
onilton: This code is not ok. You a declaring a variable inside an if. What will happen when we reach line 153 and the `average` var was never even created? BOOOM :boom:  :smile: 

You have two options here:

```
var average = 0
if (numSamples > 0) {
    average = sumSamples / numSamples
}
```

or

```
var average = (numSamples > 0) ? sumSamples / numSamples : 0
```

The last one is smaller, but you should always be careful with the ternary operator, when too big, it can make things more complicated. (Just a future advice lol)

sebastienvercammen: @onilton Ironically, Javascript does this thing called [_hoisting_](https://scotch.io/tutorials/understanding-hoisting-in-javascript) (although you should [read this](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) as well to clear up a common misconception), so `average` would exist (using ES6' `let` instead of `var` fixes this), but since it would be `undefined` it would indeed go _boom_ because `undefined` does not have a property called `toFixed`.

Just some interesting details. Hope it teaches something to someone.

static/js/status.js
```diff
@@ -2,11 +2,24 @@
...
 
+function getFormattedDate(unFormattedDate) {
+// Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting
+    var formattedDate = unFormattedDate.getFullYear() + '-' +
+    ('0' + (unFormattedDate.getMonth() + 1)).slice(-2) + '-' +
+    ('0' + unFormattedDate.getDate()).slice(-2) + ' ' +
+    ('0' + unFormattedDate.getHours()).slice(-2) + ':' +
+    ('0' + unFormattedDate.getMinutes()).slice(-2) + ':' +
+    ('0' + unFormattedDate.getSeconds()).slice(-2)
+    return formattedDate
```
sebastienvercammen: This entire function (`getFormattedDate`) can be replaced. Someone else was playing around with custom formats of dates, so I recommended importing Moment.js (https://momentjs.com/) instead.

We have to play around with dates, so we might as well do it in a proper and readable way. Make sure to look around for usage and custom formatting of dates, so we can stay consistent.

Alderon86: https://momentjs.com/ *reminder comment*

Arengo: Have an example for usage here: https://github.com/RocketMap/RocketMap/pull/1876/files

.eslintrc.json
```diff
@@ -46,6 +46,7 @@
     "light2StyleNoLabels": true,
     "pGoStyleNoLabels": true,
     "pGoStyleDay": true,
-    "pGoStyleNight": true
+    "pGoStyleNight": true,
+	"moment": false
```
neskk: Inconsistent spacing...

static/js/status.js
```diff
@@ -181,7 +175,7 @@ function createHashTable(mainKeyHash) {
         RPM Left
       </div>
       <div class="status_cell">
-        Average
+        Usage
```
neskk: I saw what @sebastienvercammen said in Discord, but this is a real average, maybe it doesn't average the values it should but never the less it's a real average. If it feels better to rename it usage, sure do it, but IMHO semantics were right before...

templates/status.html
```diff
@@ -83,5 +83,6 @@
 </div>
 <script src="{{ url_for('static', filename='dist/js/status.min.js').lstrip('/') }}"></script>
 <script src="{{ url_for('static', filename='dist/js/app.min.js').lstrip('/') }}"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.17.1/moment.min.js"></script>
```
neskk: Using a 20kb Javascript library to output a date seems like using a cannon to kill a mosquito.

static/js/status.js
onilton: @sebastienvercammen yeah, I know about hoisting, but even if we would rely on it, imho it is a bad feature, it brings more trouble than benefits (specially with variable declaration). Makes code hard to follow and can lead to nasty bugs. Since scoping is also crazy in JS it only make things worse. I am really glad they introduced let and const, I only use those nowadays.

But when they are not available, for me it still valid to not rely on hoisting as much as possible.

I am a static type guy, so a lot of those things looks like code smell to me. :stuck_out_tongue_closed_eyes:  :man_shrugging:

templates/status.html
Alderon86: thats what seb requested

sebastienvercammen: @neskk I'm not sure if you've seen how people have been _trying_ to manipulate dates so far, but it hasn't been very pretty. Considering the amount of times I've had to tell people to rewrite their date manipulation properly (e.g. not as a garbage of splitting strings, using substring, sometimes even creating multiple Date objects just because they didn't want to make a variable to store one in it), using Moment.js is a very cheap price to pay, especially considering any proper webserver (e.g. nginx reverse proxy) or the new webserver being written all support gzip, which makes things considerably smaller.

For now, Moment.js is a necessity for people to start using Dates properly. Its file size is also nothing compared to the issues created by legacy and spaghetti code, so until then we'll use it.

When we come to a point where we've fixed the most important issues and the file size of Moment.js is highest on our list, that'll be a good day, and a day I'll be prepared to rewrite all Date use myself without Moment.js.

static/js/status.js
sebastienvercammen: @neskk Usage is what it's reported as by BossLand servers, which should be the used RPM over the last 60 seconds, which isn't an average.

If you then don't average those reported values, you'll never get an average, you'll get usage during the last 60 seconds.

neskk: I've helped @Alderon86 with the averaging script so I'm just telling you that there's a averaging going on, if you look at the code you will see that it collects X samples and then averages them out. If status updates occurs exactly every 1sec then the average is for the last minute because the sampling buffer has 60 "slots".
Previously it was storing Peak values in the average buffer, now it's doing: `hashkey['maximum'] - hashkey['remaining']`. It's all RPM values so it's "per minute" but it's actually averaging "usage" or whatever you want to call `Maximum - Remaining RPM`.

sebastienvercammen: @neskk Then it's just miscommunication because I asked in #pr and the answer was that "Average" was just "usage" as reported by BossLand. BossLand's data isn't an average, as I've explained.

If our own code is actually averaging over time, then I agree it needs to be called "Average". But with the information I was given, assuming it's the usage as reported by BossLand, it should be called "Usage" (or any similar alternative).

I hope that clears it up.

neskk: o0
It's not on my own code, I just said I helped write the averaging code as it is currently in this PR. As I already said... this PR does the average of several samples and each sample is calculated from Maximum-Remaining RPM. Column could be called average because it's an average of 60 samples/values.

sebastienvercammen: @neskk You misread my previous message, we're agreeing.

tomballgithub:
I started this with 4 keys, 2 good and 2 expired.   The expired hash keys are not being removed by db-cleaner.  They also show a 1969 expired date.

image

Alderon86:
ok normally it should be removed one day after expiration date, will look at it when back home(regarding 1696 date)

Alderon86:
ok i checked back what @tomballgithub wrote. I will post here what i wrote to him on discord:

Normally you shouldnt be able to make a proper call with an expired key, which will end up in a BadHashRequestException and you also shouldnt get a date as a result of that.
If you add a Key which expires during use it will be displayed/removed properly (1 day after expiration to remind you its expired).
I also added again 2 expired keys myself to check for the date, but 1969 didnt show up.

SiteEffect:
69 is just something special

tomballgithub:
Looks like 1969 vs 1970 is just UNIX epoch time (1/1/1970) adjusted for adjusted for time zone.  The real issue is that I launched with expired keys and this PR does not accommodate that.  My concern is that this PR does acknowledge that people will be running with keys that expire and it will track all that in a database.  If a user clears the database and restarts without removing keys from the config.ini, they will be launching with expired keys.  My recommendation is to find a way to cover this condition.  These keys are invalid.

jagauthier:
The title of the PR "Hash Key manager" is probably a misnomer. It's more like "Hash Key Reporter".
What do you expect this PR to do with your expired hash key?    

In your condition, if the user clears the database and starts with an invalid key, that is exactly what will happen.  Scans will fail, exceptions will be thrown.  I'm not sure the API calls for key information will work on an expired key.   There is a whole world of hash key management that could truly be implemented, but this PR is not that one. This is just to bring visibility to the  mapper. (IMO)

Alderon86:
I am working on that, the problem is that expired keys wont report back any values. and BadHashRequestException is not only called with an invalid key, im thinking a few days back where hashing server reported valid keys as invalid, in this case all keys would be removed and scanning completley aborted. Thats the problem where i am at right now. Also i think checking for a valid key would need an extra api call/request.

sebastienvercammen:
@tomballgithub @jagauthier I've renamed the PR to be clearer about what it currently is.

@Alderon86 You just said "I am working on that": do I take that to mean that you'll be adding new things to that PR? If so, I'll wait with my review until you've completed everything and you've pinged me.

Alderon86:
@sebastienvercammen with the new front end coming i would want to wait for that before adding new things here. Maybe a follow up pr with a couple of other modifications i have in mind right now.
So please review it now!

tomballgithub:
My concern was with this part of the description: "expired Hash Keys are also removed via db-cleaner".  This nuance about keys are only handled if they are valid and then become expired during a run won't be documented anywhere and might cause confusion.  Fixing with a future PR is OK with me but not optimal.  At a minimum, I would have recommended showing INVALID for these 1969/1970 dates to avoid confusion.  Or maybe a warning to check your hash keys if a -cd is performed.

sebastienvercammen:
@Alderon86 I agree with @tomballgithub's points. At least make it clearer for the users, and perhaps update the documentation to include the specifics of what they should watch out for if they use the `/status` page.

pogom/search.py
```diff
@@ -1183,6 +1206,10 @@ def map_request(api, position, no_jitter=False):
         response = clear_dict_response(response, True)
         return response
 
+    except HashingOfflineException as e:
+        log.warning('Hashing Server offline.')
+    except BadHashRequestException as e:
+        log.warning('Invalid/Expired Hash Key check your config.')
```
sebastienvercammen: Reword & capitalization:
```
    except HashingOfflineException as e:
        log.warning('Hashing server is unreachable, it might be offline.')
    except BadHashRequestException as e:
        log.warning('Invalid or expired hashing key: ' + api._hash_server_token + '.')
```

docs/extras/status-page.md
```diff
@@ -1,6 +1,6 @@
 # Status Page
 
-Keeping track of multiple processes and accounts can be a pain. To help, you can use the status page to view the status of each of your workers and their accounts.
+Keeping track of multiple processes, accounts and Hash Keys can be a pain. To help, you can use the status page to view the status of each of your workers, their accounts and also your Hash Keys and their values.
```
sebastienvercammen: Capitalization & phrasing:
```
Keeping track of multiple processes, accounts and hashing keys can be a pain. To help, you can use the status page to view the status of each of your workers, their accounts and your hashing keys.
```

docs/extras/status-page.md
```diff
@@ -9,7 +9,8 @@ There are two steps to enable the status page:
 2. Give each of your workers a unique "status name" to identify them on the status page by setting the `-sn` argument.
 
 ## Accessing
-To view your status page, go to `<YourMapUrl>/status` (for example, `http://localhost:5050/status`) and enter the password you defined. The status of each of your workers will be displayed and continually update.
+To view your status page, go to `<YourMapUrl>/status` (for example, `http://localhost:5050/status`) and enter the password you defined. The status of each of your workers and Hash Keys will be displayed and continually update.
+But beware to double check your Hash Keys before starting up your Instances. Invalid Hash Keys resulting out of Typos are not cleared from the db, you want to drop HashKeys table manually then.
```
sebastienvercammen: Capitalization & phrasing (second part only):
```
Remember to double-check your hashing keys before starting your instances. Invalid or expired hashing keys currently won't be cleared from the database or status page automatically.
```

pogom/models.py
```diff
@@ -2412,6 +2444,13 @@ def clean_db_loop(args):
                              (datetime.utcnow() - timedelta(minutes=2)))))
             query.execute()
 
+            # Remove expired HashKeys
+            query = (HashKeys
+                     .delete()
+                     .where(HashKeys.expires <
+                            (datetime.utcnow() - timedelta(days=1))))
```
tomballgithub: I am concerned about time zones issues here.  when I run datetime.utcnow it gets me UTC time without a time zone offset.  My hashkey expiration time showing in the status window does have a time zone offset.  Indeed, further down in the code the bossland expiration is swapped out for a local time version that is ultimately shown in /status.  Need to compare values here without time zone impact.

For example, /status says key expires at 2017-04-29 23:59:26
Bossland says key expires at 4/30/2017 4:59:26 AM

pogom/search.py
```diff
@@ -1113,19 +1116,39 @@ def search_worker_thread(args, account_queue, account_failures,
...
 
                     if key_instance['peak'] < peak:
                         key_instance['peak'] = peak
 
+                    if key_instance['expires'] is None:
+                        expires = HashServer.status.get(
+                            'expiration', None)
+
+                        if expires is not None:
+                            expires = datetime.utcfromtimestamp(expires)
```
tomballgithub: Per my previous comment, it seems that you are taking the bossland expiration date received and changing it to local time.   This makes your compare to UTCNOW above invalid and it be tracking an incorrect expiration time

static/js/status.js
```diff
@@ -2,11 +2,18 @@
...
 var groupByWorker = true
+var showHashTable = true
+var showWorkersTable = true
+var hashkeys = {}
 
 // Raw data updating
 var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
 var lastRawUpdateTime = new Date()
 
+function getFormattedDate(unFormattedDate) { // eslintrc no-undef
```
tomballgithub: Does this comply with the specific was Seb wants date/times to be shown?

tomballgithub:
I approve +1

static/js/status.js
```diff
@@ -92,17 +106,109 @@ function processWorker(i, worker) {
...
+    var numSamples = hashkeys[key].samples.length
+    var sumSamples = 0
+    for (var j = 0; j < numSamples; j++) {
+        sumSamples += hashkeys[key].samples[j]
+    }
+
+    var usage = sumSamples / Math.max(numSamples, 1) // Avoid division by zero.
+
+    var lastUpdated = getFormattedDate(new Date(hashkey['last_updated']))
+    var expires = getFormattedDate(new Date(hashkey['expires']))
```
sebastienvercammen: Agreed, it would be clearer.

pogom/search.py
```diff
@@ -1183,6 +1206,11 @@ def map_request(api, position, no_jitter=False):
         response = clear_dict_response(response, True)
         return response
 
+    except HashingOfflineException as e:
+        log.warning('Hashing server is unreachable, it might be offline.')
+    except BadHashRequestException as e:
+        log.warning(('Invalid or expired hashing key:' +
+                    (api._hash_server_token + '.')))
```
sebastienvercammen: You can remove those extra brackets and still keep it split on multiple lines:
```
log.warning('Invalid or expired hashing key:' + api._hash_server_token + '.')
```

static/js/status.js
```diff
@@ -2,11 +2,18 @@
...
+var showHashTable = true
+var showWorkersTable = true
+var hashkeys = {}
 
 // Raw data updating
 var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
 var lastRawUpdateTime = new Date()
 
+function getFormattedDate(unFormattedDate) { // eslintrc no-undef
+// Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting
```
sebastienvercammen: Indentation of comment, and end with punctuation for consistency.

static/js/status.js
Alderon86: My Js skills are not good enough to pull that, please provide help! :(

tomballgithub: Per Aspirax:
After line 139 in status.js, add this 
if (moment(expires).isBefore(moment.utc())) {
            expires = "Unknown/Invalid"
    }
image

pogom/search.py
```diff
@@ -39,9 +39,10 @@
 
 from pgoapi.utilities import f2i
 from pgoapi import utilities as util
-from pgoapi.hash_server import HashServer
-
-from .models import parse_map, GymDetails, parse_gyms, MainWorker, WorkerStatus
+from pgoapi.hash_server import HashServer, BadHashRequestException, \
+                               HashingOfflineException
+from .models import parse_map, GymDetails, parse_gyms, MainWorker, \
+                    WorkerStatus, HashKeys
```
sebastienvercammen: For consistency, if you need multi-line imports, use:
```
from .models import (parse_map, GymDetails, parse_gyms, MainWorker,
                    WorkerStatus, HashKeys)
```
(Make sure to run it through flake8, I haven't checked if the indentation is correct, just editing on Github.)

pogom/search.py
sebastienvercammen: @tomballgithub Has this been checked or updated since all the recent commits? Is your comment still valid?

pogom/search.py
```diff
@@ -1096,19 +1098,39 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
+
+                        if expires is not None:
+                            expires = datetime.utcfromtimestamp(expires)
+
+                        key_instance['expires'] = expires
+
+                    log.debug(
+                        ('Hash key {} has {}/{} RPM ' +
+                         'left.').format(key, key_instance['remaining'],
+                                         key_instance['maximum']))
```
sebastienvercammen: Don't use:
```
log.debug('string with {} some {} stuff in it'.format(foo, bar))
```
Use the logger's functionality instead:
```
log.debug('string with %s some %s stuff in it', foo, bar)
```

pogom/search.py
```diff
@@ -1166,6 +1188,11 @@ def map_request(api, position, no_jitter=False):
         response = clear_dict_response(response, True)
         return response
 
+    except HashingOfflineException as e:
+        log.warning('Hashing server is unreachable, it might be offline.')
+    except BadHashRequestException as e:
+        log.warning('Invalid or expired hashing key:' +
+                    api._hash_server_token + '.')
```
sebastienvercammen: We need to move manual concatenation to properly using the logger's functionality (don't forget punctuation):
```
log.warning('Invalid or expired hashing key: %s.', api._hash_server_token)
```

static/js/status.js
```diff
@@ -2,11 +2,18 @@
...
+var showHashTable = true
+var showWorkersTable = true
+var hashkeys = {}
 
 // Raw data updating
 var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
 var lastRawUpdateTime = new Date()
 
+function getFormattedDate(unFormattedDate) { // eslintrc no-undef.
+// Use YYYY-MM-DD HH:MM:SS formatted dates to enable simple sorting.
```
sebastienvercammen: Missing indentation for the comment.

static/js/status.js
```diff
@@ -92,17 +106,117 @@ function processWorker(i, worker) {
...
+     </div>
+   </div>`
+
+    hashtable = $(hashtable)
+    $('#status_container').prepend(hashtable)
+    $(hashtable).find('.status_row.header .status_cell').click(sortHashTable)
+}
+
+function sortHashTable() {
+    var hashtable = $(this).parents('.status_table').eq(0)
```
sebastienvercammen: jQuery has `.first()`, which is more readable:
```
var hashtable = $(this).parents('.status_table').first()
```

static/js/status.js
```diff
@@ -92,17 +106,117 @@ function processWorker(i, worker) {
...
+    var hashrow = hashtable.find('.status_row:gt(0)').toArray().sort(compareHashTable($(this).index()))
```
sebastienvercammen: For readability purposes, you can restructure this:
```
var comparator = compareHashTable($(this).index())
var hashrow = hashtable.find('.status_row:gt(0)').toArray()

// Sort the array.
hashrow.sort(comparator)
```

static/js/status.js
```diff
@@ -202,14 +202,17 @@ function createHashTable(mainKeyHash) {
 }
 
 function sortHashTable() {
-    var hashtable = $(this).parents('.status_table').eq(0)
-    var hashrow = hashtable.find('.status_row:gt(0)').toArray().sort(compareHashTable($(this).index()))
+    var comparator = compareHashTable($(this).index())
+    var hashrow = createHashTable.find('.status_row:gt(0)').toArray()
```
sebastienvercammen: Why did you change
```
hashtable.find
```
to
```
createHashTable.find
```
?

static/js/status.js
```diff
@@ -202,14 +202,17 @@ function createHashTable(mainKeyHash) {
...
+
+// Sort the array.
+    hashrow.sort(comparator)
     this.asc = !this.asc
     if (!this.asc) {
         hashrow = hashrow.reverse()
     }
     for (var i = 0; i < hashrow.length; i++) {
-        hashtable.append(hashrow[i])
+        createHashTable.append(hashrow[i])
```
sebastienvercammen: Same as above?

pogom/search.py
tomballgithub: My commently previous about datetime.utcfromtimestamp converting to local time is incorrect.  All of the functions are using UTC time for consistency.  Status.js uses moment() which displays the UTC time stored for expiration date as a local time which is desired

static/js/status.js
```diff
@@ -202,7 +202,8 @@ function createHashTable(mainKeyHash) {
 }
 
 function sortHashTable() {
-    var hashtable = $(this).parents('.status_table').eq(0)
+    var hashtable = $(this).parents('.status_table').first()
+    hashtable = $(this).parents('.status_table').eq(0)
```
tomballgithub: This doesn't look right.  Left the old line in there also after editing

static/js/status.js
```diff
@@ -92,17 +106,120 @@ function processWorker(i, worker) {
...
+
+    var usage = sumSamples / Math.max(numSamples, 1) // Avoid division by zero.
+
+    var lastUpdated = getFormattedDate(new Date(hashkey['last_updated']))
+    var expires = getFormattedDate(new Date(hashkey['expires']))
+    if (!moment(expires).unix()) {
+        expires = 'Unknown/Invalid'
+    }
+
+    if (moment().isSameOrAfter(moment(expires))) {
```
Obihoernchen: This throws the following JS warning for me:
```
Deprecation warning: value provided is not in a recognized ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: Unknown/Invalid, _f: undefined, _strict: undefined, _locale: [object Object]
Error
[...]
```

pogom/search.py
```diff
@@ -1096,19 +1098,38 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                             del gym_responses
 
                 if args.hash_key:
-                    key_instance = key_scheduler.keys[key_scheduler.current()]
+                    key = HashServer.status.get('token')
```
sebastienvercammen: It's not guaranteed that `HashServer` will have a `status`, and that `status` will have a `token`. Capture possible problems:
```
key = HashServer.status.get('token', None)

if key:
    ...
```

pogom/search.py
```diff
@@ -1096,19 +1098,38 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
-                    if key_instance['maximum'] == 0:
-                        key_instance['maximum'] = HashServer.status.get(
-                            'maximum', 0)
+                    key_instance['maximum'] = (
+                        HashServer.status.get('maximum', 0))
 
-                    peak = key_instance['maximum'] - key_instance['remaining']
+                    peak = (
+                        key_instance['maximum'] -
+                        key_instance['remaining'])
```
sebastienvercammen: Variable name `peak` is a misnomer: `maximum - remaining` is usage, not peak.

pogom/search.py
```diff
@@ -1096,19 +1098,38 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
                         key_instance['peak'] = peak
 
+                    if key_instance['expires'] is None:
+                        expires = HashServer.status.get(
+                            'expiration', None)
+
+                        if expires is not None:
+                            expires = datetime.utcfromtimestamp(expires)
+
+                        key_instance['expires'] = expires
```
sebastienvercammen: This line should be in the same code block as the `if expires is not None:` just above. We're checking anyway, might as well only set it if we actually have one.

pogom/search.py
```diff
@@ -1096,19 +1098,38 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                             del gym_responses
 
                 if args.hash_key:
```
sebastienvercammen: Add some commenting right above this line to explain what you're about to do, e.g.
```
Update hashing key stats in the database based on the values reported back by the hashing server.
```

sebastienvercammen:
Two major points I mentioned on Discord:

- If peak is always included in the update instead of only updating if the instance's peak is higher than the value stored in DB, it will overwrite the highest value because the instance doesn't know about the other instance's values.
- We should consider resetting the peak to 0 at a certain time (e.g. at instance startup, or every x time), otherwise it'll never lower the peak and you'll never be able to see if your newly added keys have reduced the peak per key.

bbdoc:
Adding to Seb's comment regarding peak, I noticed that the peak is always found within the first few minutes of running the map, because that's when all accounts are logging in and thus using more RPM. If not resetting it, it's pretty useless as it doesn't give any information regarding peak usage during a normal run. Resetting on instance launch is a minimum, and we should consider a "last hour peak" or something similar.

pogom/search.py
```diff
@@ -1095,19 +1097,46 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
+                    if key_instance['peak'] < usage:
+                        key_instance['peak'] = usage
+
+                    if key_instance['expires'] is None:
+                        expires = HashServer.status.get(
+                            'expiration', None)
+
+                        if expires is not None:
+                            expires = datetime.utcfromtimestamp(expires)
+                        key_instance['expires'] = expires
```
sebastienvercammen: This line is still not indented properly. Needs one more, to be inside of the `if expires is not None:` block, otherwise you're just setting it to `None` again.

pogom/models.py
```diff
@@ -1753,6 +1754,48 @@ def get_valid(limit=15):
...
+        hashkeys = HashKeys.get_all()
+        for i, s in enumerate(hashkeys):
+            hashkeys[i]['key'] = s['key'][:-9] + '*'*9
+        return hashkeys
+
+    @staticmethod
+    # Retrieve the last stored 'peak' value for each hashing key.
+    def getStoredPeak(key):
+            query = HashKeys.select(HashKeys.peak).where(HashKeys.key == key)
+            if len(query):
```
sebastienvercammen: Needs to be checked for type (I'm not sure what the returned type of peewee is), but:
```
if len(query):
```
in python-esque is:
```
if query:
```
and `query` should be renamed to `result` if it's instantly executed without requiring `.execute()`, `.get()` or anything else.

pogom/search.py
```diff
@@ -1095,19 +1097,46 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
 
-                    if key_instance['peak'] < peak:
-                        key_instance['peak'] = peak
+                    # Prepare hashing keys to be sent to the db. But only
+                    # sent latest updates of the 'peak' value per key.
+                    hashkeys = {}
+                    hashkeys[key] = key_instance
+                    hashkeys[key]['key'] = key
+                    hashkeys[key]['peak'] = (max(key_instance['peak'],
+                                             HashKeys.getStoredPeak(key)))
```
sebastienvercammen: Remove the extra set of parentheses:
```
hashkeys[key]['peak'] = max(key_instance['peak'], HashKeys.getStoredPeak(key))
```

Alderon86:
@sebastienvercammen Just to approve last changes:
expired

tomballgithub:
I have no questions or remaining changes.

PR #1815, opened 2017-02-10
Fix issue that made TooManyLoginAttempts to never be raised

onilton:
Also, make the code follow what --login-retries really means

## Description
The current code is written in a way that TooManyLoginAttempts is be raised. It goes out of the while loop before the exception has a chance to be raised.

This causes the code to continue and next api call raises NotLoggedInException.

Since ---login-retries means the number of retries and not the tries (small difference, but important one) I also updated the code and arg description to respect that meaning

## How Has This Been Tested?
In my own map, It has beeing working fine and now TooManyLoginAttempts is raised again.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -27,7 +27,7 @@ def check_login(args, account, api, position, proxy_url):
 
     # Try to login. Repeat a few times, but don't get stuck here.
     i = 0
-    while i < args.login_retries:
+    while i <= args.login_retries:
         try:
             if proxy_url:
                 api.set_authentication(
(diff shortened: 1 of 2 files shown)


pogom/account.py
```diff
@@ -27,7 +27,7 @@ def check_login(args, account, api, position, proxy_url):
 
     # Try to login. Repeat a few times, but don't get stuck here.
     i = 0
-    while i < args.login_retries:
+    while i <= args.login_retries:
```
sebastienvercammen: `i` counts from zero while `login_retries` starts at one. `<` was correct. It did cause `TooManyLoginAttempts` not to be raised, but since you've now separated that part of the code, that is now fixed.

onilton: That is the semantic thing I pointed out.

If you are using `--login_retries`, them is the first try plus the retries (1 + num_of_retries)

Thinking in this way, this code is correct: in the first fail it will be 
```
while 0 <= args.login_retries: // First try (not retry)
Exception...
while 1 <= args.login_retries: // will run the first retry
Exception...
while 2 <= args.login_retries: // will run the second retry
Exception...
while 3 <= args.login_retries: // will run the third and last retry
Exception...
while 4 <= args.login_retries: // will exit
```

If you want to keep the old behavior, we should rename  `--login-retries` to `--login-tries` or `--login-attemps`.

I hope I could explain what I thought.

Does that makes sense?

sebastienvercammen: @onilton You're correct, I wrongly assumed that `login_retries` only became relevant after it had already tried to login once but failed, which is not the case (it uses it as total tries).

Rather than using `>=`, it might be clearer to change:
```
while i < args.login_retries:
```
to:
```
# One initial try + login_retries.
while i < (args.login_retries + 1):
```
for clarity.

Thoughts?

onilton: It looks good for me.

I also thought about renaming `i` to `tries` or `try_count`, what do you think?

sebastienvercammen: @onilton Sounds good. I'm used to `num_tries`, but that's personal preference and they all look good to me, so go ahead.

onilton:
@sebastienvercammen all done!

PR #1750, opened 2017-01-25
Add ability to hide columns on -ps status screen and /status web page

tomballgithub:
## Description
This PR adds the ability to hide columns on -ps status screen and /status web page.  There is a command line option for each independently.  The config.ini.example and commandline.md have also been updated to ensure clarity.  I decided to make the column numbering 1-based and not 0-based to prevent confusion.  Default behavior is unchanged.

**Usage - Command Line:**
      -hcps COLUMN, -hide-column-ps COLUMN
                            Hide column(s) on -ps status screen.  Use one -hcps
                            per column to hide.  (1 = first column)
      -hcs COLUMN, --hide-column-status COLUMN
                            Hide column(s) on /status web page.  Use one -hcs
                            per column to hide. (1 = first column)

**Usage - Config.ini:**
#hide-column-ps:        # Column number to hide on -ps status screen.  Use a list for multiple columns e.g. [2,3,4]. (1 = first column)
#hide-column-status:    # Column number to hide on /status web page.   Use a list for multiple columns e.g. [2,3,4]. (1 = first column)

## Motivation and Context
This change was requested on #PR discord to remove columns that some may not find necessary.  It also allows more columns to be added in the future (hash stats, etc) with the knowledge that others can be hidden to compensate.

## How Has This Been Tested?
This was tested on Windows 7 with the Chrome browser.
All error conditions were checked (negative #'s, 0, #'s greater than max columns)
Removing all columns was tested (it works)
Command Line and Config.ini were tested

## Screenshots (if appropriate):
-PS screen with columns 1,2,7,8 removed
image

/status page with columns 4,5,7 removed
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

PR #239, opened 2016-08-09
Multiply max db connections by number of accounts (threads) used

invisiblek:

PR #1820, opened 2017-02-14
Account Statistics Logging and Page

myntath:
This PR helps provide information about the performance of your worker accounts

## Description

Logs each bad scan (failed, empty, captcha) to the database
Logs a total of success, fails, empty or captcha for each account
Creates a /accounts page which provides a summary of the performance of each account

This is my first attempt at a PR on github so help/criticism would be appreciated.

## Motivation and Context

Often users will be using many worker accounts. It is useful to get details about these to help weed out problem accounts. These might have been banned or flagged by Niantic. By having easily accessible statistics which are maintained across sessions and instances this process becomes easier.

This could also assist with future options to share an account pool between multiple instances, only use accounts with certain statistics and also provide much more detailed stats/history on each account and the overall performance of all accounts at different times.

## How Has This Been Tested?

Tested on local machine

accountstats

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -423,6 +423,163 @@ def get_spawnpoints_in_hex(cls, center, steps):
         return filtered
 
 
+class BadScans(BaseModel):
+    name = CharField(null=False)
+    scan_type = CharField(null=False)
+    scan_number = IntegerField(null=False)
+    lat = CharField(default='-1')
+    lon = CharField(default='-1')
```
j16sdiz: besides `lat,lon`, please add the `status_name` of the scanning instance. This can help identifying problemtic instance.

myntath: Good idea

Alderon86:
Are you on discord somewhere? Or how can i reach you?

myntath:
My name is mynt on discord

FizzBuzz791:
Does this fix the captcha cost calculations that appear when grouping? I'm not 100% sure but they currently seem to be based on live stats, i.e. the Captchas count goes up/down based on when they get solved, rather than showing a total. This then seems to affect the /hr and /month rates.

myntath:
No this doesn't change the captcha cost statistics. I believe those statistics are based only on how many captchas are encountered in any particular session (which is what you call live stats). They go up whenever a captcha is encountered, not when solved. They go down over time as no captchas are encountered.

To determine more accurate cost stats would need other data to be kept, such as time running, number of accounts running and ensure only unique captcha encounters are recorded.

I'm not sure if that is something I can add to this pr.

FizzBuzz791:
Ah, no worries, thanks for the clarification. Great PR anyway, keen to see it get merged in.

M4d40:
i'm not sure how much work it would be, but adding the level of the account could be interesting as well.
(To see how the different lvl accounts behave)
if i remember right, Arengo already wrote a function for the account-level on his PR for the pokestops: #1857

SiteEffect:
@M4d40 as a regular RocketMap user you will only encounter level 1 or soon^(TM) level 2. We already know, that level 1 is reCaptcha hell and level 2 is not if running hashing service. This might change in the future, but for now it would not provide any information.

static/js/accstatus.js
```diff
@@ -0,0 +1,247 @@
...
+
+// Raw data updating
+var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
+var lastRawUpdateTime = new Date()
+
+function addWorker(mainWorkerHash, accountHash) {
+    var row = `
+     <div id="row_${accountHash}" class="status_row">
+       <div id="name_${accountHash}" class="status_cell"/>
+       <div id="password_${accountHash}"  class="status_cell"/>
```
myntath: You are right, I will remove these being returned by the server.

myntath:
I have been thinking about adding levels and have implemented it in my own server. It is useful for me as many of my accounts were levelled by using ditto mode in pokealert. The code already fetches the players level as part of other api calls. I think given the new level two spin it is now worth adding.

pogom/models.py
```diff
@@ -427,6 +427,183 @@ def get_spawnpoints_in_hex(cls, center, steps):
...
+        # check login type valid
+        validTypes = ['ptc', 'google']
+
+        if account_type not in validTypes:
+            log.error('Invalid account type: {}'.format(account_type))
+            return False
+
+        # add account
+        account = {}
+        account[0] = {'name': name,
```
M4d40: Why do you first create an empty dict and then just fill it with one value/dict?
Why not creating it with the value?
account = {
  0: {
    'name': name,
    'password': password,
    'login_type': account_type
  }
}

myntath: Why? because I'm a complete amateur. Thank you though, I will fix that up.

richardfoltin:
It would be good if the beehive number like (aka PR#1819) was also shown. So we could see which beehive needs more/less workers.

M4d40:
@richardfoltin 
> It would be good if the beehive number like (aka PR#1819) was also shown. So we could see which beehive needs more/less workers.

This would be an feature for a later Pull Request, but it has nothing to do with this PR.

Does someone use this PR already and can confirm this has no issues so far?

unholygod91:
I use this PR and haven't had any issues.  I mainly use it for the database updates for accounts. when used with -tut, you can make sure all of your accounts are level 2.

pogom/models.py
```diff
@@ -2365,7 +2544,7 @@ def drop_tables(db):
                     GymDetails, GymMember, GymPokemon, Trainer, MainWorker,
                     WorkerStatus, SpawnPoint, ScanSpawnPoint,
                     SpawnpointDetectionData, LocationAltitude,
-                    Token, Versions], safe=True)
+                    Token, Account, BadScans], safe=True)
```
j16sdiz: Do not remove  `Versions`

runserver.py
```diff
@@ -234,6 +234,18 @@ def main():
...
+    if args.accountdb:
+        accounts = Account.get_all()
+        args.accounts = []
+        for account in accounts:
+            if account['enabled']:
+                args.accounts.append({'username': account['name'],
+                                      'password': account['password'],
+                                      'auth_service': account['login_type']})
+    db = False
+    db = init_database(app)
```
j16sdiz: Here, it is leaking database connections.

When you call `init_database(app)`, it calls `playhouse.FlaskDb` which add a hook to connect/disconnect database when there are http requests.
Doing it twice cause leakage there.

pogom/search.py
```diff
@@ -972,10 +981,32 @@ def search_worker_thread(args, account_queue, account_failures,
...
+                                       .get('inventory_delta')
+                                       .get('inventory_items'))
+                    level = filter(lambda i:
+                                   'player_stats' in i['inventory_item_data'],
+                                   inventory_items)
+                    level = (level[0]
+                             .get('inventory_item_data')
+                             .get('player_stats')
+                             .get('level'))
+                    log.error(level)
```
j16sdiz: `log.debug(' barblabrlbar is level {}'.format(level))`  ?

myntath: thanks forgot to remove this debug log, will adjust.

pogom/models.py
myntath: versions is duplicated in the lines (2362/2543) above, I figured no one had noticed so I may as well take it out in this pr. I could do this separately as a bug fix but hardly seems worth it as i'm changing the line anyway. Do you know what preferred course of action is here?

runserver.py
myntath: Ok I'll look into this. I realise this was a terrible solution I had meant to revisit it but forgot. My problem is when we init_database we set the number of connections based on len(args.accounts). For me that is unknown until we have already accessed the database to retrieve the accounts.

I'm guessing either that I should just increase maxconnections instead or doing init_database again. However, is there any reason why we use len(args.accounts)? Is the number of workers is more appropriate?

j16sdiz: See #239

(Fyi, this is leaking one connection on each http request. Very serous if one want to use this patch in webserver

static/js/accstatus.js
```diff
@@ -0,0 +1,243 @@
...
+        addTable('global')
+    }
+
+    if ($('#row_' + hash).length === 0) {
+        addWorker(mainWorkerHash, hash)
+    }
+
+    var lastModified = new Date(account['last_active'])
+    lastModified = lastModified.getFullYear() + '/' +
+        lastModified.getMonth() + '/' + lastModified.getDate() + ' ' +
```
j16sdiz: Month is wrong. `getMonth()` would return 0 for Janurary.

bbdoc:
I see it's been discusses but with the new change from Niantic and the need for lvl 25+ account fir encountering pokemons, I think it would really be needed to add player level in this PR so we can identify performance of those accounts as well. Thanks

myntath:
This PR does store and track level, however it was implemented before the level 2 pokestop spin. There is now better ways to do this. I will have a look at updating this PR soon if I have a bit of time.

j16sdiz:
@myntath   consider use moment.js too. 
seb was asking other PR use it (c.f. https://github.com/RocketMap/RocketMap/pull/1806#discussion_r112305385 )

myntath:
@j16sdiz agreed. I have been watching 1806 closely, I would also prefer to implement this table by toggle switch rather than separate page, similar to that PR. Would be nice to have that merged first.

myntath:
Updated a few things. There is no longer a separate status page, just a toggle switch on the main status page similar to how hashkeys is presented. A few other changes to take advantage of other code that is now part of RM. Plus some terrible git skills meant a whole lot of stuffing around with merges, rebases, forced pushes etc... Have finally got things looking reasonable but seem to have messed up the commit history timing, oops.

bbdoc:
Just installed this PR. Looks good, but doesn't seem to give any status for lvl 30+ accounts, probably something that would need to be added. Maybe have a different toggle for it, or show them as a separate block so they can be easily identified.

Also, would be useful as well to have some info whether the accounts received a warning, or appear to be banned (could not log in).

Lastly, could you please keep accounts grouped by instance ? I'm running 5 instances, and get a big list of accounts, making it quite difficult to read.

Thanks

bbdoc:
PS: Adding to my previous comment, might be interesting to add # of login attempts failed/success. Would also be good to add sub-toggles so we can show/hide columns as this is already a lot of columns, some of which I don't care about. I wouldn't mind getting some colors as well, like get accounts that couldn't login (so seem to be banned) in red ;-)

myntath:
@bbdoc thanks for your comments. I think I agree with most of your suggestions and will work on implementing them. As far as grouping by instance I will add a column to show last instance, you can then sort by that provided you have named your instances.

There is another pr for hiding columns #1750 so I probably won't touch that in this pr. I agree though that it would be important.

Colours might be a bit a bit much but I will add the 'warning' flag and see what I can do about a ban/possible ban flag as well.

bbdoc:
Hi. Did you have some time looking into this ? More specifically I think with Niantic blinding warned accounts since this morning, I do think this is really an information most of us would like to see on our status page so we can better understand what's going on on our maps ;-)

myntath:
Yeah am working on this now, because of this recent problem. Hoping also to track, time since last rare sighted, along with distance walked and pokemon caught and stop spun, this might provide some insight into the shadow bans or at least give a way to weed out poorly performing accounts.

I am looking for a list of pokemon that get shadowbanned. So if anyone knows....

j16sdiz:
@myntath missing spawns count should be enough, i guess?

myntath:
yes I will add missing spawn count, and it might be go to indicator except when spawn points are changed. This may also build up over time for older accounts and recent changes may not be immediately obvious.

myntath:
This is really outdated by now and I think others are working on this area recently so will close.

PR #1840, opened 2017-02-19
Respect ng,nk,np and enc switch's on the web interface (os) instances also.

elincognito:
This PR makes the web interface respect the **ng**, **nk** and **np** switch's

## Description

Removes from the user interface the respective menu elements if one or multiple of the switch's are used, also removes gyms, pokes or pokestop for each respective switch.
## Motivation and Context

Currently if one specify for example the **ng** switch the gym related menu controls still are available to the user even tough they will do nothing.

## How Has This Been Tested?
Tested on my map with each individual switch turned on, one scanning instance and tree **os** instances.

## Types of changes

## Checklist:


SiteEffect:
What happens if I have a dedicated Gym-Scanner next to my standard Pokémon instances? Like
```
-os 
-ns -nk -ng
-ns -np
```

elincognito:
In that case since your dedicated server instance doesn't specify none of
the flags it would show the regular interface, with pokestops,  gyms and
pokemons included

neskk:
Looks nice, I'll be testing this tomorrow. Good job 👍

elincognito:
@neskk Thanks (obrigado) 👍  
Let me know any improvement required or suggestion.

SiteEffect:
This can be extended by showing gym-sidebar only with ``--gym-info`` enabled, too.

elincognito:
@Arengo i will extend with `-gi` as soon as possible thank you for the feedback

elincognito:
@neskk `-ng` is already included it removes the gym switch from DOM if `-ng` flag is on. 
What   @Arengo means is that if gyms are on but `gym-info` is not then then the gym sidebar is of no use and the switch to show the sidebar can be removed also.
Is that correct @Arengo  ?

Edit: @neskk I totally missed out your change suggestions i will check them later  thanks :+1:

neskk:
@forcewill The changes I was talking about were the ones that Arengo suggested, I just wanted to do a github "official" request for change :D
I may have not expressed myself correctly: even if `-gym-info` is set to true, `-ng` should prevent gyms from being parsed. I said `-ng` should also affect gym-side bar and by this I meant hide gym sidebar based on `-ng` too.
Your recent changes make the PR complete and in my opinion ready to merge.

Plop:
Tested, works really fine. Two only-server instances used, one with -ng, one with -np -nk, it's running really really fine. Thanks a lot :)

claydy:
Would be nice if -enc is also respected!

SiteEffect:
@claydy the only link I see is ``Notify by IV``

claydy:
@Arengo I don't understand what you mean?!

neskk:
As I've said in Discord, instead of having N different variables to control the display of certain elements, I would prefer that a single dictionary was used. Template engine can easily access dictionary contents and the code remains clean.
`show_elements['gyms'] = not args.no_gyms`
`show_elements['pokestops'] = not args.no_pokestops`
...
Pass `show_elements` dictionary to the template: `render_template('map.html', ..., show=show_elements)`.
Then use `{% if show.gyms %}`, `{% if show.pokestops %}`, etc. in the template to hide/show sections of the content.
The PR works great and it already had my approval, but since the number of flags grew, I think the approach I described above keeps things tidy.

elincognito:
@neskk and @Arengo  do you mind review the changes again ?
I've included all related visibility flags in the dictionary.
`search_display` is included for mindset but a find in all project files matched no reference in code,
so i believe that is dead code i can remove it. If you guys can confirm it also :+1: 

@neskk My bad related to not included straight away the dictionary its just on work days i have little time :)

pogom/app.py
```diff
@@ -145,11 +145,22 @@ def fullmap(self):
...
+        scan_display = (args.only_server or args.fixed_location or
+                        args.spawnpoint_scanning)
+
+        visibility_flags = {
+            'gyms': not args.no_gyms,
+            'pokemons': not args.no_pokemon,
+            'pokestops': not args.no_pokestops,
+            'gym_info': args.gym_info,
+            'encounter': args.encounter,
+            'scan_display': scan_display,
```
neskk: Here it should be `search_control` or change the template to look for `scan_display`.
Try to do things step by step, when you accepted to move flags to a dict you didn't need to include the other flags (in the same commit at least). It looks ok but I still have to test. Good work 👍

elincognito: Nice catch, i changed to `scan_display`.
And also fixed a bug with still using the values `inline` or `none` in the template.
Thks for feedback :+1:

bbdoc:
I'm not using -gi which does disable gym sidebar, but it still shows all additional switches like Team / Open Spot / ... which are useless if -gi is not enabled... I know this PR is merged already but that might require another small one ;-)

PR #1842, opened 2017-02-20
API Watchdog HTTP instead of HTTPS

dev042:
changed https to http to avoid any SSL Errors

## Description
This PR change the version check website to use http instead of https to avoid any ssl error

## Motivation and Context
I think is an overhead for a check to use SSL. To avoid any SSL change it HTTP only
`SSLError(SSLError(SSLEOFError(8, u'EOF occurred in violation of protocol (_ssl.c:661)'),),)`

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1257,15 +1257,14 @@ def get_api_version(args):
 
     try:
         s = requests.Session()
-        s.mount('https://',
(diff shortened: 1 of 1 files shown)


blueberry0101:
On the http version you will get an 404 Not Found error.

PR #1844, opened 2017-02-20
Major refraction of config.ini.example and adding missing config options

Arengo:
The file ``conf/config.ini.example`` now fully represents configuration of RocketMap.

## Description

Configuration options were retrieved out of ``pogom/utils.py`` and grouped in understandable batches in order of appearing when configuring a RocketMap process

## Motivation and Context

The current config.ini.example barely holds all needed configuration options while on the other hand contains unused options for a lot of users. The *misc* category was totally mixed up. In discord we had to answer too often which configuration flag should be used or if such an option is implemented at all. This should make them all more visible and reduce the questions in the discord help channel.

## Types of changes

- [x] Issue fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My change is part of the documentation.

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -1,62 +1,150 @@
 # DO NOT USE NOTEPAD TO EDIT CONFIG FILES!! USE NOTEPAD ++ INSTEAD.
-# Authentication settings
-#auth-service:          # ptc (default) or google
-#username:
-#password:
+# Uncomment a line when you want to change its default value (Remove # at the beginning)
+# Please ensure to leave a space after the colon if a value is needed ("example-setting" "example-setting-with: value")
+# Ot least one account with "username", "password" and "location" and "gmaps-key" are required
+# When your basic config is done, rename this file by removing the "example" extension
+# Other config-files can be included in processes by "--config-file" or "-cf" and path
+
+
+# Basic settings
+#gmaps-key:             	# Your Google Maps API key
+#auth-service:          	# ptc (default) or google
+#username:		
+#password:		
+#accountcsv:			# Load accounts instead from a CSV file containing "auth-service,username,password" lines
+
 
 # Database settings
-#db-type: sqlite        # sqlite (default) or mysql
-#db-host:               # required for mysql
-#db-name:               # required for mysql
-#db-user:               # required for mysql
-#db-pass:               # required for mysql
-#db-port:               # default 3306
+#db-type: sqlite        	# sqlite (default) or mysql
+#db:				# Database filename for SQLite.
+#db-host:               	# Required for mysql
+#db-name:               	# Required for mysql
+#db-user:               	# Required for mysql
+#db-pass:               	# Required for mysql
+#db-port:               	# Required for mysql (default=3306)
+#db-max_connections:		# Max connections (per thread) for the database.
+#db-threads:			# Number of db threads; increase if the db queue falls behind.
+#clear-db			# CAREFUL! Deletes the existing database before starting the Webserver.
+#purge-data: 			# Clear Pokemon from database this many hours after they disappear
+
+# Scan method (speed-scan preferable, (default is hex-scan)
+#speed-scan             	# Use speed-scan as the search scheduler
+#beehive			# Use beehive configuration for multiple accounts, one account per hex
+				# 	Make sure to keep -st under 5, and -w under the totalamount of accounts available
+#spawnpoint-scanning		# Use spawnpoint scanning (instead of hex grid). 
+				#	Scans in a circle based on step_limit when on DB.
+				#	A spawnpoint-file can be added optinal as value for "spawnpoint-scanning"
+#dump-spawnpoints		# Dump the spawnpoints from the db to json (only for use with -ss).
+
+
 
 # Search settings
 #location:
-#speed-scan             # To use speed-scan as the search scheduler
-#no-gyms:               # disables gym scanning (default false)
-#no-pokemon:            # disables pokemon scanning (default false)
-#no-pokestops:          # disables pokestop scanning (default false)
-#scan-delay:            # default 10
-#step-limit:            # default 12
-#gym-info:              # enables detailed gym info collection (default false)
-#max-failures:          # Maximum number of failures to parse locations before an account will go into a sleep for #account-rest-interval seconds (0 to disable), (default 5)
-#max-empty:             # Maximum number of empty scans before an account will go into a sleep for #account-rest-interval seconds. Reasonable to use with proxies (0 to disable), (default 0)
-#min-seconds-left:      # time that must be left on a spawn before considering it too late and skipping it (default 0)
-#status-name:           # enables writing status updates to the database - if you use multiple processes, each needs a unique value
+#china				# Coordinates transformer for China.
+#no-jitter:			# Don't apply random -9m to +9m jitter to location.
+#altitude: 			# Default altitude in meters. Retrieve this value regarding your scan location from the web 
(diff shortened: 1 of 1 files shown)


config/config.ini.example
```diff
@@ -1,62 +1,175 @@
...
+# Scan method (speed-scan preferable, (default is hex-scan)
+#############
+
+#speed-scan				# Use speed-scan as the search scheduler
+#beehive 				# Use beehive configuration for multiple accounts, one account per hex
+					# 	Make sure to keep -st under 5, and -w under the totalamount of accounts available
+#spawnpoint-scanning			# Use spawnpoint scanning (instead of hex grid).
+					#	Scans in a circle based on step_limit when on DB.
+					#	A spawnpoint-file can be added optinal as value for "spawnpoint-scanning"
+#dump-spawnpoints			# Dump the spawnpoints from the db to json (only for use with -ss).
```
sebastienvercammen: None of the indentation lines up.

config/config.ini.example
```diff
@@ -1,62 +1,175 @@
...
+#db:					# Database filename for SQLite.
+#db-host:				# Required for mysql
+#db-name:				# Required for mysql
+#db-user:				# Required for mysql
+#db-pass:				# Required for mysql
+#db-port:				# Required for mysql (default=3306)
+#db-max_connections:			# Max connections (per thread) for the database.
+#db-threads:				# Number of db threads; increase if the db queue falls behind.
+#clear-db				# CAREFUL! Deletes the existing database before starting the Webserver.
+#purge-data:				# Clear Pokemon from database this many hours after they disappear.
```
sebastienvercammen: Any data purging or database clearing shouldn't be in the default config, it's not something that should be repeated at each startup.

config/config.ini.example
```diff
@@ -1,62 +1,175 @@
...
+#bad-scan-retry:			# Number of bad scans before giving up on a step. (default=2, 0 to disable)
+#skip-empty				# Enables skipping of empty cells in normal scans - requires previously populated database
+					# 	(not to be used with -ss)
+#min-seconds-left:			# Time that must be left on a spawn before considering it too late and skipping it (default 0)
+#no-gyms 				# Disables gym scanning (default false)
+#no-pokemon 				# Disables pokemon scanning (default false)
+#no-pokestops 				# Disables pokestop scanning (default false)
+#gym-info 				# Enables detailed gym info collection (default false)
+#complete-tutorial			# Complete ToS and tutorial steps on accounts if they haven't already.
+#on-demand_timeout:			# Pause searching while web UI is inactive for this timeout (in seconds).
```
sebastienvercammen: Default values should be added to all entries, or removed from all of them and placed in the "commandline"/config documentation (I vote for adding to all entries). Consistency.

config/config.ini.example
```diff
@@ -1,62 +1,175 @@
...
-#ssl-privatekey:        # path to ssl private key
-#encrypt-lib:           # path to encrypt lib to be used instead of the shipped ones
-#status-page-password:  # enables and protects the /status page to view status of all workers
+####################
 
-# Uncomment a line when you want to change its default value (Remove # at the beginning)
-# Please ensure to leave a space after the colon (:) (example setting: value)
-# Username, password, location and gmaps-key are required
+#host:					# Address to listen on (default 127.0.0.1)
+#port:					# Port to listen on (default 5000)
```
sebastienvercammen: Host & port should be much higher up, they're some of the most common options. Most common should be grouped together at the very top of the file.

config/config.ini.example
```diff
@@ -1,62 +1,175 @@
...
+#manual-captcha-refresh:		# Time available before captcha page refreshes.(default=30)
+#manual-captcha-timeout:		# Maximum time captchas will wait for manual captcha solving.
+					#	On timeout, if enabled, 2Captcha will be used to solve captcha. (default=0)
+
+
+# Misc
+######
+
+#no-version-check			# Disable API version check.
+#version-check-interval			# Interval to check API version in seconds (Default: in [60, 300]).
```
sebastienvercammen: Reword & include `:`:
```
#version-check-interval:			# Interval to check API version in seconds (Default: in range [60, 300]).
```

config/config.ini.example
Arengo: Those ones are intended since it continues the text above. But I can fit it in one column, too.

config/config.ini.example
Arengo: Was not sure, too. Since these options can destroy peoples happiness.

config/config.ini.example
Arengo: I agree, will add the missing defaults.

config/config.ini.example
Arengo: Might move them to the basic settings.

config/config.ini.example
sebastienvercammen: Use spaces.

config/config.ini.example
```diff
@@ -1,62 +1,172 @@
...
+#manual-captcha-timeout:		# Maximum time captchas will wait for manual captcha solving.
+					# On timeout, if enabled, 2Captcha will be used to solve captcha. (default=0)
+
+
+# Misc
+######
+
+#no-version-check			# Disable API version check. (default=False)
+#version-check-interval:		# Interval to check API version in seconds (Default: in range [60, 300]).
+#hash-key:				# Key for hash server (default=None)
```
sebastienvercammen: Change "Basic settings" to "Required settings", and have a "Common settings" as second group. `hash-key` shouldn't be at the very bottom of the file.

Also, CTRL+F "tab", replace with four spaces.

config/config.ini.example
```diff
@@ -1,62 +1,179 @@
...
+#hash-key:                      # Key for hash server (default=None)
+#accountcsv:                    # Load accounts instead from a CSV file containing "auth-service,username,password" lines.
+#speed-scan                     # Use speed-scan as the search scheduler.
+#location:                      # Location, can be an address or coordinates.
+#step-limit:                    # Steps (default=10)
+#scan-delay:                    # Time delay between requests in scan threads. (default=12)
+#no-gyms                        # Disables gym scanning (default false)
+#no-pokemon                     # Disables pokemon scanning. (default=False)
+#no-pokestops                   # Disables pokestop scanning. (default=False)
+#print-status                   # Enable ENTER to switch between log view and a status view.
```
k4n30: should this include a : i.e `#print-status:                  # Enable ENTER to switch between log view and a status view.`

Just for consistency

k4n30:
You should also include
`#verbose:                      # Show debug messages from RocketMap and pgoapi.  Optionally specify file name to log to.
#very-verbose:            # Like verbose, but show debug messages from all modules as well.  Optionally specify file name to log to.`

config/config.ini.example
Arengo: A colon is wrong because we do not submit a value here and internally it is handled as ``True`` or ``False``. We have to submit values for all config flags which carry a colon, though.

Jesterium:
I mentioned this in #1869 but is slightly more relevant here. The changes to the config.ini example so far are exactly what we need **_but_** could we split it into **-os variables**, **-ns variables**, and **both** so users can better understand which variables to specify where?

Perhaps #1869, #1846, and #1844 should be merged?

SiteEffect:
@Jesterium splitting into -ns, -os  categories will not make it far better, imo. I tried the best to squeeze with each other related stuff into fitting categories. You are right, that it should be made clear, which flag affects which par of the scanner - that could be added in the general docs or in the python help, though.

PR #1849, opened 2017-02-21
Improve SpeedScan queue refresh performance

Obihoernchen:
## Description
- The old get_by_locs sql query was like: .`.. WHERE ((latitude IN (...) AND longitude IN (...))`
  - this is `O(n^2)` because there are two INs. This is really slow for big -st values
- ~~Use a temporary memory table to store lat,lon and join this one with scannedlocation to have `O(n)+C` behaviour which is way faster~~ 
- New: Use the already available cellids and use them for the DB query which is `O(n)` because it's only a single condition a22528d
- Replace SpeedScan.bands_filled() with a way faster version as well: ccff895
- besides the big sql performance improvement this PR reduces python cpu usage as well

Benchmark with -st 50:
```
OLD: [ schedulers ][ INFO ] New queue created with 8483 entries in 156.809612 seconds
NEW: [ schedulers ][ INFO ] New queue created with 8483 entries in 2.591794 seconds
```

## Motivation and Context
using SpeedScan with a big -st was super slow

## How Has This Been Tested?
My map and several other big maps (~100-1000 accounts).
CPU usage since 17:45 with this PR.
Map completly stopped scanning for >2min every 5min without this PR.
image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1020,14 +1020,14 @@ class ScannedLocation(BaseModel):
 
     @classmethod
     def get_band_count_by_cells(cls, cells):
-        return int(ScannedLocation
-                   .select(fn.SUM(fn.IF(ScannedLocation.band1 == -1, 0, 1)
-                                  + fn.IF(ScannedLocation.band2 == -1, 0, 1)
-                                  + fn.IF(ScannedLocation.band3 == -1, 0, 1)
-                                  + fn.IF(ScannedLocation.band4 == -1, 0, 1)
-                                  + fn.IF(ScannedLocation.band5 == -1, 0, 1))
+        return int(cls
+                   .select(fn.SUM(fn.IF(cls.band1 == -1, 0, 1)
+                                  + fn.IF(cls.band2 == -1, 0, 1)
+                                  + fn.IF(cls.band3 == -1, 0, 1)
+                                  + fn.IF(cls.band4 == -1, 0, 1)
+                                  + fn.IF(cls.band5 == -1, 0, 1))
                            .alias('band_count'))
-                   .where(ScannedLocation.cellid << cells)
+                   .where(cls.cellid << cells)
                    .scalar())
 
     @classmethod
@@ -1655,7 +1655,7 @@ class Token(flaskDb.Model):
                 if tokens:
                     log.debug('Retrived Token IDs: {}'.format(token_ids))
                     result = DeleteQuery(Token).where(
-                                 Token.id << token_ids).execute()
+                        Token.id << token_ids).execute()
                     log.debug('Deleted {} tokens.'.format(result))
         except OperationalError as e:
             log.error('Failed captcha token transactional query: {}'.format(e))
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -851,11 +851,10 @@ class ScannedLocation(BaseModel):
                 'step': scan['step'], 'sp': sp_id}
 
     @classmethod
-    def get_by_locs(cls, locs):
-        cellids = [cellid(loc) for loc in locs]
+    def get_by_cells(cls, cells):
         query = (cls
                  .select()
-                 .where(ScannedLocation.cellid << cellids)
+                 .where(cls.cellid << cells)
                  .dicts())
 
         d = {}
@@ -875,7 +874,7 @@ class ScannedLocation(BaseModel):
     def get_by_loc(cls, loc):
         query = (cls
                  .select()
-                 .where(ScannedLocation.cellid == cellid(loc))
+                 .where(cls.cellid == cellid(loc))
                  .dicts())
 
         return query[0] if len(list(query)) else cls.new_loc(loc)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1020,7 +1020,7 @@ class ScannedLocation(BaseModel):
 
     @classmethod
     def get_band_count_by_cells(cls, cells):
-        return (ScannedLocation
+        return int(ScannedLocation
                 .select(fn.SUM(fn.IF(ScannedLocation.band1 == -1, 0, 1)
                                + fn.IF(ScannedLocation.band2 == -1, 0, 1)
                                + fn.IF(ScannedLocation.band3 == -1, 0, 1)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -851,10 +851,10 @@ class ScannedLocation(BaseModel):
                 'step': scan['step'], 'sp': sp_id}
 
     @classmethod
-    def get_by_cells(cls, cells):
+    def get_by_cellids(cls, cellids):
         query = (cls
                  .select()
-                 .where(cls.cellid << cells)
+                 .where(cls.cellid << cellids)
                  .dicts())
 
         d = {}
@@ -914,12 +914,12 @@ class ScannedLocation(BaseModel):
 
     # Return list of dicts for upcoming valid band times.
     @classmethod
-    def get_cell_to_linked_spawn_points(cls, cells):
+    def get_cell_to_linked_spawn_points(cls, cellids):
         query = (SpawnPoint
                  .select(SpawnPoint, cls.cellid)
                  .join(ScanSpawnPoint)
                  .join(cls)
-                 .where(cls.cellid << cells).dicts())
+                 .where(cls.cellid << cellids).dicts())
         l = list(query)
         ret = {}
         for item in l:
@@ -1019,7 +1019,7 @@ class ScannedLocation(BaseModel):
         return scan
 
     @classmethod
-    def get_band_count_by_cells(cls, cells):
+    def get_band_count_by_cellids(cls, cellids):
         return int(cls
                    .select(fn.SUM(fn.IF(cls.band1 == -1, 0, 1)
                                   + fn.IF(cls.band2 == -1, 0, 1)
@@ -1027,7 +1027,7 @@ class ScannedLocation(BaseModel):
                                   + fn.IF(cls.band4 == -1, 0, 1)
                                   + fn.IF(cls.band5 == -1, 0, 1))
                            .alias('band_count'))
-                   .where(cls.cellid << cells)
+                   .where(cls.cellid << cellids)
                    .scalar())
 
     @classmethod
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -852,15 +852,10 @@ class ScannedLocation(BaseModel):
 
     @classmethod
     def get_by_locs(cls, locs):
-        lats, lons = [], []
-        for loc in locs:
-            lats.append(loc[0])
-            lons.append(loc[1])
-
+        cellids = [cellid(loc) for loc in locs]
         query = (cls
                  .select()
-                 .where((ScannedLocation.latitude << lats) &
-                        (ScannedLocation.longitude << lons))
+                 .where(ScannedLocation.cellid << cellids)
                  .dicts())
 
         d = {}
@@ -878,10 +873,10 @@ class ScannedLocation(BaseModel):
     # Return value of a particular scan from loc, or default dict if not found.
     @classmethod
     def get_by_loc(cls, loc):
+        cellid = cellid(loc)
         query = (cls
                  .select()
-                 .where((ScannedLocation.latitude == loc[0]) &
-                        (ScannedLocation.longitude == loc[1]))
+                 .where(ScannedLocation.cellid == cellid)
                  .dicts())
 
         return query[0] if len(list(query)) else cls.new_loc(loc)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -881,8 +881,8 @@ class ScannedLocation(BaseModel):
                          'JOIN t_scannedlocation '
                          'USING (latitude,longitude);').dicts())
 
-        # Delete temp table
-        #db.execute_sql('DROP TABLE `t_scannedlocation`;')
+        # TODO: Delete temp table
+        # db.execute_sql('DROP TABLE `t_scannedlocation`;')
 
         d = {}
         for sl in list(query):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1019,7 +1019,7 @@ class ScannedLocation(BaseModel):
         return scan
 
     @classmethod
-    def get_band_count_by_cellids(cls, cellids):
+    def get_bands_filled_by_cellids(cls, cellids):
         return int(cls
                    .select(fn.SUM(fn.IF(cls.band1 == -1, 0, 1)
                                   + fn.IF(cls.band2 == -1, 0, 1)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -873,10 +873,9 @@ class ScannedLocation(BaseModel):
     # Return value of a particular scan from loc, or default dict if not found.
     @classmethod
     def get_by_loc(cls, loc):
-        cellid = cellid(loc)
         query = (cls
                  .select()
-                 .where(ScannedLocation.cellid == cellid)
+                 .where(ScannedLocation.cellid == cellid(loc))
                  .dicts())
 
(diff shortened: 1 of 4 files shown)


pogom/models.py
```diff
@@ -851,17 +851,38 @@ def _q_init(scan, start, end, kind, sp_id=None):
...
+                         '`t1`.`longitude`, `t1`.`last_modified`,'
+                         '`t1`.`done`, `t1`.`band1`, `t1`.`band2`,'
+                         '`t1`.`band3`, `t1`.`band4`, `t1`.`band5`,'
+                         '`t1`.`midpoint`, `t1`.`width` '
+                         'FROM `scannedlocation` AS t1 '
+                         'JOIN t_scannedlocation '
+                         'USING (latitude,longitude);').dicts())
+
+        # TODO: Delete temp table
+        # db.execute_sql('DROP TABLE `t_scannedlocation`;')
```
onilton: DROP TABLE IF EXISTS `t_scannedlocation` ?

pogom/models.py
```diff
@@ -851,17 +851,38 @@ def _q_init(scan, start, end, kind, sp_id=None):
...
             lons.append(loc[1])
 
-        query = (cls
-                 .select()
-                 .where((ScannedLocation.latitude << lats) &
-                        (ScannedLocation.longitude << lons))
-                 .dicts())
+        # Create temp table with lat,long because this is
+        # way faster than using IN for two columns for a lot of parameters
+        db.execute_sql('CREATE TEMPORARY TABLE `t_scannedlocation` ('
```
onilton: Remember that some can be running multiple instances of RocketMap on the same database, so I am not sure if you can have a temporary table with a constant name. 

Maybe adding a random string to the table's name?

Obihoernchen: mysql doc states: 
> A TEMPORARY table is visible only to the current session, and is dropped automatically when the session is closed. This means that two different sessions can use the same temporary table name without conflicting with each other or with an existing non-TEMPORARY table of the same name. 
So this should be fine.

pogom/models.py
Obihoernchen: Doesn't work because this drops table before executing the query above. I dunno why... at least it didn't work in my tests.

pki0:
Great fix. Went down from 100% CPU with -st 25 and 7 runserver.py down to 30%.
Now with 15 runserver.py:
cpu

Great, thanks al lot

onilton:
Just ignore me if I am saying something stupid here.

I wonder that instead of creating a temporary table and all this db things, why we could not just call cellid, like in:

```python
cellids = [cellid(loc) for loc in locs]
query = (cls                                                                                                                                                     
                 .select()
                 .where(ScannedLocation.cellid << cellids)
                 .dicts())
```
Is `cellId()` slow? If it is we could cache it, or pre-calculate it in the calling method.

pogom/models.py
onilton: Yeah, I imagined that it was a session/connection thing, but just wanted to make sure.

onilton:
Maybe @trew  can help us with this. What you think, man?

trew:
I'm done with this. That's why I requested to be removed from #pr.

Obihoernchen:
Mhh didn't see this one. Might be a good idea.
I'll benchmark it.

Obihoernchen:
@onilton haha so much easier and cleaner thank you :+1: 
`using cellid: [    schedulers][    INFO] New queue created with 8483 entries in 2.591794 seconds`

SiteEffect:
Works fine.
``Doing 411,768 distance calcs to assign spawn points to scans``
``New queue created with 1361 entries in 0.516286 seconds``

onilton:
@Obihoernchen if you are into refactor mood, you can also remove the cellid transformation since we already have this info where the method is called:

https://github.com/RocketMap/RocketMap/blob/develop/pogom/schedulers.py#L692

Here `self.scans.keys()`. already got cellid info because it was calculate/written in:

https://github.com/RocketMap/RocketMap/blob/develop/pogom/schedulers.py#L518

I don't know how fast cellid is, I think it is not really is slow, so this probably won't make a big difference in refresh time, but anyway is nice to no to recalculate something that is already in memory.

pogom/models.py
```diff
@@ -851,16 +851,10 @@ def _q_init(scan, start, end, kind, sp_id=None):
                 'step': scan['step'], 'sp': sp_id}
 
     @classmethod
-    def get_by_locs(cls, locs):
-        lats, lons = [], []
-        for loc in locs:
-            lats.append(loc[0])
-            lons.append(loc[1])
-
+    def get_by_cells(cls, cells):
```
onilton: `get_by_cellids(cls, cellids)` is probably better (more explicit)

Obihoernchen: I used `cells` because of `def get_cell_to_linked_spawn_points(cls, cells):` for consistency. But I can change to `cellids` because I agree with you.

TheCrabity:
Quick test on st190, gave a "not enough memory to store" error, but otherwise fired up and started working quickly (which wouldn't' of happened without).

pogom/models.py
rocketrobot99: Let me start off by saying this is great! I actually had pretty much exactly the same change. I have done similar changes to ScannedLocation.get_cell_to_linked_spawn_points, Spawnpoint.select_in_hex and I have replaced ScannedLocation.bands_filled with a new  ScannedLocation.get_band_count_by_cells routine. I have attached them in files if you are interested, as they are along the same lines. Feel free to ignore them if you want, they are part of a bigger PR I am working on. However, they do improve the refresh speed, like your change. I just thought you might be interested.

The changes to ScannedLocation.get_cell_to_linked_spawn_points and Spawnpoint.select_in_hex also fixes what I believe to be some incorrect logic. When you picture the scannedlocations as circles, some spawnpoints can be in the overlaps and currently they can get selected into multiple queues. Consequently, they are also counted more than once. I have altered the SQL so that a spawnpoint is only associated with 1 Cell only. Thanks.
[Scannedlocation.get_band_count_by_cells.txt](https://github.com/RocketMap/RocketMap/files/792759/Scannedlocation.get_band_count_by_cells.txt)
[spawnpoint.select_in_hex.txt](https://github.com/RocketMap/RocketMap/files/792760/spawnpoint.select_in_hex.txt)
[get_cell_to_linked_spawn_points.txt](https://github.com/RocketMap/RocketMap/files/792761/get_cell_to_linked_spawn_points.txt)

Obihoernchen: Diff of your changes applied to my branch for easier viewing:
```diff
diff --git a/pogom/models.py b/pogom/models.py
index be2ef1b..be69b75 100644
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -915,11 +915,30 @@ class ScannedLocation(BaseModel):
     # Return list of dicts for upcoming valid band times.
     @classmethod
     def get_cell_to_linked_spawn_points(cls, cells):
+
+        # Get all spawnpoints from the hive's cells
+        sp_from_cells = (ScanSpawnPoint
+                         .select(ScanSpawnPoint.spawnpoint)
+                         .where(ScanSpawnPoint.scannedlocation << cells)
+                         .alias('spcells'))
+        # Allocate a spawnpoint to one cell only
+        one_sp_scan_loc = (ScanSpawnPoint
+                           .select(ScanSpawnPoint.spawnpoint,
+                                   fn.MAX(ScanSpawnPoint.scannedlocation).alias(
+                                       'cellid'))
+                           .join(sp_from_cells, on=sp_from_cells.c.spawnpoint_id
+                                 == ScanSpawnPoint.spawnpoint)
+                           .group_by(ScanSpawnPoint.spawnpoint)
+                           .alias('maxscan'))
+        # As scan locations overlap, spawnpoints can belong to up to 3 locations
+        # This sub-query effectively assigns each SP to exactly one location.
+
         query = (SpawnPoint
-                 .select(SpawnPoint, cls.cellid)
-                 .join(ScanSpawnPoint)
-                 .join(cls)
-                 .where(cls.cellid << cells).dicts())
+                 .select(SpawnPoint, one_sp_scan_loc.c.cellid)
+                 .join(one_sp_scan_loc, on=(SpawnPoint.id ==
+                                            one_sp_scan_loc.c.spawnpoint_id))
+                 .where(one_sp_scan_loc.c.cellid << cells)
+                 .dicts())
         l = list(query)
         ret = {}
         for item in l:
@@ -929,6 +948,18 @@ class ScannedLocation(BaseModel):
 
         return ret
 
+    @classmethod
+    def get_band_count_by_cells(cls, cells):
+
+        return (ScannedLocation
+                .select(fn.SUM(fn.IF(ScannedLocation.band1 == -1, 0, 1)
+                               + fn.IF(ScannedLocation.band2 == -1, 0, 1)
+                               + fn.IF(ScannedLocation.band3 == -1, 0, 1)
+                               + fn.IF(ScannedLocation.band4 == -1, 0, 1)
+                               + fn.IF(ScannedLocation.band5 == -1, 0, 1)).alias('band_count'))
+                .where(ScannedLocation.cellid << cells)
+                .scalar())
+
     @staticmethod
     def visible_forts(step_location):
         distance = 0.45
@@ -1327,38 +1358,33 @@ class SpawnPoint(BaseModel):
                     3600) / 15 / 60)
 
     @classmethod
-    def select_in_hex(cls, center, steps):
-        R = 6378.1  # KM radius of the earth
-        hdist = ((steps * 120.0) - 50.0) / 1000.0
-        n, e, s, w = hex_bounds(center, steps)
-
-        # Get all spawns in that box.
-        sp = list(cls
-                  .select()
-                  .where((cls.latitude <= n) &
-                         (cls.latitude >= s) &
-                         (cls.longitude >= w) &
-                         (cls.longitude <= e))
-                  .dicts())
+    def select_in_hex(cls, center, steps, cells):
+        # Get all spawnpoints from the hive's cells
+        sp_from_cells = (ScanSpawnPoint
+                         .select(ScanSpawnPoint.spawnpoint)
+                         .where(ScanSpawnPoint.scannedlocation << cells)
+                         .alias('spcells'))
+        # Allocate a spawnpoint to one cell only
+        one_sp_scan_loc = (ScanSpawnPoint
+                           .select(ScanSpawnPoint.spawnpoint,
+                                   fn.MAX(ScanSpawnPoint.scannedlocation).alias(
+                                       'Max_ScannedLocation_id'))
+                           .join(sp_from_cells, on=sp_from_cells.c.spawnpoint_id
+                                 == ScanSpawnPoint.spawnpoint)
+                           .group_by(ScanSpawnPoint.spawnpoint)
+                           .alias('maxscan'))
 
-        # For each spawn work out if it is in the hex (clipping the diagonals).
+        query = (cls
+                 .select(cls)
+                 .join(one_sp_scan_loc,
+                       on=(one_sp_scan_loc.c.spawnpoint_id == cls.id))
+                 .where(one_sp_scan_loc.c.Max_ScannedLocation_id << cells)
+                 .dicts())
+        log.info(query.database.get_cursor().mogrify(*query.sql()))
         in_hex = []
-        for spawn in sp:
-            # Get the offset from the center of each spawn in km.
-            offset = [math.radians(spawn['latitude'] - center[0]) * R,
-                      math.radians(spawn['longitude'] - center[1]) *
-                      (R * math.cos(math.radians(center[0])))]
-            # Check against the 4 lines that make up the diagonals.
-            if (offset[1] + (offset[0] * 0.5)) > hdist:  # Too far NE
-                continue
-            if (offset[1] - (offset[0] * 0.5)) > hdist:  # Too far SE
-                continue
-            if ((offset[0] * 0.5) - offset[1]) > hdist:  # Too far NW
-                continue
-            if ((0 - offset[1]) - (offset[0] * 0.5)) > hdist:  # Too far SW
-                continue
-            # If it gets to here it's a good spawn.
+        for spawn in list(query):
             in_hex.append(spawn)
+        log.info('done')
         return in_hex 
```

pogom/models.py
```diff
@@ -1019,14 +1019,15 @@ def update_band(cls, scan):
...
-    def bands_filled(cls, locations):
-        filled = 0
-        for e in locations:
-            sl = cls.get_by_loc(e[1])
-            bands = [sl['band' + str(i)] for i in range(1, 6)]
-            filled += reduce(lambda x, y: x + (y > -1), bands, 0)
-
-        return filled
+    def get_band_count_by_cells(cls, cells):
+        return (ScannedLocation
```
onilton: Legilibity suggestion: change `ScannedLocation` to `cls` (less repetittion).

Obihoernchen: See 0a9153a

SiteEffect:
Cuts CPU in half.
(not like that https://c1.staticflickr.com/3/2220/2322199702_94a61474d6_z.jpg?zz=1)

pogom/models.py
sebastienvercammen: @Obihoernchen Yup, change to cellids if you can for clarity.

Obihoernchen: @sebastienvercammen done 0a6d0e8

pogom/models.py
```diff
@@ -1026,14 +1019,16 @@ def update_band(cls, scan):
...
     @classmethod
-    def bands_filled(cls, locations):
-        filled = 0
-        for e in locations:
-            sl = cls.get_by_loc(e[1])
-            bands = [sl['band' + str(i)] for i in range(1, 6)]
-            filled += reduce(lambda x, y: x + (y > -1), bands, 0)
-
-        return filled
+    def get_band_count_by_cellids(cls, cellids):
```
smc87: get_bands_filled_by_cellids would be abit clearer for me.
With the current name i had to double check to see if it was totalcount or filled count being returned.

Anyone else?

Obihoernchen: I agree with you. Will change this.

Obihoernchen: @smc87 done bde7b67

rhaj:
something is probably wrong in this PR. After pulling it I received errors from performance statistics and removing this PR solved the problem. Tested on Linux.

2017-02-23 20:34:52,796 [ search-overseer][    schedulers][   ERROR] Performance statistics had an Exception: ValueError("Unknown format code 'f' for object of type 'str'",)
  File "xxxx/pogom/schedulers.py", line 846, in schedule
    self.active_sp - self.tth_found)
ValueError: Unknown format code 'f' for object of type 'str'

onilton:
@rhaj can you grep your log for 

```
Exception in band_status: Exception message
```

and paste the result here ?

rhaj:
2017-02-23 21:55:04,179 [ search-overseer][    schedulers][   ERROR] Exception in band_status: Exception message: OperationalError('no such function: IF',)

Obihoernchen:
@rhaj do you use sqlite?

rhaj:
yes, standard instalation with included sqlite

Obihoernchen:
That's the problem. There is no IF in sqlite. I'll push a fix. 
But please use mysql with speedscan! Sqlite is not made for this.

onilton:
@sebastienvercammen what should we do about this? This totally breaks sqlite.  We have decided to not do anything to keep supporting sqlite. Or is this an exception?

rhaj:
OK, thank you for clarification. I will try to migrate to mysql. But I recommend you to make a big note about this because there is no "warning" that mysql is mandatory for this feature.

sebastienvercammen:
sqlite supports `CASE`, replace `IF` w/ `CASE` to solve the issue. We can also rewrite the IFs to get the same result but w/o IFs.

TheCrabity:
On large sizes, there is a bug related to the delay between when a worker has moved to a position and Scanning has been triggered, to when they are suppose to be Searching. As you add more workers, this delay  between Scanning and Searching increases, in my testing, to upwards of 600 seconds. Easily repeatable by throwing a lot of workers at any size -ST.

Obihoernchen:
Yes I noticed the same. There is still a lot of stuff to do to scale SpeedScan successfully. The SQL part is fine now, but the python part needs some more optimization/fixes.

PR #1860, opened 2017-02-23
Add non-copyrighted icons

onilton:
## Description
This will make RocketMap work out-of-the-box.

These icons are just circles with the monster number in it.
The color of the circle is the same color of the first type
of the monster (from pokemon.json).

## Motivation and Context
Currently, RocketMap needs external files to work.
``
*Why numbers? I don't like them. How about this cool/custom icons someone did that I found online?*

**We can't use any icon that resembles original pokemon icons. Legal issues, period.**

## How Has This Been Tested?
In my map

## Screenshots:
screenshot

## Types of changes

## Checklist:


titanicboyke:
I do not know all the numbers of pokemon out of my head, so I am not going to use it! but nice work!

M4d40:
May be we could use these "CC-BY" licenced icons:
http://theartificial.nl/pokemonicons/

They actually updated it a week ago to have all 251 Pokemon.

The creator is active on reddit
(https://www.reddit.com/r/pokemongodev/comments/5u8pyf/251/)

Although i'm not sure how the copryight is handled world wide .__.

titanicboyke:
@M4d40 
We also use these icons for our Worldopole ;-)
image

onilton:
@titanicboyke  @M4d40  they resemble the original pokemon icons, we can't use them. That's why we have this PR with only numbers.

SiteEffect:
@titanicboyke I am always saying, some are copyright, some are trademarks. Even when we may use them under licence CC BY 3.0 in regards to the author, you may not link them to any activities regarding trademarked content since they are looking like actual Pokémon.

In general, double type color would have been nice, too. Actually I think we could do everything with a single < div > and CSS which we dynamically change via JS (or is this a runtime issue)

PR #1863, opened 2017-02-24
Support for high-resolution sprites for gym members and stats

sLoPPydrive:
Allows for higher resolution spritesheets for gym- and stats-icons.

## Description
Currently if you have gym info enabled or look at the Pokémon statistics and you are using a high resolution (retina) screen you will notice that Pokémon icons are somehow ugly and blurred. The reason is that the spritesheet uses 30x30 icons but the screen displays them in a 60x60 pixel area. Replacing the spritesheet by a higher resolution one does not help. What is missing is a background-size property which defines the size as if it were actually still a 30x30 spritesheet.

Now you can replace your spritesheet by a higher resolution one.

As a bonus I have replaced the standalone scss files for each spritesheet with a mixin. Might make things more readable and maintainable.

## Motivation and Context
Blurry icons make my eyes hurt. ;-)

## How Has This Been Tested?
Tested on my local machine with Chrome for Mac.

Note that it is somehow tricky to load the new spritesheet on an iOS device. You cannot clear your cache without also clearing (and syncing!) your history which is something I don't want, so I just have to wait until Safari decides to load the sheet again.

Before, using 30x30 icons:
before-30

Before, using 60x60 icons (completely broken):
before-60

After, using 30x30 icons (still works):
after-30

After, using 60x60 icons (works, too, but looks far better):
after-60

## Types of changes
- [X] Cosmetics

## Checklist:


fantasy4u:
you pic seems is 3d icon...@@, different from the orginal

MARTINI1:
Tested and works well, now the sprites look great

On the other hand, do not improve the sprites in the gym-sidebar and stats?

In stats does not matter much but in gym-sidebar I think it is important that they looking good, although the correct thing is that all sprite show correct on all parts of the map

Here I leave a screenshots of how they look in chrome for android




Good PR ;)
eyes appreciate it, the HD mode for rocketmap lol xD

sLoPPydrive:
I found it impossible to force my iPhone to reload the smaller sprites but after a while they were reloaded and all was good. So just try again tomorrow. :-D

static/sass/libs/_mixins.scss
```diff
@@ -37,4 +37,31 @@
...
+/// @param {string} $filename The filename of the spritesheet, including file extension, excluding path
+/// @param {number} $icon_size The size of one icon, e.g. 80px
+@mixin pokemon-sprites($filename, $icon_size) {
+  /// These are constant for now
+  $icons_per_row: 28;
+  $total_icons: 493;
+
+  display: inline-block;
+  background: {
+    image: url(../../../#{$filename});
```
Arengo: Wouldn't ``/#{$filename}`` already work since this will address project root dir?

sebastienvercammen: Yes, `/` should be used if we want to start from project root directory, unless the relative position is more important. @sLoPPydrive should decide.

sLoPPydrive: Hmm, shouldn't it be `/static/#{$filename}` then? I'm not an expert but referencing project root dir sounds wrong when the actual image file is in the `static` subdir. Please correct me if I'm wrong.

sLoPPydrive: Just tested... it doesn't work with `/#{$filename}` and as `url(../../../` is also used in other places it should stay like this. For now. :) So I decide for the already working version.

PR #1877, opened 2017-02-26
Optimize mysql data types for pokemon and gympokemon table

Obihoernchen:
## Description
- let's not waste space for each pokemon
- Use float for height and weight instead of double
  - API uses float as well
  - App displays only 2 digits after dot
- Use smallint for gender
  - no need for integer for such a few genders
  - we can't use tinyint because peewee doesn't support it

## Motivation and Context
- it makes some difference for huge pokemon tables

## How Has This Been Tested?
- tested with mysql and sqlite but please test before merge

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2505,7 +2505,3 @@ def database_migrate(db, old_ver):
                            'MODIFY COLUMN `height` FLOAT NULL DEFAULT NULL,'
                            'MODIFY COLUMN `gender` SMALLINT NULL DEFAULT NULL'
                            ';')
-            db.execute_sql('ALTER TABLE `gympokemon` '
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -102,8 +102,8 @@ class Pokemon(BaseModel):
     individual_stamina = IntegerField(null=True)
     move_1 = IntegerField(null=True)
     move_2 = IntegerField(null=True)
-    weight = DoubleField(null=True)
```
sebastienvercammen: `DecimalField` w/ fixed precision.

Obihoernchen:
_Note: Actually float and decimal(6,2) are both 4 bytes._

Obihoernchen:
I changed it back to Float because I can't see a real advantage in decimal for us.

PR #1882, opened 2017-02-27
Hash key status page fix

walaoaaa1234:
## Description

if `hash_key` not given, using the hash key status page will cause status printer failed.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -254,7 +254,7 @@ def status_printer(threadStatus, search_items_queue_array, db_updates_queue,
             status = '{:21} | {:9} | {:9} | {:9}'
             status_text.append(status.format('Key', 'Remaining', 'Maximum',
                                              'Peak'))
-            if hash_key != None:
+            if hash_key is not None:
                 for key in hash_key:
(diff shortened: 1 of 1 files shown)


pogom/search.py
```diff
@@ -254,19 +254,19 @@ def status_printer(threadStatus, search_items_queue_array, db_updates_queue,
             status = '{:21} | {:9} | {:9} | {:9}'
             status_text.append(status.format('Key', 'Remaining', 'Maximum',
                                              'Peak'))
+            if hash_key != None:
```
M4d40: if you check for None, it's more pythonic to use:
`if hash_key is not None:`

neskk:
What error are you fixing, can you show an example please?
I'm asking this because you put an `if` in a place where it doesn't really fix anything... if there aren't hash_keys the loop will run 0 times, just the same as with your `if hash_key is not None:`

sebastienvercammen:
@neskk `'NoneType' object is not iterable` http://ideone.com/sMwPSX

PR #1888, opened 2017-02-28
Improved IP Check

neskk:
Checks if X-Forwarded-For is present and uses that IP to check against the blacklist.

## Types of changes

## Checklist:


sebastienvercammen:
👎 

Client headers shouldn't be trusted, and nginx and apache can rewrite the IP header to be the X-Forwarded-For address when it's coming from a trusted proxy.

FrostTheFox:
if you wanted to check if it's from the reverse proxy, you'll probably have to have a flag to set the reverse proxy IP, only then check the X-Forwarded-For

sebastienvercammen:
@FrostTheFox Not necessary, both nginx and apache support this functionality by themselves. Define a trusted proxy, rewrite client address to be X-Forwarded-For when it's from a trusted proxy, otherwise force drop the header (because it means the client set one him-/herself).

FrostTheFox:
So if it's not behind a reverse proxy, what's to stop someone from sending a false X-Forwarded-For then?

sebastienvercammen:
@FrostTheFox Nothing, that's why I don't want to use it ;) Just use the request's IP and use nginx/apache to rewrite the request IP to `X-Forwarded-For`'s value when it's coming from a trusted proxy.

FrostTheFox:
That was more my point, if we wanted this to work both ways I'm pretty sure that'd be the way we have to do it

neskk:
I really think that some tweaking can be done both in Apache and NGINX to override the header 'X-Forwarded-For' to make sure clients don't intentionally supply it.
@sebastienvercammen you may not like this but as it stands my instances are doing hundred of thousands of comparisons against the black list with the same (wrong) IP address because I'm behind a reverse proxy. I think this is a problem and a solution is required, otherwise I might just disable the blacklist and stop doing redundant work.

sebastienvercammen:
@neskk In case you didn't understand what I said earlier, I will reclarify:

1. You use a reverse proxy. Flask gets the proxy's IP rather than the real client's address.
2. **Both apache and nginx** support trusted reverse proxies and header rewriting.
3. Rather than use X-Forwarded-For in Flask, rewrite the client's IP address in apache2 and nginx.

This makes it safe, no fiddling possible by any client, and no use of client headers not set by third parties. I'm closing the PR, as it serves absolutely no use if you use apache2/nginx properly for reverse proxying.

neskk:
@sebastienvercammen 
2. and 3. How? From what I understand: https://rocketmap.readthedocs.io/en/develop/extras/apache2.html doesn't enables this mode of "rewriting" client IP address which causes the useless IP comparisons.

sebastienvercammen:
As explained in Discord, I dropped the ball on this one, I should get some more sleep.

We need this PR, but it should be rewritten to take a list of trusted servers (e.g. `--trust [a, b, c]` or `--trust a --trust b`, there can be a discussion about the format) and use the `X-Forwarded-For` header *if* we're receiving the request from a trusted server and the header is set. If not, default to IP.

neskk:
Don't use this yet, I think it has problems parsing some weird IPs
```
File "../pogom/app.py", line 97, in validate_request
    if self._ip_is_blacklisted(ip_addr):
  File "../pogom/app.py", line 112, in _ip_is_blacklisted
    return start <= dottedQuadToNum(ip) <= end
  File "../pogom/utils.py", line 831, in dottedQuadToNum
    return struct.unpack("!L", socket.inet_aton(ip))[0]
error: illegal IP address string passed to inet_aton
```

pogom/utils.py
```diff
@@ -406,6 +406,11 @@ def get_args():
     parser.add_argument('--disable-blacklist',
                         help=('Disable the global anti-scraper IP blacklist.'),
                         action='store_true', default=False)
+    parser.add_argument('-tp', '--trusted-proxies', default=None,
```
sebastienvercammen: Default should be an empty list rather than `None`, because None is not iterable (so `ip_addr in args.trusted_proxies` will cause an error if it's `None`).

neskk: I did this and pushed without even testing it, thanks for the notice will fix it right away 👍

hokiepokedad:
What about users behind both Cloudflare and NGINX?

PR #1891, opened 2017-03-01
Redirect logging messages to stderr or stdout depending on the logging level

lluisd:
Redirect logging messages to stderr or stdout depending on the logging level

## Description
Redirect messages lower than WARNING to stdout and redirect messages equal or higher than WARNING to stderr

## Motivation and Context
Currently all logging messages are redirected to stderr without take into account the logging level, so DEBUG and INFO levels are redirected to stderr. This is so annyoing when you are using some tool to run your python processes which logs stdout and stderr to distinct log files (like Docker, PM2...) and as a result all of them go to the error log file.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -33,11 +33,35 @@ from pogom.proxy import check_proxies, proxies_refresher
 # Currently supported pgoapi.
 pgoapi_version = "1.1.7"
 
+
+class LogFilter(logging.Filter):
+
+    def __init__(self, level):
+        self.level = level
+
+    def filter(self, record):
+        return record.levelno < self.level
+
+
 # Moved here so logger is configured at load time.
-logging.basicConfig(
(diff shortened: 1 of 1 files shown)


lluisd:
@sebastienvercammen why do you set "enchancement"?

I'm a noob with Python. This is a quick fix but it works. If someone can give a feedback...

I placed them to runserver.py (without any new py file or inside some existing file like utils.py since it's a simple init logic to set the proper log handlers like before).

I forced the stdout to be from debug to info and the stderr from warning to the max since I see that as a standard (not editable)

elincognito:
I use *pm2* so this is really appreciated.
Testing it right now will provide feedback in a few hours, until now working fine
Good job @lluisd  👍

elincognito:
Working fine 5+ hours using `pm2`
👍

Issue #1822, opened 2017-02-14
Speed scan slowly grinds to halt, misses scans, 100% cpu - logs inside


dajoker17:
## Expected Behavior

Speed scan should continue to run without requiring restarting every day or so, or using 100% cpu

## Current Behavior

On some instances it will run for a number of hours, then slowly start pegging the CPU, fail to schedule new scans, start missing scans (too late for scan). 
This seems to occur randomly on different instances over time.  I have tried moving co-ordinates and this restart the initial scan but will eventually fail again.  I am almost always guaranteed to have multiple instances doing this a few days after a nest migration, making me wonder if its a spawn point issue?  Mostly it goes away if the database is dropped and recreated.  on this instance the initial scan has not completed, but I have seen the same behaviour on 100% too.

Sorry for the massive log file, but I thought a complete log might be handy to find out whats going on. 

[well1f.zip](https://github.com/RocketMap/RocketMap/files/775218/well1f.zip)

Summary:
The instance last scanned at line 36,9473
It then seems stuck in this loop:
2017-02-15 06:46:22,801 [    db-updater-3][        models][   DEBUG] Inserting items 0 to 1.
2017-02-15 06:46:22,816 [    db-updater-4][        models][   DEBUG] Inserting items 0 to 60.
2017-02-15 06:46:23,598 [    db-updater-0][        models][   DEBUG] Upserted to WorkerStatus, 60 records (upsert queue remaining: 0).
until I restarted it restarted on line 36,9937 Where it continued to run and quickly updated the map with the missing scans.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
Latest, develop build, windows server 2012 hyper-v, dedicated mysql database,

* Environment name and version (e.g. Python 2.7):
* Operating System and version (desktop or mobile):

myntath:
I was just having something similar to this happen to me yesterday. Noticed number of pokemon scanned was log after investigation I saw that CPU was at 100% all from one instance (out of four). After a restart still 100% until all scans were found then it seemed to be doing alright. I went to bed but this morning I see that the scans dropped off again.

I will try investigate if I can repeat this.

dajoker17:
Its repeated at every nest migration for me.  It sometimes takes days to show up.
I have tried picking a new location on the map and starting the initial speed scan again with the same results.
[...]

PR #1893, opened 2017-03-01
Fix bug when finding TTH for hour long spawns

myntath:
## Description

Fix bug when finding TTH for hour long spawns. This arose because the algorithm (around 1490 to 1499) expected data sorted by date. It was receiving data sorted by id.

## Motivation and Context

I noticed my speedscan was struggling to find TTH for some hour long spawnpoints. After working through the code I realised it expected spawns sorted by date. Spawns were being returned sorted by the id key which is not necessarily ordered by date.

## How Has This Been Tested?

Local machine. My remaining TTHs are being found

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1411,6 +1411,7 @@ class SpawnpointDetectionData(BaseModel):
         # Get past sightings.
         query = list(cls.select()
                         .where(cls.spawnpoint_id == sp['id'])
+                        .order_by(cls.scan_time.asc())
                         .dicts())
 
         if sighting:


onilton:
Can you send a copy of your database dump? Maybe just a select from spawnpointdetectiondata where spawnpoint = X proving this situation really happens?

onilton:
Also, we seen to sort the entries after retrieving them, as you can see in:

https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L1424

myntath:
My database has resolved the problem now so I don't have all the problem spawns (previously had 20/3100 so its not super common). Here is a link to data from one spawn (I may try restore a database later to dump some info). This one wasn't in fact problematic yet but was short enough for me to run through manually. You can see the error that is being made. This won't always lead to an issue and most of the time you will still randomly find the TTH, but not always. http://pastebin.com/E3cv96bw

You will see I have put 'x' against the table (the table is the same select query as the code) to show the rows that were actually selected by the if in 1496. The rows deselected are pretty much random because the data is not sorted by date.

Yes there is a sort later but that is a sorted list by seconds after the hour. That should not be the case here, for hour long spawns we need the data sorted by date so that all the same encounter_ids remain together. There might be nicer ways to do this but I just went with the minimal fix given what appears to be the original intention. 1490 uses query directly, which hasn't been sorted, and this is where the problem arises becuase the for loop just runs through the data in order checking against the next one in the list. This produces two problems, some of the data is not selected from the if statement when it should be and the deltas are sometimes negative values.

myntath:
The other thing I should mention is that the problem spawns show up as having the difference between latest_seen and earliest_unseen be a very small amount, often less than 10 seconds. Given the server will show us TTH if < 90 this should not be possible.

Here is a dump of a table spawnpoint and spawnpointdetectiondata (called spawnpointbak and spawnpointdectiondatabak) containing some problem spawns, https://mega.nz/#!L8YUkCBB!ZhHl83J1ePHuTiSz_1faho-OJS2kjyeF2BIcqiGkVbs

Here is another pastebin with relevant database queries, focusing on spawn 6acd4c8c161 from that table which exhibits the bug (There are other spawnpoints which also face this issue). http://pastebin.com/d3JDZNug

At the end of the paste is a copy of the new data for the same spawnpoint showing where the correct TTH is. Whether this bug shows itself is completely random and depends on what time and spawn_id occurs, so this particular point had resolved once I cleared the db.

onilton:
Ok, I think now I got what the problem is. :slightly_smiling_face: 

Looking at the data you provided I could not find one encounter_id that is not together if it others. So the first thing you said:

> That should not be the case here, for hour long spawns we need the data sorted by date so that all the same encounter_ids remain together. 

Is already true, so that is not really the problem.

Looking at the code I saw this:

https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L1774

Maybe the intent of the original author is that adding the now_secs was to make the ids respect and other (so no need to sort)? What he forgot was to pad the values with 0, maybe? Because otherwise 1000 comes first than 200.

So that's why your PR work. As you said, that code expected sorted stuff, but it isn't really sorted:

https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L1489

Sorry if you already know that, but I was trying to understand and maybe this will help people and other reviewers to understand and approve the code.

klatterhans:
Thanks! @myntath 
This fixes the issue of #1822

vincent-v-c:
I had some user complaints about 60 minutes spawns with wrong tth info. Implemented this and the TTH's are finally being found.

myntath:
@onilton I think you post above is right but also not grasping the full problem. Also my explanations are probably to blame, I struggle to clearly express my thinking on this type of issue.

Yes the id's must be together and already are together. But the sets of id's must also be in date order. This is because of the 'complement trick' that is used. When speedscan detects the id has changed in less than an hour it inserts the opposite start and end points as valid data (i.e. if we know a spawn just changed  in the last 5 minutes then it must have not changed during the other part of an hour). Which is perfect to speed up detection. However, if the date goes backward (as some of my example data does) then we start doing these complement points where we shouldn't causing problems. We could have added a check to make sure the time is < 1 hour (as it currently is) and > 0 but this is necessary if the dates are in order it can't be less than zero.

So padding the id's with zeros would help (providing we still keep id groups together). It would not fix all the problems though. 

Even though this is already merged I appreciate your clearly worded comments and am just hoping to get my thoughts down for possible future reference and maybe someone else will see this as well. Particularly because it looks like there is a bit more work being done on speedscan.

PR #1857, opened 2017-02-23
Completing tutorial with a final Pokéstop spin

Arengo:
This PR focuses on implementing a single spin of a Pokéstop in the end of the RocketMap account tutorial, resulting in a level 2 *ding* of the worker. 
## Description

The PR adds functionality to the ``--complete-tutorial`` flag. After the basic tutorial, the worker is checked with the first map-call if it is still level 1, if that is the case a nearby Pokéstop is spun. With this single Pokéstop spin, the Pokéstop-of-the-day bonus will result in an immediate rise to level 2 beucause of the the basic tutorial before. With the correct settings a big number of workers can complete the tutorial and level up to level 2 fast. How it works is detailed in the ``docs/extras/tutorial.md`` documentation.

## Motivation and Context

With the introduction of reCaptcha to Pokémon GO a new goal was set, to minimize those occurrences on your workers. While this is achievable with a good scanning method, timing and speed-limit, you won't get rid or almost rid of them. Running the newest API together with a not described hashing service and slightly level accounts (level 2) will result in almost no reCaptcha for your worker.

## How Has This Been Tested?

By myself and various Discord users, leveling up their new accounts. Here, a fresh installation was used together with a pre-configured ``config.ini``. Here, we were able to excel in more than 25 accounts per 90 seconds.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project. (we will see, dear flake8)
- [x] My change requires a change to the documentation. 
- [x] I have updated the documentation accordingly. (I set up a basic documentation already)

PR #1894, opened 2017-03-01
Fixed logging alignment issues

FizzBuzz791:
## Description

Adjusted the logging padding to account for the extra characters that the captcha-solver can introduce (0...999), which throws out the alignment.

## Motivation and Context

I'm a little bit OCD and it makes me feel better. Not sure if it solves any problems for anyone, it's just a lot more visually appealing. Potentially redundant when #1857 gets merged as it's unlikely that hundreds of captchas will need to be solved.

## How Has This Been Tested?

Tested locally - See screenshots.

Before:
image

After:
image

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -35,7 +35,7 @@ pgoapi_version = "1.1.7"
 
 # Moved here so logger is configured at load time.
 logging.basicConfig(
-    format='%(asctime)s [%(threadName)16s][%(module)14s][%(levelname)8s] ' +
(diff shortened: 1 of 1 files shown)


Arengo:
Only redundant for those who run hashing service.  I do think that this can be useful in the future and will be already useful for those not running hashing service. You should focus to add indexes X as a form of logging somewhere else instead of [captcha-solver-X].

PR #1895, opened 2017-03-01
Updated API to 0.57.3

Skorfius:
Pgoapi update


pogom/search.py
```diff
@@ -353,7 +353,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
-    api_version = '0.57.2'
+    api_version = '0.57.3'
```
SkOODaT: pretty sure you only change this when they force the new version.... doing this now would cause scanners to not work

BrownSlaughter: It is changed to what ever version we have up to date, this is to stop it using one we do not have yet.

myntath: I can confirm this currently stop scanners working. I tried these changes now and it paused because the api version didn't match.

BrownSlaughter: ah yes you are correct because of the check it does online against forced, this probably wont get merged until forced anyway

Obihoernchen: What's about changing the watchdog to check whether api_version is >= forced api?

FrostTheFox:
a167384815c1c9a03068ce74ba3396979e1fb166

pogom/search.py
Skorfius: ^that would be nice

PR #1896, opened 2017-03-01
Improve DB Schema

Obihoernchen:
## Description
This PR improves the current DB schema. It adds some missing indexes and changes some int columns to smallint to save some space. This should improve the performance and lower the sql cpu usage slightly.

**Please be patient, the migration can take really long!**

It adds missing indexes to the following columns:
- gym.last_scanned
- gymmember.last_scanned
- gymmember.pokemon_uid
- gympokemon.trainer_name
- pokestop.last_updated
  - included already but my DB missed it. It looks like a previous migration step missed that.
  - So let's check whether the index exists and add it
- pokestop.active_fort_modifier
- spawnpointdetectiondata.spawnpoint_id
- token.last_updated

Furthermore the following columns were changed from `int` to `smallint`, because that's more than sufficient:
```
pokemon
    pokemon_id
    individual_attack
    individual_defense
    individual_stamina
    move_1
    move_2
		
gym
    team_id
    guard_pokemon_id
	
scannedlocation
    band1
    band2
    band3
    band4
    band5
    midpoint
    width
				
spawnpoint
    latest_seen
    earliest_unseen

spawnpointdetectiondata
    tth_secs

versions
    val

gympokemon
    pokemon_id
    cp
    num_upgrades
    move_1
    move_2
    stamina
    stamina_max
    iv_defense
    iv_stamina
    iv_attack
			
trainer
    team
    level
```

## Motivation and Context
missing indexes are bad and I saw that int is unnecessary

## How Has This Been Tested?
test map with mysql and sqlite
and my map

## Screenshots
I added an additional message because this DB schema upgrade can take some time:
```
[      MainThread][        models][    INFO] Detected database version 15, updating to 16...
[      MainThread][        models][    INFO] This DB schema update can take some time. Please be patient.
```

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2580,7 +2580,7 @@ def database_migrate(db, old_ver):
             migrator.add_index('gymmember', ('pokemon_uid',), False),
             migrator.add_index('gympokemon', ('trainer_name',), False),
             # was missing in a previous migration
-            migrator.add_index('pokestop', ('last_updated',), False),
+            #migrator.add_index('pokestop', ('last_updated',), False),
             migrator.add_index('pokestop', ('active_fort_modifier',), False),
             migrator.add_index('spawnpointdetectiondata', ('spawnpoint_id',),
                                False),
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2506,3 +2506,19 @@ def database_migrate(db, old_ver):
                            'MODIFY COLUMN `height` FLOAT NULL DEFAULT NULL,'
                            'MODIFY COLUMN `gender` SMALLINT NULL DEFAULT NULL'
                            ';')
+
+    if old_ver < 16:
+        # add some missing indexes
+        migrate(
+            migrator.add_index('gym', ('last_scanned',), False),
+            migrator.add_index('gymmember', ('last_scanned',), False),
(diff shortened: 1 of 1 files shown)


mpw1337:
> pokestop.last_updated
included already but my DB missed it (very old DB, maybe missed in previous migration steps)?
If you have an older DB can you check whether you have this index in your DB please?

My db doesn't have this index either.
image

Obihoernchen:
@mpw1337 thank you for your report! I included a simple check to add this index if it's missing.

jangrewe:
Ah, indexes! My DB server is churning quite a bit, mostly because of other stuff creating quite a bit of DB IO, so i'm very looking forward to this.

3my:
Hey I am a bit late to this but I am just wondering if it would be a good idea to conditionally add the index i.e. if index does not exist then create? Additionally it would be good if there was some output to the console to notify of update completing.

Thanks for the time you guys put into this btw you guys are awesome

Obihoernchen:
For a status report during the upgrade run the following mysql command: `show processlist;`
The index addition isn't the big part here. Modifying the column types takes a lot of time though.

PR #1688, opened 2017-01-09
Add geofencing support (WIP)

birdstream:
## Description

Added the ability to add a geofence to the map. You can create your geofence coordinates with an online tool like [this one](http://codepen.io/bgus/full/dXxLjp/)
Copy and paste to a text file and add the -gf (or --geofence-file) parameter to the command line like:
`-gf your_geofence_file.txt`
or if you want to exclude a certain area use
`-fa your_geofence_file.txt`

This PR currently only takes one -gf and -fa argument, but you can add more than one polygon in the same file

Be aware! This is very, very WIP. My programming sucks and it could probably be much better done by someone else. I'm not sure what impact it may have on scanning logic etc. But it seems to work well with speedscan. As things are now, dont use geofencing with HexSearch unless together with some speed-limiting PR, as the resulting path will skip around (path logic needs to be more clever here)
skarmbild fran 2017-01-10 04 31 32
skarmbild fran 2017-01-10 04 31 15

## Motivation and Context

I've always wanted geofencing so i added it myself :)
Could be really useful if you're scanning around a beach or whatever but don't want to scan the sea for fish ;)

## How Has This Been Tested?

On my own map

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project. - Probably not?
- [ ] My change requires a change to the documentation. - I guess!
- [ ] I have updated the documentation accordingly. - No...

PR #1900, opened 2017-03-02
Speedscan bug fixes

rocketrobot99:
Many bug fixes and improvements to Speedscan

## Description
This PR has many bug fixes and changes to the Speed scan scheduler. 
 **While this will work with an existing database, it will identify TTHs more accurately if the DB is wiped** 

## Motivation and Context
Reviewing the logs showed many warning messages, complaining about Pokemon not being where they should. I investigated these and fixed bugs to make the warnings go away. TTHs should also be correctly identified eventually. (As this can still take several hours).

2 routines that have been changed significantly, get_cell_to_linked_spawn_points and select_in_hex
The changes are because spawnpoints can occur within the overlaps of scan locations:
image

However, when we queue a spawn point, we only need it queued once, so the changes attribute each spawnpoint to one scan location only.

Another significant change was to use the 'current_timestamp_ms' from the GET_MAP_OBJECTS response. A lot of problems were caused by using the current time, rather than the time the scan took place.

Performance gains were also included and uncaught exception handled.

## How Has This Been Tested?
Testing will be documented per change relating to an issue. Some tests will run on a clean database, others may have data already fully populated. Some may require data manipulation to force senarios. All SQL used will be put here:
[put into mem.txt](https://github.com/RocketMap/RocketMap/files/836393/put.into.mem.txt) - This SQL loads the tables into memory and creates history tables for scannedlocation and spawnpoints. Triggers are also created to populate the history tables in order to help track what happened and when.
[check_kind.txt](https://github.com/RocketMap/RocketMap/files/852341/check_kind.txt) - This SQL checks the 'kind' of the existing spawnpoints that have a TTH and the associated cell is done.
[running.txt](https://github.com/RocketMap/RocketMap/files/836493/running.txt) - This SQL displays information about how the map is progressing

Clearing the data base is performed by runing the map with -cd, then stopping it:
image

**CPU Usage**

Here are some CPU graphs created using netdata. They were taken with the server running 2 instances (-st 35 and -st 40) A total of 18,000 spawnpoints using a total of 480 workers. 
This was with the new code:
2 instances 480 workers

This was where I stopped the new code and started using the current code again:
2 instances current

And finally with the current code after an all the workers have settled in:
image

As far as I can see the overall CPU usage may actually be slightly lower with the new code, but there is not much in it.

** Update after commit to fix CPU usage, now looks like **
image

I would say 40% lower on my setup than current dev

# Issue 1
**Speedscan delay in scanning Spawns is too high:**

Current behavior:
Running with an already populated, TTH'ed database using the current code, here is the output from an -ST 30 Map:
image

On average the scan delay is about 100 seconds.

Changes:
Using the actual timestamp from the scan (see issue 5), this is returned from parse_map
image
It is then sent into task_done and used to calculate the actual difference between the start of the scan item and the actual scan time.
image
This ensures that the correct values are recorded and reported. However, another change was to have the workers target the items 'start' value accurately. Here we have the Next_item, looking for when the worker will arrive at the cell.
image

New behavior:
Running on same db with an already populated, 67% TTH'ed database using the current code, here is the output from an -ST 30 Map:

On average the scan delay is now about 50 seconds (after the queue has caught up).
image

# Issue 2
**Speedscan distance calculations take too long:**

Current behavior:
This is the log from a typical -ST 50 map at startup when it does the distance calcs:
image
The time taken is 1min 19secs.

Changes made:
The routine link_spawn_points has some additional lines to limit the calls to the CPU intensive call to 'in_radius'. It limits the possible Spawns that could in radius by seeing if they are in an area equivalent to a square box around the Cell location.
image

New behavior:
This is the log from the same settup, using the new code:
image
The time taken is 12 seconds

Logs:
[issue2_logs.zip](https://github.com/RocketMap/RocketMap/files/836411/issue2_logs.zip)

# Issue 3
**Large -ST hexes have distorted Cells at the map edges:**

Current behavior:
This is pic from a typical -ST 100 map showing the 'Squashing' effect on the left and the 'spacing out' effect on the right:
left and right

Changes made:
The routine _generate_locations was re-written and simplified. The cells no longer need to be created in a particular order. The routine works by building each 'ring' of the hex, one at a time, from the center outwards. For each ring, first 6 points vertices are calculated then the cells along each edge of the ring are filled in starting the East and working clockwise.
image

New behavior:
The same locations on the left and right of the map are no longer distorted
left_after and after_right

# Issue 4
**RocketMap can stop with an uncaught exception:**

Current behavior:
image

Changes:
In the seach.py, add an exception handler to catch any problems while copying the queue items.
image

New behavior:
image
The exception is caught and the Map continues.

# Issue 5
**Scannedloaction bands don't match the Spawnpoint detection data and don't correspond to when the scan actually took place:**

Current behavior:
Looking at a typical scannedlocation record:
image

then comparing it with the spawnpointdetectiondata:
image

In this example the _919 detection data doesn't match with the band4 time of  918 on the cell. This is not a major difference, but unifying the time will help solve a lot of problems.

Change:
In the parse_map routine, the timestamp of the request output is used where available.
image
This is also passed to other routines and used when storing scannedlocation and SPs
image
and
image

New behavior:
image
image

# Issue 6
**Unable to locate a TTH with a scan, just after the latest seen scan:**
If a TTH scan does not produce a TTH, then a scan of the same SP should still produce a pokemon if scanned within 89 secs of the previous scan.

Current behavior:
image
Something is not right here, taking a look at the scannedlocation history shows nothing out of the ordinary:
image

Changes:
See changes in issues 9,10 and 5. It is this combination of changes that fixes this issue.

New Behavior:
image
This particular warning should now be eliminated.

# Issue 7
**Spawnpoints have an incorrect 'kind':**

Current behavior:
From the 'running' SQL we can see we have 3695 SPs so far.
image

Running the check_kind SQL compares the stored kind of a spawnpoint with a calculated one. This is only for SPs with a TTH and where the cell scanning is complete. The output show that 336 SPs have an incorrect kind:
image

Changes:
This includes all changes from issue 5. Plus changes to classify:
image

New behavior:
After running an -ST 60, the check_kind SQL shows no incorrect records:
image

# Issue 8
**Spawnpoints are not counted correctly':**

Current behavior:
For a single ST 30, running a count of spawnpoints reads:
image
From the logs it is:
image

The map is reporting lower than expected SPs

Also for a multiple - ST 30, running a count of spawnpoints reads:
image
However from the logs:
image
4611+919+190+469+2026+4059+1771 = 14045
We actually have more SPs being reported than there actually are because of the SPs in the overlaps.

Changes:
I have modified cell_to_linked_spawn_points so that an SP is assigned to only one Cell, and so is only counted, searched for and reported once.
image
I similar logic was applied to create a new Spawnpoint.Select_in_hex_by_cellids
image

The existing Spawnpoint.Select_in_hex was renamed to Spawnpoint.Select_in_hex_by_location and is still needed upon startup in case the location has changed

New behavior:
We have 11568 SPs
image
From the logs..
image
3699 + 803 +140 +293 + 1781 + 3255 + 1597 = 11568

# Issue 9
**The queue can refresh between an item from the queue being scanned and when it is sent to Task_done:**

Current behavior:
Currently, there is no code to detect if the queue has been refreshed after an item has been scanned and before it is handled in task_done. All that is stored is the index to the item in the queue and can lead to problems if the queue has changed

Changes:
Added a means to detect if we are still dealing with the same item or not.
image. Thanks to @onilton for this suggestion.

New behavior:
Now this problem is at least detected and dealt with:
image

# Issue 10
**SPs are not being re-classified when a cell is done, because the last scan was a 'bad' one:**

Current behavior:
SPs are only finally classified when a cell has completed its scanning, which makes sense. However if the scan that completed the cell is a bad one, i.e. no pokemon, then it is not sent to classify and stays 'hhhs'. See issue 7

Changes:
image
This code was removed, as we don't want the processing to return until it is dealt with further down. 
image

New Behavior:
See results from issue 7, no incorrect 'kind'

# Issue 11
**Queue is continuously refreshed if there are no associated SPs within that area:**

Current behavior:
Running a map with -St 1 at a location that I know has no pokemon:
image

Changes:
Added a new variable for each hex called 'empty_hive'. When a queue is refreshed the number of SPs within the hex is counted and 'empty_hive' is set when the count is 0.
image
The variable is now added to the refresh check
image

New behavior:
image
Fro the logs that queue is never refreshed

# Issue 12
**SPs taking forever to find a TTH:**

Current behavior:
With the default value for --spawn-delay being 10 seconds, when trying to find a TTH, the start value only moves along 20 seconds.
image

Changes:
So if no TTH is found the gap moves along at least 45 sec, giving a gap of 44 seconds for the next TTH search
image

New behavior:
image
The _<secs> part of the ID shows that TTH scan is increasing by at least 50 seconds until it is found.

## Types of changes

## Checklist:


jaake77:
image

jaake77:
image

myntath:
Speedscan to me seems key to a useful RocketMap so I'm glad to see work to improve it. I plan to have a proper look through and test when I have time. Just wondering if you know if this is likely to fix issues 1897, 1898 and 1900, these just recently popped up? (1898 is mine, and at first glance this may not be fixed). If not perhaps we could have a look at some of these at the same time while there is a few of us considering speedscan.

pogom/models.py
```diff
@@ -1420,6 +1419,11 @@ def classify(cls, sp, scan_loc, now_secs, sighting=None):
         # if it changes.
         old_kind = str(sp['kind'])
 
+        tth_found = False
+        for s in query:
+            if s['tth_secs'] >= 0:
```
j16sdiz: `if s['tth_secs'] is not None:`  ?

pogom/models.py
```diff
@@ -1787,7 +1795,8 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 d_t_secs = date_secs(datetime.utcfromtimestamp(
                     (p['last_modified_timestamp_ms'] +
                      p['time_till_hidden_ms']) / 1000.0))
-                if sp['latest_seen'] != sp['earliest_unseen']:
+                if (sp['latest_seen'] != sp['earliest_unseen'] or
+                        not sp['last_scanned']):
                     log.info('TTH found for spawnpoint %s.', sp['id'])
                     sighting['tth_secs'] = d_t_secs
```
j16sdiz: Do we need something like `elif (sp['latest_seen'] == sp['earliest_unseen'] and d_t_secs != sp['earliest_unseen']): ScannedLocation.reset_bands(scan_loc)`   to allow rescan of tth  ?

rocketrobot99: Probably, I have not looked into how we should deal with spawnpoints/locations when the spawns migrate. This PR was really only concerned with fixing bugs and narrowing down the TTH. Plus it would be pretty tricky to test. It might be better in a separate PR, designed to identify if a spawn migration has occurred. Maybe have certain thresholds, then reset all the bands / TTHs?

Obihoernchen: I agree. But let's keep it separate. Already more than enough changes in this PR.

rocketrobot99:
Ok, so when the location changes, a whole bunch of new scannedlocations are created, there is no easy way to associate these with the existing ones (not in this PR anyway). So all locations are scanned again, this is existing functionality. I have reverted the select_in_hex routine back to using location when it changes, however it will use the new select_in_hex_by_cellids for normal operation. I have also fixed the spawnpoints from all being reset to hhhs (because of the location change). It the SP has a TTH, it will keep its kind until its scannedlocation is done and not be reset to hhhs.

rocketrobot99:
When the location changes a whole load of scannedlocations are created. So when you try and determine which cell an SP should belong to, how do you determine which cells are being scanned and which are old ones? I have used last_modified > the date when the location changed to identify all current cells (this works across hives now) then every SP is assigned to a single arbitrary max(cellid) and is only treated and counted once.

TheNewDarkOne:
Was there any update on https://github.com/RocketMap/RocketMap/issues/1897 ? with some kind of give up threshold? maybe limit number of times a band can be re-queued before it finally says "ok set band to empty/complete" each time it's been put back in queue it's already been scanned 3x, so maybe cap number of times it's put back to ensure it's gotten enough scans at different points of hour but ultimately is flagged complete and not left in a permanent state of incomplete. I'm also confident trouble spots mine are stuck on are yielding a lot of captcha's because they are illogical geo locations (such as in ocean off coast of an island. anecdotal for sure but i swear when it goes out to that band the accounts throw up captcha's almost immediately)

j16sdiz:
@thenewdarkone this pull actually fix that issue. But there are some details still need polishing

pogom/models.py
```diff
@@ -920,12 +920,37 @@ def linked_spawn_points(cls, cell):
...
+                                       'cellid'))
+                       .join(sp_from_cells, on=sp_from_cells.c.spawnpoint_id
+                             == ScanSpawnPoint.spawnpoint)
+                       .join(cls, on=(cls.cellid ==
+                             ScanSpawnPoint.scannedlocation))
+                       .where((cls.last_modified >= location_change_date) | (
+                             cls.cellid << cellids))
+                       .group_by(ScanSpawnPoint.spawnpoint)
+                       .alias('maxscan'))
+        # As scan locations overlap,spawnpoints can belong to up to 3 locations
```
j16sdiz: Don't fully understand this logic.
Does it cause problem if I:

Run an instance till initial scan compete
Move the instance to another overlapping location, wait till initial scan compete.
Move back to orginial location.
I feel the sp will be bounded to the "new" (but no longer scanning) cells created in step 2. It will never be scanned again.

Please correct me if wrong

pogom/models.py
```diff
@@ -1296,25 +1303,35 @@ def get_times(cls, cell, scan, now_date, scan_delay,
...
 
-            cls.add_if_not_scanned(
-                'TTH', l, sp, scan, start, end, now_date, now_secs, sp_by_id)
+            # So if the gap between start and end < 89 seconds make the gap
+            # 89 seconds
+            if ((end > start and end - start < 89) or
+                    (start > end and (end + 3600) - start < 89)):
+                end = (start + 89) % 3600
+            # So we move the search gap on 45 to within 45 and 89 seconds from
+            # the last scan. TTH appears in the last 90 seconds of the Spawn.
```
rocketrobot99: So basically what can happen is that an SP is identified as a hhss and after scanning the TTH window moves ever closer together, say latest_seen = 1128 and earliest_unseen = 1162. Now if all the workers are busy it is likely that either the workers will miss this gap, or if they are not busy then they will hit the 1128 + <scan_delay> and the window will only move on 10 secs or so. This cahnge will make sure the window moves closer by at least 45 secs, and also make the window at least 44 seconds wide. If you DM me your details, I can test this out for you.

pogom/models.py
rocketrobot99: It all comes down to trying to match an SP with only one cell (one scanning location), as they could appear in many (overlaps). Also, when the location changes, this introduces even more cells and an SP could appear in many more. To identify an SP with one cell I have limited scanned_locations to just active cells (determined by last_modified). So if you are using multiple instances an SP will only be scanned in one of those active instances. If you are using -bh with multiple hives an SP will only be scanned in one of those active hives. 

Now when the location changes, a whole load of new cells are created (scannedlocations). It then uses the original location based logic to map any existing SPs to the new cells, so the SPs don't have to be re-calculated. Every time the location changes the current cells become the current cells used to do the scanning and all the other ones become 'old'. However in your scenario, if you went back to a location that is exactly the same as an old one, then no new cells are created (as they already exist) and all the SP are already linked (via scanspawnpoint links). Those 'old' cells become the active set. 

The logic is fairly complicated, however eliminating SPs being scanned multiple times by multiple cells is quite crucial to correct TTH identification. 

So just to clarify :
"I feel the sp will be bounded to the "new" (but no longer scanning) cells created in step 2. It will never be scanned again.

Please correct me if wrong"

At step 3, the original cells are used and the SPs will use the original binds to the original cells. Not the ones created in step 2.

j16sdiz: its the line 940 that worries me:

```
 +                       .where((cls.last_modified >= location_change_date) | (
 +                             cls.cellid << cellids))
```

why we need  `(cls.last_modified >= location_change_date)`  here?

Isn't `( cls.cellid << cellids)` sufficient? 
Do you means `(cls.last_modified >= location_change_date) & (cls.cellid << cellids)` ?

j16sdiz: > The logic is fairly complicated, however eliminating SPs being scanned multiple times by multiple cells is quite crucial to correct TTH identification.

Don't think this is crucial. Have been running multiple  instances (speed or hex or mixed) on the same database without problem.  

a side note:
 Niantic have changed the sememtic when they roll out gen 2. 
In older semetic, only 30 minutes spawns have  `time_till_hidden_ms`.
In new semetic, all spawn point have `time_till_hidden_ms` in the last 90s.

rocketrobot99: The | is needed for when the location first changes, or when the map first starts up. Sat we have 2 instances Green and Yellow, and there is an overlap of cells. The red cells refer to a previous location:
overlaps

Now at start the first instance is Green, the only cells known to green are the green ones (cls.cellid << cellids). As the instance has just started cls.last_modified >= location_change_date will yield nothing. So all the SPs in the overlap will be linked to Green cells. The green instance scans for 5 minutes or so.

So now the yellow instance starts up, the cells known to it will be all the yellow ones (from the << cellids part) + all the green ones (from the last_modified part). But not any of the red ones as the last_modifed date will be before the change in location. SPs in the overlaps between will either be scanned by green or yellow cells, but not both.

After another 5 minutes green refreshes its queue and the cells known to it are now, all the yellow and the green cells, but again not the red ones. Some SPs in the overlaps may move from being scanned by green cells to being scanned by yellow cells. However they will now be fixed to those cells, until the location changes, and only scanned by 1 associated cell. Excuse my 'diagram', but sometimes it's easier with a picture. I hope it helps.

Where I have used 'multiple instances' this of course applies to using multiple hives in one instance.

As to your other point, perhaps this is not crucial, however having multiple cells updating the same SP was causing problems, it could lead to the an SP switching 'kind', especially if one cell had completed its 5 bands and the just starting. It also lead to some SPs being counted multiple times in the stats.

j16sdiz: >  After another 5 minutes green refreshes its queue and the cells known to it are now, all the yellow and the green cells, but again not the red ones. Some SPs in the overlaps may move from being scanned by green cells to being scanned by yellow cells. However they will now be fixed to those cells, until the location changes, and only scanned by 1 associated cell. Excuse my 'diagram', but sometimes it's easier with a picture. I hope it helps.

What if I terminate one instance after this?
The ` (cls.last_modified >= location_change_date)` still include the cells from the terminated instance.  Some of the sp may be missed until I change the location / restart the map.

> As to your other point, perhaps this is not crucial, however having multiple cells updating the same SP was causing problems, it could lead to the an SP switching 'kind', especially if one cell had completed its 5 bands and the just starting. It also lead to some SPs being counted multiple times in the stats.

Yes, it reset the kind to `hhhs` if one instance is not `done`. But it should stablize when all locations are marked as `done`.

rocketrobot99: This is true, I did not cater for the fact some SPs will not be scanned if they exist in an overlap between 2 instances and you stop one instance. However the reason they won't be scanned is because those SPs in the overlap are associated with that instance and it has stopped. Anyway to cater for this I can add another caveat, that :
Cells will only be considered if they have not only been updated after the location changed, but they have been updated within the last hour. This way, when the instance terminates, the SPs in its overlaps with the other instances will be allocated to the running instances.

"Yes, it reset the kind to hhhs if one instance is not done. But it should stablize when all locations are marked as done." - But they would still be counted incorrectly. Plus in this modern world of hashing servers, we want to save as many scans as possible.

PokemonCatcher:
Love your work on this PR. 
While you're working on this section of code, there is another problem with a seemingly simple solution you might be able to implement.
When scanlocations are generated, the latitude and longitude are arbitrary-length floating point numbers which are then stored in the database.
The implication of this is that if you move a starting location of a hex exactly 70 metres to the east, due to inherent inaccuracies in arbitrary-length floating point numbers, the coordinates generated from the new starting location do not exactly match the previously generated coordinates. So in the new hex perhaps 5 scanlocations match up and the others are all treated as new locations and the initial scan for bands takes place. Even though almost all of the scanlocations (except for, in this example, the strip down the east edge) have been fully scanned.
A simple solution to this would be to round the coordinates to a fixed number of decimal places in python when scanlocation coordinates are generated. These rounded values would be stored in the database and could be compared accurately with new scanlocation coordinates even if starting location moves. 9 decimal places is better than 1 mm of accuracy and in my tests up to ST=12 allowed almost all coordinates to be matched.
The change would be to scheduler.py line 585:
`results.append(float("{0:.9f}".format(loc[0])), float("{0:.9f}".format(loc[1])), 0))`
Are there any potential problems that could be caused by this change?

j16sdiz:
@pokemoncatcher the database type is single precision float. It only provide 9 significant figures.

Obihoernchen:
@j16sdiz no it's not. All lat/long columns are double precision in DB.
@PokemonCatcher I wouldn't do this. It could be a source of potential problems. And actually I don't see a real problem with the current solution. Scheduler will keep Spawnpoints + TTHs for spawns in the overlapping area (old and new), so all the scheduler does is a new initial scan which is kinda fast anyways.

Btw didn't test a location change yet, but everything else seems to work perfectly with the latest commit 👍

TheCrabity:
Looks like a good set of fixes. However, the delay between Search --> Searching is still there. More accounts you add to  a single process, the longer this delay gets. Not sure if you can have a chance to look or not, but would be much appreciated. Thanks and good work!

rocketrobot99:
@PokemonCatcher I think there needs to be a separate PR concerning location changes. At the moment as @Obihoernchen points out, the map will cope with it, but it is not ideal. I think a solution might be to introduce some sort of 'snap to grid', which the starting location will have to conform to. If you have any existing scanned locations it automatically adjusts the starting location to the nearest cell. BTW I did try to round the coords and it still generates new locations when the location changes (even just 1 cell to the left). I went down as low as 5 decimal places and still no joy, so this needs a bit more thought.

rocketrobot99:
@PKMTracker I think @onilton is looking into this in a separate PR, certainly in a similar area.

pogom/models.py
```diff
@@ -896,6 +896,9 @@ def link_spawn_points(cls, scans, initial, spawn_points, distance,
                 continue
 
             for sp in spawn_points:
+                if (abs(sp['latitude'] - scan['loc'][0]) > 0.0008 or
```
azotov: This assumption is wrong. The "distance" in longitude between centers of neighbor cells is variable over the globe. It is equal to 0.00108913 degrees at equator and 360 degrees at poles. Nobody will be scanning at poles :), but for example at Oslo (latitude=60) distance between cells centers is equal to 0.00217826 and your code starts missing pokemons. The "distance" in latitude is constant over the whole globe and approx equal to 0.00094481 degrees. So "abs(sp['latitude'] - scan['loc'][0]) > 0.0008" check is correct. But only for pokemon scan with 70m distance between cells centers. For gym scans it is equal to 200m, so this code may have problems too.

rocketrobot99: Ok, the circumference of the earth is 40,075,000m. I am looking at 70m cell radius only and we are dealing only with SPs here, not forts. So at the equator (lat 0), 70m would be the equivalent of 70m/40075000m x 360degrees = 0.000628 degrees. If we move up to lat 51, where I am, then the ring around the earth is 40,075,000 x cos (51) = 27,888,705m which would be 0.000904 degrees. Oslo would, in fact, be 0.00107. However, I get your point that the higher the latitude the closer the longitudinal lines are. I'll alter the code to use 0.0007 / cosine (latitude) to be safe. 
Anyway before the change:
before_trig
After change:
after_trig
Both pick up the correct number of SPs and no difference the calculation time.

rocketrobot99: Also, regarding _generate_locations. The bug in old version is why I made this simplified version. I have shown the reason behind it above in the comments. Along with the before and after pics:
image
The original was way more complicated because the steps had to be in the order the worker would walk around the hex. With speedscan this is not needed.

azotov: Thank you, using 0.0007 / cosine (latitude) should solve the problem. Your simplified _generate_locations is definitely better than the old code. While I think that my _generate_locations is better in some ways I support your version of _generate_locations - it should do a good work and can be optimized (with my code or some others) later after pushing this PR into base branch.

rocketrobot99:
@azotov Thanks for your comments. I think as you say TTH finding is ok now. Also manually rescanning (done=0,band*=-1) is a great idea, it will re-classify all SPs when the band completes. I have some SQL attached that is not the best looking, but it will identify any SPs with an incorrect 'kind'. I ran an -ST 60 from start to finish and ALL SPs were correct.
[check.txt](https://github.com/RocketMap/RocketMap/files/836143/check.txt)

(MySQL only)

sirlindqvist:
Tried latest commit with no issues at all. I feel spawn times got better in time. On 30 mins spawns i do get 29m45s most times and never under 29m. Workers seems to move very effective and all TTH's found on an st50.

pogom/schedulers.py
```diff
@@ -976,13 +966,25 @@ def next_item(self, status):
...
+                return -1, 0, 0, 0, 0, messages
 
             if best['score'] == 0:
                 if cant_reach:
                     messages['wait'] = ('Not able to reach any scan'
                                         + ' under the speed limit.')
-                return -1, 0, 0, 0, messages
+                return -1, 0, 0, 0, 0, messages
+
+            if best['step'] != item['step']:
```
onilton: This is really happening? If it is, I think the solution may be a lock instead of this check?

sirlindqvist:
Pulled latest commit, still no issues at all with this. Should be ready for merge into dev?

tobias86aa:
Do this improve the speed on big steps use? like -st 60 and above?

SiteEffect:
I am running this PR on my main setup right now without any flaws and improved speed as intended. Sadly, I do not have the time to fully go through each issue with the logs, DB and testing methods at the moment to fully approve it.

Imoq:
Testing this on -st 25 instance. It just started (without a new DB or without -cd) on the same setup than current development and for some reason it restarts the initial scan. Will report how it goes tomorrow.

Obihoernchen:
@lmoq that's expected because previous location calculation was slightly wrong

Imoq:
I have some feedback on this PR, I have logs of everything in case they are required to pull more information from. Apologies if I provide too much (or too little) information.

The server I tested all of this is a 6-core, 12-thread Linux (Ubuntu 16.04) with 64 GB of RAM which doesn't execute anything else than the maps. The DB is on another dedicated server with SSD and high RAM and innodb optimizations, on the same local network.

Last night, at (my) 11 PM I updated to latest devel + checked out this PR, so nothing else or different is applied. I first started with 4 -st 25 zones on the existing database (didn't -cd or use a new one). As I said, the initial scan was triggered but it was already explained it was expected. Whenever this gets merged to main develop branch, I believe people should be very aware of it and plan accordingly.

On those 4 -st 25 -w 90 instances the CPU usage went to the roof (sever load from usually 4.0-5.0 to 25-30) for about 1-1.5 hours while the initial scan was (almost) completed, but nothing to really worry about. Since most TTHs were already found (even if on the same previously scanned areas, an important number, around 300-500 of new TTHs were marked as needed to re-get, which I'm only blindly assuming they were incorrectly identified before) after about 3 hours the scans continued well.

The queue for "total waiting" went down from a constant 90-ish on the current develop branch to around 30-ish on this PR, and the identified pops on alerts are now closer to 29-30 minutes than to 25-28 which is great. I left these 4 instances running up to this point, even if in general they take a bit more of CPU than the current develop does. It's a good trade off in my book.

Because of this success, I went ahead and updated my friend's big map, which runs an -st 35 -w 230 and an -st 40 -w 250 with exactly the same setup as described above, and started initial scan at around 1 AM.

At 6 AM that I woke up to see the status, the initial scan was like this:

> Scanning status: 4140 total waiting, 3 initial bands, 135 TTH searches, and 4002 new spawns
> Initial scan: 86.50%, TTH found: 90.13% [1194 missing], Spawns reached: 11.38%, Spawns found: 100.00%, Good scans 100.00%
> Total active: 250 | Success: 27447 (5320.6/hr) | Fails: 335 (64.9/hr) | Empties: 1118 (216.7/hr) | Skips 0 (0.0/hr) | Captchas: 0 (0.0/hr)|$0.00000/hr|$0.000/mo

Didn't look that well after 5 hours, but I left it and came back almost at 9 AM, on which initial scan was like this:

> Scanning status: 4509 total waiting, 3 initial bands, 6 TTH searches, and 4500 new spawns
> Initial scan: 99.85%, TTH found: 99.04% [101 missing], Spawns reached: 42.12%, Spawns found: 100.00%, Good scans 95.05%
> Total active: 250 | Success: 7017 (2833.9/hr) | Fails: 162 (65.4/hr) | Empties: 52 (21.0/hr) | Skips 0 (0.0/hr) | Captchas: 0 (0.0/hr)|$0.00000/hr|$0.000/mo

Which looked better for sure, but I noticed a very strange behavior on the queue and this is what I want to focus on. It looked like the workers were "stuck" for several minutes at the same point ("scanning step XXX for a spawn") and even if the /status screen refreshed exactly each 3 seconds (and the time on the workers did change) they kept saying they were scanning that step until about 3-5 minutes after when they were changing to another step, without providing the info "Search at x,y completed with 2 finds. Sleeping xs until HH:mm.". So, the status was like this.

08:53:20
image

08:54:25
image

08:55:29
image

Finally it changed at 08:57:18!
image

So, if you see, the workers kept "stuck" scanning one point without advancing.

At around 10 AM I gave up because so many Pokémon were being missed (reaching not even 30% of them with initial scan practically completed) and I gave up and got back to develop branch on the -st 35 and -st 40 ones. Again, on the other map's -st 25, they are still running this PR.

Please let me know if you need any logs provided or any more information.

dsorc:
From what I've seen average delay is about 40s on ST60 (5500 spawn points), which is great compared to the main branch (150s).
But CPU load is quite worrisome.
It keeps my i5 3570k at the overclocked frequency of 4.4Ghz and 25% usage on all cores. If I set CPU affinity to 1 core it reaches 100% for it.

Edit:
After dropping from 280 to 200 accounts the CPU load decreased from 25% to 15-20%. Average delay seems to remain 40s. I might try dropping even more accounts (128s per scan per worker).

Edit2:
Nevermind, it's back to 25%.

Stereolab42:
Increased CPU usage is not an option, it's the single most expensive thing about running a map (now that captchas can be avoided). Please do not merge this unless CPU usage is flat or goes down.

pogom/models.py
```diff
@@ -1280,7 +1293,7 @@ def start_end(cls, sp, spawn_delay=0, links=False):
         start = sp['earliest_unseen'] - (4 - plus_or_minus) * 900 + spawn_delay
         no_tth_adjust = 60 if not links_arg and not cls.tth_found(sp) else 0
         end = sp['latest_seen'] - (3 - links.index('-')) * 900 + no_tth_adjust
-        return [start % 3600, end % 3600]
+        return [start % 3600, (end - 1) % 3600]
```
PKMTracker: return [start % 3600, (end - 1) % 3600] or return [start % 3600, (end + 1) % 3600] ? As you mention both

rocketrobot99: @j16sdiz Can I ask your reasoning for this suggested change? If I am not mistaken, the 2 lines produce exactly the same result.

j16sdiz: @rocketrobot99 forget it. I don't realize they are the same in python. (They are different in C/C++ when `end - 1` is `0`).

rocketrobot99: @j16sdiz No worries, you are not alone on this matter...

azotov:
@rocketrobot99, I have found the bug in classify method. After rescanning all scan points I checked ~35K spawnpoints with your SQL and it did not find any problems with spawnpoints kinds. But when I run my own validator script in php it did find 11 points with incorrectly detected kind. I have checked one of them. According to spawnpointdetectiondata it had spawns at 1486, 1571, 2139, 2262 and 2383 seconds. 2383-1486=897 and it was classified as hhhs by RocketMap and your SQL script. But when a pokemon was spawned at 2383 it got a TTH indicating that it will be visible till 2467 (tth_secs=2467). So its real spawn time is not from 1486 till 2383 but from 1486 till 2467. 2467-1486=981. The correct kind for 981 interval is not hhhs but hhss. I suggest to add all not null tth_secs values to seen_secs list in classify method for correcting this bug.

pogom/models.py
rocketrobot99: @dsorc , @Stereolab42 I have added some CPU graphs to the top. If you could provide more details of how exactly the CPU usage has increased. Before and after, please. Also some more details on your setup etc..

rocketrobot99: @azotov Ah, finally the last piece to the puzzle. I have updated the code and tested. Thanks. Although I couldn't find any of these rare examples in my setup.

rocketrobot99: I have also updated the check_kind script

dsorc:
I have some spawn points on the map with negative seconds
"May appear as early as 38:-8"
There was a Larvitar in such a spawn point some hours ago and it didn't get scanned at all. (a friend running the main branch scanned it)

I'm not sure if this problem was also present in the main branch. I also did the initial scan with a 150RPM key (while the needed RPM was about 350), not sure if that would cause issues.

There was also a problem with a Snorlax
His map (main line) spawn point and notification:
Every hour from 28:48 to 43:48
May appear as early as 58:12
Snorlax, 40.0% IV (9/6/3), Zen Headbutt/Hyper Beam! Expires 07:43:48

My map (PR 1900) spawn point and notifications!
Every hour from 59:21 to 14:21
May appear as early as 29:21
Snorlax, 40.0% IV (9/6/3), Zen Headbutt/Hyper Beam! Expires 08:14:20
Snorlax, 40.0% IV (9/6/3), Zen Headbutt/Hyper Beam! Expires 07:14:20

rocketrobot99:
@dsorc This is generated from the following function:
image
I don't really know .js, but it looks like it doesn't cater for all the different kind of spawns. There could also be some logic that might produce '38:-8'. I think it is outside of this PR though. As for the larvitar, can you find out the Spawnpoint ID, and see if you have that in your DB. Contact me on Discord if you want some help with that.

dsorc:
I can confirm that the negative seconds spawn point timers have nothing to do with this PR. (I checked the spawn points on someone else's map running on develop and they had the same timers)
Created issue https://github.com/RocketMap/RocketMap/issues/1941 for this

The larvitar spawn point is also not an issue of this PR. The area just cannot be scanned (pogom\models.py", line 1776, in parse_map     cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells'])
I have 18 such scan locations with the PR and 16 with the main branch dev.

The snorlax spawn point might be incorrectly identified as per azotov's comment.
I don't know how to check if the pushed fix works.

And finally, the CPU load is 5% at 2Ghz with the Dev and 20% at 4Ghz with the PR. Same settings. i5 3570 @ 4.4Ghz. ST 60, 5500 spawn points, 200 accounts. Don't know what other info is needed.

j16sdiz:
@dsorc are you on mysql or sqlite?

AEtHeLsYn:
I've been running this PR for about a week, no issues so far. CPU usage with st 30, 2600 spawn points and 200 accounts is about 5%. Using local mysql DB

dsorc:
@j16sdiz 
I'm on mysql (mariadb), since sqlite can't handle a large ST (or many spawnpoints, I'm not sure).

db-type: mysql
db-host: 127.0.0.1
db-threads: 21

tobias86aa:
Nice. Do you notes any improvements @FrostTheFox ?

ObsessedGamer:
Been using this for about a week and updated to dev import when available, running a Step 70 with 1600 workers without a problem, noticeable speed increases & CPU Usage stays around 30% & Ram Usage around 1400 which is about the same as the main branch.  Thanks for the hard work!

FrostTheFox:
Running for over 3 hours constant now, even with accounts occasionally going into hold state, they come right back out and get to work. No captcha when combined with lv2 accs.

pogom/schedulers.py
```diff
@@ -992,7 +992,12 @@ def next_item(self, status):
                 messages['wait'] = 'Moving {}m to step {} for a {}.'.format(
                     int(distance * 1000), step,
                     best['kind'])
-                return -1, 0, 0, 0, messages
+                # So we wait while the worker arrives at the destination
+                # But we don't want to sleep too long or the item might get
+                # taken by another worker
+                if secs_to_arrival > 179 - self.args.scan_delay:
+                    secs_to_arrival = 179 - self.args.scan_delay
```
j16sdiz: what if scan_delay > 180 ?
time.sleep(-1) would give IOError.

dsorc:
The CPU fix has dropped the CPU load from 20% at 4Ghz to like 5% at 2Ghz (similar to develop!).
Average scan delay is still great (45s).

pogom/models.py
```diff
@@ -1438,7 +1512,7 @@ def classify(cls, sp, scan_loc, now_secs, sighting=None):
         # An hour minus the largest gap in minutes gives us the duration the
         # spawn was there.  Round up to the nearest 15 minute interval for our
         # current best guess duration.
-        duration = (int((59 - max_gap / 60.0) / 15) + 1) * 15
+        duration = (int((60 - max_gap / 60.0) / 15) + 1) * 15
```
j16sdiz: (this is a minor error, i prefer to leave this to other PR. Just a note here to explain the situration)


I have got some `hhss` misclassify as `hsss`. In this case, the error occur are they are on the  "60"-th  second:

```
+-----------------------------------+------------------------------+---------------+---------------------+----------+
| id                                | encounter_id                 | spawnpoint_id | scan_time           | tth_secs |
+-----------------------------------+------------------------------+---------------+---------------------+----------+ 
| MzM0Nzk0MzEzMzMxOTU2MDM2MA==_1473 | MzM0Nzk0MzEzMzMxOTU2MDM2MA== | 3404063930b   | 2017-03-04 19:24:33 |     1560 |
[...]
| MTE5MzA0ODY5MTI3MzkzMDIwNDc=_1560 | MTE5MzA0ODY5MTI3MzkzMDIwNDc= | 3404063930b   | 2017-03-15 05:26:01 |     NULL |
<----
| MTU1NjU4NzgzMjQzNzg4Mjg2MzU=_3361 | MTU1NjU4NzgzMjQzNzg4Mjg2MzU= | 3404063930b   | 2017-03-06 03:56:01 |     NULL |
```

On the second row, you saw  `05:26:01` have the id suffix`1560`. This is because the `now_date` have millisecond part (and *rounded* to nearest second when stored), but the `now_secs = date_secs(now_date)` *truncates* the milliseconds..

The easiest fix is to have some safty margin and leave this as `59`.
The *hard* fix to migrate to millisecond, but the benifit is too small to justify the cost....

j16sdiz: On a second thought, the cause seems seems to be not that simple. But the inconsistency in millisecond handling is real.
Charging it back to 59 give a large margin , but at least it is correct

j16sdiz: Again, leaving this to another pr maybe better. Don't want to delay this further

pogom/schedulers.py
rocketrobot99: Ok, I guess if you ar maybe searching for gyms. Anyway, I have pushed on update just in case.

rocketrobot99: @j16sdiz The problem is, with the last CPU fix, some maps, with lower specs would have struggled and ultimately fail. It could not have gone in without it.

rocketrobot99:
@dsorc That is good news.

rocketrobot99:
I still think that the scan_delay could be tweaked, but definitely outside of this PR. 30-40 seconds is pretty good, though:
image

tobias86aa:
Should this dont be ready to be merge now then? @sebastienvercammen

TheCrabity:
error
After the last few changes to this PR.
On the bright side, CPU load went down around 80%.

rocketrobot99:
@PKMTracker It is because you are running a large -st. If you edit your my.ini and add the following variable:
range_optimizer_max_mem_size= 16000000.
BTW, it didn't seem to make any difference on the performance.

TheCrabity:
In the config.ini ? I didn't even know that was an option to set.

rocketrobot99:
@PKMTracker No, sorry, in the config file for mysql. On windows it can found in:
image
On unix it is:
image

Imoq:
After my feedback from some days ago (https://github.com/RocketMap/RocketMap/pull/1900/#issuecomment-287108914), I've been working with @rocketrobot99 who very diligently and patiently helped me to debug the issues I was having on larger instances (-st 37, -st 40), and after his last commits (specially https://github.com/RocketMap/RocketMap/pull/1900/commits/39536a8d21975e1feb0057444763d781e8d34ac3) everything is working just as expected from my side.

All the Pokémon are being encountered and reported usually after few seconds after pop. Sometimes there is something up with my DB that the occasional pop takes up to 2 minutes to appear, but it's nowhere near as slow as current develop branch is. My "total waiting" items in the queue has drastically reduced to around 30, when I was all the time above 200 before. The CPU usage has been drastically reduced too (not that it was an issue for me before, but the load average for example has going down from around a constant 8-9 on current develop to 2-3 only on this PR).

Only as a tester and not being able to judge the code, but I can see really a great improvement over the current scans in develop branch.

sebastienvercammen:
@tokar86a Don't ping us when you believe something is ready to get merged. In 99% of cases, you're most likely wrong.

@Imoq Thanks for the extra detailed info, I was waiting for that :)

As always, we're on it, we're reviewing it, give it time. As it is, it's not finished yet, and it's not the only (or single highest) priority (although I do like the PR).

Demokdawa:
This is just a suggestion, but I think it would be a good idea to try to add this feature (https://feedback.userreport.com/9eec75e1-1e84-4f02-b254-5e239d896805/#idea/157043) to the next speed PR ^^ It would be very efficient. 

Thanks for you work anyway !

sebastienvercammen:
@FrostTheFox Can you remove @Demokdawa from the repository? Posting feature requests, for which we already intentionally set up a platform, on a completely separate PR, is not what Github PRs are for and they're not what I want to wake up to.

sirlindqvist:
I run 2 maps and on commit 83c5b80 cpu load on map1 was 35-40% and 25-30% on map2. Updated to 71e1237 and now both are around 10-15% - Good! Also got the very last % of init/tth because i was missing a few 0.0x% on both init/tth on both maps. Now everything is 100%. Good work! Timers are also good. If looking at 30mins spawns (as they are most common) i never (almost!) get any under 29m. Most of the times the spawn is 29m30-45s. Like @Imoq I'm only a tester and can't judge the code but I can also see really a great improvement.

sebastienvercammen:
@sirlindqvist Thanks, this is good feedback. Keep it up. :)

sirlindqvist:
Any new commits needed or is this considered as final and working? :)

myntath:
I had originally planned to review this code in detail, but it quickly grew beyond may capabilities. Instead I will add my voice and a mostly thumbs up as a tester. My CPU usage on the old code was already low so I can't say much in regards to that, but there has been no significant increase.

My map covers ~3100 spawn points. I backed up my old tables to enable comparison after clearing the database.

```SQL
mysql> select kind,links, count(*) from spawnpointbak group by kind, links;
+------+-------+----------+
| kind | links | count(*) |
+------+-------+----------+
| hhhs | ????  |      234 |
| hhhs | hh??  |      190 |
| hhhs | hhh?  |        8 |
| hhss | hh??  |     2375 |
| hsss | h???  |        1 |
| ssss | +++-  |      326 |
+------+-------+----------+
6 rows in set (0.01 sec)
```

From the new code:
```SQL
mysql> select kind,links, count(*) from spawnpoint group by kind, links;
+------+-------+----------+
| kind | links | count(*) |
+------+-------+----------+
| hhss | hh??  |     2802 |
| ssss | +++-  |      323 |
+------+-------+----------+
2 rows in set (0.00 sec)
```
On the assumption we only have 30 and 60 minute spawns the new code did a better job of classifying the spawns. The new code did not find 10 spawnpoints compared to the old but did find 1 new spawnpoint. All lost spawnpoints were on the extreme East and the new spawn on the extreme West. This seems acceptable and expected. 

All my old 'hhhs' spawns became 'hhss' spawns and the one 'hsss' spawn became an 'ssss' spawn. This appears to be an excellent result.

I used 50 workers, which is probably well under the suggested workers. The initial scan and about 95% of TTH were found within 6 hours. Approximately 24 hours after there is 1 TTH to find. This seems possibly bugged as it looks like this point should have resolved already as there is a tth_secs entry in its spawnpointdetection data. Some dumps are here http://pastebin.com/39dTHtCF. I expect it will still resolve eventually.

P.S. I note that the scans still become 'fixated' on objectives sometimes which is not particularly efficient. For me this was especially around when 95% of TTH were found. Only scans for TTH's were being assigned, often a long time in the future when there was plenty of capacity to check ordinary spawns in the mean time. Previously I have played with altering the variables 1e12,1e6 and 1 to something like 4,2,1 this gives you a lot more spawns but I presume slows down overall discovery of TTH somewhat. This might be worth tweaking in the future, but not this PR.

pogom/schedulers.py
```diff
@@ -924,14 +909,15 @@ def next_item(self, status):
...
-
+                # If we are going to get there before it starts then ignore
                 loc = item['loc']
                 distance = equi_rect_distance(loc, worker_loc)
                 secs_to_arrival = distance / self.args.kph * 3600
+                secs_waited = (now_date - last_action).total_seconds()
+                secs_to_arrival = (secs_to_arrival - secs_waited) if (
+                    secs_waited < secs_to_arrival) else 0
+                if ms + secs_to_arrival < item['start']:
+                    continue
```
onilton: Why continue instead of break?

Since the queue is sorted by start, there is no point in keep searching, right?

rocketrobot99: @onilton Thanks for taking a look at this. The queue is sorted by start, however, not by the distance from the worker's location. There may an item closer, but later on in the queue. 

BTW, I have already got some alterations for another PR to follow this one. It would basically define a limit that the worker can travel. e.g. if you have a -st 60, the maximum distance a worker could travel is 70m*118*sqrt(3) = 14km. You then calculate how long it would take to get there, take away how long you have waited already and that is the max 'start' it would be worth considering. Because anything after that, you would get there too soon and be waiting. I would also consider having a max distance. You don't really want the worker to travel more than maybe 2km?? 

Anyway, along with a whole load of other mods, it will be for the follow up PR, as this has too much in already.

rocketrobot99:
@myntath Some great feedback there. There are some locations that just won't scan at the moment, so any  TTH not found could be due to that. I have some tweaks for a follow up PR.

myntath:
@rocketrobot99 thanks. I'm not sure if this relates to unscannable locations. I did encounter that initially because of an old pgoapi hanging around. But those locations didn't even pass the initial scan. Updating pgoapi to latest commit fixed that. 

I might be wrong but my interpretation of my linked pastebin for the location shows it successfully scanned many times and one even with a TTH found. I just checked again and it is still not resolved... I will investigate further. It is making horrible guesses at the spawn time, now it just alternates between ~400 and ~3000 (I can see it despawns at 825). I will investigate further soon...

edit: this looks related to azotov comment earlier, not sure the best way to link it so just quoting it below:

>azotov 6 days ago

>I have found a problem which is observed after spawnpointdetectiondata table cleanup. Not null tth_secs value is written to db only once when TTH is found. If you delete spawnpointdetectiondata for this point but left the point in spawnpoint table and mark scan location of this point for rescan, then all new records for this point in spawnpointdetectiondata table will have NULL in tth_secs field (even when d_t_secs < 90 seconds is observed) and the point will not be able to be reclassified forever. This problem should not be corrected in this PR (more than enough changes in this PR), just don't forget about it when making new PR for point migration support.

So maybe fix for future pr. However, my error didn't relate to a migration or table clean. Maybe the data got borked some other way.

kinar:
Hey guys, I'm just posting this here because it seems like the right place.  I experienced extremely high MySQL CPU usage when I enabled this PR.  Last night I brought everything down to upgrade RM and decided to test out this PR as well.

I'm running a lot all on one machine and I understand that isn't ideal but it's worked for me so far.

I'm also running the geofencing PR #1688.  Here is my setup in terms of the code I'm running

```
$ git show-ref
71e1237bea613352d7163d84f9f45a7b3053f9d4 refs/heads/Speedscan_bug_fixes
eac532da8c86cb48696d183dfe72eb904c378f2d refs/heads/develop
93c42d2704e20f7ec63be76970a9d36e7c9cfa27 refs/heads/geofencing
cae7c30aecd3a7259a2948b50f24671f922dc892 refs/heads/multiple_pr
```

the multiple_pr branch merges both the geofencing and Speedscan_bug_fixes on top of develop.  There were no conflicts I had to manually resolve.

Here is my setup in terms of operation:
I've got 9 geofenced instances which make up an area roughly equal to 5x st32 instances.  
This all runs on a single virtualbox virtual machine (Win7 64bit) along with my mysql 5.7.17 (not maria) database.  
The machine it's running on is a 3.0Ghz i7 with HT 
It has 2x 500GB SSD with one SSD dedicated to the Virtual machine and other to the host OS
It has 16GB RAM with 12GB of that dedicated to the VM (nothing significant is running on the host at this time).

Of note is that I did NOT drop and start over with a new database when I integrated this PR.  The spawnpoint table in my database has ~19K rows.  The pokemon table has ~400K (I truncated it a couple days ago).  
Pretty much immediately after restarting everything, MySQL started taking 90% CPU (so saturating more than one core) and all the RM instances (even the smallest ones) started having DB queues.  I rolled back the PR and everything went back to normal.

I don't know if this is a conflict with the geofencing PR or if its just something unique to my setup but I know that PR and this one are both very popular with potential to be merged fairly soon (this one probably before that one) so I wanted to get this out there in case others have similar issues.  Unfortunately, I don't have the expertise to effectively troubleshoot MySQL to determine the specific cause of the high CPU usage.  I'm happy to run any queries or post any db settings here if there is someone here with better mysql knowledge and you tell me what to execute.

voxx:
Haven't had a chance to dig too much into this PR, but I've been running it on a few nodes for the past 48 hours and I'm actually experiencing more invalid timers than without the PR. Most of them appear to be 1 hour timers that when initially scanned a mon is found but it disappears before a subsequent encounter can even be made to pull iv's and other details.

All three nodes im running on were fresh databases and allowed to run for a full 24 hours before I started alarms back up. Initial scan lenth appears to be shorter than it was before, but TTH's phase has increased probably 50%.

Probably roughly 1/3 of the alarms contain no encounter information on alarm, and when I manually rescan that area, the pokemon that should be present for another 30-60 minutes is no longer present, indicating the TTH/Spawn Type value was inaccurately identified.

I did notice a fairly decent CPU usage decrease across all the hives running this PR, but I'd rather have high CPU usage and accurately identified spawns, than vice versa. I can try to dig in and compare old spawn dumps to current spawn dumps, but in my limited testing this PR still needs some work before I'd opt for it over the existing codebase.

j16sdiz:
@voxx are you sure it is unrelated to recent hash server issue?

dsorc:
I also have a small percentage of incorrectly identified spawn point timers.
The initial scan was done using a 150RPM key but max requests were reaching 300.
Assuming that is the reason for the incorrect timers is there any way to identify and correct them, other than wipe the database?

TheCrabity:
With IV's enabled, CPU usage skyrockets on this PR, compared to same scan area on develop with IV's. Can anyone confirm?

rocketrobot99:
@voxx Hi. I think we all have been experiencing some difficulties with the hashing server recently. This would certainly explain why a Mon is not being encountered. Would it be possible to wipe an instance and try again. If you are still having problems contact me on discord and I'll see what I can do.

voxx:
@rocketrobot99 @j16sdiz The encounter isn't failing due to hash server, the encounter is failing because the mon is not there when it's supposed to be, its de-spawned already. I've confirmed this is the case with multiple improperly identified spawn locations by manual rescans of the coordinates with a secondary script. They just simply aren't there anymore.

Had quite a few reports today from my feed users that many mons people went to track down with 45+ min timers that there was absolutely nothing to be found. So again, wrong de-spawn timer. 

The hashing issues can impact scans/encounters failing of course, but previously to the best of my knowledge they did not result in spawn types/timers being misidentified. It's possible your new logic is not quite so gracefully handling these fail scenarios and is wrongly marking type/despawn because of it?

I've gone ahead and wiped all three cities again, and am running a fresh scan in these areas as a final test. Hash is behaving a bit better tonight, but even under poor conditions the logic needs to be able to have graceful degradation, so we can ensure failed locations are tried again, and not wrongly identified because of some fail state.

rocketrobot99:
@voxx I agree, it should be robust enough to cope with anything. However, I am just wondering if this might have something to do with the recent nest migration (number 17) a couple of days ago. I haven't seen any changes where I am though. Anyway, let me know how you get on, and contact me on Discord with the location etc... if it is still not working properly. Without digging deeper I can't really help and I haven't come across this problem on the maps I have tested.

bluemode:
48 hours, no problems. I did not use encounter tho.

ObsessedGamer:
I have been running this for a while now with encounters enabled & encountering everything & several instances (Step 70, 39, 28, 45, 50) & no issues with timers ext.  Just wanted to give some feedback, also thanks for the recent CPU Usage fixes as it reduced my overall CPU Usage by 50%!

voxx:
Dumped all cities last night, rebuilt them completely today. 100 initial scan complete 100% tth's found. Again Initial scan phase was a reduction from dev branch, which is nice. TTH phase took a few hours longer than usual, but thats fine. CPU usage seems overall less than dev branch.

Still seeing same issues on many spawn point locations though. Some spawns are flagged, and when secondary script checks them, they are not actually there. (Despawned already). These are primarly 1 hour spawns, as I've mentioned previously.

Heres an example.
33.793824731105836,-117.85948570734597

1) RM Scanned and found pokemon at spawn point then passed pokemon to PokeAlarm
2) Pokealarm filter triggered because no iv's could be pulled on encounter 15-20 seconds later.
3) Secondary script queried same location, the pokemon id PA alarmed on was not present in GMO response object.
4) Pokealarm sent alarm to discord with no IV's.

screen shot 2017-03-25 at 10 35 39 pm

Output from secondary script that attempted to validate if the pokemon from alarm was present or not at the location 20-30 seconds after RM GMO call found it.
```
[25/Mar/2017 21:55:27] "POST /vsnipe/ HTTP/1.1" 200 32
Using location [33.7938247311, -117.859485707, 141.33894490936905] for this request.
Map request returned [{'last_modified_timestamp_ms': 1490504164445L, 'longitude': -117.85948570734597, 'pokemon_data': {'pokemon_display': {'gender': 2}, 'pokemon_id': 19}, 'latitude': 33.793422453150846, 'spawn_point_id': u'80dcd761e45', 'encounter_id': 6030834014311899708L, 'time_till_hidden_ms': -150512734}, {'last_modified_timestamp_ms': 1490504164445L, 'longitude': -117.85948570734597, 'pokemon_data': {'pokemon_display': {'gender': 1}, 'pokemon_id': 183}, 'latitude': 33.7937442756476, 'spawn_point_id': u'80dcd761ef1', 'encounter_id': 12703016358927657904L, 'time_till_hidden_ms': -150512734}, {'last_modified_timestamp_ms': 1490504164445L, 'longitude': -117.85957727662986, 'pokemon_data': {'pokemon_display': {'gender': 1}, 'pokemon_id': 195}, 'latitude': 33.793721903344796, 'spawn_point_id': u'80dcd761efb', 'encounter_id': 14849081349477972620L, 'time_till_hidden_ms': -150512734}, {'last_modified_timestamp_ms': 1490504164445L, 'longitude': -117.85985198420505, 'pokemon_data': {'pokemon_display': {'gender': 1}, 'pokemon_id': 19}, 'latitude': 33.79405706213706, 'spawn_point_id': u'80dcd761f2f', 'encounter_id': 15479537108209349847L, 'time_till_hidden_ms': -150512734}].
```
There are many of these in all 3 cities. More so than same results in development branch.

rocketrobot99:
@voxx  Hi. thanks for this, I have kicked off a build at that location of -st 45. Are you on discord? It would be better if you could contact me (rocketrobot) there.

j16sdiz:
I got quite a few wrong end time with tth_secs=451 .. all introduced yesterday. 

Not a 1900 specific problem.

dsorc:
Note that the few spawn time issues I've encountered were also tied to 1hr spawns.
But I've set pokealarm to not send a notification if the IV is missing, so it's possible there are more than a few.

voxx:
Additional details from database.

scanned at 4:54:59
screen shot 2017-03-26 at 3 04 57 am

gone at 4:55:19 (less than 30 seconds later), but despawn timer says it should be there until  5:54:58 (for another hour)
screen shot 2017-03-26 at 3 04 38 am

Secondary script then attempted it's validation moments later, and mon was still not there.
[25/Mar/2017 21:55:27] "POST /vsnipe/ HTTP/1.1" 200 32

voxx:
I've also passed numerous other examples to @rocketrobot99 via discord chat, and will happily share with any others who wish to troubleshoot and identify the underlying issue.

rocketrobot99:
@voxx @dsorc I have pushed a fix. The problem set SP's disappear time incorrectly. What would happen is an SP would be picked up right at the very end of its spawn, by the scan of another SP. This is fine, and a simple lookup would normally show that the pokemon for the spawn already exists. However, the lookup was using the current time (now just after the SP has despawned) rather than the time of the scan. Thanks to Voxx for his help, I have been running all day with this fix and not found any more examples.

voxx:
I've been running the latest patch https://github.com/RocketMap/RocketMap/pull/1900/commits/4fc0fd7ea37efd6e996adbef3cc0dc9830a50a26 for about 12 hours on three fairly large instances now.

I spent some time working with @rocketrobot99 testing a handful of previously erroneous spawn points and it appears the bogus alarms for 1 hour spawns that were occurring are now alerting at proper times.

While I did still experience a small handful of bogus timers appearing, these particular issues seem to not be related to this same bug, and I think can be considered out of scope for this PR. Likely the infamous "rare" spawn timers that have been alluded to in the past that the existing code base, and this PR don't account for.

I'll report back in another 12 hours or so, but thus far it looks good on my end. If no additional weirdness creeps up this evening you've got thumbs up to merge from me.

^^

pogom/models.py
```diff
@@ -1705,6 +1779,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     # and a list of forts.
     cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells']
     for cell in cells:
+        # If we have map responses then use the time from the request
+        now_date = datetime.utcfromtimestamp(
```
friscoMad: This should be better outside the loop, just get the first item timestamp (or the last as it is working now) but there is no reason for doing this for all cells

friscoMad:
I have been using this PR for almost 2 weeks (updating as needed) and when I went back to the dev version today to make a new PR I did notice how fast the startup is in this PR and that I just need 1 db worker with PR and 3 without it.

I am not sure about CPU usage but mem usage is similar.

pogom/models.py
rocketrobot99: Got you. Will be in the next commit. Thanks.

pogom/schedulers.py
```diff
@@ -924,14 +909,15 @@ def next_item(self, status):
...
-                # since queue sorted by start time.
-                if ms < item['start']:
-                    break
-
+                # If we are going to get there before it starts then ignore
                 loc = item['loc']
                 distance = equi_rect_distance(loc, worker_loc)
                 secs_to_arrival = distance / self.args.kph * 3600
+                secs_waited = (now_date - last_action).total_seconds()
+                secs_to_arrival = (secs_to_arrival - secs_waited) if (
```
rocketrobot99: Thanks, again in next commit

rocketrobot99:
@onilton I understand your point, Although I have only heard of cases where this improves the scanning. I Just don't want to make any more changes in this PR. If your system is set up so that the queue is minimal, then, in theory, all SPs will be scanned as soon as possible. The situation you describe will not happen, if you have enough workers as another worker will pick up the missed SP and that other  worker may even be closer. The current dev logic just goes for the closest, but only considers moving once the spawn has already started. I have been experimenting with some amendments to next_item and will be included in a follow-up PR. They are a combination to minimise secs after spawn and distance to travel.

FrostTheFox:
Putting this on the "ready for merge" list, any other issues please @ me, otherwise will probably be merged tonight

ObsessedGamer:
For some reason when I switched to your most recent commit I get huge DB ques now upwards of 10,000 and it stops scanning until restart.  Switched back to the one before your last recent commit and it went back to working fine.

pogom/search.py
```diff
@@ -537,7 +537,15 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
...
-        update_total_stats(threadStatus, last_account_status)
+        # Added exception handler as dict items change
+        try:
+            update_total_stats(threadStatus, last_account_status)
+        except Exception as e:
+            log.error(
+                'Update total stats had an Exception: {}.'.format(
+                    repr(e)))
+            traceback.print_exc(file=sys.stdout)
+            time.sleep(10)
```
friscoMad: There is no need to sleep here.
Also the proper fix is to catch this in overseer as here we can't do anythin special and there is no exception handling there and it is needed as if that thread dies everything dies.

rocketrobot99: @friscoMad Thanks for taking a look at this, the exception handling was literally just copied from the lines above:
image
If you have a better way of coping with the exception, it would probably need to be applied to both parts. Please send me the changes and I will have a look.

sebastienvercammen:
I've been getting poked about merging this PR so often, by so many people (most of which have never coded in their life), that I'd like to say congratulations: you know how to annoy a person. And not just random people, but people that have contributed in some way to the project (#pr) and that I consider to be on the team.

I'm used to people poking us constantly for stuff they want us to do, but this time the "request a review" feature on Github has been used for the first time ever by @FrostTheFox to poke me towards this PR.

This makes it very clear to me that y'all don't want an actual review, because that would take more time - as I've explained several times (especially to the people in #pr on Discord). You just want it merged.

So here you are, I'm merging it. No review, reduced code quality.

jagauthier:
No one here has any idea of what the actual process is for code review outside of the button in github.  But this PR is extremely popular, and obviously many people wanted you to review it.  So why didn't you?  We don't know your timetable for review, you don't provide us with a roadmap or milestones.
I thought your goal was transparency?
"To stay transparent about everything we do (and why we do it),..." 
"We believe it’s better to be transparent... "
"The only way to let everyone’s voice/opinion count, is to be transparent about the situation and to let everyone participate fairly."
(all quotes from the most recent devkat blog)

Since you  haven't reviewed this, why do you assume it will be reduced code quality?  Are you immediately assuming that the code written here is inferior to your own?

Please, open up to us.. be transparent, about this review process, the timetable, who is doing it, the roadmap ahead... heck, even what you're working on right now... Because this is an open source project, and an open source community.. right?

sebastienvercammen:
@jagauthier Except you're oblivious to everything that's been happening in #pr.

I have been reviewing it. I have been replying to questions about where I was and how it was going, I addressed that in my earlier comment.

Everyone knew. I just needed more time, because proper review isn't something you *force* to finish. And I know it's bad code because it's the only thing I've been putting my time towards for the past few weeks. Even rocketrobot knew it was pretty bad and was open to being patient for an actual review - but no one else was.

I'm not even given the time by #pr to finish a review, even when I communicated properly with them, so I can tell you with absolute certainty that we don't have the time to put towards communicating _everything_ to _everyone_. It doesn't work and it increases the amount of time we need to act like customer support to hold everyone's hands and to make sure we don't "offend" anyone with our decisions, actions, or delays because we're trying to communicate first rather than coding first.

For everything that the community has against commercial organisations (e.g. the negative feedback towards pogodev/pgoapi after BossLand entered the scene to solve the hashing problem), a commercial organisation with strict rules and nearly no transparency is actually what everyone seems to need. And of course that's going to annoy me, it's hypocritical.

As for your "are you assuming it's inferior to your own", that wasn't necessary.

The 6 people who upvoted your comment are all people who don't understand the consequences of some decisions, and the time and effort everything takes. Besides bluemode _(you troll)_, I don't recognize anyone for any effort they've put towards RocketMap. And before I get another reaction of "then tell us, explain it all to us!", I'm sorry but I'm not here to hold your hands and teach you how this part of the world works.

It's open source. The thing you don't seem to get, is that the equivalent of the time and effort we've been putting in for the past 9+ months, is for you to start learning from scratch and to try to catch up to where we are in whatever way you can rather than complaining about us. We don't expect you to suddenly catch up, but the effort of trying will move you forward and into a position where you can contribute (small parts) to the code and the project rather than having to place some form of vague idealistic comment about communication.

As for this PR, everyone could comment and participate. But that doesn't mean everyone will suddenly have the required experience to properly evaluate software code, and whether it's one person without the required skills/experience approving the PR or twenty, it doesn't make any of them more valid. Feedback is _always_ welcome, but that doesn't mean it'll be enough to use as a single point of reference for approval or rejection of a PR.

Someone PM'ed me this, and I'm finding myself agreeing more lately: the community has been spoiled.

Perhaps it's time for us to take a step back as RocketMap and act more as an organization rather than your local pub where everyone can enter to improve the work and code quality. To set up formal processes of recruitment, development and communication. But I'm convinced that'll also bring complaints along with it.

Either way, we have things to discuss and we're discussing them. For now, these are just thoughts, and I've always been transparent in any way with the relevant people: #pr knows everything I've been up to, including where I went for the holidays and for how long. I've always included everyone that was necessary in any decision we've made so far, and in the cases where that wasn't always possible (e.g. a ban on Discord), I've always been open to receiving feedback about them and adjusting if necessary (FrostTheFox). So to use my own quotes to turn a lack of time into a way of saying I haven't been transparent enough, is kind of insulting.

jagauthier:
"Except you're oblivious to everything that's been happening in #pr." 

Youre 100% correct. I've submitted PRs and requested access (March 3, 27, 29) three times.

sebastienvercammen:
> Youre 100% correct. I've submitted PRs and requested access (March 3, 27, 29) three times.

So rather than find a solution to a practical problem (not getting a reply at the times you tried to contact us), you complain that we're not being transparent enough, rather than trying to understand our situation, without even once having contacted us about it? _(What did you do, who did you talk to, in what timezone were you when sending your messages, was it when I was at Disneyland or snowboarding, what did you try when you noticed I didn't/couldn't reply, ...?)_

I'd like to share people's ideas and thoughts with the community, but if it keeps coming back to the fact that people haven't even tried to put in the effort of understanding the other party, then it makes it a bit more difficult for me.

It only reinforces the suggestion of being a managed organization rather than "your local bar where everyone can enter".

jagauthier:
@sebastienvercammen 
" (What did you do, who did you talk to, in what timezone were you when sending your messages, was it when I was at Disneyland or snowboarding, what did you try when you noticed I didn't/couldn't reply, ...?)"

It's not my responsibility to send you a message at the perfect time for you, or @FrostTheFox .
Which I have done both.  

"and I've always been transparent in any way with the relevant people: #pr knows everything I've been up to,". 

Being transparent in behind closed doors (#pr) is exclusivity, not transparency.

I would like access to #pr.  I would like to discuss RM and it's development will all the fine people who contribute.  I would like to have a friendly relationship with you (which is very achievable, because I am certainly willing, and respect you).

So, this is my 4th format request for access to #pr.  You can search my name for PRs, you know who I am on discord.  And I've tagged both you and Frosty in it.

Thanks! Looking forward to further discussions about RM!

FrostTheFox:
@jagauthier I'm pretty sure I would remember if you direct PMd me recently while I was on and active, sometimes I wake up to a hundred messages (yes, not even lying, not all from RM, but from all the servers I'm in) and sometimes I don't really feel like reading every single one of them. 

I don't even know your discord name.

sebastienvercammen:
> It's not my responsibility to send you a message at the perfect time for you, or @FrostTheFox .

Yes it is. That's what you get when _everyone_ is a volunteer, no one is paid, and all we have is what we put into it. You cannot choose not to be proactive and then complain we didn't help you out enough.

Either you want to be in #pr and you do the absolute minimum to make it work (which really doesn't take much, you can ask all the other #pr members), or you don't but then don't complain about it.

> Being transparent in behind closed doors (#pr) is exclusivity, not transparency.

Transparency to _everyone_ contributes nothing and wastes time, which we already don't have. Transparency to relevant people, **especially when #pr is a public rank that anyone with minimal coding experience can access, making it a publicly accessible rank rather than an exclusive one**, keeps transparency and takes up less time.

> So, this is my 4th format request for access to #pr. You can search my name for PRs, you know who I am on discord. And I've tagged both you and Frosty in it.

The first thing I did was search for your name on Discord, I got zero search results. I have absolutely no idea who you are.

And this is the last comment I'll post about this here. This isn't showing us anything new, and it's not winning us any time.

smartcuc:
Well, the RM core team should be close together. Please do not get upset or anything else. Stay for this great software. Quality instead of speed. And, when an #pr comes up, take your time, then at the end, it's your baby you've made public.

jchristi:
Trying to think of takeaways from this and I think the obvious one is to encourage smaller pull requests. The larger the PR gets, the longer a code review will take (and the increase in turnaround time is not linear). This is kind of a microcosm for the argument of agile vs waterfall in software development. While this PR wasn't gargantuan, it wasn't small and it addressed _multiple_ bugs which probably could have been broken up and submitted as separate PRs. The maintainers should encourage and/or enforce smaller PRs but outside contributors would do well to heed this advice if they want to see their code merged in a timely fashion.

PR #1904, opened 2017-03-03
Making api check work when we are ahead of forced api

Buncertainc:
Update api check to make it possible to be ahead of forced api.
## Description

Check if forced api is newer than the api currently using rather than checking if they are different.
## Motivation and Context

If we are ahead of forced api and when we change the api version, the scanner will pause.
## How Has This Been Tested?

Tested on personal computer with window 7
## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -353,7 +353,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_queue = Queue()
     threadStatus = {}
     key_scheduler = None
-    api_version = '0.57.2'
+    api_version = '0.57.3'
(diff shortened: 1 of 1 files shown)


sebastienvercammen:
👎 

You're comparing strings with an inequality operator (`<`). While this will end up working in this case because of the ASCII values of your strings, it's much too primitive to merge into the codebase. It doesn't convert to integers, it doesn't handle unexpected values.

Using `!=` at least makes it strict, while `<` makes it much too lenient.

Buncertainc:
OK.
I may be thinking it too simple. I will close the PR then.

PR #1914, opened 2017-03-06
Support variations of Unown's form

Arengo:
The Pokémon Unown may have different forms of appearances which are now also available in the Pokémon info window on the map.
## Description

We are currently already retrieving informations about ``pokemon_display`` from every encounter request on the API. These informations are now being parsed in ``models.py`` and stored in the database. This has to be done to access the information alike all other Pokémon data for the Javascript map. An additional line carrying the Unown's form has now be added to the map.js as well, only triggered when it is an #201 Unown and the form is correctly retrieved.

## Motivation and Context

Unown is a very mythic and extremely rare Pokémon right now. It definitely has collection value and adds an unique entry to the corresponding medal. So, adding the parsing of this Pokémon's form has value and looking in the future enables some more playing with it - i. e. adding notifications just for the forms which are missing for each user.

## How Has This Been Tested?

There has not been one Unown seen in my area. Also, I need some more data when the form information is actually provided form the API and if there is the standard ``unset`` value for every Pokémon or just for failed calls etc. to handle possible exceptions since this can't directly be compared with gender/height/weight which is provided for every Pokémon.
**Update:** Received some user-testing. 

- Front-end seems fine
- Back-end is fine if you have correct db types of latest commits ;-) 
- There are some Unowns with forms displayed now in some live maps :-) 

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -2507,3 +2540,9 @@ def database_migrate(db, old_ver):
                            'MODIFY COLUMN `height` FLOAT NULL DEFAULT NULL,'
                            'MODIFY COLUMN `gender` SMALLINT NULL DEFAULT NULL'
                            ';')
+
+    if old_ver < 16:
+        migrate(
+            migrator.add_column('pokemon', 'form',
+                                DoubleField(null=True, default=0)),
```
Obihoernchen: SmallIntegerField and no default please

pogom/models.py
```diff
@@ -38,7 +38,37 @@
...
+UNOWN_S = 19
+UNOWN_T = 20
+UNOWN_U = 21
+UNOWN_V = 22
+UNOWN_W = 23
+UNOWN_X = 24
+UNOWN_Y = 25
+UNOWN_Z = 26
+UNOWN_EXCLAMATION_POINT = 27
+UNOWN_QUESTION_MARK = 28
```
Obihoernchen: ^ not needed I think

SiteEffect:
Someone has "frequent" Unown spawns to test? Would be perfect to see it live or at least a successfully written DB value.

Obihoernchen:
I've found an Unown and added the following code:
```diff
diff --git a/pogom/models.py b/pogom/models.py
index 8f0f921..6510e58 100644
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1934,6 +1934,10 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'weight': pokemon_info['weight_kg'],
                     'gender': pokemon_info['pokemon_display']['gender'],
                 })
+                if (pokemon_info['pokemon_id'] == 201 and 'form'
+                        in pokemon_info['pokemon_display']):
+                    log.info(pokemon_info['pokemon_display']['form'])
+

             if args.webhooks:
```
It outputs `6`

So the python code should work just fine.

SiteEffect:
Nice to see that it works on a real Unown in general, next step, have a successful value in the DB.

SiteEffect:
Awaiting a pending review result by a test user to confirm it working with the DB. Front-end seems fine.

twoopi:
image
Confirmed working for me.

SiteEffect:
@twoopi awesome, thanks for sharing. May I ask you, at which commit you pulled?

SiteEffect:
General question: Do you guys like it as ``Form: XYZ`` at the Gender/Height/Weight stuff or more like ``Unown (XYZ) - #201 (Ultra Rare)``. Happy about feedback.

pogom/models.py
```diff
@@ -1878,7 +1878,6 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 'height': None,
                 'weight': None,
                 'gender': None,
-                'form': None
```
Obihoernchen: This shouldn't cause an error. Which error did you get?
Because without setting it to None I think i fails if one scan finds an Unown and some other pokemon.

pogom/models.py
```diff
@@ -1894,9 +1895,16 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
-                    'gender': pokemon_info['pokemon_display']['gender'],
+                    'gender': pokemon_info['pokemon_display']['gender']
                 })
 
+                # Check for Unown's alphabetic character
+                if (pokemon_info['pokemon_id'] == 201 and 'form'
+                        in pokemon_info['pokemon_display']):
+                    pokemon[p['encounter_id']].update({
+                        'form': pokemon_info['pokemon_display']['form']
+                    })
```
Obihoernchen: This whole block should be in the if condition above I think.
This one:
```python
            if (encounter_result is not None and 'wild_pokemon'
                    in encounter_result['responses']['ENCOUNTER']):
```

Arengo: Yes.

343334:
For those of us who rely on the webhooks of RM will this also send the correct form to the webhook server? So that maybe we can talk to Deadly to have those forms added to PokeAlarm.

SiteEffect:
@343334 thanks for pointing that out. Will have a chat with Deadly and Seb about those webhooks.
**Update:** Seems like adding to webhook is supported by default already, well coded I guess - kidding. PokeAlarm would support it, by the way.

To actually push it a bit:
```General question: Do you guys like it as "Form: XYZ" at the Gender/Height/Weight stuff or more like "Unown (XYZ) - #201 (Ultra Rare)". Happy about feedback.```

**Comment by Jesterium** [comment]
I would prefer "`Unown (XYZ) - #201 (Ultra Rare)`" as I've trained my eyes to skim and ignore the gender/height/weight lines.

**Comment by SiteEffect** [comment]
Changed the displaying style in the info window. Another screenshot would be nice by you testers.

**Comment by Arengo** [comment]
@twoopi did you change the position of

**Comment by SiteEffect** [comment]
bildschirmfoto 2017-03-18 um 14 55 03
Always missclicking on "Close pull request", sorry :-/

File: pogom/models.py [hunk 5]
```diff
@@ -2598,3 +2615,11 @@ def database_migrate(db, old_ver):
                 migrator.add_index('pokestop', ('last_updated',), False)
             )
         log.info('Schema upgrade complete.')
+
+    if old_ver < 17:
+        migrate(
+            migrator.add_column('pokemon', 'form',
+                                SmallIntegerField(null=True)),
+            migrator.add_column('pokemon', 'shiny',
+                                BooleanField(null=True))
```
onilton: I think here it would better to add `shiny null=False` and  default to false, makes more sense.

pogom/models.py
```diff
@@ -105,6 +105,8 @@ class Pokemon(BaseModel):
     weight = FloatField(null=True)
     height = FloatField(null=True)
     gender = SmallIntegerField(null=True)
+    form = SmallIntegerField(null=True)
+    shiny = BooleanField(null=True)
```
onilton: `shiny null=False` and  default to false, makes more sense. (If missing, not shiny)

static/js/map.js
```diff
@@ -67,7 +67,9 @@ var gymTypes = ['Uncontested', 'Mystic', 'Valor', 'Instinct']
 var gymPrestige = [2000, 4000, 8000, 12000, 16000, 20000, 30000, 40000, 50000]
 var audio = new Audio('static/sounds/ding.mp3')
 
-var GenderType = ['♂', '♀', '⚪']
+var GenderType = ['♂', '♀', '⚬']
```
onilton: I know this is not really your work here, but looking at `gymTypes` it would better to keep consistency and convert all this variables to lowercase, like `genderType`, `shiny` and `form` (or `unknownForm`) ?

pogom/models.py
Arengo: It would add a ``false`` bit to multi-million of lines of existing databases, got feedback for the original Unown table from @Obihoernchen guiding me to set ``null=True`` :-/

pogom/models.py
Arengo: Depends, we don't know yet if info is only sent in encounter or before. If only in encounter ``False`` would be a "false positive" in some cases, if already on map_cell data, we can set it directly.

static/js/map.js
Arengo: Thought the same, wanted to consistent to the latest additions. Will change. Thanks for your review :-)

pogom/models.py
```diff
@@ -1894,9 +1898,22 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                        in pokemon_info['pokemon_display']):
+                    pokemon[p['encounter_id']].update({
+                        'form': pokemon_info['pokemon_display']['form']
+                    })
+
+                # Check for shiny Pokemon
+                if 'shiny' in pokemon_info['pokemon_display']:
+                    pokemon[p['encounter_id']].update({
+                        'shiny': pokemon_info['pokemon_display']['shiny']
+                    })
```
sebastienvercammen: Don't use `update`, just assign the value directly, it's much faster. The entire block (including the if statement) can be rewritten to:
```
pokemon[p['encounter_id']]['shiny'] = pokemon_info['pokemon_display'].get('shiny', None)
```

static/js/map.js
sebastienvercammen: Agreed, consistency fixes.

pogom/models.py
sebastienvercammen: Leave null as default. Null means no data, `True` or `False` mean we know what it was.

pogom/models.py
sebastienvercammen: Leave null as default. Null means no data, `True` or `False` mean we know what it was.

static/js/map.js
```diff
@@ -424,8 +426,14 @@ function pokemonLabel(name, rarity, types, disappearTime, id, latitude, longitud
...
         <div>
-            <b>${name}</b>
-            <span> - </span>
+            <b>${name}</b>`
+    if (id === 201 && form != null && form > 0) {
+        contentstring += ` (${Form[form]})`
+    }
+    if (shiny === true) {
+        contentstring += ` ${Shiny}`
+    }
```
sebastienvercammen: What will this show up as? Will this show a number for form, and `true` or `false` for shiny? If so, that's ugly and no one will understand.

pogom/models.py
sebastienvercammen: Same as my other edit, this entire block can be rewritten to the more efficient form:
```
if (pokemon_info['pokemon_id'] == 201):
     pokemon[p['encounter_id']]['form'] = pokemon_info['pokemon_display'].get('form', None)
```

static/js/map.js
Arengo: ...don't tell me JS is not case sensitive?

343334:
will this pr only work with -enc enabled ?

pogom/models.py
Arengo: Thanks, much appreciated, also for the Unown line.

SiteEffect:
There are rumors that shinies are unique per player encountering the spawned Pokémon. Therefore, the shiny-part of this PR maybe won't work  as source for public or personal maps. So it could perhaps just be used to gather data related to a shiny-percentage for example.

sebastienvercammen:
@343334 This isn't customer support, get out.

@Arengo Speculation for now, we'll have to wait either way.

sadyc1:
It is confirmed that shinies are unique per player.

SiteEffect:
The shiny part of this PR can now be seen as collecting research data. Actually I am happy that shinies are unique to players.

realindiahotel:
Keen for this, especially if it'll be accepted and displayed by the Webhooks

krosenvold:
I can confirm that this PR works, at least that the form is submitted to the webhook. I'm not running a map

realindiahotel:
@krosenvold did you have to put form in your alarms.json for PokeAlarm? Because I've left my PA setup as is and not seeing form.

Bart274:
Could the requested changes pleased be addressed so this could get merged?

343334:
@Thashiznets  Pokealarm needs to be updated as well for this PR to work -   def pokemon(data): in WebhookStructs.py would need to be updated to include form so it know how to handle that info from the webhook.  This is something for the PokeAlarm devs not RM ones. If you have the text output of what data is being sent to the webhook - im sure that would definitely help Deadly in incorporating it into PokeAlarm.

SiteEffect:
@bart274 all requested changes have been addressed.

PokemonCatcher:
The shiny code is cruft. Fork it out into another patch for people wanting to perform research on shiny spawns. The rest of us are better off without it.

SiteEffect:
@PokemonCatcher yup, you're right. Was hyped in the minutes Shinies touched the ground and haven't touched it since.

SiteEffect:
@Thashiznets GitHub is no place for support questions. Next time please use a PR only database and not deploy non-merged code to live maps if you don't know how to handle issues. However, this is not really a bad issue. You can just delete the ``shiny`` column if the code does not need it anymore via a DB editor like phpmyadmin and it won't affect anything else.

realindiahotel:
I don't have an issue and not asking for support, I'm just saying that as the original PR created an extra DB column, and now it's removed you've created two version 17 DB schemas one with shiny column one without which I thought would be a bad idea as some people will have the column and v 17 DB and others wont with v 17 DB.

Obihoernchen:
@Thashiznets oh no you are really wrong. It's your job to fix that. It's not a bad idea. It's a PR! Ofc he doesn't add migration code to remove this column. Everybody using PRs should be able to do this otherwise he shouldn't use PRs in the first place. This is not a place about how to properly apply PRs or react to changes of a PR. What @Arengo did here is totally fine.

realindiahotel:
Ok, I didn't realise that was the case, but if that's how it is then that's how it is.

SiteEffect:
@Thashiznets you can still fix your table by removing this single column.

realindiahotel:
Yes all good I was thinking about others however I guess you guys are right it's a PR bolt on so that's what we get for playing early.

ghost:
PokeAlarm PR to compliment this PR and Display Unown form now available https://github.com/ifindpokemons/PokeAlarm/tree/PR-Form

static/js/map.js
```diff
@@ -393,59 +394,62 @@ function openMapDirections(lat, lng) { // eslint-disable-line no-unused-vars
...
+    var pMove2 = (moves[item['move_2']] !== undefined) ? i8ln(moves[item['move_2']]['name']) : 'gen/unknown'
 
-    $.each(types, function (index, type) {
+    $.each(item['pokemon_types'], function (index, type) {
         typesDisplay += getTypeSpan(type)
     })
     var details = ''
-    if (atk != null) {
-        var iv = getIv(atk, def, sta)
+    if (item['individual_attack'] !== null && item['individual_defense'] !== 0 && item['individual_stamina'] !== 0) {
```
sebastienvercammen: I like that you simplified the function to its object as parameter instead of all individual values, but instead of referencing the object every time like you do here (`item['individual_attack']`), you should define the individual values at the top of the function, e.g.:
```
var atk = item['individual_attack'];
```
Computationally, this is entirely unnecessary, but it will improve readability and consistency.

It would also have reduced time spent reworking the functions, because you wouldn't have had to edit the previous variable names.

SiteEffect:
Since Travis is still not happy by default, here is flake8, build and lint outputs:
```
flake8 pogom
> Found a total of 15 violations and reported 0 (That's ok for Travis)
npm run lint
> Done, without errors.
npm run build 
> Done, without errors.
```

SiteEffect:
``Location`` was intendedly cropped out of the screenshot.
bildschirmfoto 2017-04-19 um 21 26 17

SiteEffect:
Close/open to restart Travis.

pogom/models.py
```diff
@@ -1988,9 +1990,14 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'move_2': pokemon_info['move_2'],
                     'height': pokemon_info['height_m'],
                     'weight': pokemon_info['weight_kg'],
-                    'gender': pokemon_info['pokemon_display']['gender'],
+                    'gender': pokemon_info['pokemon_display']['gender']
                 })
 
+                # Check for Unown's alphabetic character
```
sebastienvercammen: End comment with punctuation. We've had people put in time in the past to correct this.
```
# Check for Unown's alphabetic character.
```

imp3tus:
The latest change made this work 100% better.

pogom/models.py
Arengo: Point, dot, period. My true dedication since ArengoPR#1.

ghost:
Cheers @Arengo for this!

PR #1915, opened 2017-03-07
Make title customizable

WaveHack:
## Description
Adds a `--title` argument and a `title` template variable to override the default 'Rocket Map' title.

## Motivation and Context
I wanted to change the title from 'Rocket Map' to 'Pokemon Go map of $cityname', to make it more clear to my users.

## How Has This Been Tested?
1. Made changes
2. Use no config.ini with only req params (gmap key, location, username, password)
3. Observe title 'Rocket Map'
4. Cp config.ini.example config.ini
5. Change title in config.ini
6. Observe my custom title

rocketmap-title-pr-1
Without using title

rocketmap-title-pr-2
WIth using config: `title: My custom title`

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -56,6 +56,7 @@
 #ssl-privatekey:        # path to ssl private key
 #encrypt-lib:           # path to encrypt lib to be used instead of the shipped ones
 #status-page-password:  # enables and protects the /status page to view status of all workers
+#title:                 # title to use (default Rocket Map)
 
 # Uncomment a line when you want to change its default value (Remove # at the beginning)
 # Please ensure to leave a space after the colon (:) (example setting: value)
--- a/docs/extras/commandline.md
+++ b/docs/extras/commandline.md
@@ -35,6 +35,7 @@
                         [-spp STATUS_PAGE_PASSWORD] [-hk HASH_KEY] [-tut]
                         [-el ENCRYPT_LIB] [-odt ON_DEMAND_TIMEOUT]
                         [-v [filename.log] | -vv [filename.log]]
+                        [--title TITLE]
 
     Args that start with '--' (eg. -a) can also be set in a config file
     (default: <RocketMap Project Root>/config/config.ini or specified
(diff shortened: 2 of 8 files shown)


SiteEffect:
So, it's ``RocketMap`` either way. Either you fork and change code in ``templates/map.html`` or you install it and let it run, I think we should not provide a general support for that.

Alderon86:
Personally i like this idea, instead of changing the code everytime in your template. If its RocketMap or ArengoMap, who cares? if you dont wont to edit it you dont have to and easy custom settings on the user end are always nice and appreciated.

SiteEffect:
Agreeing, that it's actual useful if you provide multiple web servers and can easily distinguish them.

WaveHack:
For a bit more context, I wanted to put the name of the city I'm scanning in there. E.g. 'Pokemon Go map of $cityname', instead of a generic 'Rocket Map'. To make it more clear for my users.

onilton:
I think a PR like this was already submitted but it was decided to close it, I don't remember why. 

Unfortunately I was not able to locate it. If anyone can help with this...

lluisd:
I won't encourage to merge this PR. It adds a new commandline option which is not to enable/disable some functionaly/behaviour on the map. It's only a text/translate setting for the frontend part. Check at the current command line options, there isn't any text/constant replace setting.

A proper way will be to place some settings like that in some new json file related to the frontend website.

WaveHack:
@lluisd You make a good point. I can work on your suggestion.

I've also had some ideas regarding:

1. Hiding the 'Location & Search Settings' all together in options when the location is fixed (or its own setting)
2. Ability to set default options under settings. Notably a hidden search icon marker when the location is fixed, but could be used for anything (including default pokemon to hide, like pidgey, rattata, and sentret).

Something like a /config/frontend.ini.example would work? No CLI options, purely configuration file based. Could also use JSON, but I'd like to keep it consistent with the current configuration style.

Please let me know what you think of this.

MARTINI1:
@lluisd @WaveHack 100% according to hide options: location options as you say, also the options of the gym-sydebar, pokestops, the default search icon marker (Only obtrusive) and something else that I  forget xD

Good idea also to make a configuration file with the features that we want to have or disable, the list of pokemon hidden by default I also totally agree

Is something that I have always seen that needed the rocketmap and it is good that we talk about it here

Congrats for ideas  ;)

WaveHack:
Closing this PR then. I'll be starting a new branch on my fork in a few days.

I'll open up a new PR once I have something significant.

PR #1938, opened 2017-03-16
Added a method to force quit the server

EagleWorks:
## Description
Added a trigger for 'Q' to force quit the server.

## Motivation and Context
This change adds a method to (force) quit the server.  This solves the problem of having to spam Ctrl+C to quit the server, which in turn can cause things to be aborted mid-routines.

## How Has This Been Tested?
Ran the server, confirmed text formatted correctly on screen, confirmed pressing 'q<enter>' caused the server to quit.
This method _is_ unclean (should really signal from thread to process and have the process react appropriately with e.g. quit), yet many times cleaner than spamming Ctrl+C.

## Types of changes

## Checklist:

( I actually couldn't find where e.g. 'h' is documented!  If this hasn't been documented, I'd be happy to add documentation or otherwise modify the existing documentation. )

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -96,7 +96,8 @@ def switch_status_printer(display_type, current_page, mainlog,
         elif command.lower() == 'h':
             mainlog.handlers[0].setLevel(logging.CRITICAL)
             display_type[0] = 'hashstatus'
-
+        elif command.lower() == 'q':
+            os._exit(0)
 
 # Thread to print out the status of each worker.
 def status_printer(threadStatus, search_items_queue_array, db_updates_queue,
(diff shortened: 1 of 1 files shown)


EagleWorks:
Checked the auto-check, looks like style issues (well shoot, so much for "My code follows the code style").  Easy enough to resolve.  If I missed something in that output, let me know :)

EagleWorks:
Yep, just wanted to make sure that was all it was tripping over - thanks! :)

EagleWorks:
Yeah, could move it out, but given the other key interactions all fall under -ps as well, it seemed like a logical place.  Cool, thank you!

sebastienvercammen:
`os._exit(0)` forces the process to exit without calling cleanup handlers etc. It's not an improvement for anything and it looks like it's only trying to be a replacement for CTRL+C (or SIGINT), but CTRL+C already fulfills that job.

We do need graceful exits, but that'll be for after some other, more important stuff.

EagleWorks:
Thanks for taking a look, and fine by me :)  For what it's worth, it is an improvement to an extent.  This method, while unclean (which I mentioned), does exit the process reliably.  Ctrl+C can end up with things just hanging about.  E.g. right now, after a single Ctrl+C:

    2017-03-20 21:09:10,734 [switch_status_printer][     runserver][   ERROR] Uncaught exception  
    [...]
    command = raw_input()  
    EOFError

It then sits in the console waiting for...no idea what.  Takes a second Ctrl+C to actually exit.  Anecdotally, it beats spamming Ctrl+C, in lieu of a graceful exit :)

PR #1943, opened 2017-03-20
Fix #1929 Status page setTimeout() added whenever 'Group workers' is toggled

### LINKED ISSUE #1929: Status page setTimeout() added whenever 'Group workers' is toggled
## Expected Behavior
When toggling 'Group workers', only the display of the page should change, not the number of times the page is refreshed.

## Current Behavior
When toggling 'Group workers', updateStatus() is called.  This also adds another window.setTimeout(), causing each refresh to be called an additional time for every toggle.

## Possible Solution
clearTimeout(), but that would require it to be stuck in a variable first.  Alternatively, separate the setTimeout() from updateStatus().

1. Open /status
2. Hit F12 and check the Network tab to see when new data is loaded.  Keep an eye on the number of requests (Chrome), it should increase by 1 for each data load.
3. Go to the settings menu, and toggle "Group by workers"
4. The number of requests now increases by 2.
5. Toggle "Group by workers again", now the number of requests increases by 3.
etc.

## Context
This affected me more when the status page was updating as fast as it possibly could, already straining some browsers (FireFox, for example), exacerbating even further.  
I was trying to address that when I noticed this separate issue.

## Your Environment
RocketMap + FireFox / Chrome.  Did not test IE/Edge/Safari.

Alderon86:
When toggling 'Group workers', only the display of the page should change, not the number of times the page is refreshed.

## Description

quick fix for the timeout issue, status page is refreshed every second, so no need to call updateStatus again

## How Has This Been Tested?

My own server
## Types of changes

## Checklist:


myntath:
Taking this out fixes bug #1929, we have to wait up to 1 second before the data is refreshed, but that's seems acceptable to me.

PR #1948, opened 2017-03-23
Update how we get spawnpoints from the database and fix spawnpoint labels.

### LINKED ISSUE #1944: Spawn point labels have incorrect time span
When you enable the display of spawn points under Marker Settings and hover over a spawn point, the times displayed are at least partially incorrect.  (Note that I'm assuming the use of speed scan after it has learned all of the spawn times.)

## Expected Behavior
The label for a spawn point should indicate the correct appearance and disappearance times for the spawn point.
In addition, the text stating "May appear as early as XX:XX" is confusing and should be removed or corrected (if it has any meaning in the current spawn environment).

## Current Behavior
An example label for a 30-minute spawn that is available from 21:03-51:03 is as follows:

>Every hour from 36:03 to 51:03
>May appear as early as 06:03

Here, the appearance time (36:03) is incorrect, the disappearance time (51:03) is correct, and the "may appear as early as" time makes no sense.

Every spawn point label seems to have a 15-minute time duration, which makes no sense in the current spawn environment where every(?) spawn is either 30 or 60 minutes.

## Possible Solution
It looks like the spawnpointLabel() function in map.js is hard-coded to display a 15-minute spawn duration based on an offset from a single input time value.  The function should be modified to receive two inputs and display the correct appearance and disappearance times.  Depending on the database structure, this could be the either the appearance and disappearance times directly, or just one of them and a duration.

I don't know the purpose of the "May appear as early as" text.  Maybe this was supposed to be used when a Pokemon had discontinuous available times (such as being shown for 15 minutes, hidden for 15 minutes, and then shown for 15 minutes)?  I'm not sure if discontinuous spawns exist any more (everything seems to be 30 or 60 minutes).  If they don't, this text can be removed.  If they do, this text should be reworded to be clearly specify what it means.

## Steps to Reproduce
1. Run speed scan until all the timers are learned
2. Go to Options -> Marker Settings and enable Spawn Points
3.  Place your mouse cursor over a spawn point

## Your Environment
* Version used:  Latest develop branch
* Environment name and version:  Python 2.7.12, MySQL/MariaDB 10.1
* Operating System and version:  Windows 7

### LINKED ISSUE #2099: Shown spawnpoints are retrieved from Pokemon table instead from spawn points table
I was trying to restore the old spawnpoints from a backup DB, when I realized that, even when I delete all the spawn points from the current DB, the interface is able to show the spawn points.

I read the code a little and I think that the show spawn points in the web interface are retrieved from the pokemon table, instead from the spawnpoint one.

I think that with the probable incoming changes in the game, with some spawn points migration, it will be better to retrieve the spawns from its own table instead from the pokemon database.

To reproduce it (assuming MySQL DB)

1.- DELETE FROM  scanspawnpoint;
2.- DELETE FROM scannedlocation;
3.- DELETE FROM spawnpointdetectiondata;
4.- DELETE FROM spawnpoint;

Then, launch a web server and enable the spawn points: you will see a lot of spawns, even with that tables empty

myntath:
## Description

Previously we were getting a list of spawnpoints for the map by querying potentially every previous pokemon spawn. Since speedscan we keep a separate database of spawnpoints (even for non-speed scan instances). It is much faster to query this. We can also access accurate information about the appearance and disappearance times from the new databases. Previously we used a hardcoded 15 minute estimation of appearance times. This is pull request fixes this and displays correct spawnpoint labels.

Previously we used wording which said 'may appear as early as' on spawnpoitn labels. This was somewhat confusing. Now if we are not certain about a spawn point we state, 'spawn times not yet determine' and then give a best guess based the same methods as are currently used for guessing speedscan appear and disappear times.

## Motivation and Context
Fixes #1944 
Fixes #2099

## How Has This Been Tested?
Local Machine

## Types of changes

## Checklist:


Acadia1337:
+1 Working. Seems to reduce cpu usage.

PhDeez:
Do you need to drop your database to use this, or can I use an existing database that has all TTHs determined?

myntath:
@PhDeez no need to drop your database. The PR doesn't affect the way any data is stored.

pogom/models.py
```diff
@@ -315,73 +315,69 @@ def get_appearances_times_by_spawnpoint(cls, pokemon_id,
     def get_spawn_time(cls, disappear_time):
         return (disappear_time + 2700) % 3600
 
-    @classmethod
-    def get_spawnpoints(cls, swLat, swLng, neLat, neLng, timestamp=0,
+    @staticmethod
+    def get_spawnpoints(swLat, swLng, neLat, neLng, timestamp=0,
```
Obihoernchen: Shouldn't we move this to SpawnPoint class?

pogom/models.py
```diff
@@ -315,73 +315,69 @@ def get_appearances_times_by_spawnpoint(cls, pokemon_id,
...
+            del sp['done']
+            del sp['kind']
+            del sp['links']
+            del sp['latest_seen']
+            del sp['earliest_unseen']
 
         return list(spawnpoints.values())
 
     @classmethod
     def get_spawnpoints_in_hex(cls, center, steps):
```
Obihoernchen: What's about this method? We should change it to use spawnpoint table as well, shouldn't we?

pogom/models.py
myntath: Alright, finally got around to this...

pogom/models.py
myntath: Yes I could update this if we even still want it... get_spawnpoints_in_hex is only used in the SpawnScan scheduler and then only as a fallback if a json file is not provided. I can't think why you would want to use such an option now, surely you would just speedscan instead? Loading from a json still has its uses but perhaps if a json is not provided we should just throw and error and leave it at that.

myntath: In fact the code to do this is already there it's just commented out!

michikrug:
@myntath Should we then not also change the "get_spawnpoints_in_hex" method?

Obihoernchen:
@michikrug yes see my comment https://github.com/RocketMap/RocketMap/pull/1948#discussion_r109771468

myntath:
@michikrug @Obihoernchen I had previously indicated I didn't think there was any point fixing that code but I have reviewed this again and even after #2017 it appears that function will still exist for the purpose of dumping spawn points to file. I'm not quite sure why they functionality will remain but I will update this pull shortly to address the issue you have raised.

SiteEffect:
Can you resolve the conflicts, please?

myntath:
Conflicts resolved

friscoMad:
@myntath  Could you rebase pls, it should be easy.

Issue #1941, opened 2017-03-19
Some spawn points have timers with negative seconds


dsorc:
## Expected Behavior
Spawn points on the map should display the correct time.

## Current Behavior
Some spawn points on the map display an incorrect time (negative seconds)

Spawn Point
Every hour from 05:58 to 20:58
May appear as early as 35:-2

Every hour from 06:56 to 21:56
May appear as early as 36:-4

Every hour from 24:59 to 39:59
May appear as early as 54:-1

Every hour from 00:52 to 15:52
May appear as early as 30:-8

## Possible Solution
See https://github.com/RocketMap/RocketMap/pull/1900#issuecomment-287567008

## Steps to Reproduce (for bugs)
Check your map for any such points. Can be found relatively fast.

## Context
I don't think the scan is affected.
But the information provided to the user about the spawn point is confusing. Also, not sure why all spawn points seem to hint at being 45 minutes long (I think there are only 30 and 60 minutes).

## Your Environment
* Version used:
Develop branch 16 march
* Environment name and version (e.g. Python 2.7):
Python 2.7.13
* Operating System and version (desktop or mobile):
Windows 10 x64

EagleWorks:
Yours is the last Issue I'm looking at before calling it a night, and it's fairly recent, so just dropping a quick comment here to note that I noticed this as well.  It's a combination of calculation and subsequent display error.

The display error is that it's prepending the value with a zero which, in case of negative numbers, leads to "0-n".  It then takes the last 2 characters.  If 'n' is single-digit, you end up with the negative value string "-n".  When it's double-digit, all is well, as '0-nn' just yields 'nn'.  This could be fixed by looking for the sign and prepending the digit with the 0 (if appropriate) and leaving the sign in front.  It's moot, however, because...

The calculation error is the real culprit.  I don't know enough about spawn times to say one way or the other whether the input value is correct (if it's not, that's a different Issue), but the output value doesn't take into account the negative seconds, leading to some strange time measures.  This part is fairly easily fixed.  In the function `formatSpawnTime()`, replace the calculation with (the only thing that changes is the `seconds % 60` bit):

`return ('0' + Math.floor(((seconds + 3600) % 3600) / 60)).substr(-2) + ':' + ('0' + (seconds >= 0 ? seconds % 60 : 60 + (seconds % 60))).substr(-2)`

[...]

PR #1949, opened 2017-03-23
Fixes Issue 1941, negative seconds

EagleWorks:
Spawn point details can currently show negative seconds.  These should
be substracted from 60 to get the correct second count.
Note: This only fixes a formatting issue, not any issues with regard to
actual spawn times.  For discussion on the latter, see Issue #1944

## Description
Adds a ternary operator to subtract negative seconds from 60 to correct the negative seconds issue.

## Motivation and Context
Negative seconds don't make sense in display formats, especially not when they're only negative if they're single-digit.  
This fixes Issue #1941  

## How Has This Been Tested?
Three-fold: In-map, using Calc, and using a javascript iteration over a range of numbers.  
This change only affects displayed times in the spawn point details on the map.

https://cloud.githubusercontent.com/assets/3775662/24132943/989aeda0-0dcf-11e7-9335-42e971672d6a.png

## Types of changes

## Checklist:

The last two points are just a reminder to self that the front-end needs documentation, period :)

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -578,7 +578,7 @@ function pokestopLabel(expireTime, latitude, longitude) {
 function formatSpawnTime(seconds) {
     // the addition and modulo are required here because the db stores when a spawn disappears
     // the subtraction to get the appearance time will knock seconds under 0 if the spawn happens in the previous hour
-    return ('0' + Math.floor(((seconds + 3600) % 3600) / 60)).substr(-2) + ':' + ('0' + seconds % 60).substr(-2)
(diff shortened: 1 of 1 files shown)


SiteEffect:
The question is, is that a front-end issue or if there is something going wrong in the python part!?

EagleWorks:
@Arengo as per the summary, this fixes purely a display issue (a function that takes a value and returns a string for incorporation in the html), and doesn't touch any underlying issues with regard to whether or not the supplied value is actually correct.  For that, see Issue #1944 and its discussion thread :)

EagleWorks:
Yep - I vote the same if the underlying cause is coming up on fixes.  Might be more appropriate to put checks for incorrect values earlier in the code (from the calling function, say) and display errors (even if just in dev console) explicitly?

SiteEffect:
Imo, while this is a well PR code-wise it just fixes a symptom and not the bug itself. So I vote no, too :-/

PR #1950, opened 2017-03-24
Config file for map interface defaults

jagauthier:
This will allow end users to edit a special config file to modify their map front-end option defaults.

## Description
The idea behind this PR is to give the end users control over their map defaults.  With over 200 pokemon, it can get tiresome to configure the notifications,  and exclusions, for each web browser.  (Personally, I use web browsers on 2-3 PCs, and that doesn't include the other people in my house.)
So, I've created a file that will allow a person to specify what they want the map to do.

The following options are supported:
map_style - This is the look of the map
remember_select_exclude - exclude certain pokemon
remember_select_notify - notify certain pokemon
remember_select_rarity_notify - rarity notification
remember_text_perfection_notify - %IV notification
playSound - true/false if bounced pokemon go 'ding'
showPokestops 
showPokemon
showGyms
startAtUserLocation - start at the users location (of course this only works with SSL enabled map)
zoomLevel - the default zoom level

I tried very hard to make it so that map.common.js can be modified in the repository without worrying about stepping on people using this code.
So, basically, map.common.js has the changes merged into it, but it stored as map.common-custom.js.
The grunt process will use the custom file, instead.  Also, if there are no customizations done then
the process will continue to work.  

The new config file is called map-defaults.ini (the example is attached and commented)
I felt putting this in a different config file made sense, since we have no idea what an installations
config file might be called, we can't easily figure it out.   This way the map-defaults.ini will be the only config file needed.

I added a grunt process, called 'shell' that copies the map.common.js file, and then also runs the included map_default,py file.  It's a fairly crude method to accomplish this, but it should be effective.

If the user has not made any customization, it will exit without error.

## Motivation and Context
The primary motivation for this change was driven by the end users I see asking how to change their defaults.  I don't think it's unreasonable for an end-user of an application to want to change it's defaults.
The issue comes into play when the main repo gets an update and then those end-users have merge issues because they've modified a local file that the repo handles versioning for.  This will not be an issue moving forward.

## How Has This Been Tested?
I tested this on a local development environment by changing all of the settings mentioned.

No screenshots.

## Types of changes

## Checklist:

If this PR is accepted, I will create a new page dedicated to the configuration of this.


MARTINI1:
This pr looks like #1915 of @WaveHack, which we talked about (besides modifying the title of the page) to hide sections like locations, options of the gymsidebar, default search icon marker, etc., so it was closed to make another PR with these modifications.

Have you considered doing this or only what you have done here?

Still this is already a principle to personalize the frontend well without having to modify the js files

Good contribution ;)

jagauthier:
When I began work on this, I considered the ability to configure multiple things.    My initial plan was to help people who are constantly asking to exclude pokemon by default. I then added several other 'easy' options.

I am not opposed to  more options.  This was just a stepping stone.

WaveHack:
@MARTINI1 About PR #1915, I haven't found time yet to work on it. :( I'll take another look at it in the coming weeks, now that I have some time off work.

I'll probably check out and work off this branch, since some work has already been done.

jagauthier:
I am gong to close this PR, and remove it from my local fork. 
With the impending front-end changes, it's not relevant in it's current state.

PR #1953, opened 2017-03-26
Added exception handler for update_total_stats in search.py

voxx:
This is a patch pulled out of #1900 that adds an exception handler for update_total_stats when size of dict changes. Detailed in that PR as ISSUE4

## Description
RocketMap can stop with an uncaught exception:

Current behavior:

Changes:
In the seach.py, add an exception handler to catch any problems while copying the queue items.

New behavior:

The exception is caught and the Map continues.

## Motivation and Context
I've plucked this patch from #1900 and am submitting it as it's own PR, as it has been a fairly commonly reported problem in help chat on Discord recently and it may be a while before the original larger PR is fully vetted/tested/approved. This smaller issue was bundled with that PR and is probably more appropriate as a smaller patch.
This fixes ISSUE 4 outlined in #1900 

## How Has This Been Tested?

Deployed on 3 VM's for past 3 days with no issues found aside from catching the previously uncaught exception and allowing the script to continue without a hard failure.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -537,7 +537,15 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
                     i].get_overseer_message()
 
         # Let's update the total stats and add that info to message
-        update_total_stats(threadStatus, last_account_status)
+        # Added exception handler as dict items change
+        try:
+            update_total_stats(threadStatus, last_account_status)
+        except Exception as e:
+            log.error(
+                'Update total stats had an Exception: {}.'.format(
+                    repr(e)))
+            traceback.print_exc(file=sys.stdout)
+            time.sleep(10)
         threadStatus['Overseer']['message'] += '\n' + get_stats_message(
             threadStatus)
 


pogom/search.py
```diff
@@ -537,7 +537,15 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
...
-        update_total_stats(threadStatus, last_account_status)
+        # Added exception handler as dict items change
+        try:
+            update_total_stats(threadStatus, last_account_status)
+        except Exception as e:
+            log.error(
+                'Update total stats had an Exception: {}.'.format(
+                    repr(e)))
+            traceback.print_exc(file=sys.stdout)
+            time.sleep(10)
```
voxx: Is this sleep actually required?

friscoMad:
I got the same exeption today, I wouldn't fix here as there is nothing to do here and real problem is that there is no exception handling on overseer and without that thread the map just hangs.

pogom/search.py
friscoMad: Not needed

voxx:
Without handling the exception the thread just hangs and doesnt do anything, creating a memory leak, usually leading to additional problems for other instances running on the same server.

This problem occurs quite often for me, and without the above fix I was constantly having to restart instances that got hung up and wouldnt continue. With the fix I no longer have to do that at all.

Catching the exception allows the script to continue, and it recovers and functions normally.

-voxx

friscoMad:
I know I got the exception but there is no need to handle here, just handle it on the function calling this one as that will make it continue in case of an exception in this function or any other part of the overseer and you will do the same, just trace and continue.

friscoMad:
My bad I did not read correctly the diff, but I still think that the try/catch should be for the whole while (true) instead of just fixing 1 exception point.

voxx:
Closing this PR out as 1900, which contains this fix has finally been merged.

-voxx

Issue #1829, opened 2017-02-16
[FEATURE REQUEST] Make pokemons with high IV bigger.


clahs:
If the Pokemon has an IV of 90+, then make them more visible on the map, perhaps. larger than other pokemons on the map.

SiteEffect:
+/- 50% size for IVs from 0% to 100% would be an idea.

sebastienvercammen:
Feature requests: http://vote.devkat.org.

PR #1706, opened 2017-01-12
Gym opacity

springjools:
#Fade out gyms that have been scanned a long time ago

## Description
Gyms scanned less than 2 hours ago: 100% opacity
Gyms scanned more than 2 hours ago but less than 12 hours ago: 75% opacity
Gyms scanned more than 12 hours ago but less than 24 hours ago: 33% opacity
Older gyms: 17 % opacity

Can be configured in  map.common.js
## Motivation and Context
Makes it easier to distinguish old gym data from new one

## How Has This Been Tested?
Tested on local machine

## Screenshots (if appropriate):
Normal map: http://prntscr.com/dup4jx
Dark map: http://prntscr.com/dup6nu

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1956, opened 2017-03-30
Add Pokémon transparency based on IV

Arengo:
This PR introduces the feature to distinguish Pokémon by their IVs by making the low IV-Pokémon more transparent. 
## Description

A new toggle option is available in the Map menu, wrapped below the Pokémon marker switch. When enabling (default is ``false``) marker opacity get's weighted with Pokémon IV, ranging from 33 % for no-IV Pokémon or 0% IV Pokémon to an opacity of 100% for 100% IV Pokémon.

## Motivation and Context

This feature is introduced through requests to better distinguish Pokémon of interest on the map. Also there was a Feature Request in #1829 as well as on [DevKat](https://feedback.userreport.com/9eec75e1-1e84-4f02-b254-5e239d896805/#idea/159800).
Instead of a toggle it was also considered to introduce this as default, when ``-enc`` is enabled.

## How Has This Been Tested?

On my personal map.

## Types of changes

## Checklist:

- [x] My code follows the code style of this project. (Awesome JS code style)

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2294,7 +2294,6 @@ $(function () {
             lastpokemon = false
             wrapper.hide(options)
         }
-        Store.set('showPokemon', this.checked)
         buildSwitchChangeListener(mapData, ['pokemons'], 'showPokemon').bind(this)()
     })
     $('#pokemon-opacity-switch').change(function () {
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -349,6 +349,8 @@ function initSidebar() {
     $('#max-level-gyms-filter-switch').val(Store.get('maxGymLevel'))
     $('#last-update-gyms-switch').val(Store.get('showLastUpdatedGymsOnly'))
     $('#pokemon-switch').prop('checked', Store.get('showPokemon'))
+    $('#pokemon-wrapper').toggle(Store.get('showPokemon'))
+    $('#pokemon-opacity-switch').prop('checked', Store.get('showPokemonOpacity'))
     $('#pokestops-switch').prop('checked', Store.get('showPokestops'))
     $('#lured-pokestops-only-switch').val(Store.get('showLuredPokestopsOnly'))
     $('#lured-pokestops-only-wrapper').toggle(Store.get('showPokestops'))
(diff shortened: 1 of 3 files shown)


static/js/map.js
```diff
@@ -2276,8 +2283,23 @@ $(function () {
         buildSwitchChangeListener(mapData, ['gyms'], 'showGyms').bind(this)()
     })
     $('#pokemon-switch').change(function () {
+        var options = {
+            'duration': 500
+        }
+        var wrapper = $('#pokemon-wrapper')
+        if (this.checked) {
+            lastpokemon = false
```
Arengo: Not sure if this line and [L#2294](https://github.com/RocketMap/RocketMap/pull/1956/files#diff-052cfee44ae462a28560613abd7fdd14R2294) are needed for a more robust code-style. Any deeper knowledge available?

friscoMad:
I think that no IV should be left as full opacity as there is no way for a web instance to know if the mapper instance has -enc or not, it is not great having a slider that does not do anything but and having all pokemon at 33% make it worse.
Otherwise it should be under webserver flag to disable it when it does not do anything.

SiteEffect:
@friscoMad thought about it while implementing. Came to the conclusion for myself that either you are interested in IVs, so you only add ``-enc`` for IVs to those which you want to see, so this option is something for you to turn on **or** you don't scan for IVs at all and therefore this option could be just hidden like in #1840. This could be implemented as soon as the other PR is merged. Happy about more feedback, though :-)

friscoMad:
If PR #1840 is merged before this and this PR is updated, then it would be perfect.
This PR is really nice, it is just the degraded UX for users of maps without encounters that bogged me.

springjools:
I tested various tranparencies with gyms, and I noted that it's not a good indicator for a scale that is larger than say 2 or 3 steps. The reason is that you cannot see various transparencies well on the map, or how they differ. Did you also test with dark, road and light map? It has to work with all of these. 

Just my thoughts, but maybe a iv bar (like energy bar in some games) would suit better?

And similar with #1706 : would it be better to store the opacity levels in map_common.js? To be consequent...

SiteEffect:
Regarding consistency I could move the min-opacity to map.common.js to make it easier customizable, any other value is assigned during runtime and dependent on that. Nice you mention downsides with opacity, so I did not test this on a very large scale, however when you have a crowded area, there are general distinction problems with overlapping and non-linear sized markers.

SiteEffect:
(There are thoughts to change this PR to highlight high IV Pokémon another way.)
This PR is currently paused but might be opened in the future.

Issue #1487, opened 2016-10-08
[Feature Request] dump-pokestops


kbtbc:
Now that active scanning is dead (temporarily I hope) ... I would like to use the spawnpoint and pokestop data I have saved.   I saw this tool, which might make some of this data very useful: https://www.reddit.com/r/pokemongodev/comments/50cxls/what_is_the_best_path_i_could_do_for_xp/

We can currently dump spawnpoints (constrained with the -st flag) ... I'm looking to do the same w/ pokestops (and why not gyms, too?)  

What I'm missing is a pokestop.json file -- I could probably work this out in the code, but I would be starting from scratch since I'm not a dev.    A simple side script to do this would be awesome :-)

shadi1989:
Think this site will do it for you : http://www.pokemongomap.info/

kbtbc:
Thanks for the tip! It's took a little learning, but in Chrome using the Developer Tools and copying the response of mapdata.php from that site, and then feeding it into a json formatter, I was able to get this info.

F1delis:
I have a spawns.json with > 40.000 spawnpoints, but I see now way to load it to the map.

StairyNight:
@shadi1989 Do you suggest that it can be relatively simple for a noob to pull the webpage and then somehow feed the data in local mysql database to show the gym and pokestop locations on the map?

m3g4mInD:
Yeap its simple, i have made that , only php/mysql data ,
Pokestops  : http://prntscr.com/ctbc2e
Gyms :   http://prntscr.com/ctbck2

This is cool for the people who whant atack gyms for coins :)

StairyNight:
@m3g4mInD  Mind sharing the procedure how you did that?

m3g4mInD:
For Gyms:  https://gist.github.com/m3g4mInD/7e981603a852b099675bacb396e1b9c9

For Pokestops: https://gist.github.com/m3g4mInD/b5e1a7e022976765e465601f87b1368f

I'am working on spawnpoints ....
I know the code is not clean and i have some mistakes,but i gonna improve better code.
Thanks.

SiteEffect:
We could add a --dump-pokestops and --dump-gyms and make it being used by -ss scanning to enable spawnpoint scan for Gym trackers and lured Pokémon trackers.

FrostTheFox:
Feature Requests are no longer are taken via GitHub. Please submit FRs using http://vote.devkat.org

Issue #1321, opened 2016-09-17
[Feature Request] Show gym scan radius


SoonDead:
Would be great to see how far a scan is from specific gym. Would be easier to spread out scanners just for gyms.

beccasafan:
Is this what you're going for? #861

SoonDead:
Yep

FrostTheFox:
Feature Requests are no longer are taken via GitHub. Please submit FRs using http://vote.devkat.org

Issue #1154, opened 2016-09-05
[Feature Request] Gyms scan only


cyan:
heya,
i'm looking for a way to get a gym scan with fresh info, thanks to some people in the discord i have a scan with -np & -nk wich make it a lot faster but i still have a lot of useless area scanned (i live in a town the kinda the size and form of central park so a hexa would get a lot of point outside).

i've been looking in -ss and ---spawnpoints-only but could find a way to do it :/.
it would be nice to have a --gyms-only ^^

thanks again for your work and good luck :)

hundasupa:
gym range is 900m so you can just put a worker with -st 1 in a decent spot with a long enough -sd for whatever you want the refresh rate to be. 

or do you want to cover the entire town for just gyms with a single worker?

filippounitn:
there's already gym only scan i guess.. tried some week ago just put -nk -np and the scan will be very very wide
Try again with those 2 parameters and you should ahve what you are looking for if they didn't change it

cyan:
@hundasupa yea i would like to scan the whole area since it thin but long, so i would need something like 10-15 spot to be scanned to get all info i think :/

@filippounitn as i said i tried that already, it work fine, but lot of useless area are scanned like half the spots

jchristi:
@cyan, you could probably dump a list of gym locations and reformat it into -ss formatted json and run a worker with -ss set to that file. Kinda hackish but it should only scan the gym locations.

cyan:
@jchristi i tought about that but when i got my hand on the -ss json it had a "timer" value for each spot and couldn't remove it.
i wouldn't mind having to create manualy a list of spawn point to scan (would even be better since i could scan an area and not each gyms).
i'm not that much into coding, i know how to change some little stuff but i don't know the json formated or what's really needed for a -ss json :/

jchristi:
@cyan this would be a cool feature though. I'm worried code base is getting bloated with all the features being added (search.py is getting to be a mess). I may look into working on this later but I've got a lot of other things I'd like to do first.

cyan:
@jchristi , if you could do something like simple scan but from coords and i add -np -nk, it would scan listed in a file for everything but those 2 ;) (maybe the code would be easier that way for a first time, and could be used in other way like scanning for specific spawn spot ;) ) thanks

Timbo925:
[...]

PR #861, opened 2016-08-20
Scan circles appear the correct size on the map when no-pokemon is

wesleytsai:
## Description

Scan circles were stuck at 70m regardless of whether we're scanning gyms/pokestops only, or for Pokemon. Changed so that it dynamically switches circle size based on scan type from #748 
## Motivation and Context

Scan circles were improperly sized when scanning for only pokestops/gyms 

## How Has This Been Tested?

I see that the circles change size between pokemon scanning and non pokemon scanning

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

enabled now

PR #1957, opened 2017-03-30
Add Gym and Pokéstop scanning from DB and requested file dumps

Arengo:
Implemented the ability to dump data of gyms and Pokéstops directly to a json file. Additionally a scan mechanic was introduced to scan forts (i. e. gyms and Pokéstops) based on the ``skip-empty`` scan method since a real spawnpoint scanning is not relevant when there is no timer. Gym scanner are now better to track by vision, adding a old and never merged PR feature which increases scanned location radius on map.

## Description

The implementation was done with respect to the existing features of dumping spawnpoints as well as the skip empty cell scans. This resulted in the introduction of a new scan flag ``--fort-scan`` (``-fs``) for so called "gym scanners". Hereby, valid gyms and Pokéstops inside the step-limit are pulled from the DB and stored inside a file if ``--dump-gyms``or ``--dump-pokestops`` is specified together with a file. If not file is added for scanning or dumping, the new scan sub-class ``HexSearchFort`` pulls all otherwise dumped forts from the DB directly and the generated hex area is thinned out when no fort is found inside each cell.
All these features are best used with the ``--no-pokemon`` (``-np``) flag to increase the cell size from standard Pokémon size of ``70 m`` to a respectable gym scan size of ``450 m``.
To make the bigger cells when using ``-np`` finally visible I implemented PR #861 to todays DB structure. All credits to @wesleytsai.

## Motivation and Context

This feature has been requested various times since gym only scanners are often used to expand the gym map while not increasing ressource costly Pokémon scan areas. Most lately the highly voted request of [DevKat](https://feedback.userreport.com/9eec75e1-1e84-4f02-b254-5e239d896805/#idea/151939) and in Issue #1487, #1321 and #1154 

## How Has This Been Tested?

On my personal map.

## Types of changes

**Note:** This feature can be extended with Lured Pokéstop support - finding a lure with a fort scanner locks scan-position for one worker and switches ``scan-delay`` to each 3 min. to retrieve the newly spawned Pokémon until the lure expires.

## Checklist:

- [ ] I have updated the documentation accordingly. (To-Do)


bbdoc:
Hi Arengo,

Looks like a very interresting feature. Anyway, I'm a bit surprise this wan run with only -np set. I thought you can scan with a 450m range only if running -np -nk ?

Could you also give more details on how this would work ? Will this PR do some kind of initial scanning to discover gyms in the same way -speed is doing it and then automatically switch to skip-empty ? What if gyms have been scanned already ? How will workers be assigned to different hives ? In other word, will this work with a very big -st as workers will automatically calculate ideal path the closest gyms ? I fear as distances can be quite big in gym-only scanning my workers might start travelling 100km to get to a gym on the other side of the map.

Thanks

SiteEffect:
@bbdoc ``-np`` deactivates the 70m search cell and activates the 450m search cell since quite some time now. (``-nk`` is not needed). You can compare this on your phone, when travelling above speed limit of 38 km/h that Pokémon will not appear anymore, sightings remain empty but you still see Gyms and Pokéstops.
Regarding to this ability to travel above speed limit, people have already implemented Gym scanners  for a while. I know metropolitan gym scanners where worker travel with 700 km/h without loss in functionality or banned accounts over time. Therefore, as viable technique to scan forts, ``--skip-empty`` was chosen - skipping all cells without any fort in it improves the general load of your workers. You use it on an existing database, where forts are already all known. This can be achieved during regular Pokémon scanning with speed or ``-np`` standard HexScan. When you collected your forts you can switch to ``-fs gyms.json/pokestops.json`` and **either** ``--dump-gyms/pokestops`` them once into a file for re-use (dumping was generally requested due to other reasons, too) and get your scan-informations of where the forts to scan are **or** you just leave it the scan method to grab them for each scan from the DB what is a slight increase of load. Compare this behaviour with the old ``-ss`` from file or directly from the DB. 
The scan mechanism itself is just HexScan with skipped cells because we don't have to account timing of spawnpoints, since forts can change anytime, and we don't care about an optimal worker speed related path since we have technically not a real speedlimit. With this, all needs are already fulfilled while still having a efficient way to cover as much area as needed/possible. I wan't to point out that it would be great to address leap frogging of multiple worker in HexScan in the future - preventing a start from one hex corner at once but distributing the area initially somewhat equally between the available worker.

bbdoc:
ok so my workers will simply teleport between the different hives that contain a gym ? How do you define the time between the scans for each gym ? Is there some sort of formula to know how many workers you need based on the number of hives that contain some gyms ? Btw, will status screen report on the number of gyms found and number of hives that need to get scanned ?

SiteEffect:
It's fully configuration dependant. You simply decide with ``-sd`` how often scans happen.

SubandiK:
Works like a charm thanks so much!

pogom/models.py
```diff
@@ -524,6 +524,42 @@ def get_stops(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None,
...
+        s = list(query.dicts())
+
+        step_distance = ((steps - 1) * 779.4231) + 450
+        filtered = []
+
+        for pokestop in s:
+            if geopy.distance.distance(
+                    center,
+                    ((pokestop['lat'], pokestop['lng'])).meters <=
+                    step_distance):
```
michikrug: you messed up the brackets in this one.
should be:
```
if geopy.distance.distance(
                    center, (pokestop['lat'], pokestop['lng'])).meters <= step_distance:
```

Arengo: Oh, you're right. Thank you, will fix it! Flake8 persuaded me to write shitty code :-P

moropg:
It seems that this PR updates the db to a newer schema version. Beware of that while checking this out.

For the rest, this PR is quite gamechanging while scanning gyms. It has issues, e.g. with massive step distances (like st 50 or 100) it takes forever to fire up the threads (if it ever starts), and I still get some "empty" scan. This thing asks for a step up, something like clustering the dumped points and feeding them in a -ss like fashion. A massive improvements from a regular hex nonetheless!

SiteEffect:
@leccarelepalle When it comes to huge ``-st``, RocketMap struggles in general. This is not related to this PR.
The empty scans might come from complete soft-bans when you really travel with enormous speed like in a ``-st 100`` with ``-w 100``and ``-np`` so you travel +45 km in one single ``-sd 15``. This is probably too much and asks for an additional speed-limitation implemented into the code (I do travel with around max. 600 km/h without any issues for forts, though). In your case, the exact numbers of your configurations would be interesting, when you get these full-empty responses.
Regarding clustering, I do not know if this would really improve that scanning mechanism at all since we do not have a specific time to scan a gym because we do not know when it changes and the areas are already filtered. What can happen is, that let's say 2 gyms are close to each other but fall into two different cells and no other gym would be neglected when only these two are covered by a "clustered" scan. It happens sometimes, but I do not think this is a game changer like clustering Pokémon spawns, where the additional time information is the main reason why to cluster.

moropg:
@arengo about empty scans, I'll tell you more once I did more testing, I already have a reasonable -sd though . I've yet to fire up a st100, because the hex schedulers gets stuck for ages. I think the problem is in the skip-empty algorithm which needs to calculate which cells are empty. I agree that clustering is totally optional, but a better approach for wide areas would be the aforementioned "-ss like" approach in which only non-empty points are used (and directly fed by the DB/Json). This way the whole "which cells are empty" calculations are avoided (worth mentioning that some of those routines are also pretty broken and unreliable for distances like a st100 gym scanner) and you may actually go big with gym scans. Using this approach the scanning points clustering (just spatial clustering) may be really nice to avoid to scan many times the same area if they are full of forts. Would also be a pretty trivial modification of the already existing tool.

SiteEffect:
@leccarelepalle from a computational point of view, I totally agree. It will be much resource lighter in the start up. In this case, clustering makes sense instead multiply scan one area with multiple forts. Thanks for your feedback! Let me hear about your findings regarding the empty scans.

BeriBGood:
I have been using this for 2 days+ now, working fine and no bug or exceptions detected for now.

moropg:
@arengo empty scans seems quite sporadic, and we don't have any on smaller step sizes. Using st < 20 is quite fast to fire up (seems faster pulling directly from DB instead of Json), even when dinamically changing location. I can use up to st 40ish with a reasonable startup time, anything above seems to get stuck forever...Can anything be done about it? I would love to scale up to st 100 hexes

bbdoc:
Am I the only one who cannot get this to work ? I'm getting `Schedule creation had an Exception: UnboundLocalError("local variable 'fort_locations' referenced before assignment",)` on startup. I started with a fresh clone of PGM and only included this PR to make sure it was not linked to a dependency with something else, but still the same error. Did anyone else tested this PR using beehives ? Thanks

SiteEffect:
@bbdoc I guess beehives probably produces the error. While this should be addressed by me to support all viable scanning methods, it is not recommended to use beehives when you can cover the same area without ``-bh`` in this PR.

moropg:
The problem is, @arengo , I can't really exceed  st 40ish or the threads never start, so I too could make some use of the -bh flag

SiteEffect:
Close/Open to restart Travis.

bbdoc:
Any chance to get an update working with -bh ?  ;-)

bbdoc:
I'm giving this PR a try without -bh at the moment ;-)

Anyway, getting a lot of those error messages : 

```
Exception message: UnboundLocalError("local variable 'abandon_loc' referenced before assignment",)
  File "/usr/src/app/pogom/search.py", line 987, in search_worker_thread
    scan_date, account, account_sets)
  File "/usr/src/app/pogom/models.py", line 1967, in parse_map
    if abandon_loc:
UnboundLocalError: local variable 'abandon_loc' referenced before assignment
```

No big deal as probably linked to some accounts being banned, but should not throw this error...

michikrug:
As I researched on those errors, it seems that this is a problem with old / left-over code that should not be there anymore. I removed those parts (line 1962-1978) and it is working fine without errors.

bbdoc:
@michikrug So you removed lines that were added by this PR on models.py ? Would prefer to let @Arengo check this will not result in any side effect.

[EDIT] Removed those lines, but now getting another error :  `2017-05-05 12:24:36,817 [   search-overseer][        search][   ERROR] Schedule creation had an Exception: TypeError("unsupported operand type(s) for +: 'dict' and 'list'",).`

michikrug:
@bbdoc since I have merged a lot of PRs I can not really tell, what is also changed in my code in comparison to this current PR
But your error seems strange and more likely a mistake on your side?!

bbdoc:
I didn't merge any other PR. Fresh clone of dev + merged only this PR + deleted those lines from models.py. I'll wait for @Arengo to remove those lines and push another commit if he confirms they are not needed and will try again.

SubandiK:
I have the same problem as bbdoc after getting a more recent version of this, with the original it worked flawlessly, didn't change anything else.

SubandiK:
Actually, I just "solved" it for me by inserting 
 abandon_loc = False at #1953 
because the error was basically complaining that it wasn't set. Seems to run again, dunno about long-term consequences though.

SiteEffect:
Will probably rebase that soon and change a lot. Let's use the deprecated ``-ss`` #2010 or ``-speed`` for this. Closing it for now.

rpotpogo:
@Arengo - Any chance of this PR coming back? I need this for my gym only scanner.

michikrug:
@rpotpogo I have it still maintained in my new-api and raids branch. But I have to say, there is also a lot of customized code and content included.

SiteEffect:
Yes, chance for a re-write with -ss for raids and --skip-empty for forts is there.

SkOODaT:
I know this is closed but I need to say THANK YOU for this, I adapted it to work with my map and I have the ability to lure stops on my map, so now i have a dedicated lure placement scanner .....my community thanks you :)

blakewenloe:
@SkOODaT Can you PM me how you accomplished that?

ghost:
@SkOODaT How did you manage that?

PR #1960, opened 2017-03-30
Improve memory usage

friscoMad:
## Description
Free variables as soon as possible to let Python GC free the memory, this was done on all thread but focused on search thread.
Minor changes where needed in pokespin function to allow an early del of inventory data in map request.

## Motivation and Context
One of the main memory hugs right now is keeping all the reponses while the thread is paused.
This is due to the lack of scopes in python  outside the function scope and the fact that the current code has enormous functions.
So the first thing to reduce memory is free it as soon as possible in the thread functions as otherwise it will be kept referenced until a new value is assigned to it.
Request responses are fairly big specially for accounts that have non empty inventory, so I did review all threads and functions used in those threads trying to delete the variables so Python GC can free them.

For that I did include a new function that clears all the extra info that is not needed but it is requested every time (badges, buddy, settings...) I am not even sure if it is needed in every request, probably it would be enough if made in map requests, anyway it is freed as soon it is returned from api.

In DB and WH threads I did the same thing for the objects that can be freed, those changes are probably minor as those threads are almost never paused, not like search threads.

I did choose to use `del var` over `var = None` as it seems more obvious the reason of that line and would make future commiters aware. Another option would be leaving just the main loop in the thread function and then make everything else in another function that way local variables would be cleared on exit. 

## How Has This Been Tested?

This was tested for 7h using develop branch and then with the same config with this changes, the only minor change I did was add another database worker as the baseline test was not able to cope with data (so there gain is a bit less than shown), still in previous tests the results were similar, around 30% decrease in memory consumption and almost 90% decrease in dict objects in memory (responses have a bunch of dicts inside dicts).
The test setup did not have web server, st 39, 230 workers, 1 webhook with gym info and not encounter.

My accounts are leveled so their inventory has a lot more items, setups with lvl 1 and lvl 2 accounts will see smaller improvements.
The only modification I did aside of the PR was adding a hook for pdb-clone so I could connect and get some insight of the running instance, only used after the memory screenshoots were taken.

**Not tested:** Tutorial usage

Current develop branch, after 7h of work, around 1.1Gb memory

PM2 info (test app is the one your are looking for):
2017-03-30 11_54_28-bitvise xterm - pokemad bscp - pokemad neoseo es_81

Top:
2017-03-30 11_55_47-bitvise xterm - pokemad bscp - pokemad neoseo es_81

Objgraph info:
2017-03-30 11_57_04-bitvise xterm - pokemad bscp - pokemad neoseo es_81

After PR, 7h working 800Mb
2017-03-31 01_13_43-bitvise xterm - pokemad bscp - pokemad neoseo es_81

Top:
2017-03-31 01_10_13-bitvise xterm - pokemad bscp - pokemad neoseo es_81

Objgraph:
2017-03-31 01_11_40-bitvise xterm - pokemad bscp - pokemad neoseo es_81

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X ] My code follows the code style of this project.

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -207,8 +207,7 @@ def complete_tutorial(api, account, tutorial_state):
 # Complete tutorial with a level up by a Pokestop spin.
 # API argument needs to be a logged in API instance.
 # Called during fort parsing in models.py
-def tutorial_pokestop_spin(api, map_dict, forts, step_location, account):
-    player_level = get_player_level(map_dict)
+def tutorial_pokestop_spin(api, player_level, forts, step_location, account):
     if player_level > 1:
         log.debug(
             'No need to spin a Pokestop. ' +
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1766,7 +1766,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     skipped = 0
     stopsskipped = 0
     forts = []
+    forts_count = 0
     wild_pokemon = []
+    wild_pokemon_count = 0
     nearby_pokemon = 0
     spawn_points = {}
     scan_spawn_points = {}
@@ -1794,9 +1796,20 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # necessarily need to know *how many* forts/wild/nearby were found but
         # we'd like to know whether or not *any* were found to help determine
         # if a scan was actually bad.
-        wild_pokemon += cell.get('wild_pokemons', [])
+        if config['parse_pokemon']:
+            wild_pokemon += cell.get('wild_pokemons', [])
+        else:
+            wild_pokemon_count += len(cell.get('wild_pokemons', []))
 
-        forts += cell.get('forts', [])
+        if config['parse_pokestops'] or config['parse_gyms']:
+            forts += cell.get('forts', [])
+        else:
+            forts_count += len(cell.get('forts', []))
+
+    if wild_pokemon:
+        wild_pokemon_count = len(wild_pokemon)
+    if forts:
+        forts_count = len(forts)
 
     del map_dict['responses']['GET_MAP_OBJECTS']
 
@@ -1988,7 +2001,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 })
                 wh_update_queue.put(('pokemon', wh_poke))
 
-    wild_pokemon = len(wild_pokemon)
+        del wild_pokemon
 
     if forts and (config['parse_pokestops'] or config['parse_gyms']):
         if config['parse_pokestops']:
@@ -2102,7 +2115,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         f['last_modified_timestamp_ms'] / 1000.0),
                 }
 
-        forts = len(forts)
+        del forts
 
     log.info('Parsing found Pokemon: %d, nearby: %d, pokestops: %d, gyms: %d.',
              len(pokemon) + skipped,
@@ -2170,7 +2183,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # After parsing the forts, we'll mark this scan as bad due to
         # a possible speed violation.
         return {
-            'count': wild_pokemon + forts,
+            'count': wild_pokemon_count + forts_count,
             'gyms': gyms,
             'sp_id_list': sp_id_list,
             'bad_scan': True,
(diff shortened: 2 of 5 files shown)


springjools:
Nice! I can see from this documentation that you have done a thorough work.

jagauthier:
I am testing this out, After running for about 10 minutes, I am seeing this:

```
Map parse failed at x,y, abandoning location. username may be banned.. Exception message: TypeError("unsupported operand type(s) for +: 'int' and 'list'",)
  File "/home/jgauthier/PokemonGo-Map/pogom/search.py", line 991, in search_worker_thread
    dbq, whq, api, scan_date, account)
  File "/home/jgauthier/PokemonGo-Map/pogom/models.py", line 2117, in parse_map
    'count': wild_pokemon + forts,
TypeError: unsupported operand type(s) for +: 'int' and 'list'
```

I keep all my instances logs. I went back through, and this has never occurred before.
It's fairly frequent on my instance with 150 workers.

The offending code looks like it was changed:
```
-        'count': len(wild_pokemon) + len(forts),
+        'count': wild_pokemon + forts,
```

I confirmed the other PR I am testing does not have any other changes with these variables either.

friscoMad:
@jagauthier Thanks for such detailed report, it had an error when no forts (gyms or pokestops) where parsed.

I did change the logic a bit to make it clear when we are using a count or a list and also avoid generating lists that we are not going to use.

MARTINI1:
I'm trying it from yesterday and for the moment I notice the performance of the map better but it keeps going up and up the memory until I'm near the end

On the other hand, yesterday before turning it off I looked for the status and I found that the map was left in this loop error:

abr 01 01:29:34 raspberrypi python2.7[794]: 2017-04-01 01:29:34,882 [      db-updater-6][        models][ WARNING] OperationalError(2002, "Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)")... Retrying...
abr 01 01:29:34 raspberrypi python2.7[794]: 2017-04-01 01:29:34,891 [      db-updater-7][        models][ WARNING] OperationalError(2002, "Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)")... Retrying...
abr 01 01:29:34 raspberrypi python2.7[794]: 2017-04-01 01:29:34,894 [      db-updater-1][        models][ WARNING] OperationalError(2002, "Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)")... Retrying...

Right now was making mysql verifications and scares a little that consumes 45% of the ram xD

Well, I will continue testing and see if it is with who is causing so much memory leak

Nice PR @friscoMad ;)

jagauthier:
RM is very resilient when it comes to SQL.So, if your DB can't be connected to, it queues up the requests. Those requests will chew up memory.    I am still running this, and will report after several hours of use.

MARTINI1:
I forgot to say, today I used the rocketmap without updating with the latest update in this PR and I did not get the mysql error that said, it might have been something punctual because I had never seen it

jagauthier:
Before (unsorted, sorry)
image
After (sorted):
image


There does seem to be a memory decrease in all instances. The larger of them goes from 18.2% to 14.9%, which is an 18% decrease.

friscoMad:
@MARTINI1 as jagauthier said this PR can not do anything with the DB or WH queues they are unbounded so any error or stagnation in wh or db threads will make those thread take all the memory available, I am thinking on fix that in another PR but if your logs show a very high DB queue then thats the reason of the memory usage. 
In my setup right now I have 5000 elements in queue and that makes almons 100Mb of extra mem usage.

MARTINI1:
@friscoMad ok no problem 👍 

Now I have optimized the rpi and mysql and with the map (40 workers and st 23 ) with this PR spends mysql 335 MB RAM and 2.4% CPU and rocketmap 132 MB RAM and 69% CPU, fluctuating up and down

Not bad, it tempts me to put more workers xD

captura de pantalla 2017-04-02 a las 16 25 27

I need to do this same analysis with the original branch of the rocketmap to see what has improved as it has done @jagauthier to be able to compare better

MARTINI1:
Here I bring my comparative of the current branch develop and this pr, just when the 40 workers are connected and begin to work on the map

current branch
captura de pantalla 2017-04-05 a las 11 52 26

this pr
captura de pantalla 2017-04-05 a las 12 18 31

In the long run, memory and cpu is more stable using this pr and I've been testing it since the day it came out and it has not given me any problems

PokemonCatcher:
Needs minor update after merge of #1900.

friscoMad:
@PokemonCatcher It is already rebased and have been working in my instance for several hours without issues.

PR #1962, opened 2017-04-01
Ignore specific pokemon (do not store to DB)

jagauthier:
The idea is to completely ignore 'trash' pokemon as much as we can.  This code will basically strip
out the pokemon that come from the API, and pretend they were never found

## Description
In parse_map, there are two calls that get pokemon from the virtual environment.
`cell.get('nearby_pokemons', [])`
and
`cell.get('wild_pokemons', [])`

wild_pokemon will contain all the found pokemon, that have spawned in the workers range.
Typically, what happens after that is each pokemon is queried in the database to see if it's already been found (because the area can be passed over multiple times), they are encountered, and inserted into the database, and then pushed to webhooks (if all that configured).  There is a PR to not send specific pokemon to webhooks, and this takes that one step further by not even allowing these pokemon in the database (and they won't be encountered or sent to webhooks either)

## Motivation and Context
Why do people run maps?  One of the major reasons is to find pokemon.  But, we don't need to search and find pokemon that are easily found. With this enhancement, it will allow us to completely (as possible) ignore specific pokemon. This will reduce DB size, load, and overall performance.  As an example, I purge pokemon once per week.  My current DB has the following records:
```
mysql> SELECT pokemon_data.pokemon_id,pokemon_name,count(pokemon.pokemon_id) as count  FROM pokemon,pokemon_data  WHERE pokemon.pokemon_id=pokemon_data.pokemon_id   GROUP BY pokemon.pokemon_id order by count desc limit 10;
+------------+--------------+-------+
| pokemon_id | pokemon_name | count |
+------------+--------------+-------+
|         16 | pidgey       | 62781 |
|        167 | spinarak     | 57792 |
|         19 | rattata      | 54500 |
|        198 | murkrow      | 45687 |
|        129 | magikarp     | 44188 |
|        161 | sentret      | 41556 |
|        165 | ledyba       | 41276 |
|        163 | hoothoot     | 38190 |
|        177 | natu         | 31643 |
|         13 | weedle       | 28162 |
+------------+--------------+-------+
```
The last 3 hours alone reveals this many:
```
mysql> SELECT pokemon_data.pokemon_id,pokemon_name,count(pokemon.pokemon_id) as count  FROM pokemon,pokemon_data  WHERE disappear_time > DATE_SUB(UTC_TIMESTAMP(), INTERVAL 3 HOUR) and pokemon.pokemon_id=pokemon_data.pokemon_id   GROUP BY pokemon.pokemon_id order by count desc limit 10;
+------------+--------------+-------+
| pokemon_id | pokemon_name | count |
+------------+--------------+-------+
|         16 | pidgey       |  2587 |
|        167 | spinarak     |  2339 |
|        161 | sentret      |  2303 |
|         19 | rattata      |  2191 |
|        198 | murkrow      |  1342 |
|        177 | natu         |  1270 |
|        165 | ledyba       |  1232 |
|         13 | weedle       |  1226 |
|         48 | venonat      |  1096 |
|        133 | eevee        |   786 |
+------------+--------------+-------+
```

For me, this is a lot of unnecessary DB action. Not only at backend, but also many of these rows are being returned by the frontend.  The frontend itself does not limit the DB queries, it just filters based on the exclude.  

The goal is to reduce DB overheard in both the backend and frontend by eliminating thousands of 'trash' pokemon records.

## How Has This Been Tested?
In a clean database, I ran an -st 4 at a local park which is famous for spawning commons.  
I included a large list of junk pokemon in the config file:
`ignore-list: [10,11,12,13,14,15,16,17,18,19,20,21,22,46,47,48,49,161,162,163,164,165,166,167,168,198,177,178,187,188,189]`

You can see some are found, and ignored.  
The found pokemon shows as 0, as there were none that were not ignored in this request.

```
2017-04-01 16:15:16,957 [   search-worker-1][        models][   DEBUG] Ignoring pokemon id: 161
2017-04-01 16:15:16,958 [   search-worker-1][        models][   DEBUG] Ignoring pokemon id: 165
2017-04-01 16:15:16,958 [   search-worker-1][        models][   DEBUG] Ignoring pokemon id: 187
2017-04-01 16:15:16,964 [   search-worker-1][        models][    INFO] Parsing found Pokemon: 0, nearby: 32, pokestops: 29, gyms: 3.
```

In this example, several pokemon are discovered, and 4 of the ignored ones are filtered out.
One of these was on on the list, and it was reported that 1 was found.
```

2017-04-01 16:19:59,025 [   search-worker-3][        models][   DEBUG] Ignoring pokemon id: 187
2017-04-01 16:19:59,025 [   search-worker-3][        models][   DEBUG] Ignoring pokemon id: 167
2017-04-01 16:19:59,025 [   search-worker-3][        models][   DEBUG] Ignoring pokemon id: 177
2017-04-01 16:19:59,025 [   search-worker-3][        models][   DEBUG] Ignoring pokemon id: 177
2017-04-01 16:19:59,033 [   search-worker-3][        models][    INFO] Parsing found Pokemon: 1, nearby: 33, pokestops: 29, gyms: 3.
```

After running for awhile:
```
MariaDB [rocketmapdb]> SELECT pokemon_data.pokemon_id,pokemon_name,count(pokemon.pokemon_id) as count  FROM pokemon,pokemon_data  WHERE pokemon.pokemon_id=pokemon_data.pokemon_id   GROUP BY pokemon.pokemon_id order by count desc;
+------------+--------------+-------+
| pokemon_id | pokemon_name | count |
+------------+--------------+-------+
|         69 | bellsprout   |     3 |
|         32 | nidoran-m    |     2 |
|         23 | ekans        |     1 |
|        220 | swinub       |     1 |
|        116 | horsea       |     1 |
|        133 | eevee        |     1 |
|         43 | oddish       |     1 |
|        191 | sunkern      |     1 |
|        209 | snubbull     |     1 |
|         90 | shellder     |     1 |
|        216 | teddiursa    |     1 |
|        102 | exeggcute    |     1 |
+------------+--------------+-------+
12 rows in set (0.00 sec)
```

image

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -1710,15 +1710,22 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # necessarily need to know *how many* forts/wild/nearby were found but
         # we'd like to know whether or not *any* were found to help determine
         # if a scan was actually bad.
-        wild_pokemon += cell.get('wild_pokemons', [])
-
+        temp_wild_pokemon = cell.get('wild_pokemons', [])
+        # go through this and determine if any of these are ignorable
+        # we'll rewrite the list without them, and the subsequent code
+        # will never know they were present.
+        if temp_wild_pokemon:
```
Arengo: A temporary list is unneeded when you directly ``list.remove(element)`` from actual list. Also you should not only address the wild Pokémon but also those in nearby to keep consistency - maybe we will use these nearby Pokémon one day.

jagauthier: Can you explain how you would achieve that considering list.remove(element) will only remove a single element when it's invoked?  In the case there are 5 pidgies that spawn, then there is considerably extra logic to remove all 5 pidgies.

I considered multiple ways to do this, based on your suggestion, but I am not happy with any of them.  Your input is welcomed.

I considered removing from nearby_pokemon, but I decided that was a bad idea right now.
Since it's not used yet, it doesn't cause any extra DB load.  But a few line down there is some logic that uses it:

 ```
   if not wild_pokemon and not nearby_pokemon:
        # . . . and there are no gyms/pokestops then it's unusable/bad.
        abandon_loc = False
        if not forts:
            log.warning('Bad scan. Parsing found absolutely nothing.')
            log.info('Common causes: captchas or IP bans.')
            abandon_loc = True
        else:
            # No wild or nearby Pokemon but there are forts.  It's probably
            # a speed violation.
            log.warning('No nearby or wild Pokemon but there are visible gyms '
                        'or pokestops. Possible speed violation.')

```

Leaving nearby reduces the risk of the location being abandoned. We definitely don't want that.
And, we don't want a bunch of errors "No nearby or wild Pokemon but there are visible gyms                         or pokestops. Possible speed violation", because we already see a lot of those questions in #help

Comment (friscoMad) [inline pogom/models.py]: Yup probably you can skip the if and the var at the same time, just iterate over `cell.get('wild_pokemons', [])`.
I don't think nearby will be used any time, so I think it is good as is.

File: pogom/utils.py [hunk 2]
```diff
@@ -169,6 +169,13 @@ def get_args():
                                 default='', help='File containing a list of '
                                                  'Pokemon to NOT encounter for'
                                                  ' more stats.')
+    ignore_list = parser.add_mutually_exclusive_group()
+    ignore_list.add_argument('-ign', '--ignore-list',
+                             action='append', default=[],
+                             help=('List of Pokemon to ignore.'))
+    ignore_list.add_argument('-ignf', '--ignore-list-file',
```
friscoMad: I am not sure if this is needed you can use the config file to "free" the command line anyway

jagauthier: I'm not sure I follow you.

friscoMad: To me the "--ignore-list-file" option to me it is almost the same of using the general "--config-file" option to set the ignore list. 
There are a hell of a lot command line options and I don't see the point of adding 2 more when 1 is enough.

pogom/utils.py
```diff
@@ -644,6 +651,13 @@ def get_args():
                                         args.encounter_blacklist]
             args.encounter_whitelist = [int(i) for i in
                                         args.encounter_whitelist]
+        if args.ignore_list_file:
+            with open(args.ignore_file) as f:
+                args.ignore_list = [get_pokemon_id(name) for name in
```
friscoMad: Use a set instead of a list, as order does not matter and `in` checks are a lot faster.

pogom/models.py
Arengo: Nearby could be used to directly math them to Pokéstop where they are attached. However, regarding this PR nearby in general is not needed but the check you mentioned deeper in the code should be made before anything else imo to detect wrong scans. Also filtering the nearby does not have to be done since it uses additional ressources but there could be out-commented code which addresses it, when it will be used one day. You can remove them from the list based on there encounter IDs like this pseudo python code (watch out for actually needed structure)
```python
for p in wild_pokemon:
    if p['pokemon_id'] in exclude_list:
        wild_pokemon.remove(p)```

File: pogom/utils.py (hunk 2 above)
Comment (Arengo) [inline pogom/utils.py]: Regarding consistency in comparison to ``encounter`` or ``proxy`` a additional command for a ``ignore`` file is correct. However, by myself I do always only use files for everything because re-usability. Other people tend to not use files at all. It's a question of personal taste.

File: pogom/models.py (hunk 1 above)
Comment (Arengo) [inline pogom/models.py]: Thanks for pointing this out, you are right. Remove only deletes the first occurence and list indices change of course. However, imo there still is a way via pseudo-code
```python
for index in range(0, len(list)):
    remove(list[index]) if list[index] ['pokemon_id'] in exclude_list
    index = index - 1
```

Holzberg:
If all the Pokemon found in a scan are filtered and it says 0 Pokemon found, does RM see that as an empty scan?  If it does, then that is an issue.

jagauthier:
It is an issue, which is why I fixed that (https://github.com/RocketMap/RocketMap/pull/1962/commits/4b96a6775170802e7869c1fe43a4bafe5aa406c8)

It will always "find" them now, but not insert into DB/webhook.

tomballgithub:
Can you resolve the conflicts, I would like to review/test this?

pogom/utils.py
```diff
@@ -713,6 +724,14 @@ def get_args():
                                       args.webhook_blacklist]
             args.webhook_whitelist = [int(i) for i in
                                       args.webhook_whitelist]
+        if args.ignore_list_file:
+            with open(args.ignore_file) as f:
```
cliff-svt: I think this should be "args.ignore_list_file" instead of "args.ignore_file".

jagauthier: You reviewed this as I was making updates. :) Incidentally, I agreed with you.

pogom/models.py
```diff
@@ -1932,11 +1932,20 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
                 timedelta(seconds=seconds_until_despawn)
 
-            printPokemon(p['pokemon_data']['pokemon_id'], p[
+            # Store this and reuse
+            pokemon_id = p['pokemon_data']['pokemon_id']
+
+            # if this is an ignored pokemon, skip this whole section
+            # We want the stuff above or we will impact spawn detection
+            # but we don't want to insert it, or send it to webhooks
+            if pokemon_id in args.ignorelist:
```
cliff-svt: I don't think args.ignorelist exists unless args.ignorelist_file is provided (it's only set within a conditional block).

pogom/utils.py
```diff
@@ -160,6 +160,14 @@ def get_args():
                         help=('Time delay between encounter pokemon ' +
                               'in scan threads.'),
                         type=float, default=1)
+    ignore_list = parser.add_mutually_exclusive_group()
```
cliff-svt: Do you still need this mutually exclusive group?

pogom/models.py
```diff
@@ -1932,11 +1932,20 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
                 timedelta(seconds=seconds_until_despawn)
 
-            printPokemon(p['pokemon_data']['pokemon_id'], p[
+            # Store this and reuse
+            pokemon_id = p['pokemon_data']['pokemon_id']
+
+            # if this is an ignored pokemon, skip this whole section
+            # We want the stuff above or we will impact spawn detection
+            # but we don't want to insert it, or send it to webhooks
+            if args.ignorelist_file and (pokemon_id in args.ignorelist):
```
friscoMad: Maybe we could merge this condition with the condition on 1924 or at least before 1930 as it does seem that calcs are not needed (not sure thought).

Acadia1337:
Are both blacklist and whitelist supported with this? Also are blacklist and whitelist file supported?

jagauthier:
@Dcstegg228 I think you've missed the objective of this modification if you're asking those questions.
No, there is no "whitelist/blacklist".  It's a blacklist, of sorts.  The pokemon specified will not be saved or send to webhooks.   Many people don't want to store commons. This is what this idea targets.
It only supports file, as that is the new methodology for pokemon list items.

Acadia1337:
Ummm... no I'm not missing the objective of this. I've used a pr like this before back on RocketMap. Why not have a whitelist in addition to the blacklist like every other blacklist feature? Instead of blacklisting 200 pokemon to be ignored why not create a whitelist of the only 10 that I want to be stored? I think this makes a great deal of sense to implement and it will follow the blacklist/whitelist style of the rest of the features in RocketMap. My users only care about maybe 20 of the total pokemon that exist. A whitelist would be perfect.

docs/first-run/commandline.md
```diff
@@ -52,326 +53,330 @@
...
+                          Maximum time captchas will wait for manual captcha
+                          solving. On timeout, if enabled, 2Captcha will be used
+                          to solve captcha. Default is 0. [env var:
+                          POGOMAP_MANUAL_CAPTCHA_TIMEOUT]
     -ed ENCOUNTER_DELAY, --encounter-delay ENCOUNTER_DELAY
-                        Time delay between encounter pokemon in scan threads.
-                        [env var: POGOMAP_ENCOUNTER_DELAY]
+                          Time delay between encounter pokemon in scan threads.
+                          [env var: POGOMAP_ENCOUNTER_DELAY]
+    -ignf IGNORE_LIST_FILE, --ignore-list-file IGNORE_LIST_FILE
```
tomballgithub: Please document the format of this file here.  Most/all other files referenced here have an example or description of file format.

pogom/utils.py
```diff
@@ -160,6 +160,13 @@ def get_args():
                         help=('Time delay between encounter pokemon ' +
                               'in scan threads.'),
                         type=float, default=1)
+    parser.add_argument('-ignf', '--ignorelist-file',
+                        default='', help='File containing a list of '
+                        'Pokemon IDs to ignore. Pokemon will ' +
```
tomballgithub: Please document the format of this file here. Most/all other files referenced here have an example or description of file format.

tomballgithub:
I found the '0 pokemon found' messaging confusing until I saw you had documented it as intent and others will find it confusing also.  It will send them into #help all day long.

**2017-04-01 16:15:16,964 [   search-worker-1][        models][    INFO] Parsing found Pokemon: 0, nearby: 32, pokestops: 29, gyms: 3.**

Is there a way to make the printed message have more clarity?  For example, by adding an 'filtered: XXX' to the below line.

**2017-04-01 16:15:16,964 [   search-worker-1][        models][    INFO] Parsing found Pokemon: 0 [14 filtered], nearby: 32, pokestops: 29, gyms: 3.**

or maybe

**2017-04-01 16:15:16,964 [   search-worker-1][        models][    INFO] Parsing found Pokemon: 0, filtered: 14, nearby: 32, pokestops: 29, gyms: 3.**

Anything would be an improvement...  I'd prefer to _avoid_ just 'documenting' it because again I think it will invariably cause questions in #help.  And if real issues come up, how do you differentiate a '0' from this feature or a true '0'?

jagauthier:
I like your suggestion.  I will add that in the next few days.

Acadia1337:
I have been testing this for the past week on one of my maps thato is scanning 4-5million pokemon per day. It works as intended and I like it a lot. I would however like to note a few things. This does not reduce database load by an substantial ammount. The #of queries per second is about the same before and after this PR. It does however lower the ram requirements as the pokemon table is substantially smaller. I also noticed increased front end speeds due to the smaller tables.

Acadia1337:
Also, a whitelist would be amazing as my blacklist is gigantic but my whitelist would be much much smaller and I could use the same whitelist as the encounter whitelist.

tomballgithub:
Nice improvement on the log as requested.

```
> 2017-05-19 22:12:54,893 [   search-worker-0][        models][    INFO] Parsing found Pokemon: 0 (6 filtered), nearby: 17, pokestops: 5, gyms: 2.
> 2017-05-19 22:12:54,913 [   search-worker-0][        search][    INFO] Search at XXX completed with 13 finds. Sleeping 9.967689s until 22:13:04.
> 
```

tomballgithub:
I have been using this for 2 months without issue or complaint.

SiteEffect:
Same here.

AEtHeLsYn:
Using this PR for a month without complain, working great

VYRNACH:
would like to add that I have been using this since May without any issues

tomballgithub:
I think in general for RM there are many different types of input files and it is not really documented what the format of the input file is.  The one says a list of IDs so it's logical to try a list of numeric IDs one per line but someone might try to use a list of names or all on one line, etc.  Food for thought in future for better documentation in general.

The code here looks good, I know the code works, the commandline.md is updated, and the config.ini.example is updated.

I approve +1

pogom/models.py
```diff
@@ -1996,6 +1997,15 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 timedelta(seconds=seconds_until_despawn)
 
             pokemon_id = p.pokemon_data.pokemon_id
+
+            # if this is an ignored pokemon, skip this whole section
+            # We want the stuff above or we will impact spawn detection
+            # but we don't want to insert it, or send it to webhooks
```
sebastienvercammen: Capitalization and punctuation.

pogom/models.py
```diff
@@ -1996,6 +1997,15 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+            if args.ignorelist_file and (pokemon_id in args.ignorelist):
```
sebastienvercammen: Unnecessary parentheses.

pogom/models.py
```diff
@@ -1996,6 +1997,15 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                log.debug("Ignoring Pokemon id: %i", pokemon_id)
```
sebastienvercammen: Missing dot. Use single quotes `'` for consistency.

pogom/models.py
```diff
@@ -2399,9 +2409,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # Helping out the GC.
         del forts
 
-    log.info('Parsing found Pokemon: %d, nearby: %d, pokestops: %d,' +
-             ' gyms: %d, raids: %d.',
-             len(pokemon) + skipped,
+    log.info('Parsing found Pokemon: %d (%d filtered), nearby: %d, ' +
+             'pokestops: %d, gyms: %d, raids: %d.',
+             len(pokemon) + skipped, filtered,
```
sebastienvercammen: Move `filtered` to a separate line below for readability. It was originally done like that intentionally. Watch code style.

pogom/utils.py
```diff
@@ -156,10 +156,17 @@ def get_args():
...
+                        'not be added to DB, not sent to ' +
+                        'webhooks, and not encountered. ' +
+                        'Will still be used to ' +
+                        'determine spawnpoints. One line per ID.')
```
sebastienvercammen: Rephrase to make it easier to read:

> File containing a list of Pokemon IDs to ignore, one line per ID. Spawnpoints will be saved, but ignored Pokemon won't be encountered, sent to webhooks or saved to DB.

pogom/utils.py
```diff
@@ -156,10 +156,17 @@ def get_args():
...
     parser.add_argument('-encwf', '--enc-whitelist-file',
                         default='', help='File containing a list of '
                         'Pokemon IDs to encounter for'
-                        ' IV/CP scanning.')
+                        ' IV/CP scanning. One line per ID')
```
sebastienvercammen: Missing dot.

PR #1977, opened 2017-04-18
Snap to grid

rocketrobot99:
# Snap to grid of cells and functionality to clear all existing cells.
If you are intending to use this new functionality it is advised to clear all existing cells first (-clc). Please note that this will cause a re-scan of bands, but not TTHs.

Usage: -stg <longitude reference>.
When you use the -stg param you now have to supply a longitude reference, where all measurements are taken from. So typically if you are scanning 39.656726, -84.223432, then I would use -84 as the reference point. In my case, as I am scanning near to London (51.559628, -1.781075) I will just use 0.

## Description
The main new functionality here is the ability to have all cells locations conform to exact multiples of the width and height of the cells. This essentially means the following:
1. If you move your map a tiny distance to the left, the location will 'snap' back to where it was. It will not initiate a re-scan or create any new cells.
2. If you decide to move your large -st map a bit to the left, it will not generate hundreds of new cells and initiate a re-scan. Rather it will just create new cells for any new unscanned areas.
3. Similar to above, if you decide to run 7x -st 20 instead of one big -st 40, then the cells will be re-used.

A consequence to this is that some users have been using the 'Move a tiny distance' to force a rescan. This was not the best way to achieve this, as it creates loads of new cells. With this in mind, I have added an option to clear all the existing cells (scannedlocations) from the database. This will force new cells to be created an initial rescan to occur. However, all TTHs will be preserved.

## Motivation and Context
If the base location of the Rocketmap changed even slightly, then this not only produce a whole new set of cells, it would also cause an initial re-scan. 

## How Has This Been Tested?
The snap to location functionality has been tested on an RM with -st 50. Firstly letting the map build overnight then changing then moving the base location. Each time monitoring the cells on the db to ensure only the correct new cells are created.

pic1
The initial load
pic2
Count of cells.
pic3
Log after moving the grid, note the scan is 95% complete as the existing cells are re-used
pic4
The new count of cells, with 299 added
pic5
Representation of the new cells created.

## Types of changes

## Checklist:
- [x ] I have updated the documentation accordingly.


sebastienvercammen:
I took a look at why Travis is failing, and you can ignore this for now. It's unrelated to your code; I've contacted the author of the dependency that's causing it, it'll be fixed soon.

jagauthier:
General comment on testing.  I set up an a small area, and scanned it to 100%.  I then changed the location slightly, and also changed the step size.  The scan picked up and continued as if nothing changed.

I did notice the scanned area representation shows an overlap.
image

rocketrobot99:
@jagauthier Hi, I will try and contact you on discord for your Map setup. This does not look right at all. I will also try and bring it up to date with the latest Dev.

rocketrobot99:
@jagauthier Can you contact me on discord (rocketrobot). I want to test out your location etc.. as your map does not look right.... The circles should meet at one point, with no gaps:

image

I have pushed a more up to date commit.

Arengo:
@rocketrobot99 why should they meet up when you are jittering all the locations?

rocketrobot99:
@Arengo As far as I can see, Jitter is only applied when the call to GET_MAP_OBJECTS is done. The scannedlocations are fixed once generated at the start. The point being that you don't want to leave any gaps. Anyway @jagauthier I have managed to recreate the problem, so I will have a look later:
image
It is supposed to look like this (but it doesn't, yet):
image

rocketrobot99:
@jagauthier and @Arengo I have pushed a new commit that should fix the problems above. When you use the -stg param you now have to supply a longitude reference, where all measurements are taken from. So typically if you are scanning 39.656726, -84.223432, then I would use -84 as the reference point. In my case, as I am scanning near to London (51.559628, -1.781075) I will just use 0. The result below is from the updated version:
image

AEtHeLsYn:
@rocketrobot99 Wouldn't be better to set a fixed longitude reference instead of setting on command line? If by mistake you set a different -stg argument in two different instances the commit would be useless as we are using two references.

rocketrobot99:
@AEtHeLsYn The problem is, what do you set it to? It doesn't have to be that close. e.g for the whole of the UK you could use longitude 0. But you can't use 0 when you are scanning USA, as you start to get noticeable changes in the distance when you move just 70m north or south (this is what was happening above).  I did think about using the nearest whole number, i.e. If you are scanning 39.344, -84.530 then use -85 as the reference. That would also cause problems though for someone if they then created an instance a few miles West, say 39.344, -84.495, then the nearest whole number would be -84. Then things would look really odd for them where the 2 instances meet. This is why I chose to let the user decide.

Alderon86:
@rocketrobot99 where do you round things? im not quite getting why -1.781075 refers to 0 (isnt -1 closer to that?) and -84.2 refers to -84.

friscoMad:
As I understand the need for rounded latitude comes from the vicenty algorithm, maybe switching to haversine even with less precission could fix the issue, otherwise maybe you can check https://geographiclib.sourceforge.io/html/python/code.html#module-geographiclib that uses karney algo that seems to fix some issues in vicenty algorithm.
I don't see why there is such deviation in latitude and almost no deviation in longitude as it should have the same problem.

rocketrobot99:
@Alderon86 Actually -2 is closer to -1.78 than -1. However, the point is it doesn't matter, you just need to choose some fixed frame of reference that is fairly close. The (excluding Ireland) goes from longitude -5.6 to +2, roughly. So choosing longitude 0 will be fine for all maps within the UK.

rocketrobot99:
@friscoMad I am not exactly sure what you mean. As far as I know there is nothing wrong with vincenty. The point of this point of this PR is to have a kind of 'snap to grid' functionality so that users can play around with different sizes and numbers of beehives without having to constantly create new cells and rescan all the bands. 

The idea is that for the area you are going to cover, there is already a predefined grid of cells. I bit like graph paper, but with circles. Now the need for the user to specify a longitude point as a point of reference is just because it is impossible to define one ourselves. I originally tried using longitude 0, and this works perfectly for the UK, however not so much for USA. The reason is as follows:

At latitude 40, longitude -84.45 the distance to longitude 0 is 6907.914 km. however, if we move up just 70m to latitude 40.00063, the distance to longitude 0 is now 6907.879 km. The difference is 35m! This is not a problem with Vincenty, it is because the earth is a sphere. This distance will continue to get smaller and smaller until you reach the north pole where it will be 0.

Now the same example using longitude -84 as our point of reference:
At latitude 40, longitude -84.45 the distance to longitude -84 is 38.42719 km. however, if we move up just 70m to latitude 40.00063, the distance to longitude -84 is now 38.42708 km. A difference of only 11cm.

I hope this explains it and I understood your comment correctly. If not, please comment again.

friscoMad:
Thanks for the explantaion @rocketrobot99 
I think I understand the whole algorithm you are using and I think that the problem is not solved just that using a nearer reference does minimize the error but it is still there.

What I think it is wrong or can be improved:

- It is always trying to snap to the previous cell not the nearest one (previous = the nearest one with lat < original_lat , long < original_long) not a big issue but it will displace at max step meters from supplied point instead of 1/2 step.
- After getting the new y we are discarding the new x, while getting the new x we take both, that should help converging faster.
- x offset is being recalculated in every pass but only for even y if while itereating we switch from even to odd y the wrong offset would be applied (if it is impossible to happen then there is no need to calculate it for every iteration)
 - x offset is being divided alongside with long dist, so the offset really applied is lower as the long increases (that seems to be the long problem to me, and explains why we don't see the same issue for the lat), it should be applied after the divission, just adding half a step when the y is even, maybe it could even be done after the loop.

All of this is my understanding and purelly theoretical so probably I am missing something.

rocketrobot99:
@friscoMad Yes, indeed, the problem is not solved, but the near reference makes the error minimal.

Thanks for your comments: 
"It is always trying to snap to the previous cell not the nearest one (previous = the nearest one with lat < original_lat , long < original_long) not a big issue but it will displace at max step meters from supplied point instead of 1/2 step." - Yes, this is because with large -STs of 80 or more it becomes a problem. If we calculated one of the outer points of our -ST 80 Hex, then snapped it to a cell, it might not snap to the correct cell, simply because of this adjusting we have to do. It is rare, and it also depends on the origin location and its relation to the center of the snapped cell grid. e.g. 

If the original location lies right between 2 cells then this location will be snapped to one or other of those cells, the hex will then be built up one cell at a time until we get to the outer ring. There will be no gaps. However, if we were to take our original point and calculate a point 80 cells away on a bearing of 30 degrees. Then the calculated cell location, when snapped to the grid may end up snapping to a location on the 79th,80th or 81st ring due to any tiny difference in calculations.

"After getting the new y we are discarding the new x, while getting the new x we take both, that should help converging faster." - It makes no difference, the new Loc_y is an adjustment 180 degrees south, so the new x stays the same. I did, of course, try it to make sure and the iterations were identical. FYI a lot of the time the loop only runs once, because we are moving an exact cell distance each time from the previous 'snapped' cell:

image


"x offset is being recalculated in every pass but only for even y if while itereating we switch from even to odd y the wrong offset would be applied (if it is impossible to happen then there is no need to calculate it for every iteration)" - Thanks, I have changed the code.

"x offset is being divided alongside with long dist, so the offset really applied is lower as the long increases (that seems to be the long problem to me, and explains why we don't see the same issue for the lat), it should be applied after the divission, just adding half a step when the y is even, maybe it could even be done after the loop." - I have thought about this and it would be possible to take this out of loop, however this would require another call to Vincenty, do determine the longitude adjustment (as it varies depending on the latitude). As the loop seems to only iterate 2 times on average I don't think that the call is needed. And (a+b)/c uses the same CPU as (a/b)+c. In fact, taking it outside the loop may actually cause harm, as the locations are already adjusted by 1/2 a cell when they are sent to the function. If we were to take it out, for the even rows of cells, the location would have to be snapped to the grid (1/2 a cell away) and then adjusted by 1/2 cell back. 

I did a run to see if my assumption was right and the average iterations went up from 2.05 to 2.34. So I don't think a change is needed there.
image

Arengo:
I had issues with distortioned hexes with .5° difference to the reference.

rocketrobot99:
@Arengo Any chance of getting a bit more detail, please. Location and reference, the size of hex etc..

friscoMad:
I did some more test yesterday, I did not get any improvement in the results but I thought that after all there is no need to snap every cell as long as you snap the center position, snapping every cell is better but not sure if the difference is worth the issue.
Aside for that I did thought that there is no need for the offset, just snap to half xdist, the cells will naturally go to the nearest point that will be offseted or not depending on the initial location provided the downside is that depending on how far you are moving all cells will be snapped half cell to the side (so will need to rescan everything).
Probably you have already thought on this but maybe not.

pogom/transform.py
```diff
@@ -71,6 +73,145 @@ def get_new_coords(init_loc, distance, bearing):
...
+    while x_diff != prev_x_diff and x_diff > 1e-12 and loop_counter < 20:
+        prev_x_diff = x_diff
+        num_ysteps = vincenty((0, loc[1]), (loc[0], loc[1])).km / ydist
+        y_diff = num_ysteps - float("{0:.0f}".format(num_ysteps))
+        # If the y multiple is an even number then we must calculate the
+        # x multiple to the nearest half number. If we imagine the gird, as we
+        # move up the cell's offset alternates between 0 and a half of a cell.
+        loc_y = get_new_coords((loc), ydist*y_diff, 180)
+        if float("{0:.0f}".format(num_ysteps)) % 2 == 0:
+            num_xsteps = (vincenty((loc_y[0], snap_to_grid), (loc_y[0], loc[1])
```
friscoMad: The last part should be (loc_y[0], loc_y[1])

pogom/transform.py
```diff
@@ -71,6 +73,145 @@ def get_new_coords(init_loc, distance, bearing):
...
+                                   ).km + x_offset) / xdist
+        else:
+            num_xsteps = vincenty((loc_y[0], snap_to_grid), (loc_y[0], loc[1])
```
friscoMad: Same as above

PR #1978, opened 2017-04-20
Switch control logic to prevent unintended control from outside

Arengo:
## Description

This PR changes how the arguments ``fixed-location`` and ``no-search-control`` work in their logic by switching from positive and negative logic to the other as default while maintaining a good to read code (where those arguments are used in ``pogom/app.py``) by usage of the arguments parse parameter ``dest``. 
Hereby, additional functionality is enabled by setting these arguments instead of forcing beginners to add arguments which they do not consider at the first glance of this program and are mostly compromising easy setups.

The new arguments are now:
- ``-sc`` or ``--search-control``
- ``-nfl`` or ``--no-fixed-location`` (yeah, US-Football fans love me, now)

## Motivation and Context

From the very beginning of my own map and recent requests I am bugged that, by default, people can change your set-up location. Especially with -speed this is a not-wanted feature by default. Also the search location feature, intended for distributed big shared maps, is not a default behaviour for the standard use-case of a neighbourhood map.

## How Has This Been Tested?

On my local machine. Tests from you guys are welcomed, who still use on demand hex on variable locations or general a non-fixed map.

## Types of changes

- [x] Change default configuration to the better

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -112,7 +112,7 @@
 #only-server                    # Server-Only Mode. Starts only the Webserver without the searcher.
 #locale:                        # Pokemon translation
 #search-control                 # Enables search control.
-#no-fixed-location              # Shows the search bar for use in shared maps.
+#no-fixed-location              # Disables the fixed map location and shows the search bar for use in shared maps.
 #cors                           # Enable CORS on web server.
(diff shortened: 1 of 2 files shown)


Alderon86:
So if i dont run -sc its still the same like using -fl + -nsc?

PR #1980, opened 2017-04-21
Add support for IV scanning and CP scanning

krosenvold:
Added two additional account pools, for a total of three pools. One regular pool, one for level 25 iv scanners and another for level 30 cp/iv scanners.

Added additional abstractions to handle the increased complexity.

To reduce overall risk, I did not introduce the new abstractions in the existing code. I believe the safe approach is to do this once the new feature is properly battle proven. Hence the impact on existing functionality if running without IV/CP scanning is practically nonexistant

## Description
Added Account, AccountManager, Worker and WorkerManager to facilitiate multiple account pools and safe & simple IV/CP scanning.

## Motivation and Context
Continued support of IV scanning as well as CP scanning.

## How Has This Been Tested?
A superset of this code has been run on my 250km2 scanner rig. The CP scanning code has been running for 4 months

## Types of changes

## Checklist:

### DIFF
--- a/README.md~
+++ b/README.md~
@@ -10,7 +10,7 @@ Live visualization of all the pokemon (with option to show gyms and pokestops) i
 ## Features:
 
 * Shows Pokemon, Pokestops, and gyms with a clean GUI.
-* Notifications.
+* Notifications
 * Lure information
 * Multithreaded mode
 * Filters


Alderon86:
this will need a lot of testing, also i can see alot of unneeded changes in the code, prepare for adjustments :)

343334:
without looking at the code, is this designed to be where the iv/cp accts are restricted to a particular hex? or those who run bh -st 5 etc, will they be able to service all the hexes available ?

krosenvold:
I'm not familiar with beehive, I use speed scan. Maybe someone who understands the entire code base can find out how this would work ? I suspect it will work. The workers will currently speed block to avoid high kph. I am planning to implement geo-localization of the workers, but might not include this in first version (The worker manager would take location as input and find the "closest" available free worker, worker still blocks to obey speed restrictions like today). Right now your IV checker pool needs to be sufficiently large to avoid too much travel-blocking issues (Currently it appears 
something like 30-40 IV workers should be sufficient for a -st 33 area as long as you're not IV scanning the full pokedex)

343334:
well how --behive works with -speed is lets say i do -st 6 -wph 3 -w 108  that creates 36 st6 hexes - which if defined by hex would require 36 lvl 25 accts to handle it to keep one locked to each hex if needed.  Your geo-loc would overcome that situation or that multi webhook as well that we talked about so the lvl 25 accts arent restricted to a certain region like those who are running multiple hives or bh. Though maybe the recommended way to run would be like -st10 wph 4 -wphiv 1 -w 144  etc so that in a st10 4 accts are normal and 1 is set as the iv checker but i think that would be way more work to go that route as many prefer the large st path.

bbdoc:
Wonderful. I'll test this PR asap. Thankd for it.

Just one question : how does it work wrt webhooks ? As far as a "normal" account will probably find the pokemon before another account from the 25+ pool is sent for encountering, will it wait for sending the webhook until pokemon has been encountered or will it send 2 webhooks ?

Also will pool accounts be used for "normal" scanning as well or will they wait until they are needed for an encounter ?

And finally are you also sending CP to webhook ?

Thanks

krosenvold:
All encountering is done before anything is sent to the webhook. cp is sent to webhook. cp & iv scanning pools are only used for cp & iv scanning

JoniValles:
I don't know if i'm the only one with this error but when CP-IV accounts try to log in they fail. The accounts are working 100%.
My .csv, as its said in the code, is in the format: PTC, username, password.
Good job!!

SiteEffect:
Review inc. soon by me. It's a huge PR and my time feels like its reciprocal.

pogom/models.py
```diff
@@ -1760,9 +1767,60 @@ def hex_bounds(center, steps=None, radius=None):
     return (n, e, s, w)
 
 
+def do_cp_iv_scan(cpWorkerManager, p, step_location):
+    worker = cpWorkerManager.get_worker_for_location(step_location)
```
j16sdiz: put this line inside `try` block?

config/config.ini.example
```diff
@@ -1,183 +1,197 @@
...
+#spawnpoint-scanning            # Use spawnpoint scanning (instead of hex grid).
+                                # Scans in a circle based on step_limit when on DB.
+                                # A spawnpoint-file can be added optinal as value for "spawnpoint-scanning".
+#dump-spawnpoints               # Dump the spawnpoints from the db to json (only for use with -ss).
+
+
+# Search settings
+#################
+
+#china                          # Coordinates transformer for China.
```
j16sdiz: All end-of-line changes?  
Please reduce the noise

krosenvold: I think the line endings in this file are incorrect in the RocketMap repo

pogom/models.py
krosenvold: I am definitely not a python developer, but the try/finally block is not to guard against allocation failures (which should be handled inside get_worker_for_location), but from failures when invoking the actual logic inside the block. Please explain to me if python somehow works differently from other langs I'm used to :)

accounts.txt
```diff
@@ -0,0 +1,308 @@
+ptc,irkopampcr0kysk,aaaaaaA1!
```
walaoaaa1234: @krosenvold remove this file

krosenvold: Done :)

bbdoc:
Hello. Trying to test this PR, but I'm immediately getting an error when starting 

`InternalError: (1054, u"Unknown column 't1.cp' in 'field list'")`

Looks like my DB was not updated as it should...

Any idea why ? What columns should I add to my DB ?

Also 2 more questions : 
- If providing only a CP scanning account file, will those accounts also be used for IV (I only want to add lvl30+ accounts, no specific lvl25+ for IV).
- If providing no list of pokemons to scan for CP, will it by default scan for CP all pokemons that are in my encounter list ? Again, if working only with 30+ accounts, I have no reason to specify different lists, so everything that's in my encounter whitelist (or not in my encounter blacklist) should be scanned for both IV and CP using my 30+ accounts.

Thanks

Bart274:
A group of us have been testing for 6 hours yesterday and noticed the CP was different with the same spawn for all their levels and they're all lvl30+ trainers? They think the CP is now depended for each individual trainer level and is no longer the same for evere trainer level from lvl 30?

SiteEffect:
@bbdoc Happy to see people in GitHub, fiddling around PRs etc. but you probably see reasons for direct errors pretty fast. So, in this PR the cp column was not integrated into the DB as well as there is a wrong DB version and migration only for ``level`` implemented. Please keep in mind, that this is not support forum for PRs but constructive Feedback. If you are really comfortable with the code-base, please ask the PR creator directly via Discord e. g. when you can't figure out an error at all and it is for sure related to the PR. In your case, how about adding manually a ``cp`` column with the correct type seen in models.py ``Class Pokemon`` regarding CP?

pogom/models.py
```diff
@@ -2625,6 +2723,12 @@ def database_migrate(db, old_ver):
 
         db.drop_tables([ScanSpawnPoint])
 
+    if old_ver < 12:
```
Arengo: Pretty sure it has to be ``17``.

pogom/models.py
```diff
@@ -2625,6 +2723,12 @@ def database_migrate(db, old_ver):
...
+        migrate(
+            migrator.add_column('pokemon', 'cp',
+                                IntegerField(null=True, default=0)),
```
Arengo:
```py
SmallIntegerField(null=True)
```

pogom/models.py
```diff
@@ -109,6 +114,8 @@ class Pokemon(BaseModel):
     gender = SmallIntegerField(null=True)
     last_modified = DateTimeField(
         null=True, index=True, default=datetime.utcnow)
+    cp = IntegerField(null=True)
```
Arengo:
```py
SmallIntegerField(null=True)
```
This is valid until we reach Pokémon with CP ~16.000 and can be extended by ``PositiveSmallIntegerField`` to ~32.000. The we can still switch IntegerField if this will expand further, what I do not believe.

pogom/models.py
```diff
@@ -2667,6 +2771,7 @@ def database_migrate(db, old_ver):
                 'NULL DEFAULT NULL,'
                 'MODIFY COLUMN `move_1` SMALLINT NULL DEFAULT NULL,'
                 'MODIFY COLUMN `move_2` SMALLINT NULL DEFAULT NULL;'
+                'MODIFY COLUMN `cp` SMALLINT NULL DEFAULT NULL;'
```
Arengo: Remove this, we are at 
```sql
Versions.schema_version = 16
```

pogom/models.py
```diff
@@ -2738,4 +2843,9 @@ def database_migrate(db, old_ver):
             migrate(
                 migrator.add_index('pokestop', ('last_updated',), False)
             )
+    if old_ver < 19:
```
Arengo: Add this to the
```py
if old_ver < 17
```

pogom/models.py
```diff
@@ -37,6 +40,8 @@
 log = logging.getLogger(__name__)
 
 args = get_args()
+
+
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
```
Arengo: Can't set a comment below, don't know why. So regard this as comment for [line 48](https://github.com/RocketMap/RocketMap/pull/1980/files#diff-541e2b217ba4a7d7859dae2cb443b6b7R48):
Set the new schema version one higher than current value since we modify a table (if it'd be only a new table created we wouldn't have to higher it, btw)
```py
db_schema_version = 17
```

pogom/utils.py
```diff
@@ -86,6 +86,12 @@ def get_args():
     parser.add_argument('-ac', '--accountcsv',
                         help=('Load accounts from CSV file containing ' +
                               '"auth_service,username,passwd" lines.'))
+    parser.add_argument('-cpac', '--cp-accountcsv',
+                        help=('Load CP-accounts from CSV file containing ' +
+                              '"auth_service,username,passwd" lines.'))
+    parser.add_argument('-ivac', '--iv-accountcsv',
```
Arengo: Just a minor thing, order IV before CP throughout the PR.

pogom/utils.py
```diff
@@ -429,6 +435,37 @@ def get_args():
                         help=('Pause searching while web UI is inactive ' +
                               'for this timeout (in seconds).'),
                         type=int, default=0)
+    parser.add_argument('--wh-iv-filter-whitelist', action='append',
+                        default=[], help='List of pokemon to always'
+                                         ' send through webhooks.')
+    parser.add_argument('--scan-cpall', action='append', default=[],
+                        help='List of pokemon to check CP for all IV.')
+    parser.add_argument('--scan-cp100', action='append', default=[],
```
Arengo: Start small for a new code-base introduction, especially considering new flags since we like to stop the inflation. So stash the changes regarding all additional range filters for a second PR afterwards.

static/js/map.js
```diff
@@ -424,6 +424,10 @@ function pokemonLabel(name, rarity, types, disappearTime, id, latitude, longitud
             </div>
             `
     }
+    if (cp !== null && cp > 0) {
```
Arengo: First if argument is enough since ``cp = 0`` would not be send even if it'd exist. Additionally we do not want to have Pokémon with ``cp = 0`` in our Database because of size in Millions of rows and the fact that this is "false information". If we do not know what CP it is ``NULL``, if we see ``0`` it means, it's actually 0 which is not the case.

pogom/webhook.py
```diff
@@ -171,7 +171,7 @@ def __get_key_fields(whtype):
         'pokemon': ['spawnpoint_id', 'pokemon_id', 'latitude', 'longitude',
                     'disappear_time', 'move_1', 'move_2',
                     'individual_stamina', 'individual_defense',
-                    'individual_attack'],
+                    'individual_attack', 'cp', 'level'],
```
Arengo: Stash ``level`` completely for later or fully support it - also in the web front-end.

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
-                        'seconds_until_despawn': seconds_until_despawn,
-                        'spawn_start': start_end[0],
-                        'spawn_end': start_end[1],
-                        'player_level': level
-                        })
-                    wh_update_queue.put(('pokemon', wh_poke))
+                if cpWorkers.is_scanning_active() and is_known_cp_scan(pid):
+                    pokemon_info = do_cp_iv_scan(cpWorkers, p,
+                                                 step_location)
+                    cp = pokemon_info.get('cp', 0)
```
Arengo: Set default as ``None`` instead of ``0``. This will result in a ``NULL`` entry in the DB which we would like to have if something fails or is not known for sure. Process throughout the PR (Keep in mind, for values which can actually reach ``0`` that a ``0`` will not be provided by API - so it has to be default ``0`` for ``individual_attack`` for example.

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                'level': None,
                 'height': None,
                 'weight': None,
                 'gender': None
             }
 
+            pokemon_info = None
+            cp = 0
+            pokemon_level = 0
             if (encounter_result is not None and 'wild_pokemon'
```
Arengo: Instead of a ``do_iv_scan``, ``do_cp_scan`` think about abstracting any ``encounter_result`` into one function definition ``def get_encounter_response`` where you request accordingly to given args for distinguishing the config of the user.

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                        iv = (pokemon_info.get('individual_attack', 0) +
+                              pokemon_info.get('individual_defense', 0) +
+                              pokemon_info.get('individual_stamina', 0)) \
+                             * 100.0 / 45.0
+                        if cpWorkers.is_scanning_active() and \
+                                should_get_cp(pid, iv, args):
+                            pokemon_info = do_cp_iv_scan(cpWorkers, p,
+                                                         step_location)
+                            cp = pokemon_info.get('cp', 0)
+                            pokemon_level = pokemon_info.get('level', 0)
```
Arengo: Afaik a ``level`` is not submitted, only the corresponding ``cp_multiplier`` which translates into a level with a table (I think I saw this table once in your fork)

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 'individual_stamina': None,
                 'move_1': None,
                 'move_2': None,
+                'cp': None,
+                'level': None,
```
Arengo: Until now we only store direct API values in the database - so this should be ``cp_multiplier`` and the conversion should be done somewhere after the DB instead of in the back-end

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                if pokemon_info is not None:
+                    attackIv = pokemon_info.get('individual_attack', 0)
+                    defenseIv = pokemon_info.get('individual_defense', 0)
+                    staminaIv = pokemon_info.get('individual_stamina', 0)
+                    move_1 = pokemon_info['move_1']
+                    move_2 = pokemon_info['move_2']
+                    height = pokemon_info['height_m']
+                    weight = pokemon_info['weight_kg']
+                    gender = pokemon_info['pokemon_display']['gender']
+                    if pokemon_info['pokemon_id'] == 201:
```
Arengo: Thanks for being so kind, addressing #1914 directly. While this is in general a good move to finally include it into RM, leave it out for lowering complexity since whatever will be merged first, the other one has to adjust.

bbdoc:
@Arengo Yes of course adding CP column would be a quick fix, but is it really a good solution.

I do see indeed that the code still mention `db_schema_version = 16` which is the same as current dev version. Strange as I'm pretty sure I did see 17 at some point in time. Have to dig into latest commits to see where it was rolled back.

I still guess it's worth it mentionning to PR initiator that there is something wrong in his code resulting in db not being updated. I guess people who managed to get it working started with a blank db and thus didn't get this error.

pogom/models.py
```diff
@@ -1968,47 +2027,86 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                              pokemon_info.get('individual_stamina', 0)) \
+                             * 100.0 / 45.0
+                        if cpWorkers.is_scanning_active() and \
+                                should_get_cp(pid, iv, args):
+                            pokemon_info = do_cp_iv_scan(cpWorkers, p,
+                                                         step_location)
+                            cp = pokemon_info.get('cp', 0)
+                            pokemon_level = pokemon_info.get('level', 0)
+
+                if pokemon_info is not None:
```
Arengo: Having a general dict update style across all possibilities imo it would be logically the best to get away from standard encounter, check if encounter lvl 25+ or lvl 30+, adding all values of old encounter by a first dict.update and then adding the last two values regarding cp.

pogom/models.py
krosenvold: Ok thanks, I didnt really fine-study the db update logic

krosenvold:
Note that there was a bug I fixed this morning: For pokemons that were NEITHER CP or IV scanned, the "cp" would slip through from the original map worker. This was fixed about 5 hours ago

pogom/utils.py
krosenvold: I am not sure this is entirely wise, at least cp90 and cp100 are quite useful, so you can conserve those level 30 workers until you know the pokemon is good

pokemonkiel:
Could you please add cp_mutliplier to check Pokemonlevel?
I can give you JS code for calculation but adding the column would be perfect. Thanks!

pogom/models.py
krosenvold: I'm not entirely sure I understand this comment

pogom/models.py
krosenvold: Style-wise I agree. But I also believe the current encounter logic is entirely useless ? I'd really enjoy looking at all of these together, so I'm sort-of awaiting the "encounter" issue to clear out.

krosenvold:
I have rebased this PR, but I am now not squashing to make it easier to follow the changes done in the review process.

walaoaaa1234:
@krosenvold you should remove copyrighted content

pogom/utils.py
```diff
@@ -429,6 +435,37 @@ def get_args():
...
+    parser.add_argument('--scan-cp95', action='append', default=[],
+                        help='List of pokemon to check CP for when IV > 95.')
+    parser.add_argument('--scan-cp90', action='append', default=[],
+                        help='List of pokemon to check CP for when IV > 90.')
+    parser.add_argument('--scan-cp80', action='append', default=[],
+                        help='List of pokemon to check CP for when IV >= 80.')
+    parser.add_argument('--scan-iv', action='append', default=[],
+                        help='List of pokemon to check IV and '
+                             'moveset with level 25+ account')
+
```
bbdoc: I see a --scan-iv parameter on command line but no equivalent in config.ini.example ? Did I miss it ? I also think it would be good to have both whitelist and blacklist for this parameter as it does exist for encounter, and I guess if you use an encounter blacklist you also want to use the same blacklist for scanning IV with a lvl 25+ account. Quid if you only have -cpac ? Will those accounts be used for IV scanning as well (as if they are 30+ they are also 25+) ?

bbdoc: Running on my map right now and all looking fine, but only scanning IV right now. My IV scanning accounts are lvl30+ so I think when sending an IV scanning account to a spawn and it is lvl30+, it should automatically check cp as well as it doesn't cost any additional request.

sebastienvercammen:
PR #1981 is being tested for merge rather than this PR.

Relaying feedback as mentioned in our Discord's #pr:

- The PR tries to do too much simultaneously. Abstractions are good, but unfinished abstractions are not. More about this in the other points.
- The first thing to do is to release a working implementation for our users that doesn't change the old way of working, or that doesn't break anything. The users need to be given time to adjust to major reworks.
- Careful for Travis. Run a linter.
- New abstractions try to incorporate changes to structure. This would be encouraged, if it was done properly. When writing abstractions, keep in mind the separation of responsibility between objects. As it stands, the Account is responsible for the Search (search & rest intervals, scan timing), for the API objects and requests, the hashing keys, the proxies and failure handling. This is too much.
- You've moved search_interval and rest_interval into Account but you haven't implemented it.
- A PR was recently merged to reduce memory usage, introduced some changes in how we should handle objects and requests. You've ignored these changes and went back to an older way of doing things that unnecessarily increases memory usage.
- You've redefined `equi_rect_distance` in Account, which is already in `utils.py`.
- You've mixed naming styles and you've ignored consistency with previous code: in naming, in comments, in structure.
- We're not going to say much about `__do_with_backoff`, you probably know what we mean. At least it has a comment that says it needs reworking, that's a plus.
- The amount of new config items you've added is enormous. While trying to copy the contents over to Discord, we received an error that the message we were trying to send was over the 2000 character limit of Discord. Here's a list of the config items you've added: `-ivac, -cpac, --wh-iv-filter-whitelist, --scan-cpall, --scan-cp100, --scan-cp95, --scan-cp90, --scan-cp80, --scan-iv, --iv-auth-service, --iv-username, --iv-password, --cp-auth-service, --cp-username, --cp-password`.

Arguably, you've taken time to work on things that absolutely needed to be reworked, but then you don't seem to have taken the time to wonder how the abstractions could be written properly. And then you don't seem to have talked to anyone about what you were doing, so there was no code feedback.

With the PR's size, there's no time for us to review and rework all points properly, because that would involve rewriting account handling as there's an important part in the legacy code that really needs to be reworked.

In the meantime, PR #1981 will serve as the initial implementation, and we'll rework the rest as we go. PR 1981 separates the new account pools from the old code so we can more easily rewrite account handling and the other TODOs without having to undo a new mess that a larger PR like this one would have created.

I've also been PM'ed that you've been inviting people to a new Discord of yours called _"FreeSpeechRocket"_. I can appreciate a good joke, but not a bad attitude. I don't know you and I've never worked with you, so if there's already a problem I'd love to hear about it so I can make sure someone else talks to you instead in the future. That someone PM's me about it is already too much: I'm not sure how you consider that kind of attitude constructive or productive.

krosenvold:
Thank you for commenting @sebastienvercammen. I am aware the abstractions are not perfect and may need further fine tuning. This is also the reason I did not introduce them all over the code base, their use is limited to the new cp/iv scanning. I will think a little about your comments, but overall it looks like it should be easily fixable in a couple of hours.

As for the discord, I only created it to handle the extreme amount of people who were DM'ing me, there was simply no way for me to reply them all. I'm sorry if the name offended you. I have renamed it to "G-118A Peacekeeper". I have no intention of forking neither rocketmap nor the rocketmap community, and I have only invited people that were DM'ing me directly. Presumably I'll shut it down once this all blows over. I will invite you if you want, send me a DM asking about pr1980 :)

The last thing you might want to think about is this: If you come to the discord and say "I have written some code" you will hear "shut up and make a pull request". I know this is standard operating procedure in most open source projects. So you make a pull request, putting quite some effort into it. Then you get rebuted for wanting to talk about the code you've written because you're advertising a pull request. So once the review comes along you get rebuted for not discussing the code. So, exactly where is one supposed to discuss code ?

PR #1981, opened 2017-04-22
Add IV/CP scanning support with a L30 account pool.

StevenHickson:
## Description
**Seb edit:** This PR implements IV & CP scanning with L25/L30 account pools.

**It's important** to read the included new doc, [encounters.md](https://github.com/StevenHickson/RocketMap/blob/83c07d992a600ff20471e0f8cc31644aadf090c5/docs/extras/encounters.md).

## Motivation and Context
**Seb edit:** This allows people to see correct IVs for L25 and above, and correct CP for L30 and above.

## How Has This Been Tested?
Tested on two of my own maps on my computers. It does require -cd as we are adding a new column to the database.

selection_010

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -1,0 +1,44 @@
+# Pokémon Encounters
+
+Since the IV update of April 21st which makes IVs the same for players of level 25 and above, the encounter system has been reworked and now includes CP/IV scanning.
+
+Steps for using the new encounter system:
+
+1. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
+   ```
+   25,ptc,randOMusername1,P4ssw0rd!
+   25,ptc,randOMusername2,P4ssw0rd!
+   30,ptc,randOMusername1,P4ssw0rd!
+   ```
+2. Create files for your IV and CP encounter whitelists and add the Pokémon IDs which you want to encounter, one per line.
+   iv-whitelist.txt:
+   ```
+   10
+   25
+   38
+   168
+   ```
+   cp-whitelist.txt:
+   ```
+   35
+   66
+   89
+   ```
+3. Enabled the whitelist files in your config or cli parameters (check commandline.md for usage):
+   ```
+   --iv-whitelist-file iv-whitelist.txt --cp-whitelist-file cp-whitelist.txt
+   ```
+4. Optionally set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
+   ```
+   --hlvl-kph 25
+   ```
+
+L25/L30 accounts are not being recycled and are not in the usual account flow. This is intentional, to allow for future reworks to handle accounts properly. This also keeps interaction with high level accounts to a minimum. We can consider handling them more automatically when the account handlers are properly fully implemented.
+
+Some important notes:
+
+ * The old encounter whitelists/blacklists have been removed entirely.
+ * Both the IV and CP whitelists are optional.
+ * When no L25 accounts are available but IV encounters are enabled, it will try to get a L30 account instead for the encounter.
+ * Captcha'd L25/L30 accounts will be logged and disabled in the scheduler. Having `-v` enabled will show you an entry in the logs mentioning "High level account x encountered a captcha". They will not be solved automatically.
+ * The encounter is a single request (1 RPM). We intentionally don't use the account for anything else besides the encounter.
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -4,7 +4,8 @@ Since the IV update of April 21st which makes IVs the same for players of level 
 
 Steps for using the new encounter system:
 
-1. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
+1. Enabled encounters on your map (`-enc`).
+2. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
    ```
    25,ptc,randOMusername1,P4ssw0rd!
    25,ptc,randOMusername2,P4ssw0rd!
@@ -14,7 +15,7 @@ Steps for using the new encounter system:
    ```
    --high-lvl-accounts high-level.csv
    ```
-2. Create files for your IV and CP encounter whitelists and add the Pokémon IDs which you want to encounter, one per line.
+3. Create files for your IV and CP encounter whitelists and add the Pokémon IDs which you want to encounter, one per line.
    iv-whitelist.txt:
    ```
    10
@@ -28,11 +29,11 @@ Steps for using the new encounter system:
    66
    89
    ```
-3. Enable the whitelist files in your config or cli parameters (check commandline.md for usage):
+4. Enable the whitelist files in your config or cli parameters (check commandline.md for usage):
    ```
    --iv-whitelist-file iv-whitelist.txt --cp-whitelist-file cp-whitelist.txt
    ```
-4. Optionally set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
+5. Optionally set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
    ```
    --hlvl-kph 25
    ```
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -4,11 +4,15 @@ Since the IV update of April 21st which makes IVs the same for players of level 
 
 Steps for using the new encounter system:
 
-1. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
+1. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
    ```
    25,ptc,randOMusername1,P4ssw0rd!
    25,ptc,randOMusername2,P4ssw0rd!
    30,ptc,randOMusername1,P4ssw0rd!
+   ```
+   The config item or parameter to use this separate account file is:
+   ```
+   --high-lvl-accounts high-level.csv
    ```
 2. Create files for your IV and CP encounter whitelists and add the Pokémon IDs which you want to encounter, one per line.
    iv-whitelist.txt:
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -6,7 +6,7 @@ Steps for using the new encounter system:
 
 1. Make sure initial scan has finished. Enabling encounters during initial scan is a waste of requests.
 2. Enable encounters on your map (`-enc`).
-3. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). The lines should be formatted as "25 or 30,service,user,pass":
+3. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). **Warning: read the important points below if you're scanning with only L30s!** The lines should be formatted as "25 or 30,service,user,pass":
    ```
    25,ptc,randOMusername1,P4ssw0rd!
    25,ptc,randOMusername2,P4ssw0rd!
@@ -43,6 +43,7 @@ L25/L30 accounts are not being recycled and are not in the usual account flow. T
 
 Some important notes:
 
+ * If you're only scanning with high level accounts (i.e. your regular accounts file only has L30s), the `--high-lvl-accounts` file can stay empty. The encounter code will use your regular accounts to encounter Pokémon if they're high enough level. But don't mix low level accounts with high levels, otherwise encounters will be skipped.
  * To report Unown form, Unown's Pokémon ID must be added to the IV or CP whitelist.
  * The old encounter whitelists/blacklists have been removed entirely.
  * Both the IV and CP whitelists are optional.
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -6,33 +6,27 @@ Steps for using the new encounter system:
 
 1. Make sure initial scan has finished. Enabling encounters during initial scan is a waste of requests.
 2. Enable encounters on your map (`-enc`).
-3. Add L25 (and optionally L30) accounts for IV scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). **Warning: read the important points below if you're scanning with only L30s!** The lines should be formatted as "25 or 30,service,user,pass":
+3. Add L30 accounts for IV/CP scanning into a CSV file (separate from your regular accounts file, e.g. 'high-level.csv'). **Warning: read the important points below if you're scanning with only L30s!** The lines should be formatted as "service,user,pass":
    ```
-   25,ptc,randOMusername1,P4ssw0rd!
-   25,ptc,randOMusername2,P4ssw0rd!
-   30,ptc,randOMusername1,P4ssw0rd!
+   ptc,randOMusername1,P4ssw0rd!
+   ptc,randOMusername2,P4ssw0rd!
+   ptc,randOMusername1,P4ssw0rd!
    ```
    The config item or parameter to use this separate account file is:
    ```
    --high-lvl-accounts high-level.csv
    ```
-4. Create files for your IV and CP encounter whitelists and add the Pokémon IDs which you want to encounter, one per line.
-   iv-whitelist.txt:
+4. Create files for your IV/CP encounter whitelist and add the Pokémon IDs which you want to encounter, one per line.
+   enc-whitelist.txt:
    ```
    10
    25
    38
    168
    ```
-   cp-whitelist.txt:
-   ```
-   35
-   66
-   89
-   ```
 5. Enable the whitelist files in your config or cli parameters (check commandline.md for usage):
    ```
-   --iv-whitelist-file iv-whitelist.txt --cp-whitelist-file cp-whitelist.txt
+   --enc-whitelist-file enc-whitelist.txt
    ```
 6. (Optional) Set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
    ```
@@ -43,7 +37,7 @@ Steps for using the new encounter system:
    --no-api-store
    ```
 
-L25/L30 accounts are not being recycled and are not in the usual account flow. This is intentional, to allow for future reworks to handle accounts properly. This also keeps interaction with high level accounts to a minimum. We can consider handling them more automatically when the account handlers are properly fully implemented.
+L30 accounts are not being recycled and are not in the usual account flow. This is intentional, to allow for future reworks to handle accounts properly. This also keeps interaction with high level accounts to a minimum. We can consider handling them more automatically when the account handlers are properly fully implemented.
 
 Some important notes:
 
@@ -51,7 +45,6 @@ Some important notes:
  * To report Unown form, Unown's Pokémon ID must be added to the IV or CP whitelist.
  * The old encounter whitelists/blacklists have been removed entirely.
  * Both the IV and CP whitelists are optional.
- * When no L25 accounts are available but IV encounters are enabled, it will try to get a L30 account instead for the encounter.
- * Captcha'd L25/L30 accounts will be logged in console and disabled in the scheduler. Having `-v` enabled will show you an entry in the logs mentioning "High level account x encountered a captcha". They will not be solved automatically.
+ * Captcha'd L30 accounts will be logged in console and disabled in the scheduler. Having `-v` enabled will show you an entry in the logs mentioning "High level account x encountered a captcha". They will not be solved automatically.
  * The encounter is a single request (1 RPM). We intentionally don't use the account for anything else besides the encounter.
  * The high level account properly uses a proxy if one is set for the scan, and properly rotates hashing keys when needed.
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -1,6 +1,6 @@
 # Pokémon Encounters
 
-Since the IV update of April 21st which makes IVs the same for players of level 25 and above, the encounter system has been reworked and now includes CP/IV scanning.
+Since the IV update of April 21st 2017 which makes IVs the same for players of level 25 and above, the encounter system has been reworked and now includes CP/IV scanning.
 
 Steps for using the new encounter system:
 
@@ -34,9 +34,13 @@ Steps for using the new encounter system:
    ```
    --iv-whitelist-file iv-whitelist.txt --cp-whitelist-file cp-whitelist.txt
    ```
-6. Optionally set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
+6. (Optional) Set a speed limit for your high level accounts. This is separate from the usual speed limit, to allow a lower speed to keep high level accounts safer:
    ```
    --hlvl-kph 25
+   ```
+7. (Optional) To reduce the number of times a high level account will log into the game via the API, the API objects are stored in memory to re-use them rather than recreating them. This is enabled by default to keep high level accounts *safer* but it will cause an increase in memory usage. To reduce memory usage, disable the feature with:
+   ```
+   --no-api-store
    ```
 
 L25/L30 accounts are not being recycled and are not in the usual account flow. This is intentional, to allow for future reworks to handle accounts properly. This also keeps interaction with high level accounts to a minimum. We can consider handling them more automatically when the account handlers are properly fully implemented.
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -40,5 +40,6 @@ Some important notes:
  * The old encounter whitelists/blacklists have been removed entirely.
  * Both the IV and CP whitelists are optional.
  * When no L25 accounts are available but IV encounters are enabled, it will try to get a L30 account instead for the encounter.
- * Captcha'd L25/L30 accounts will be logged and disabled in the scheduler. Having `-v` enabled will show you an entry in the logs mentioning "High level account x encountered a captcha". They will not be solved automatically.
+ * Captcha'd L25/L30 accounts will be logged in console and disabled in the scheduler. Having `-v` enabled will show you an entry in the logs mentioning "High level account x encountered a captcha". They will not be solved automatically.
  * The encounter is a single request (1 RPM). We intentionally don't use the account for anything else besides the encounter.
+ * The high level account properly uses a proxy if one is set for the scan, and properly rotates hashing keys when needed.
--- a/docs/extras/encounters.md
+++ b/docs/extras/encounters.md
@@ -4,8 +4,9 @@ Since the IV update of April 21st which makes IVs the same for players of level 
 
 Steps for using the new encounter system:
 
-1. Enabled encounters on your map (`-enc`).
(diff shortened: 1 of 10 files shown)


pogom/models.py
```diff
@@ -1986,6 +1988,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         'individual_stamina', 0),
                     'move_1': pokemon_info['move_1'],
                     'move_2': pokemon_info['move_2'],
+                    'cp': pokemon_info['cp'],
```
Arengo: Go a step further and make your PR viable for the whole RocketMap userbase, checking for account level (which is already implemented). So remove ``cp`` in this line but add below:
```python
   pokemon[p['encounter_id']]['cp'] = pokemon_info['cp'] if level >= 30
```

Also think about adding a similar if statement in front of encounter for IVs/Movesets. (Still not 100% sure about gender/height/weight)

static/js/map.js
```diff
@@ -420,6 +420,9 @@ function pokemonLabel(name, rarity, types, disappearTime, id, latitude, longitud
                 IV: ${iv.toFixed(1)}% (${atk}/${def}/${sta})
             </div>
             <div>
+                CP: ${cp}
```
Arengo: Adjust accordingly to ``models.py`` by adding an if clause:
```js
if (cp !== null){
    details += `...`
}
```

SiteEffect:
``-cd`` can be avoided, inspecting the database via phpMyAdmin or programs like SequelPro and add the new column manually. After the manual "migration" you need to update ``versions`` table to the new database version.
```sql
UPDATE versions SET val = X
```
Funny side fact: ``versions``, together with ``gymdetails`` are the the only tables we use a plural wording. Especially for ``versions`` it's cool, because it only holds one single entry.

pogom/models.py
```diff
@@ -2739,3 +2742,9 @@ def database_migrate(db, old_ver):
                 migrator.add_index('pokestop', ('last_updated',), False)
             )
         log.info('Schema upgrade complete.')
+
+    if old_ver < 17:
+        migrate(
+            migrator.add_column('pokemon', 'cp',
+                                IntegerField(null=True, default=0))
```
Arengo: SmallIntegerField()

static/js/map.js
Obihoernchen: And please add some info that this is valid for lvl 30+ only

Sttug:
U can add cp to webhook too ?

pki0:
Could you please add cp_mutliplier to check Pokemonlevel?
I can give you JS code for calculation but adding the column would be perfect. Thanks!

tobias86aa:
Can we dont use csv files for the 25 and 30 level accounts? How should the format be then? I dont see any information about that in the  encounters.md file.

sebastienvercammen:
@tokar86a This PR is for development, not customer support. Be patient until things are finalized.

pogom/utils.py
```diff
@@ -86,6 +86,11 @@ def get_args():
     parser.add_argument('-ac', '--accountcsv',
                         help=('Load accounts from CSV file containing ' +
                               '"auth_service,username,passwd" lines.'))
+    parser.add_argument('-hlvl', '--high-lvl-accounts',
```
Arengo: Personally, I like to have two different files for CP and/or for IV. This also prevents mess by an additional csv column with a huge amount of accounts.
For consistency, the argument name should be something like ``--high-lvl-accountcsv``, ``-hlvlac`` longer but easier to identify.

pogom/utils.py
```diff
@@ -155,23 +160,14 @@ def get_args():
...
-                                      'more stats.'))
-    encounter_list.add_argument('-ewhtf', '--encounter-whitelist-file',
-                                default='', help='File containing a list of '
-                                                 'Pokemon to encounter for'
-                                                 ' more stats.')
-    encounter_list.add_argument('-eblkf', '--encounter-blacklist-file',
-                                default='', help='File containing a list of '
-                                                 'Pokemon to NOT encounter for'
-                                                 ' more stats.')
+    parser.add_argument('-ivwf', '--iv-whitelist-file',
```
Arengo: With all the files, we should think about providing a config where to read from in the future. Just a side note.

static/js/map.js
```diff
@@ -435,6 +436,17 @@ function pokemonLabel(item) {
             <div>
                 IV: ${iv.toFixed(1)}% (${atk}/${def}/${sta})
             </div>
+            `
+
+        if (cp !== null && cp > 0) {
```
Arengo: ``cp == 0`` is not possible from the API and should not be possible from the code. Combining it with the related comment from models.py with a default value in python of ``None`` this second if check argument should be stashed.

bbdoc:
Why do we specify pokemon to encounter for IV/CP by their pokemon IDs in iv-whitelist-file and cp-whitelist-file, while we have to specify them by their name in webhook-whitelist-file ? Wouldn't it be more consistent to chose for one or the other ? I also guess lots of people have no reason to send pokemons for which they don't scan IV to their webhook so having similar config file would allow to maintain only one. I personally do prefer pokemon IDs as lots of maps are localized and Pokemon IDs are the same for everyone while names are not...

pogom/models.py
```diff
@@ -1995,6 +2049,11 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'gender': pokemon_info['pokemon_display']['gender']
                 })
 
+                # Only add CP if we're level 30+.
+                if encounter_level >= 30:
+                    pokemon[p['encounter_id']][
+                        'cp'] = pokemon_info.get('cp', 0)
```
Arengo: Defaut to ``None`` - ``0`` is not possible, ``0`` is missleading since it means that we know it, there can other additional checks be performed than just comparing with ``0`` in case of a fail and ``None`` is the general default value for "not knowing something``.

pogom/models.py
```diff
@@ -1930,34 +1935,82 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+
+                    responses = encounter_result['responses']
+                    # Check for captcha
+                    captcha_url = responses['CHECK_CHALLENGE']['challenge_url']
+                    # Throw warning but finish parsing
+                    if len(captcha_url) > 1:
+                        # Flag account.
+                        hlvl_account['captcha'] = True
+                        log.info('High level account '
+                                 + hlvl_account['username']
```
Arengo: Since we have a level, we should provide also for the log and maybe on which Pokemon ID it failed.

Sttug:
if some one pokemon is in both list (cp and iv) account lvl 25 is free then use for iv next use lvl 30 account for cp, i mean we need 3 list cp, iv, and cp/iv list

pogom/models.py
```diff
@@ -1930,34 +1935,82 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
-                encounter_result = clear_dict_response(encounter_result)
-
-                captcha_url = encounter_result['responses']['CHECK_CHALLENGE'][
-                    'challenge_url']  # Check for captcha
-                if len(captcha_url) > 1:  # Throw warning but finish parsing
-                    log.debug('Account encountered a reCaptcha.')
+
+                # Get account to use for IV or CP scanning.
+                if pokemon_id in args.cp_whitelist:
+                    hlvl_account = account_sets.next('30', step_location)
```
Arengo: Throughout the changes ``hlvl`` is used, this is a "symptom" but not the root of the recent changes. Niantic can basically adjust to whatever concept to distribute Pokémon details to their users as they like. Since the levels may change (maybe to "lower" one day), the act of getting the details is a bit more constant - we have to encounter it. We should think about adjusting all namings to ``encounter_ .. ``

static/js/map.js
sebastienvercammen: `cp == 0` is possible because `cp = parseInt(item['cp']) || 0`.

pogom/utils.py
sebastienvercammen: You can't fix user error: if you're scared of messing up when it's in one file, you should also be scared of messing up when it's in separate files, because you're assuming you're going to make a mistake.

pogom/models.py
sebastienvercammen: Agreed, but an AccountSet isn't a solution to our old bad account handling. We need better account handling, preferably stored in database so we can distribute/manage accounts globally.

Arengo: I am already onto [database accounts](https://github.com/Arengo/RocketMap/tree/PR-AccountModel).

pogom/models.py
```diff
@@ -1930,34 +1935,82 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                        # Flag account.
+                        hlvl_account['captcha'] = True
+                        log.info('High level account '
+                                 + hlvl_account['username']
+                                 + ' encountered a captcha.')
+                else:
+                    log.error('No L25 or L30 accounts are available, please'
+                              + ' consider adding more. Skipping encounter.')
 
             pokemon[p['encounter_id']] = {
```
sebastienvercammen: Added a check on the account's level in last commit. We throw an exception if there's user error.

myntath:
It doesn't appear to me that the high level accounts would ever call api.get_map_objects (if I'm wrong great and please ignore anything further). Although this works I would be concerned that this could cause accounts to be flagged for banning. All the high level accounts will only every be entering encounters, they never 'walk' around the map.

If the high level accounts all just directly encounter. Isn't this similar to 'sniping' methods? I know this saves on hash RPM but I would prefer to play it safe and do the redundant get_map_objects anyway  before encountering and while it may be wasteful to fully process the second get_map_objects i would propose a few sanity checks to see the pokemon is there still (it hasn't despawned, we didn't breach speed limit etc...)

awwwwwyup:
Looks as though the encounter to find unown's form is limited to a high level account now? We had one spawn last night using the latest merges and the worker account (level 2) identified it as an L, and several players in the area caught it and it was still an L (they are all 30+). Perhaps a special case could be implemented to use the low level account to encounter unown, saving the higher level accounts for other whitelisted spawns?

I do realize it introduces added complexity for something that spawns so infrequently, but just something that caught my eye.

Bart274:
Would it be possible to add the pokemon level to the db?
The pokemon levels are fixed, while the cp is different for each pokemon depending on it's level.
So it would be easier to see which pokemon has max cp for it's iv's instead of havong to look it up for each encountered pokemon.

StevenHickson:
Tested and attempted a bunch of odd things to try to break it. Worked really well. LGTM

pogom/models.py
```diff
@@ -1930,34 +1935,94 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
             encounter_result = None
-            if (args.encounter and (p['pokemon_data']['pokemon_id']
-                                    in args.encounter_whitelist or
-                                    p['pokemon_data']['pokemon_id']
-                                    not in args.encounter_blacklist and
-                                    not args.encounter_whitelist)):
+
+            if args.encounter and (pokemon_id in args.iv_whitelist or
+                                   pokemon_id in args.cp_whitelist):
                 time.sleep(args.encounter_delay)
```
friscoMad: This delay is not based on the account doing the encounter, so it is useless, different thread scanning different points could use the same hlvl account at the same time.
This should be done on next()

pogom/models.py
```diff
@@ -1930,34 +1935,94 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                            'Using key {} for this encounter.'.format(key))
+                        hlvl_api.activate_hash_server(key)
+
+                    # Log in.
+                    check_login(args, hlvl_account, hlvl_api, step_location,
+                                status['proxy_url'])
+
+                    # Setup encounter request envelope.
+                    req = hlvl_api.create_request()
+                    encounter_result = req.encounter(
```
friscoMad: no need to set encounter_result here just call req.encounter

pogom/models.py
```diff
@@ -1930,34 +1935,94 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                        player_longitude=step_location[1])
+                    req.check_challenge()
+                    req.get_hatched_eggs()
+                    req.get_inventory()
+                    req.check_awarded_badges()
+                    req.download_settings()
+                    req.get_buddy_walked()
+                    encounter_result = req.call()
+
+                    # Update level indicator before we clear the response.
```
friscoMad: Check for captcha?

pogom/models.py
```diff
@@ -1995,6 +2061,11 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'gender': pokemon_info['pokemon_display']['gender']
```
friscoMad: Add form also as unown forms seems to use the same IV/Moves logic

pogom/utils.py
```diff
@@ -639,6 +639,65 @@ def get_args():
...
+                                        + '". Type must be 25 or 30.')
+
+        # Prepare the IV/CP scanning filters.
+        args.iv_whitelist = []
+        args.cp_whitelist = []
+
+        # IV scanning.
+        if args.iv_whitelist_file:
+            with open(args.iv_whitelist_file) as f:
+                args.iv_whitelist = [int(l.strip()) for l in f]
```
friscoMad: Please use set() or even better frozenset() ```in``` comparison in list() is O(n) while in set is O(1)
If you update also wh_*list it would be great.

pogom/models.py
sebastienvercammen: A different thread won't be able to use the same high level account, `next()` is thread safe and adds a timestamp to the account so it won't be returned again until enough time has expired.

I agree that sleeping doesn't make much sense anymore because it's a different account, but just moving the sleep to `next()` also doesn't make sense since that'll have exactly the same effect as it's happening on the same thread.

Perhaps adding `args.encounter_delay` to the amount of time a high level account sleeps for (on top of the speed limit) is a better middle ground.

bbdoc:
There is something wrong in the way this PR handles accounts. It runs fine for about 5 mins, and then I start getting `[   ERROR] No L25 or L30 accounts are available, please consider adding more. Skipping encounter.`

Had a look at my tables and counted pokemons with CP > 0, got exactly 40 of them after 2 runs with 20 lvl 30 accounts. So looks like my accounts have been used once and then put aside.

I was running PR 1980 since yesterday morning without having this issue, it could continue scanning without running out of accounts.

This PR is also missing verbosity in logs. I can't find any mention of my accounts being sent for an IV/CP encounter until I hit this error message.

[EDIT] Confirmed after running with -v. I can see each account being sent exactly once for an encounter, and then start getting this error message. Accounts are not reused for next encounter.

sebastienvercammen:
@bbdoc Verbosity is addressed with an upcoming commit that also solves the problem of users who use RocketMap with only L30s in their regular account pool (which wouldn't be used for encountering because we've defined a separate pool).

Can you give some details on the distance of the encounters and the scan radius? Accounts store their last scanned coordinates and follow the km/h limit.

bbdoc:
@sebastienvercammen tested on an st 30 using 70 regular accounts and 20 lvl 30 accounts. Cannot really tell distance of the encounters as I have no info about it in logs, not even running -v. That's indeed something I could clearly see in logs using 1980. I'll test again once this commit with more verbosity is available. I think we need more details on accounts being put aside for whatever reason and encounters that are being skipped because of being too far away taking hlvl-kph into account. Would really help troubleshooting this kind of behaviour.

sebastienvercammen:
@bbdoc I've found the source of the problem you're encountering. Fixed in next commit.

bbdoc:
wonderful @sebastienvercammen You are too quick ;-)

pogom/models.py
friscoMad: Being next being thread safe doesn't maen anything, if you call next() twice with coords near enought (totally possible for spawnpoints near but in differnt steps) the same account could be returned.
AFAIK python threads only give priority on IO so doing the login would probably force a thread change, so it is very possible to violate speed limits/have login problems if the first thread does not get back the priority before any other thread gets this account.
I would say that if used with a big encounter white list and with high cpu usage that would happen easily.

friscoMad:
As it is now this PR several pokemons in the same scan would probably use the same account (if there are no threading problems) but will login for each encounter, this wouldn't be a problem for instances with a small a number of whitelisted pokemons but otherwise it is not really nice.
Possible solutions:

- Storing apis alongside with accounts to avoid logins => less cpu/requests more memory usate
- Reviewing all pokemon found creating a list of pokemon to encounter, finding the max lvl account needed for the set and the login in and encounter all pokemons in that list => more code/cpu/requests less memory

Also to avoid threading scheduler issues that could cause login and/or speed issues I suggest using some kind of 'in_use' attribute that it is only cleared when a thread has finished the encountering.

sebastienvercammen:
@myntath So the others on Github can follow, here's what I said on Discord:

>Correct on get_map_objects. On one hand it reduces requests and captchas, on the other we don't know what effect it will have.
>
>We need better account and scan handling overall with better abstractions so everything can be done properly.

Additional sanity checks would be code that we'd end up moving/rewriting/splitting when we work on better account and scan handling. For now, this method of encountering is a tried and tested method and so far it has been less strict than regular scanning.

That doesn't mean we shouldn't move to safer scanning anyway, but it's the thought process that went behind the current implementation which isn't tied directly into legacy code, making it easier to rework accounts/scanning.

sebastienvercammen:
@friscoMad The AccountSet uses a threading lock, it's thread safe. As I've said before, this means any account will only be taken by a single thread, and it will only ever be used again when enough time has passed for the speed limit.

An "in_use" flag would be redundant.

It's true that accounts log in for the encounter, but the rate at which this happens is still limited by the speed limit and the number of whitelisted Pokémon. I already added thoughts on storing the API objects in the comments in the code, but this is something that should be discussed. Even when cleared, an API object will take up space, and depending on the amount of high level accounts people intend to scan with, it can take up a considerable amount of memory.

I'm not saying we shouldn't or can't do it, but that's not something I'd decide by myself because I know it's not the best thing to do.

In my opinion, it's the legacy code that needs to be rewritten to properly and consistently handle scans and accounts, with better abstractions, so that we can have an actual workflow rather than an arbitrary amount of threads that are all competing for some work, with variables being passed all over.

sebastienvercammen:
@awwwwwyup @Bart274 Keeping Unown form and Pokémon level (based on the cp_multiplier map) for a separate PR.

sebastienvercammen:
@bbdoc 
>Why do we specify pokemon to encounter for IV/CP by their pokemon IDs in iv-whitelist-file and cp-whitelist-file, while we have to specify them by their name in webhook-whitelist-file ?

Correct, we discussed this and decided to choose for Pokémon IDs. The others should be reworked in a separate PR.

The thoughts were that IDs are more consistent to work with and that everyone will want to use Pokémon names in their own language.

Bart274:
@sebastienvercammen wasn't Unown Form already merged with https://github.com/RocketMap/RocketMap/pull/1914 ?
will this remove this again?
I don't mind having to wait for an additional PR for the pokemon level, as long as it arrives in the code at some point, I'm glad. :-)

sebastienvercammen:
@Bart274 Unown form is still shown with these new encounters, but Unown would need to be added to the IV or CP whitelist.

We're checking to confirm if Unown form is always the same for accounts of all levels, or if there is any kind of randomization like Niantic has been adding recently.

Bart274:
@sebastienvercammen gotcha :-) Thanks for the good work :-)

pogom/models.py
sebastienvercammen: @friscoMad Euh, no... Thread safety means that no two threads can **ever** be running that piece of code simultaneously.

An account would only be returned again when it's ready to scan again, which is precisely what the code needs to do.

Bart274:
@sebastienvercammen I would love to see the accounts being recycled in the future and have an option to gather cp/iv for all pokemon instead of having to whitelist them for people who run maps with only lvl 30+ accounts

wouldn't it be possible to check the account level of each account when you're using it in the normal accounts.csv with the api and just do the encounter with that account and get the right info depending on it's level?

sebastienvercammen:
@Bart274 Last commit addresses that. If you're using only high level accounts in your regular account pool, that account will be used for the encounter.

The Pokémon ID will still need to be in the IV or CP whitelist, but that's trivial.

friscoMad:
@sebastienvercammen Maybe if I give you an example you would understand why locking in next doesn't help that much.

Problem 1:

1.- Account A scans point 1 get 1 mon and need a IV then calls next() and get account X for it
2.- While in the same thread X calls for log in, this makes a request, sleep and changes threads
3.- Account B was waiting for point 2 results (point 1 and point 2 are alongside each other) parses results and needs to encounter 1 mon for IV then call next() and could possibly get account X again (thread safety is only inside that function, speed limit is ok as it is near, etc)
4.- While in thread B we call another login on X this could lead to context switch
In any case A encounter will fail due to login auth token will be invalid.

While complying with the speed limit if there is thread starvation due to high cpu usage it could even happen with points far away.

Problem 2 (only happens if context switch could happen anytime):

1.- Account A scans point 1 get 1 mon and need a IV then calls next() and get account X for it
2.- Before login there is a context switch.
3.- Some time passes (maybe a couple of minutes)
3.- Account B was waiting for point 2 results parses results and needs to encounter 1 mon for IV then call next() and could possibly get account X again (thread safety is only inside that function, speed limit is ok as it is near enough, etc)
4.- B login, encounter, etc.
5.- Thread A gets priority again tries to login, no issues but we are violating speed limit as we are traveling from B to A without any control.

Problem 3:
1.- Account A scans point 1 gets 3 mons that need encounter
2.- We call next() 3 times and log in 3 times with the same or with different accounts.

Cpu starvation is not that strange as python threading is bounded to 1 core and we are using hundreds or threads in any instance, I have seen situations where my db queue skyrockets just because the db threads doesn't get priority

sebastienvercammen:
@friscoMad That makes more sense. Fixed in next commit.

Problem 3 however would use 3 separate accounts for the scans (assuming we're not entering Problem #1), as intended. The login rate is a valid point, and one that I started a discussion about on Discord (login rate vs memory usage).

All of the threads that our legacy code uses are actually completely unnecessary. They're bad practice, slow down the process _(have you seen the number of threads we use...?!)_ and lead to unexpected behavior. Luckily they're already put on the roadmap to be reworked.

pogom/models.py
```diff
@@ -1930,34 +1935,112 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
-
-                captcha_url = encounter_result['responses']['CHECK_CHALLENGE'][
-                    'challenge_url']  # Check for captcha
-                if len(captcha_url) > 1:  # Throw warning but finish parsing
-                    log.debug('Account encountered a reCaptcha.')
+
+                # If the host has L30s in the regular account pool, we
+                # can just use the current account.
+                if level >= 30:
+                    hlvl_account = account
```
friscoMad: This is not working the rest of the ifs will need to be under else: for this to work

friscoMad:
@sebastienvercammen 
For Problem 3 I was thinking that using the same account for all encounters is the best option using the step location you can see all the mons so you can encounter them, there is no need to use 3 different accounts, the problem in that case is that you need to check the highest level needed to encounter all of them

sebastienvercammen:
@friscoMad It doesn't have to be the same account per se. If they're truly so close to one another, it will have the same result but it would spread encounters over different accounts. The problem lies in the login rate, which would be solved by storing the API object, but as mentioned before, that's being discussed.

friscoMad:
@sebastienvercammen I don't the reason why not using the same account for everything as it was done previously for all mons found in a scan using the scanning account. If the account from the pool can reach the step it can encounter all in 1 login without any violation (and we can use again the encounter delay).

pseudo code (missing a lot of checks) to explain myself:
```python
use_cp = false
use_iv = false
encounter_mons = set()
for mon in pokes:
  if mon in IV_white:
    encounter_mons.add(mon)
    use_iv = true
  else if mon in CP_white:
    encounter_mon.add(mon)
    use_cp = true

acc
if use_cp:
  acc = next(30, location)
else if use_iv:
  acc= next(25, location) or next(30, location)

login(acc)
for mon in encounter_mons:
  delay()
  encounter(acc, mon)
```

sebastienvercammen:
@friscoMad Remember you also need to consider the distance between each of the encounter_mons. It should be trivial, but needs to be checked.

In an upcoming commit, I've added a toggle to enable storing the API object to re-use it. So the host can choose to increase the login rate, or to increase memory usage.

I'll do some testing on the distances and re-using a single account for all encounters. If I can confirm our thoughts, I'll look at adding it in.

docs/extras/encounters.md
```diff
@@ -0,0 +1,53 @@
+# Pokémon Encounters
+
+Since the IV update of April 21st which makes IVs the same for players of level 25 and above, the encounter system has been reworked and now includes CP/IV scanning.
```
elincognito: Mention the year just for historical purposes

bbdoc:
@sebastienvercammen Confirmed your latest commit does solve my issue with account stopping after 1 encounter. Thanks

sebastienvercammen:
All of the previously mentioned points are now committed.

neskk:
Why so much trouble having different files and all when a simple check to get_player_level response could determine account level?
CP being included when information is sent to front-end and to the database is one thing, having separate account managers/queues, separate csv files etc. should be another thing.
I feel we really need to put accounts in database, this would drastically improve how easy we can pick an account with a specific level or close to a location instead of having accounts tied to instances.
This is a good starting point but it's a bit "clunky"... for instance, I manage 44 instance so ultimately I would need 44*3=132 csv files with accounts to enable IV scanning in all my instances which seems impractical. Also I don't know exactly how you and Seb did the lookup on the "high-level.csv" but you can't/shouldn't use the same file for all instance or accounts will be shared between instances. I saw that seb uses `'in_use'` flag for each account but it's kinda silly because another instance could have picked that account as well and since `'in_use'` is not a state shared between instances -> kaboom two logins/scans in parallel in different locations.

SiteEffect:
@neskk this is already addressed by some work on my fork. I thought exactly like you just brought up. Just, this is a first implementation to keep the map up to date with what Niantic pushs into the scanners way. Improvement is coming step by step.

sebastienvercammen:
@neskk Because the implementation we decided to go for is one that reduces the amount of integration required with legacy code. This implementation is simple and very easy to separate. This includes not reworking the legacy code to first send an additional request just for its level.

A proper rework includes reworking account handling, and ultimately scan handling. This has been talked about several times, including in this thread.

High level account files shouldn't be shared by instances, as has always been the case for account files for the past 10+ months, and I'll only repeat this one last time: we need to rework account handling to allow account file sharing and to properly distribute accounts. But that's outside of the scope of this PR.

`in_use` isn't a flag to share between instances, it's to avoid one account being used in two separate threads because of thread starvation.

And it would be 88 rather than 132 files with your 44 instances, because we only have two account files: one for accounts to scan with, one for accounts to encounter with.

neskk:
@sebastienvercammen upon login (first_login), get_player should always be requested to check warning flags, check if account is banned, check tutorial state (remove necessity for -tut flag) and perhaps try to collect level up rewards (something that current -tut with pokestop spinning doesn't do). These are all things easy to implement, I did it in less than an hour and one extra request upon first_login isn't relevant in the overall scheme of things.

sebastienvercammen:
@neskk It is relevant when it only adds more code to the list of code that needs to be scrapped, because you'd be reimplementing the same duplicate code yet once again, adding one more place to get rid of it.

And if you implement it with an abstraction, you might create an abstraction that fits for your little rework but not for the proper account handling/scanning rework.

I said I wouldn't repeat myself again but I indirectly have so I'm leaving that discussion.

neskk:
@sebastienvercammen impressive how you can write so much and tell so little.
Duplicate code? Just remove get_player_level() (account.py) and instead redo the method to parse player data into the account dict: `account['level']` -> parse_map() already receives account dict as param, so it's an improvement to remove get_player_level() there because you don't have to parse player_level in each parse_map() call as it does now.
This new method "get_player_state()" would be called only upon first login to save player level, tutorials, warnings, etc for later because it can be useful in several stuff including CP/IV scanning.

PhilConnor:
Niantic seems to have changed something recently that might affect this PR. https://www.reddit.com/r/TheSilphRoad/comments/67ypx2/different_stats_and_moves_between_level_26_and/
https://www.reddit.com/r/TheSilphRoad/comments/67yuzc/it_seems_pokemon_spawn_stats_may_now_be_entirely/

Sttug:
The resent changes make this pr more simple, just only lvl 30+ acc in the pool ?

jorgitoh:
@Sttug Actually we can use the current RocketMap code and add the column CP. The current encounter grab the CP but dont save in the DB.

SiteEffect:
Adjust title ;-)

PR #1964, opened 2017-04-04
More verbosity when clearing the database & other -cd changes

jagauthier:
Cleaning the database (-cd) can sometimes take a while.  There isn't currently any information to the end user to indicate what is happening.   

This code will report table by table as it deletes and creates them at startup.

Additionally, two other things are changed.
First, if -cd is specified the blacklist is not downloaded. This just saves time.
Secondly, when -cd is complete the process exits.

People run -cd and then stop the process.  This makes -cd more of a utility function intended for one
time run, and not to have the instance running with it enabled.

## Motivation and Context
I'm a big fan of bringing visibility to the users of the application.  I'm also a big fan of implementing code changes that answer questions I see in #help.  

## How Has This Been Tested?
Tested in my local dev environment of a copy of my production database.
```
2017-04-04 13:53:47,038 [        MainThread][           app][    INFO] Blacklist disabled for this session.
2017-04-04 13:53:47,040 [        MainThread][        models][    INFO] Connecting to MySQL database on localhost:3306...
2017-04-04 13:53:47,040 [        MainThread][     runserver][    INFO] Clearing database
2017-04-04 13:53:47,042 [        MainThread][        models][    INFO] Dropping table: Pokemon
2017-04-04 13:53:47,072 [        MainThread][        models][    INFO] Dropping table: Pokestop
2017-04-04 13:53:47,134 [        MainThread][        models][    INFO] Dropping table: Gym
2017-04-04 13:53:47,298 [        MainThread][        models][    INFO] Dropping table: ScannedLocation
2017-04-04 13:53:47,359 [        MainThread][        models][    INFO] Dropping table: Versions
2017-04-04 13:53:47,400 [        MainThread][        models][    INFO] Dropping table: GymDetails
2017-04-04 13:53:47,441 [        MainThread][        models][    INFO] Dropping table: GymMember
2017-04-04 13:53:47,483 [        MainThread][        models][    INFO] Dropping table: GymPokemon
2017-04-04 13:53:47,524 [        MainThread][        models][    INFO] Dropping table: Trainer
2017-04-04 13:53:47,565 [        MainThread][        models][    INFO] Dropping table: MainWorker
2017-04-04 13:53:47,606 [        MainThread][        models][    INFO] Dropping table: WorkerStatus
2017-04-04 13:53:47,688 [        MainThread][        models][    INFO] Dropping table: SpawnPoint
2017-04-04 13:53:47,811 [        MainThread][        models][    INFO] Dropping table: ScanSpawnPoint
[...]

PR #1983, opened 2017-04-24
Lower default startup verbosity regarding DB table creation

Arengo:
This PR prevents misleading verbosity of  ``Creating table: XYZ`` for each table, which let's some users think they just overwrote their database or stuff like that. Functionality of the quite new DB verbosity is still maintained for ``-cd`` flag as well as new Databases (referring to #1964).
With the additional ``-v`` flag all the current existing (and some of them misleading) log entries are still present, displaying the actual code work at the DB. Also, a table is only dropped if already there, otherwise pewee spits out a ugly warning (not an error) message.

## Description

The PR checks for a already existing or not existing table before outputting a log info for creation or not creation ``table.table_exists()``. 

## Motivation and Context

Some were annoyed ¯\\__(ツ)_/¯

## How Has This Been Tested?

My local setup with all kinds of empty/full, deleted Databases or single tables.

## Types of changes

- [x] Text fix and preventing site-package warning message

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2492,11 +2492,12 @@ def create_tables(db):
               Token, LocationAltitude]
     for table in tables:
         if not table.table_exists():
-            log.info("Creating table: %s", table.__name__)
+            log.info('Creating table: %s', table.__name__)
+            db.create_tables([table], safe=True)
         else:
-            log.debug("Table already exists: %s", table.__name__)
-        db.create_tables([table], safe=True)
-        db.close()
+            log.debug('Skipping table %s, it already exists.', table.__name__)
+
+    db.close()
 
 
 def drop_tables(db):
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -2491,7 +2491,10 @@ def create_tables(db):
               SpawnPoint, ScanSpawnPoint, SpawnpointDetectionData,
               Token, LocationAltitude]
     for table in tables:
-        log.info("Creating table: %s", table.__name__)
+        if not table.table_exists():
+            log.info("Creating table: %s", table.__name__)
+        else:
+            log.debug("Creating table: %s", table.__name__)
```
sebastienvercammen: If it already exists, just log that it already exists?

Arengo: Not a particular bad idea :-P

tomballgithub:
Looks good...

PR #1984, opened 2017-04-24
Fix some issues caused in #1960.

sebastienvercammen:
## Description
Conversion from list to counter for memory efficiency in #1960 was done improperly, making us send empty counters. An element that we use later on was also deleted prematurely.

## Motivation and Context
Took a look after getting reports of issues with "Nothing to scan" spam.

## How Has This Been Tested?
Tested by Smurfies @ Discord on his instances.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1802,13 +1802,14 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # if a scan was actually bad.
         if config['parse_pokemon']:
             wild_pokemon += cell.get('wild_pokemons', [])
-        else:
-            wild_pokemon_count += len(cell.get('wild_pokemons', []))
+
+        # Update count regardless of Pokémon parsing or not, we need the count.
+        # Length is O(1).
+        wild_pokemon_count += len(cell.get('wild_pokemons', []))
+        forts_count += len(cell.get('forts', []))
 
         if config['parse_pokestops'] or config['parse_gyms']:
             forts += cell.get('forts', [])
-        else:
-            forts_count += len(cell.get('forts', []))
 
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -1999,18 +2001,16 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
                         'verified': SpawnPoint.tth_found(sp),
                         'seconds_until_despawn': seconds_until_despawn,
                         'spawn_start': start_end[0],
                         'spawn_end': start_end[1],
                         'player_level': level
-                        })
+                    })
                     wh_update_queue.put(('pokemon', wh_poke))
-        # Helping out the GC.
-        del wild_pokemon
```
friscoMad: The proper fix leaving this and changing 2912 to:
`if not nearby_pokemon and not wild_pokemon_count `

pogom/models.py
```diff
@@ -1795,20 +1795,22 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
-        else:
-            forts_count += len(cell.get('forts', []))
+
+        # Update count regardless of Pokémon parsing or not, we need the count.
+        # Length is O(1).
+        wild_pokemon_count += len(cell.get('wild_pokemons', []))
+        forts_count += len(cell.get('forts', []))
 
     now_secs = date_secs(now_date)
     if wild_pokemon:
```
friscoMad: This lines until 1819 does the same that this PR is doing, there is no need to do both.

I can´t see any functional change in the changes above, as counts were properly initialized to 0 and set in lines 1816-1819 in case we are parsing forts and mons. But this change improves readability so I think it is an improvement.

PR #1985, opened 2017-04-25
Separated WH LFU key caches per message type + added gym_details cache + increased WH LFU size.

### LINKED ISSUE #1982: Send 'gym_details' webhook when gym members change only.
## Expected Behavior

Sending gym_details webhook if gym members change.

## Current Behavior

If gym is modified, webhook is send.

## Possible Solution

Check if gym member change.

1.
2.
3.
4.

## Context

Sending gym_details to PokeAlarm for notifications.

## Your Environment

* Version used:
* Environment name and version (e.g. Python 2.7):
* Operating System and version (desktop or mobile):

sebastienvercammen:
## Description
Separated WH LFU key caches per message type, added `gym_details` to the key caching and increased the key cache size.

Fixes #1982.

## Motivation and Context

1. Added `gym_details` to webhook caching.
2. Avoids conflicts when multiple message types use the same identifier field (e.g. `gym` and `gym_details` both use the gym's ID).
3. Caches are separated, so the cache size is per type rather than getting filled up with whatever item is most common and not leaving space for the other types.
4. Cache size increased to better handle setups that actually use webhooks. A cache of max. 1k items is too small for practical purposes.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -385,7 +385,7 @@ def get_args():
                         type=float, default=0.25)
     parser.add_argument('-whlfu', '--wh-lfu-size',
                         help='Webhook LFU cache max size.', type=int,
-                        default=1000)
+                        default=2500)
     parser.add_argument('-whsu', '--webhook-scheduler-updates',
                         help=('Send webhook updates with scheduler status ' +
                               '(use with -wh).'),
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -4,6 +4,7 @@
 import logging
 import requests
 from datetime import datetime
+from cachetools import LFUCache
 from requests_futures.sessions import FuturesSession
 import threading
 from .utils import get_args
@@ -45,7 +46,7 @@ def send_to_webhook(session, message_type, message):
             log.exception(repr(e))
 
 
-def wh_updater(args, queue, key_cache):
+def wh_updater(args, queue, key_caches):
     wh_threshold_timer = datetime.now()
     wh_over_threshold = False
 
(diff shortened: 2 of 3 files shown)


pogom/utils.py
```diff
@@ -385,7 +385,7 @@ def get_args():
                         type=float, default=0.25)
     parser.add_argument('-whlfu', '--wh-lfu-size',
                         help='Webhook LFU cache max size.', type=int,
-                        default=1000)
+                        default=2500)
```
friscoMad: I am running a -st 41 instance without encounters with a 300 cache size without issues
The cache was previously one of the biggest elements in memory this will increase the memory consumption for pokestops and pokemon and cache all gyms and gymdetails as I don't think there are a lot of maps with that number of gyms (my st 41 has 400 gyms), it will be more efficent to just use a dict for gyms/gym details in that case.

sebastienvercammen: > it will be more efficent to just use a dict for gyms/gym details in that case

How do you think these caches work?

Default size of a cache is 1, while 2.5k is the max size. If you don't have 2.5k unique gyms or pokéstops, you won't be using memory for 2.5k items.

friscoMad: I know how caches work, I did say that it will max the size for pokemon and pokestops and just max at the number of gyms and gym details.
I was saying a dict will be more efficient, as a dict has no LFU semantic so reading and writing an item is just that, a cache will need to, at least, store usage alongside the data adding some overhead and increasing memory usage for no gain.

sebastienvercammen: @friscoMad I'm sorry, I didn't realize anyone was trying to run this on a machine with less than 235 KiB memory (assuming an integer for counting and 4 caches all filled with 2.5k items).

PR #1987, opened 2017-04-26
Minor text fixes.

Xeirwn15:
Minor text fixes.

### DIFF
--- a/templates/map.html
+++ b/templates/map.html
@@ -241,7 +241,7 @@
             </div>
               {% endif %}
             <div class="form-control switch-container" >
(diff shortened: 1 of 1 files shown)


templates/map.html
```diff
@@ -241,7 +241,7 @@
             </div>
               {% endif %}
             <div class="form-control switch-container" >
-              <h3>Start map at me</h3>
+              <h3>Start map at my position</h3>
```
Arengo: Normally I'd vote for a ``.`` but this time, it's ok ;-)

PR #1988, opened 2017-04-26
Move get_args() outside of send_to_webhook call

jagauthier:
## Description
send_to_webhook is a frequently used function, and I realized that each call is 
parsing the arguments again.  After reviewing get_args() I couldn't see any reason
it needed to be process for each call.
The only two members of args that are referenced are args.wh_timeout and args.webooks
These two items are generated from config/command line and never change.

So, calling this inside send_to_webhook() doesn't provide the function with any new information that might be needed.

## Motivation and Context
More zooms.

## How Has This Been Tested?
Tested in my local development map environment for a few hours.


jagauthier:
Nice! I actually saw it was memoized, and read about what that means... and then completely forgot.

PR #1991, opened 2017-04-28
Add Pokémon level to front-end + save cp_multiplier to DB & send it to webhooks.

tallypokemap:
## Description
**Seb edit:** This saves `cp_multiplier` to the database and uses the `cp_multiplier > level` map to determine Pokémon level. We then show the Pokémon level in the front-end and send `cp_multiplier` to the webhooks.

## Motivation and Context
It is very nice not only to see what the CP of a given pokemon is but how high it is compared to other pokemon of that species.  Level 30 pokemon are the highest you can encounter in the wild so while a lower CP may mask that fact.

## How Has This Been Tested?
I tested this on my 7x40 step setup on windows 7.  Requires a db update.

## Types of changes

## Checklist:


SkOODaT:
This is a nice feature to add since we have CP now BUT i think this should all just be handled by the map , webhook the calculated pokemon level from CP we shouldn't need to pass cp_multiplier to webhooks unless you really want the multiplier for another reason in the webhooks :)~

relatively a small chuck of code ... then log pokemon level directly into the db

```
def calc_pokemon_level(pokemon_info):
    cpm = pokemon_info["cp_multiplier"]
    if cpm < 0.734:
        pokemon_level = 58.35178527 * cpm * cpm - 2.838007664 * cpm + 0.8539209906
    else:
        pokemon_level = 171.0112688 * cpm - 95.20425243
    pokemon_level = (round(pokemon_level) * 2) / 2.0
    return pokemon_level
```
The snippet is out of a fork im using of RM but yall get the idea
i guess on the flip side letting an external service handle it would be a small load off the map :)

Bart274:
I like the idea of @SkOODaT to calculate the pokemon level in the map itself and save that in the db instead. Also show the pokemon level on the map itself and send it to the webhook would be great.

SiteEffect:
@Bart274 normally I'd say yes. But we strictly save the values directly from the API and this should be the case for ``cp_multiplier``, too. So, @tallypokemap  please leave it that way for the DB. Also you can add cp_multipler front-end code to determine level. Either use @SkOODaT way to calculate or have a look into [this](https://github.com/Arengo/RocketMap/blob/F-SnipeCP/static/js/map.js#L697).

Bart274:
@Arengo thanks for the explanation 😊

sebastienvercammen:
Pokémon level was already going to be added to the map itself in a separate PR (since the IV/CP scan update). However, adding it as was done in this PR is not how it was intended.

The cp_multiplier should be used to determine Pokémon level, to then show on the front-end and to send it to webhooks.

@tallypokemap Are you up for changing your PR to do the above, or should someone else take over?

tallypokemap:
I think I can handle it.  I'll work on it this afternoon.

- [x] 1. Calc level for webhook but leave cp_multiplier as is in database.
- [x] 2. Display level on frontend

sebastienvercammen:
@tallypokemap Exactly 👍

tallypokemap:
Should be good!

sebastienvercammen:
Updated title, labels and PR description. Ready for review.

pogom/utils.py
```diff
@@ -973,3 +973,13 @@ def clear_dict_response(response, keep_inventory=False):
...
+
+
+def calc_pokemon_level(cp_multiplier):
+    if cp_multiplier < 0.734:
+        pokemon_level = (58.35178527 * cp_multiplier * cp_multiplier -
+                         2.838007664 * cp_multiplier + 0.8539209906)
+    else:
+        pokemon_level = 171.0112688 * cp_multiplier - 95.20425243
+    pokemon_level = int((round(pokemon_level) * 2) / 2)
+    return pokemon_level
```
sebastienvercammen: Function won't work when `cp_multiplier = None`. Add:
```
if cp_multiplier is None:
    return None
```

static/js/map.js
```diff
@@ -695,6 +699,18 @@ function getIv(atk, def, stm) {
...
+function getPokemonLevel(cpMultiplier) {
+    var pokemonLevel = 40
+    cpMultiplier = Math.round(cpMultiplier * 1000000) / 1000000
+    $.each(cpMultipliers, function (lvl, cpm) {
+        if (cpMultiplier <= cpm) {
+            pokemonLevel = pokemonLevel > parseFloat(lvl) ? parseFloat(lvl) : pokemonLevel
+        }
+    })
+
+    return pokemonLevel
```
sebastienvercammen: You use one way to calculate Pokémon level in the Python code, but another in JS? Be more consistent. It also has the same issue: if `cpMultiplier` isn't a number (null, undefined), you will get an error.

pogom/models.py
```diff
@@ -2113,7 +2118,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         'seconds_until_despawn': seconds_until_despawn,
                         'spawn_start': start_end[0],
                         'spawn_end': start_end[1],
-                        'player_level': encounter_level
+                        'player_level': encounter_level,
+                        'pokemon_level': calc_pokemon_level(wh_poke[
```
elincognito: We need to verify if encounter_level is >= 30 before call `calc_pokemon_level` otherwise we're calling it without a valid cp_multiplier

Arengo: See [Line 2124](https://github.com/RocketMap/RocketMap/pull/1991/files#diff-541e2b217ba4a7d7859dae2cb443b6b7R2124)

pogom/models.py
```diff
@@ -2113,7 +2118,9 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                            'cp_multiplier'])
                     })
```
Arengo: Remove ``'pokemon_level': calc_pokemon_level(wh_poke['cp_multiplier'])`` from the first ``.update(..)`` and insert the following as a second ``.update(..)``:
```py
if encounter_level >= 30:
    wh_poke.update({
        'pokemon_level': calc_pokemon_level(
            wh_poke['cp_multiplier'])
    })
```

pogom/utils.py
```diff
@@ -973,3 +973,13 @@ def clear_dict_response(response, keep_inventory=False):
     if 'GET_BUDDY_WALKED' in response['responses']:
         del response['responses']['GET_BUDDY_WALKED']
     return response
+
+
+def calc_pokemon_level(cp_multiplier):
+    if cp_multiplier < 0.734:
```
Arengo: Since this will get you an error in the cases @sebastienvercammen pointed out, how about using ``isinstance(var, type)``, so in this case you could use
```py
if isinstance(cp_multiplier, float):
    ...
else:
   return 0

File: static/js/map.js (hunk 2 above)
Comment (Arengo) [inline static/js/map.js]: While the python method is easier to calculate, it relies on linear regression analysis when the formula was derived obtained. The JS method is with strict API responses a bit more robust but has some design flaws, yes it is from my code and I got it like you got it from me. So, let's make it better. The whole issue is, that the mapped numbers in JS do not have the same precision like the API floats which resulted in the strange ``round(..)`` workaround. Any ideas for a more beautiful solution? Also, like @sebastienvercammen pointed out, we need a check for forbidden values:
```js
if (cpMultiplier !== null) {
    ...
}
```

tallypokemap:
Still working on this, but some things have come up today.  I'll have updates tomorrow.  Just don't want anyone to think I've abandoned this.

tallypokemap:
Task list:
- [x] Fix what happens if an invalid cpm gets passed
- [x] Make formulas consistent

tallypokemap:
I went ahead with the linear regression version of the formula.  When two formulas are correct and will universally yield the correct value, go with the simpler one.  I added only update the webhook with the level if worker is over level 30 in the python code.  I'm not sure why we need to have a check for a valid cpm in the js code since it is only called if the previous check passes 
```
if (cp !== null && cpMultiplier !== null)
```
Though I may be missing something obvious here since I am not as well versed in js as I am in python.

pogom/models.py
```diff
@@ -2137,8 +2141,13 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         'seconds_until_despawn': seconds_until_despawn,
                         'spawn_start': start_end[0],
                         'spawn_end': start_end[1],
-                        'player_level': encounter_level
+                        'player_level': encounter_level,
```
Arengo: You don't need that comma for a final entry.

tallypokemap: Good catch, I'm surprised it passed Travis

pogom/models.py
```diff
@@ -30,7 +30,7 @@
 from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, \
```
Alderon86: For consistency, if you need multi-line imports, use:

`from .utils import (get_pokemon_name, get_pokemon_rarity, get_pokemon_types,
     get_args, cellid, in_radius, date_secs, clock_between, secs_between,
     get_move_name, get_move_damage, get_move_energy, get_move_type,
     clear_dict_response, calc_pokemon_level)`

pogom/utils.py
Alderon86: i think this part is not indendet properly.

pogom/models.py
tallypokemap: I mean I can change that, but that was existing code.  I didn't write the code that way.  Would it be an appropriate change in this PR?

pogom/utils.py
tallypokemap: Which line do you think isn't indented properly?  It compiles and runs perfect on my machine and with Travis.  I don't see the bad indent unless you are talking about line 981, in which case, that is how pep8 recommends, I believe.  Where should I break the line and how far should it be indented?

pogom/models.py
sebastienvercammen: @tallypokemap Yup, go for it.

tallypokemap: Sweet, done.

pogom/utils.py
Alderon86: maybe im wrong here but shouldnt it be:
`else:
        pokemon_level = 171.0112688 * cp_multiplier - 95.20425243
        pokemon_level = int((round(pokemon_level) * 2) / 2)
        return pokemon_level`

tallypokemap: No line 984 needs to be executed whether the if statement is true or not which is why it is outside the if statement
If line 985 were indented then when cp_multiplier < 0.734 then it would never return a pokemon_level

Alderon86: gotcha! 👍

jesshub:
`var LevelCP = ['0.0939999967813','0.166397869587','0.215732470155','0.255720049143','0.290249884129',
'0.321087598801','0.349212676287','0.375235587358','0.399567276239','0.422500014305',
'0.443107545376','0.462798386812','0.481684952974','0.499858438969','0.517393946648',
'0.534354329109','0.550792694092','0.566754519939','0.582278907299','0.597400009632',
'0.612157285213','0.62656712532','0.640652954578','0.654435634613','0.667934000492',
'0.68116492033','0.694143652916','0.706884205341','0.719399094582','0.731700003147']`

```
var cpmultiplier = item['cp_multiplier']
if (cpmultiplier != null) {
        var cpLevel = `| LVL: <b>${LevelCP.indexOf(cpmultiplier)+1}</b>`
    }
```

image

Much simpler code.

pogom/utils.py
```diff
@@ -973,3 +973,13 @@ def clear_dict_response(response, keep_inventory=False):
     if 'GET_BUDDY_WALKED' in response['responses']:
         del response['responses']['GET_BUDDY_WALKED']
     return response
+
+
+def calc_pokemon_level(cp_multiplier):
```
elincognito: If there are no LVL 30 accounts the encounter will be skipped .
and when we reach the `calc_pokemon_level`  function `wh_poke['cp_multiplier']` will be `None` that will give the following exception
`
Exception message: TypeError("unsupported operand type(s) for *: 'float' and 'NoneType'",)
  File "/home/***/RocketMap/pogom/search.py", line 987, in search_worker_thread
    scan_date, account, account_sets)
  File "/home/****/RocketMap/pogom/models.py", line 2193, in parse_map
    wh_poke['cp_multiplier'])
  File "/home/*****/RocketMap/pogom/utils.py", line 981, in calc_pokemon_level
    2.838007664 * cp_multiplier + 0.8539209906)
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
`

I recommend either check for None for  `cp_multiplier` on `calc_pokemon_level` and return `None` from it or
change
[models.py:L2147](https://github.com/tallypokemap/RocketMap/blob/821c8aff956f1b8f6dcd37e4c504ad13e670bbb4/pogom/models.py#L2147) to check if `encounter_result` is None:


    if encounter_level >= 30 and encounter_result is not None:
                        wh_poke.update({
                            'pokemon_level': calc_pokemon_level(
                                wh_poke['cp_multiplier'])
                        })

tallypokemap: Should be fixed in latest commit.  Changed models.py:L2147 to:
`if wh_poke['cp_multiplier'] is not None:
                    wh_poke.update({
                        'pokemon_level': calc_pokemon_level(
                            wh_poke['cp_multiplier'])
                    })`

pogom/models.py
```diff
@@ -2132,6 +2135,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 individual_defense = pokemon_info.get('individual_defense', 0)
                 individual_stamina = pokemon_info.get('individual_stamina', 0)
                 cp = pokemon_info.get('cp', None)
+                cp_multiplier = pokemon_info.get('cp_multiplier', None)
```
Obihoernchen: This isn't needed here because you don't log cp_multiplier.
Remove it and use the following in L2166:
```python
pokemon[p['encounter_id']]['cp_multiplier'] = pokemon_info.get('cp_multiplier', None)
```

tallypokemap: Totally correct, fixed.

Oxyde59:
Just try this PR and work well.
Just a little thing : I lost the Gym Sidebar

tallypokemap:
@Oxyde59 Was this from a clean pull with only this PR merged and no other modifications?

sebastienvercammen:
@tallypokemap @Oxyde59 Sounds unrelated to this PR. This PR doesn't change any gym or gym sidebar code.

Oxyde59:
@sebastienvercammen Yep, I don't understand why too, but it's the only PR I'm testing actually, and got sidebar just before. Test on Edge and Chrome, before and after cleaning browser cache

Oxyde59:
@tallypokemap yep, only PR I test actually.
I just "git pull" this morning before fetching to #1991 for the .63 update, perhaps it's that ?

sebastienvercammen:
@Oxyde59 @tallypokemap It's this: https://github.com/RocketMap/RocketMap/commit/fffc3c2d8525868dc72cde1304adaeff1b9fe54e. It's being looked at.

sebastienvercammen:
@Alderon86 @Obihoernchen Can you both review and update your feedback (approval or request changes)? If there's nothing left to change, this can be merged.

PR #1994, opened 2017-04-29
Adding Geofence Support (re-written from PR #1688)

Arengo:
This PR introduces geofence support for RocketMap, re-written from the implementation of @birdstream in #1688 and enhanced with additional functionality.

So, first of all: **I would like to thank @birdstream** coming up with this idea of implementation and making a great approach to a important feature for this project. I have been communicating with him and he is cool guy. Since his possibilities of participation have slimmed down due to a lack of time, I overtook the idea and tried to make an implementation for proper testing with the goal of a final merge.

The following has been implemented:

- parsing from files of geofences or forbidden areas
- filtering scanning cells for all current scan methods, also ``-ss``
- ~~adding them to the DB to display them in the front-end~~ (re-adding in a different PR later on)
- ~~front-end display with toggle and random green/red colours~~ (re-adding in a different PR later on)
- optional support without ``matplotlib`` on incompatible devices

## Description

To geofence your desired area, you have to add at least on of the following flags to your configuration or command line:

- ``-gf PATH/FILE.txt`` or ``--geofence-file PATH/FILE.txt``
- ``-ff PATH/FILE.txt`` or ``--forbidden-file PATH/FILE.txt``

These files have to be formatted in this way (this is compatible with PokeAlarm geofence file format) while ``lat`` and ``lng`` and their indices are placeholders for actual coordinates, drawable with one of these tools: 
- [basic](http://codepen.io/bgus/full/dXxLjp/)
- [improved](https://codepen.io/jennerpalacios/full/mWWVeJ), thanks @imp3tus 
- [advanced](http://www.headwallphotonics.com/polygon-tool), thanks @SkOODaT
- [userfriendly, much functionality](http://geo.jasparke.net/), thanks @Boby360

```
[GeofenceName1]
lat11, lng11
lat12, lng12
...

[GeofenceName2]
lat21, lng21
lat22, lng22
...
```

Please have in mind that the header lines ``[GeofenceNameX]`` have to be present with your custom name. It does not matter how many empty lines you have in between multiple geofences in one file.

The provided files are then being parsed into an ``Geofences`` class object. Starting your desired scanning method, all possible cells are filtered against the geofences, checking them for being a valid cell.

The scanning starts as normally and no further action of this code is taking place. So the whole back-end implementation is being processed during the thread start-up.

To get an idea of which area is being geofenced, the front-end could ~~now~~ visualize your provided geofences. You can see screenshots below. (re-adding in a different PR later on)

## Motivation and Context

Geofences are a very strong tool to getting exact that area scanned, you want to. Personally, I have used #1688 a lot, being a nice first implementation. Through this, I made interesting scanning shapes possible like outlining exactly city or coast borders, preventing scans on company or dangerous areas and even making scanning shapes like lines (giving a big ``-st`` combined with the fitting geofence) which was not possible without geofences. The computational speed is rather fast, which you can watch in the log. The mighty ``matplotlib`` package has of course advantage over the custom implementation of a point in polygon problem which was found in the web. With ``matplotlib``, big ``-st`` with big and complicated geofences can be filtered quite fast but also on small devices, like Raspberry Pi, which sometimes don't support ``matplotlib`` can now operate geofence filters with the config flag

- ``-nmpl`` or ``--no-matplotlib``

## How Has This Been Tested?

On my own map, flawlessly. 
So I encourage you to find out possible weaknesses or errors. Please focus on:

- Large areas (with and without ``matplotlib``)
- Complicated geofence configurations
- ``matplotlib`` incompatible devices
- misc. stuff you will probably encounter somewhere

## Screenshots:
**Jacqueline Kennedy Onassis Reservoir** as example to prevent empty scans on a e. g. lake
**Wards Island** as an example to outline an island surrounded by water
bildschirmfoto 2017-04-29 um 20 13 24

**Manhattan** as area of interest with hotspots like the battery park
**Ground Zero** as an example of prohibited, dangerous or not suited area to play
bildschirmfoto 2017-04-29 um 20 13 41

**Scanning in defined geofence** 
bildschirmfoto 2017-04-29 um 21 41 57

**Scanning around forbidden area**
bildschirmfoto 2017-04-29 um 20 53 37

**EDIT:** Front-end was removed for including it later on in a separate PR into the all-new RocketMap front-end.

## Types of changes

## Checklist:

- [x] My change requires a change to the configuration file. (``config.ini.example``)
- [x] I have updated the configuration file accordingly. (``config.ini.example``)


birdstream:
Really nice work here! :)

pogom/geofence.py
```diff
@@ -0,0 +1,278 @@
...
+                    'lat': locations[i]['lat'],
+                    'lon': locations[i]['lng']
+                }
+                for geofence in self.geofences:
+                    if not self.geofences[geofence]['forbidden']:  # Geofences
+                        if not self.args.no_matplotlib:  # Matlplotlib
+                            if not self.point_in_polygon_matplotlib(
+                                    point,
+                                    self.geofences[geofence]['polygon']):
+                                del locations[i]
```
Arengo: Check all ``del`` if list-entry is available to delete to catch exceptions.

imp3tus:
fantastic update to the original PR, i do recommend a minor change in your geofence tool linking, https://codepen.io/jennerpalacios/full/mWWVeJ - has better layout, and lighter shades to give better view of a geofenced area, incase they want to geofence within a geofence.

SiteEffect:
@imp3tus thank you very much. I adjusted the links accordingly.

.gitignore
```diff
@@ -7,6 +7,11 @@
 *.json
 config/*
 !config/config.ini.example
+accounts/*
+!accounts/accounts.csv.example
```
BrownSlaughter: are these mentions of accounts part of this PR or left overs from something else?

accounts/accounts.csv.example
```diff
@@ -0,0 +1,3 @@
+ptc,username1,password1
+ptc,username2,password2
+...
```
BrownSlaughter: is this accounts.csv part of this PR or left overs from something else?

pogom/search.py
```diff
@@ -371,7 +371,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     account_sets.create_set('30', args.accounts_L30)
 
     # Debug.
-    log.info('Added %s accounts to the L30 pool.', args.accounts_L30)
+    log.debug('Added %s accounts to the L30 pool.', args.accounts_L30)
```
BrownSlaughter: was this part of the geofences or just a change that was made?

.gitignore
Arengo: Kind of both, was an left over but I thought: Not bad, let's try. We do the same for config.ini.example and the config directory.

accounts/accounts.csv.example
Arengo: Kind of both, was an left over but I thought: Not bad, let's try. We do the same for ``config.ini.example`` and the config directory.

pogom/search.py
Arengo: You know the answer :-P It was a clear debugging message which popped up each start-up. So we can fix it in a small fix PR or just let it be handled later.

pogom/geofence.py
```diff
@@ -0,0 +1,278 @@
...
+log = logging.getLogger(__name__)
+
+
+class Geofences:
+    def __init__(self, args, db_updates_queue):
+        self.args = args
+        self.geofences = {}
+        self.db_updates_queue = db_updates_queue
+
+        Geofence.clear_all()  # Remove old geofences from DB.
```
Arengo: Happy about feedback, here. With multiple instances and totally different geofence files only the last instance geofences are then being displayed. However, I like putting all geofences - for all instances which write to the same DB - in one file.
This would need a new flag to clear existing DB geofences, if changed. Any other ideas or feedback?

pogom/geofence.py
Arengo: How about that?
```python
try:
    array[i]
except IndexError:
    exampleHandle()
```

pogom/geofence.py
Arengo: I just pushed the ``purge-geofence-data`` flag and thinking about rewriting all of the same name while not touching others.

SiteEffect:
The index error when using multiple geofences is fixed. 
All different combinations work:
- no geofences
- only valid area
- only forbidden area
- valid and forbidden area
- combinations of above with multiple areas of each

There is a new flag:
``-pgd`` or ``--purge-geofence-data`` this is to only manually reset geofences in the DB and to be able to display all provided geofences of multiple instances in the front-end. This is not yet final state because there might be problems in the DB with changing single coordinates of same named geofences. This will be addressed later, checking for existing name in db, deleting all of that name and rewriting with the new ones. You can use -pgd, once, when making geofences changes across your instances. I'd like to remind you that this is a front-end thing. The back-end part is directly handled with the read-in files.

accounts/accounts.csv.example
Arengo: @sebastienvercammen may I leave it in or shouldn't we do it at all or should I make a different PR for that?

.gitignore
Arengo: @sebastienvercammen may I leave it in or shouldn't we do it at all or should I make a different PR for that?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def parse_geofences(self):
+        geofence_data = {}
+        lenGeofenceData = 0
+        name = ''
+        log.info('Looking for geofenced or forbidden areas.')
+
+        i = 0
+        if self.geofence_file:
+            j = 0
```
Obihoernchen: please use some descriptive names for i and j.

Arengo: If that's all ;-)

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+            coordinateTouple = (c['lat'], c['lon'])
+            polygonToupleList.append(coordinateTouple)
+
+        polygonToupleList.append(polygonToupleList[0])
+        path = Path(polygonToupleList)
+
+        return path.contains_point(pointTouple)
+
+    @staticmethod
+    def point_in_polygon_custom(point, polygon):
```
Arengo: Move this to DB Geofence model?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+                    'coordinates_id': coords,
+                    'latitude': self.geofences[key]['polygon'][coords]['lat'],
+                    'longitude': self.geofences[key]['polygon'][coords]['lon']
+                }
+
+        self.db_updates_queue.put((Geofence, db_geofences))
+        log.debug('Upserted %d geofence entries.', len(db_geofences))
+
+    @staticmethod
+    def point_in_polygon_matplotlib(point, polygon):
```
Arengo: Move this to DB Geofence model?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+
+        endTime = time.time()
+        elapsedTime = endTime - startTime
+        log.info(
+            'Geofenced to %s coordinates in %.2f s.',
+            len(geofenced_coordinates), elapsedTime)
+
+        return geofenced_coordinates
+
+    def push_db_geofences(self):
```
Arengo: Move this to DB Geofence model?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+                self.valid_areas.append(self.geofences[g])
+
+    def get_geofenced_coordinates(self, coordinates):
+        log.info('Found %d coordinates to geofence.', len(coordinates))
+        geofenced_coordinates = []
+        startTime = time.time()
+        if self.valid_areas:
+            for c in coordinates:
+                if self.args.spawnpoint_scanning:
+                    point = {'lat': c['lat'], 'lon': c['lng']}
```
Arengo: Whole RocketMap code ``lon`` or ``lng`` or full words for consistency?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Initialization of object when geofence files are provided.
+        if self.args.geofence_file or self.args.forbidden_file:
+            self.geofence_file = self.args.geofence_file
+            self.forbidden_file = self.args.forbidden_file
+
+            self.parse_geofences()
+            self.push_db_geofences()
+
+    def get_geofences(self):
```
Arengo: Do we want this as get method or just get the value via ``geofence_example_object.geofences`` Should I just leave it out if not used?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+try:
+    from matplotlib.path import Path
+except ImportError as e:
+    log.warning('Exception while importing matplotlib: %s', repr(e))
+    log.warning('Enable "-nmptl" or "--no-matplotlib" to circumvent.')
+    pass
+
+
+class Geofences:
+    def __init__(self, args, db_updates_queue):
```
Arengo: Pass ``args`` to the object or import ``args`` for the whole file?

pogom/geofence.py
```diff
@@ -0,0 +1,267 @@
...
+import time
+import logging
+
+from .models import Geofence
+
+log = logging.getLogger(__name__)
+
+try:
+    from matplotlib.path import Path
+except ImportError as e:
```
Arengo: Works, but is this the fine solution? We need some talk about matplotlib, custom calculations what hardware to really support in the end.

pogom/app.py
```diff
@@ -370,6 +370,35 @@ def raw_data(self):
                             oSwLat=oSwLat, oSwLng=oSwLng,
                             oNeLat=oNeLat, oNeLng=oNeLng))
 
+        if request.args.get('geofences', 'true') == 'true':
```
Arengo: I don't like this whole following part of the implementation, any idea to make it more beautiful?

myntath:
This is great, have started testing now, would be good to have a proper error if your geofence file is incorrectly formatted. I used my old ones which were missing the [name] on the first line. The error thrown seems unrelated and quite confusing.

myntath:
untitled

Geofence not being drawn correctly when more than one separate area. This relates to a setup with three RM instances plus a separate -os instance. Looks like all geofences were assigned geofence_id=1. 

This may be a misunderstanding of how the files should be formatted on my part. I have three separate geofence files one for each instance. Reviewing your comments above perhaps I should have all fences in the one file.

SiteEffect:
@myntath thanks for testing. That is a current present conflict with multiple instances and different files, overwriting each other. Like you said, put them all into one file.

static/js/map.js
```diff
@@ -1005,6 +1031,66 @@ function setupSpawnpointMarker(item) {
...
+        disableAutoPan: true,
+        position: markerPosition
+    })
+
+    addListeners(polygon)
+
+    return polygon
+}
+
+function polygonCenter(polygon) {
```
friscoMad: Not sure if this algorithm works in every case but check http://moduscreate.com/placing-markers-inside-polygons-with-google-maps/ or https://github.com/mapbox/polylabel for alternative solutions

static/js/map.js
```diff
@@ -1005,6 +1031,66 @@ function setupSpawnpointMarker(item) {
...
+
+    var X = 0
+    var Y = 0
+    var Z = 0
+    polygon.getPath().forEach(function (vertex, inex) {
+        var lat
+        var lng
+        lat = vertex.lat()
+        lng = vertex.lng()
+        lat = lat * Math.PI / 180
```
friscoMad: Maybe it is clearer doing all this in one line: ```var lat = vertex.lat() * Math.PI / 180```

pogom/app.py
friscoMad: Maybe:
```python
d['geofences'] = {}
for g in geofences_db:
    geofence = d['geofences'].get('geofence['name']]', None);
    if not geofence: 
      geofence = {
        'forbidden': g['forbidden'],
        'name': g['name'],
        'coordinates': []
     }

    coordinate = {
      'lat': g['latitude'],
      'lng': g['longitude']
    }
   geofence['coordinates'].append(coordinate)

   d['geofences'][geofence['name']] = geofence

File: pogom/geofence.py (hunk 7 above)
Comment (Arengo) [inline pogom/geofence.py]: Done.

File: pogom/geofence.py (hunk 8 above)
Comment (Arengo) [inline pogom/geofence.py]: Done.

File: pogom/geofence.py (hunk 9 above)
Comment (Arengo) [inline pogom/geofence.py]: Done.

File: pogom/geofence.py (hunk 10 above)
Comment (Arengo) [inline pogom/geofence.py]: @sebastienvercammen let's change it consistently throughout RM code?

File: pogom/geofence.py (hunk 11 above)
Comment (Arengo) [inline pogom/geofence.py]: Removed for now.

File: pogom/geofence.py (hunk 12 above)
Comment (Arengo) [inline pogom/geofence.py]: @sebastienvercammen what is your opinion to that?

File: static/js/map.js (hunk 15 above)
Comment (Arengo) [inline static/js/map.js]: I have not encountered an issue, yet. It's based on the weight-center of an n-dimensional object.

**Comment by SiteEffect** [comment]
@myntath I pushed changes which hopefully solve your issues. Have a try.

File: pogom/geofence.py [hunk 17]
```diff
@@ -0,0 +1,184 @@
...
+            self.upsert_geofences()
+
+    def is_enabled(self):
+        enabled = False
+        if self.geofences:
+            enabled = True
+
+        return enabled
+
+    def parse_geofences(self):
```
Arengo: Duplicated code has to be reduced here. Also change geofence_data dict to a list and adjust other code accordingly.

friscoMad:
After reviewing the code I am not sure about using DB for geofences. I don't like using DB as a message system between back and front.

So while thinking about it I think there is little value to add all the geofences to the map, as it is now I can see the forbidden geofences value but there is no value for regular ones. I think it would be a lot better if there are 2 PRs:
  * Backend: Use geofencing to improve scanning area selection
  * Fronted: We could display areas with additional info in the map (forbidden areas, closing times, special alert channels, info url...) 

Both use cases make sense and can be used independently and a bonus for that is there is no need to pass static info from back to front. As Seb is working on a new frontend maybe we can just drop all frontend from this PR.

pogom/models.py
```diff
@@ -11,7 +11,7 @@
 import geopy
 import math
 from peewee import InsertQuery, \
-    Check, CompositeKey, ForeignKeyField, \
+    Check, CompositeKey, PrimaryKeyField, ForeignKeyField, \
```
Arengo: I needed something as index, so I used PrimaryKeyField, any other idea, like using two keys ``name`` and ``coordinates_id`` since those have to be unique in combination?

pogom/models.py
```diff
@@ -1753,6 +1762,122 @@ def get_valid(limit=15):
...
+        with flaskDb.database.transaction():
+            for geofence in geofences:
+                (DeleteQuery(Geofence)
+                    .where(Geofence.name == geofences[geofence]['name'])
+                    .execute())
+
+    @staticmethod
+    def get_db_entries(geofences):
+        db_geofences = {}
+        id = 0
```
Arengo: I maybe can not get rid of ``id`` as PrimaryKeyField but do I have to explicitly upserting it when it already is a auto-increment field? I'd like to leave it away in this method by upserting a list instead of a dict with ``id`` as keys for the entries.

pogom/schedulers.py
```diff
@@ -534,7 +553,9 @@ def location_changed(self, scan_location, db_update_queue):
         self.scans = scans
         db_update_queue.put((ScannedLocation, initial))
         log.info('%d steps created', len(scans))
-        self.band_spacing = int(10 * 60 / len(scans))
+        if len(scans) > 0:
```
Arengo: Probably not needed anymore since I ``sys.exit()`` when no valid cells are found.

runserver.py
```diff
@@ -272,6 +273,9 @@ def main():
         t.daemon = True
         t.start()
 
+    # Create Geofences object and save into args
+    args.geofences = Geofences(args, db_updates_queue)
```
Arengo: Using ``geofences`` object as an argument could increase memory size. This was done as easy code-wide solution but due to the memory I'd like to use another code-wide solution but only importing it when needed. Your idea?

pogom/geofence.py
```diff
@@ -0,0 +1,184 @@
...
+
+from .models import Geofence
+
+log = logging.getLogger(__name__)
+
+
+class Geofences:
+
+    def __init__(self, args, db_updates_queue):
+        self.args = args
```
friscoMad: maybe pass and store only the used params, it will not improve memory usage but it makes more clear what are we ussing and why.

SiteEffect:
@friscoMad while I agree on the part that DB as direct communication from back-end to front-end is not the best solution, I'd like to point out that we are doing the same thing with permanently changing entries in the workerstatus. Regarding visualization, I see this as a tool to identify your true scan area while giving the user an idea what area is really covered and which is explicitly not (I get tons of user questions regarding this). 
Combining it into one feature is neat, that's why I have done it but I totally see your point and we should discuss here as developers and community. However - the new front-end will change a lot of things and it is probably the best to ... 
image

runserver.py
friscoMad: Yup probably it is best to validate and upsert (if you insist 😀) on creation and then reload (or load from DB) only for get_geofenced_coordinates().
Another option that seems even more clean is just instantiate and use the geofencing service for _generate_locations() maybe make the functions that validate args static and just instantiate where it is needed so after the creation of locations it will be deleted.

Obihoernchen:
I agree.
What's about removing the frontend and DB stuff out for now so the PR is lean and easier to review / more likely to be merged.
Once we have the new frontend we can think about showing some special areas in the frontend imo. Or open a separate PR for it.

SiteEffect:
@Obihoernchen @friscoMad I did as requested by you. I removed the entire front-end part to push for a clean back-end solution and re-adding front-end visualization later on. To anyone who is interested, you can still have a look at the visualization code via [my fork branch](https://github.com/Arengo/RocketMap/tree/PR-GeofenceVisualization).

friscoMad:
Ok another refactor idea, instead of filtering on each scan method, just add filtering on schedule() if we are getting a new location list, it is only 1 point and you don't need to handle differences between spawnscan and the others.

pogom/geofence.py
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+import time
+import logging
+
+log = logging.getLogger(__name__)
+
+# Trying to import the, not to all hardware compatible, matplotlib
+# Matlplotlib is faster for big calulations
```
Obihoernchen: Missing dots (╯°□°）╯︵ ┻━┻

pogom/schedulers.py
```diff
@@ -376,6 +385,16 @@ def _generate_locations(self):
             self.locations = Pokemon.get_spawnpoints_in_hex(
                 self.scan_location, self.args.step_limit)
 
+        # Geofence spawnpoints
```
Obihoernchen: Missing dot (╯°□°）╯︵ ┻━┻

pogom/schedulers.py
```diff
@@ -585,6 +604,15 @@ def _generate_locations(self):
                     loc = get_new_coords(star_loc, xdist * (j), 210 + 60 * i)
                     results.append((loc[0], loc[1], 0))
 
+        # Geofence results
```
Obihoernchen: Missing dot (╯°□°）╯︵ ┻━┻

runserver.py
```diff
@@ -272,6 +273,9 @@ def main():
         t.daemon = True
         t.start()
 
+    # Create Geofences object and save into args
```
Obihoernchen: Missing dot (╯°□°）╯︵ ┻━┻

SiteEffect:
@friscoMad sadly this is not possible to just use it at one location since ``schedule()`` is redefined for each scanning method children of the class BaseScheduler while the BaseScheduler does not schedule locations itself.
However, pretty much handled all other existing refactoring ideas in my latest commits and it is safe for a complete new review. I tried to abstract as much as possible and only used stuff where it is needed while trying to maintaining readability and re-use of the code. Instead of initializing the geofence object in ``runserver.py`` I do it now once for the BaseScheduler and do not pass it to ``args`` which is the better option. Also args is now not passed to the ``Geofences`` class but imported once into that file for free use.
Please review.

pogom/geofence.py
```diff
@@ -0,0 +1,162 @@
...
+
+log = logging.getLogger(__name__)
+
+# Trying to import the, not to all hardware compatible, matplotlib.
+# Matlplotlib is faster for big calulations.
+try:
+    from matplotlib.path import Path
+except ImportError as e:
+    log.warning('Exception while importing matplotlib: %s', repr(e))
+    log.warning('Enable "-nmptl" or "--no-matplotlib" to circumvent.')
```
friscoMad: Not sure but using --no-matplotlib will raise the same error here, maybe it should be checked if it was defined and don't log anything in that case.

Arengo: Yup, you're right.

friscoMad:
Ok, I am fine with it, great work 👍(not tested though)

The only improvement I can think of is that the whole Geofences class it is only needed once in the lifetime unless you move the location and need to recalculate everything, the object itself is not that big but the geofence data stored iniside could be somewhat big if there are a lot of areas and/or points.
What I tried to say before (and probably did not explain properly) is that you could instantiate the class only for filtering when it is needed and then delete it to free the memory or not save valid_areas and forbidden_areas and load them only during ```get_geofenced_coordinates```. 
It is just a memory vs cpu tradeoff but I think that the function will only be called once in most installs.

SiteEffect:
@friscoMad you are right with the possible deletion of that object. I have no idea how to best put it and delete it while having it available when needed and also only one deleting line or is a 
```python
del self.geofences
```
after the filtering in the various schedulers without harming code or causing exceptions?

Killerproof:
How can i add this pull request in my RocketMap files? Sorry but i'm new in GitHub.

accounts/accounts.csv.example
FrostTheFox: @Arengo general rule is, don't make changes that you don't need to make for your PR to work.

SiteEffect:
@Killerproof best you go through a general git / GitHub guide. GitHub itself has some good resources for that. You can mess up your setup when you just try and that's why we do not support it here. Best is you just stay away and wait until PRs are merged into ~~master~~ develop branch.

friscoMad:
@Arengo 
My proposal would be moving most of the code to ```get_geofenced_coordinates```:

```python
def get_geofenced_coordinates(self, coordinates):
  valid_areas, forbidden_areas = load_geofences()
  if not valid_areas and not forbidden_areas:
    return coordinates
  do the same thing as always..
 #this could be in the calling function just another suggestion
 if not geofenced_coordinates:
   raise exception
```

That way you can store Geofences object but the data is only store while filtering.
Assigning self.valid_areas and self.forbidden_areas  to none after filtering would do the trick but then you will need to create a gettter to load the data again in case it is needed due to relocation.

KartulUdus:
Finally upgraded from 1688 to 1994:

3x `-st 70 -nmpl` with shared fence file for city 
1x `-st 40 -npml` 
1x `-st 22 -npml` 
1x not fenced `-st 40` (just to mention there was a case without a fence used 🙂 )
1 new area `-st 50` with a narrow geofence, matplotlib and initial scan. 

Running for 12 hours by now. 
No unusual or different errors in logs than usual 
Initial scan didn't take longer than expected
Didn't have staggered/clustered scans near the edge of the fence like I did with 1688 before merge of 1900. 

all geofences were in separate files and my observation for anyone coming from 1688: 
The `[Name of geofence area]` seems to now be a requirement in the geofence file. 

Thank you for keeping this one in particular going! ❤️

SiteEffect:
@KartulUdus thanks for sharing your test results. Do you have matplotlib installed on your scanning machine? I am interesting how it is working with and without installed matplotlib and how the Exception is doing, so if you would like to test a bit more, leave ``-nmpl`` out once.

SiteEffect:
@friscoMad it's not the best option, too, my guess. On the one hand we only want to read in the files once at startup, on the other hand we would like to safe memory by having the geofence data deleted after use, all while we would like to have a usable class for other purposes (in the future). So for me, it's currently the best way to just delete the object after fencing but that would kill location changes on user demand inside the area (I can't think of another location change during runtime). So, should we just initialize the geofence object inside each ``__generate_locations`` instead of BaseScheduler initialization and set it to none after the fencing, again?

friscoMad:
Yup that would work, you don't even need to del or set to none if no reference to it is stored in the Scheduler object as it will be deleted on function exit.

We only need to del or null objects in thread functions (as they never exit) or long running functions in threads (as there is more probability of the thread losing priority and the memory will not be freed until resume and exit). 

That's why I suggested delaying loading the files until they are used as we would not need to store references to the data and it will be used in the function and then discarded and loaded again if needed and you could keep the geofence class referenced in the scheduler.

jorgitoh:
How we should use ST?. Draw an area and then calculate which ST will cover that area?.

SiteEffect:
@jorgitoh you can guess your geofence size after drawing it, define a good center point based on your geofences and have a look of the needed -st size with the [loc-Tool](https://voxx.github.io/pgm-multiloc/). You should reach any maximum outer point with the -st. Even if you throw a bit above it does not harm, just might take longer on initial calculations.

tomballgithub:
Is this working well for people?  I added this PR and started getting database issues.  I turned off the instances using geofence and things are better (but I also have 3 less instances of load I admit).   I ran 1688 for months with zero issues

SiteEffect:
@tomballgithub could you please provide details to your DB issues? The database is not used at all for this anymore. Also which configuration and scan-scheduler do you use (I'd just assume ``-speed`` but one never know)?
**EDIT:** Seems you deleted a comment below. Is this problem still persistent or wiped out for you

tomballgithub:
I suspected that would be the response I received.  I'll reproduce again tonight and look into it.  But basically I get DB locks, each instance starts getting behind on DB writes, and the the system utilization skyrockets and becomes unusable.  Only two changes I made were upgrading from April 28 code base and adding this PR.  I ran overnight with current code base and without PR (but also 3 less instances).   I have been running the total of 15 instances for 6 months without issue.  Not sure what is going on.  But again, I'll retest yet again.

tomballgithub:
I am still struggling with database issues that I never had before.  The only 2 changes are moving from 4/28 codebase to latest codebase, and this PR instead of the old geofence PR.  Everything is fine until I fire up my larger -ST instances with geofence which add 50 workers each.  At some future point, I check back and I'm at 100% utilization and getting database errors.  But the base codebase is stable so it's either my database or this PR OR it's a specific instance with specific geofence/spawnpoints that is the issue and bringing out a bug?    I am trying to fire up instances one by one and see if I can get an idea of where the problem is

tomballgithub:
The only thing unusual at all about my geofences are I have the last coordinate the same as the first in my geofence file.  The 'improved' tool listed in intro doesn't do that, so I just removed it

cptmedic:
Just wanted to report that I am running four maps with geofences and have not ran into any database issues. Two map has been running for a solid week, the other two I started a little over a day ago with no issues either. Using develop branch and this PR only.

tomballgithub:
Ran most of today without issues with 182 worker without geofences.  Added two geofence instances of 50 each and watched utilization.  CPU was in 15-25% range.  Memory usage 42%.  No problems, no problems.  Come back a bit later (< 1 hr) and boom 100% utilization.  I think the database queue getting backed up is an artifact of the utilization making system unusable. Last two things I can think of trying are reset my database (I already deleted the pokemon table) or try 300 workers without geofence

cptmedic:
Hmm, I am not sure honestly. When I started using this PR it was with new databases, my CPU has stayed in the same range which is 15-20% with the maps I currently run, never has the usage went to 100% though. Looks like the only difference may be the databases?

tomballgithub:
I am restarting everything from scratch with no geofence and no PR's whatsoever.  Will then add geofence.

ghost:
@tomballgithub are you using -bh ? That's the only thing that's causing me issue with this PR.

ghost:
@Arengo is there a way for us to incorporate -bh in the same instance that we're using the geofence in?

SiteEffect:
@blakewenloe since I have never used -bh with the introduction of -speed, I do not know how -bh behaves ad-hoc with that PR. I, ofc, see the possibility to try supporting it since my goal was to support any of the schedulers. What current issues are known with -bh and this PR? 
**EDIT:** I just tested various configurations of -bh with and without speed scheduler. They seem to work as intended. If there are explicit problems with -bh please provide a configuration and setup to test that I can bug-fix.

tomballgithub:
My issues are unrelated to this PR.  Look forward to resolving them so I can test this.

ghost:
@Arengo I just tried again with a fresh install. When running with -bh, I get:

`2017-05-15 04:41:00,417 [   search-overseer][    schedulers][   ERROR] No cells regarded as valid for desired scan area. Check your provided geofences. Aborting.`

When I disable -bh, it runs normally.

SiteEffect:
@blakewenloe again, I'd really like to address this issue, if it's the case for this PR. Unfortunately I am not able to reproduce unintended behaviour, thus I highly recommend to send me (@Arengo in Discord, as well), your personal setup for this PR test that I can address it. In my tests with -bh I specified an area where not all sub-hives of the beehive are covered, in the end it fences correctly and scanning the cells inside the geofence while sub-hives are neglected which did not fit the requirements. It might be the case that the introduced system exit prevented further scanning thus the following might solve the problem. But please, let me try your setup first to make sure what really is the problem.
```py
if not args.beehive:
   sys.exit()
```

tomballgithub:
I have been running well with this PR once I resolved stability issues with my VPS.  I do not have the expertise to review this code.

unholygod91:
@blakewenloe  @Arengo  I've seen this error before when I first tested this PR and the problem wasn't caused by -bh. I do not have time tonight to attempt to reproduce it, but if your location is set outside of the -gf it will throw this error on start. Hopefully this helps but I've been running this PR for awhile now and  none of my errors have been caused by it on any setup. Thank you for the work.

SchroedingersKater:
@unholygod91 thanks for pointing out your experience. That might be a thing, that starting location inside/outside the fence impacts a successful filtering. So, that needs further testing: How does the scheduler behave in that aspect when only a small part of the hive is inside the geofence, while it's center is outside. However, my guess is that you can always position your hive-center inside and adjust -st accordingly to still cover the desired area.

jorgitoh:
Hi @Arengo. Actually im using this setup: -bh -st 7 -w 40-sd 15 -np -nk -gi -ps. Can I use geofence if I want a bigger area like -bh -st 7 -w 100 but filtering areas?

SiteEffect:
@jorgitoh so it's just a gym-scanner. You can try it for sure. But like others wrote here, there might be an issue with root-location of some hives if they are not in already covered are. Unsure though, since untested.

kinar:
Hopefully this is useful info to help track down the issue with gymscans.  

TL;DR: Geofences work fine for gym only scans on single cell beehives but fail when adding multiple instances (I think..not fully tested).

I just added a geofence to my Gymscan instance:  relevant config info is as follows:

beehive: true
step-limit: 25
gym-info: true
#account-rest-interval: 1800
#account-search-interval: 5400
workers-per-hive: 25
workers: 200
login-delay: 15
no-pokemon: true
#no-pokestops: true

This produces a beehive consisting of the center, 1 complete ring plus one addition cell in the 2nd ring.  The farthest cell is ~35miles from the center.  Overall, it covers a lot of small rural towns.

After adding multiple geofences to a single file, to surround the towns, I got the error posted above by @blakewenloe when starting.

However, removing all but the center cell by putting all workers into a single cell and not altering the geofence file:

workers-per-hive: 200
workers: 200

This, of course, drops all of the cells except for the center cell but the geofences that exist only in the center cell load and scan just fine.

Happy to test any additional requested changes to help track this down if it's something @Arengo wants to tackle with this PR.

fufufabulous:
Hi,

i try to get this working, but i'm getting an error i probably miss something:
No handlers could be found for logger "pogom.geofence"

AEtHeLsYn:
I've been using this PR since 1st may, everything working flawlessly, I'd like to see it merged ;)

SiteEffect:
Rebased, after @sirlindqvist pitchforks ;-)

jorgitoh:
Im using this since February. Why this PR is not merged yet?

sebastienvercammen:
@jorgitoh We have a roadmap and a priority list. This is on there, but not at the top.

SiteEffect:
Rebased once more.

vincerules805:
@fufufabulous Did you that error out? I am getting the same

SiteEffect:
*Awww ma gaaaad* - sorry, just some spontaneous expression of my feelings.

geofences/forbidden.txt.example
```diff
@@ -0,0 +1,9 @@
+[ForbiddenArea1]
+lat11, lng11
+lat12, lng12
+...
+
+[ForbiddenArea2]
```
sebastienvercammen: Rename file to `excluded.txt.example` for clarity.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+
+args = get_args()
+
+# Trying to import matplotlib, which is not compatible with all hardware.
+# Matlplotlib is faster for big calculations.
+try:
+    from matplotlib.path import Path
+except ImportError as e:
+    if not args.no_matplotlib:
+        log.error('Exception while importing "matplotlib": %s', repr(e))
```
sebastienvercammen: You can remove `repr()`, it's only useful when not using the logger.

We decided to enforce `repr()` for a while because everyone made mistakes that crashed the map, but now we're at a good point where people are really learning, so we can start using it only when necessary 👍

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+        log.error(
+            'Aborting. Install "matplotlib" or ' +
+            'enable "-nmptl" or "--no-matplotlib" to circumvent.')
+        sys.exit()
+    else:
+        pass
```
sebastienvercammen: Remove the `else`, it's not doing anything.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class Geofences:
+
+    def __init__(self):
+        self.valid_areas = []
+        self.forbidden_areas = []
+
+        if args.geofence_file or args.forbidden_file:
+            log.info('Looking for geofenced or forbidden areas.')
```
sebastienvercammen: Rephrase:

> Loading geofenced and excluded areas.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+                'Loaded %d valid and %d forbidden areas',
+                len(self.valid_areas),
+                len(self.forbidden_areas))
+
+    def is_enabled(self):
+        enabled = False
+        if self.valid_areas or self.forbidden_areas:
+            enabled = True
+
+        return enabled
```
sebastienvercammen: Rework:
```
def is_enabled(self):
    return self.valid_areas or self.forbidden_areas
```

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def get_geofenced_coordinates(self, coordinates):
+        log.info('Found %d coordinates to geofence.', len(coordinates))
+        geofenced_coordinates = []
+        startTime = time.time()
```
sebastienvercammen: Use `timeit.default_timer()` for timing.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+            geofenced_coordinates = coordinates
+
+        if self.forbidden_areas:
+            for c in reversed(geofenced_coordinates):
+                for fa in self.forbidden_areas:
+                    if self.is_coordinate_in_geofence(c, fa):
+                        # Coordinate is invalid if in one forbidden area.
+                        geofenced_coordinates.pop(
+                            geofenced_coordinates.index(c))
+                        break
```
sebastienvercammen: Instead of using a separate loop and `list.pop()` which is `O(n)`, can we not have
```
self.is_coordinate_excluded(excluded_areas, coordinate)
```
which loops over all `excluded_areas` and returns `True` if the coordinate is in one of the excluded areas?

By doing this, you can call the function `self.is_coordinate_excluded()` **before** you append the coordinate to `geofenced_coordinates` on line 62, which will be more efficient.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+
+        endTime = time.time()
+        elapsedTime = endTime - startTime
+        log.info(
+            'Geofenced to %s coordinates in %.2f s.',
```
sebastienvercammen: No space:

> Geofenced to %s coordinates in %.2fs.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+    def parse_geofences_file(geofence_file, forbidden):
+        geofences = []
+        # Read coordinates of forbidden areas from file.
+        if geofence_file:
+            with open(geofence_file) as f:
+                for line in f:
+                    if len(line.strip()) == 0:  # Empty line.
+                        continue
+                    elif line.startswith("["):  # Name line.
+                        name = line.strip().replace("[", "").replace("]", "")
```
sebastienvercammen: You're doing a lot of stripping.
```
for line in f:
    line = line.strip()

    if len(line) == 0:
        continue
    ...
```

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+                        geofences.append({
+                            'forbidden': forbidden,
+                            'name': name,
+                            'polygon': []
+                        })
+                        log.debug('Found geofence: %s', name)
```
sebastienvercammen: Missing dot at end of sentence.

pogom/geofence.py
```diff
@@ -0,0 +1,168 @@
...
+                    else:  # Coordinate line.
+                        lat, lon = line.strip().split(",")
+                        LatLon = {'lat': float(lat), 'lon': float(lon)}
+                        geofences[-1]['polygon'].append(LatLon)
+
+        return geofences
+
+    @staticmethod
+    def is_point_in_polygon_matplotlib(point, polygon):
+        pointTouple = (point['lat'], point['lon'])
```
sebastienvercammen: In English, it's `tuple`.

requirements.txt
```diff
@@ -25,3 +25,4 @@ requests-futures==0.9.7
 PySocks==1.5.6
 git+https://github.com/maddhatter/Flask-CacheBust.git@38d940cc4f18b5fcb5687746294e0360640a107e#egg=flask_cachebust
 cachetools==2.0.0
+matplotlib
```
sebastienvercammen: For consistency, specify a version. If you want latest on PyPi:

> matplotlib>=2.0.2

friscoMad:
When matplotlib is not installed and -no-matplotlib is not configured it shows an error due to the log handler not configured

`No handlers could be found for logger "pogom.geofence"`

It can be fixed moving the import after the logging setup

```
# This needs to be ater logging to be able to log geofences import errors
from pogom.search import search_overseer_thread # noqa
```

#noqa is needed to make flake8 skip this a try: also works but seems more hacky

PR #2007, opened 2017-05-01
Encounter code cleanup

Obihoernchen:
Remove: `if pokemon_id in args.enc_whitelist:`
- This is checked in line 1942 already.
- https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L1942

Improve pokemon dict update
- remove unnecessary check for lvl 30 (exception is thrown if player_lvl < 30 in encounter code already)
  - https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L2037
- use safer dict .get() methods

Fix exception

Remove "unused" captcha_url
- captcha_url is used in encounter code only.
- captcha_url isn't set outside of encounter code so this check in line 2155 is wrong.
- It's only > 1 if a lvl 30 acc triggered a captcha.

## Motivation
over 9000

## How Has This Been Tested?
my map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1785,7 +1785,6 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     sightings = {}
     new_spawn_points = []
     sp_id_list = []
-    captcha_url = ''
 
     # Consolidate the individual lists in each cell into two lists of Pokemon
     # and a list of forts.
@@ -2158,7 +2157,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                      datetime(1970, 1, 1)).total_seconds())) for f in query]
 
         # Complete tutorial with a Pokestop spin
-        if args.complete_tutorial and not (len(captcha_url) > 1):
+        if args.complete_tutorial:
             if config['parse_pokestops']:
                 tutorial_pokestop_spin(
                     api, level, forts, step_location, account)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2040,7 +2040,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                                                 + ' or higher, but account '
                                                 + hlvl_account['username']
                                                 + ' is only level '
-                                                + encounter_level + '.')
+                                                + str(encounter_level) + '.')
 
                         # We're done with the encounter. If it's from an
                         # AccountSet, release account back to the pool.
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -2102,21 +2101,20 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                    'move_2': pokemon_info.get('move_2', None),
+                    'height': pokemon_info.get('height_m', None),
+                    'weight': pokemon_info.get('weight_kg', None),
+                    'gender': pokemon_info['pokemon_display'].get(
+                        'gender', None)
                 })
 
-                # Only add CP if we're level 30+.
-                if encounter_level >= 30:
-                    pokemon[p['encounter_id']]['cp'] = cp
```
sebastienvercammen: This can't just be deleted: player level is currently not stored in the database, so we can't know where the CP came from. Similarly, IVs are still stored as usual but only L30 accounts or higher will have the same results.

Adding player level (called `player_level` in webhook data, `encounter_level` as variable) to the encounter data in the database solves this.

friscoMad: Except that this cp comes from the encounter result as ivs, moves, etc. and that data does not have a check because there is a check already on the encounter parsing that raises an exception if the account is under lvl 30.
If we want to be extra sure then we should add that check to the condition in lines 2081-2082 but it does not make too much sense as it is now.

Arengo: Indeed, we have it checked after in the exception after [Line 2032](https://github.com/RocketMap/RocketMap/pull/2007/files#diff-541e2b217ba4a7d7859dae2cb443b6b7R2032)

Obihoernchen: yup right now we can't reach this part without a proper lvl30+ acc encounter because of the exception.

pogom/models.py
```diff
@@ -1955,8 +1955,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     hlvl_api = api
                 else:
                     # Get account to use for IV or CP scanning.
-                    if pokemon_id in args.enc_whitelist:
-                        hlvl_account = account_sets.next('30', scan_location)
+                    hlvl_account = account_sets.next('30', scan_location)
```
Arengo: I'd like to point out that the [no level 30 exception](exhttps://github.com/RocketMap/RocketMap/pull/2007/files#diff-541e2b217ba4a7d7859dae2cb443b6b7R2041) is causing an exception itself since encounter_level can not be concatenated, being a non-string. So we need a string cast ``str(int_value`` over there. I wanted to directly address it in code, since I can't comment unchanged lines I leave it here.

Obihoernchen: done de82a6a

Obihoernchen:
#2139 was merged :)

PR #2009, opened 2017-05-02
Added IV/CP/Level/Moves/etc. scouting via PGScout

sLoPPydrive:
## Description
This PR adds support for **IV/CP/Level/Moves/Catch Rates**/etc. scanning using the **[PGScout](https://github.com/sLoPPydrive/PGScout)** webservice and is fully compatible with the current IV/CP scanning algorithm. Consider you found a Pokemon which is part of your encounter whitelist and encounters are currently enabled:

- If you have a PGScout webservice URL configured PGScout is being queried to get the stats.
- If PGScout isn't configured but you have given a list of L30 accounts these are used to perform the encounter.
- Else no encounter will take place. You won't know the stats of the Pokemon.

Additionally PGScout will also return the **moveset rating** as determined by [GamePress](https://pokemongo.gamepress.gg/).

The result looks something like this:
image

This PR also allows you to perform an IV scan on any Pokemon that you usually don't encounter automatically. Just click the "Scout" link in the Pokemon window:
1. image
2. image
3. image

To use this service you need access to a running PGScout instance, either on your own machine or somewhere on the net. Then you have to configure RocketMap to use it as simple as this (assuming you are running PGScout locally on default port):

```
--pgscout-url http://localhost:4242/iv
```

Attention: This PR **changes the schema** by adding new columns to the found Pokemon (leve, catch probabilities and moveset ratings).

**Disclaimer:** PGScout basically works the same as the integrated IV/CP scanning in RocketMap but way more aggressive. It doesn't obey any speed limit, so you can "scout" far more often and at very different places. This sounds a lot less secure and I cannot guarantee that this might in fact be true but in the end nobody knows for sure if single encounter requests really do that much harm. Most probably the process of getting an account to L30 is way more problematic than simple "scouting" with encounter requests. My own L30 accounts even survived the last ban wave.

Anyways, use at your own risk.

## Motivation and Context
Since 2017-04-28 RocketMap is able to scan for Pokemon stats (IV, Moves, CP) using a dedicated level 30 account queue. While this is great news it lacks some essential features:

1. It only fetches CP from the Pokemon. This isn't a particular good value to rank the Pokemon strength. A better value would be the **Pokemon level** (going from 1 to 30) which can be compared among all types of Pokemon and directly relates to the amount of stardust you would have to use to max out the Pokemon.
2. Having the **catch probabilities** for ball types Poke Ball, Great Ball and Ultra Ball is also missing from the current implementation.
3. The encounter whitelist defines which Pokemon to scan. Other are never touched. I wanted to be able to **scan** any Pokemon **on demand** by clicking a link.
4. The main reason: A popular RocketMap setup is having one instances for only the web frontend and multiple scanner instances that just scan for Pokemon, all attached to the same database. While this is surely planned for a near future release currently RocketMap doesn't allow to **share L30 workers** (or "scanners") between RocketMap instances. This means you have to pre-assign your precious L30 accounts to certain instances which isn't very flexible.

As I also run other tools that like to know Pokemon stats I decided to implement PGScout as an external webservice. The only purpose of this tool is to throw basic Pokemon data (IDs and Location) at it and get back the full set of Pokemon stats for a trainer with level 30+.

## How Has This Been Tested?
On my local map and by some guys who forked my branch.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -26,6 +26,7 @@ from cachetools import TTLCache
 from cachetools import cached
 from timeit import default_timer
 
+from pogom.pgscout import pgscout_encounter
 from . import config
 from .utils import get_pokemon_name, get_pokemon_rarity, get_pokemon_types, \
     get_args, cellid, in_radius, date_secs, clock_between, secs_between, \
@@ -42,7 +43,7 @@ args = get_args()
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 18
+db_schema_version = 19
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -111,6 +112,12 @@ class Pokemon(BaseModel):
     height = FloatField(null=True)
     gender = SmallIntegerField(null=True)
     form = SmallIntegerField(null=True)
+    level = SmallIntegerField(null=True)
+    catch_prob_1 = DoubleField(null=True)
+    catch_prob_2 = DoubleField(null=True)
+    catch_prob_3 = DoubleField(null=True)
+    rating_attack = CharField(null=True, max_length=1)
+    rating_defense = CharField(null=True, max_length=1)
     last_modified = DateTimeField(
         null=True, index=True, default=datetime.utcnow)
 
@@ -1764,6 +1771,32 @@ def hex_bounds(center, steps=None, radius=None):
     return (n, e, s, w)
 
 
+def perform_pgscout(p):
+    pokemon_id = p['pokemon_data']['pokemon_id']
+    pokemon_name = get_pokemon_name(pokemon_id)
+    log.info("PGScouting a {} at {}, {}.".format(pokemon_name, p['latitude'],
+                                                 p['longitude']))
+
+    # Prepare Pokemon object
+    pkm = Pokemon()
+    pkm.pokemon_id = pokemon_id
+    pkm.encounter_id = b64encode(str(p['encounter_id']))
+    pkm.spawnpoint_id = p['spawn_point_id']
+    pkm.latitude = p['latitude']
+    pkm.longitude = p['longitude']
+    scout_result = pgscout_encounter(pkm)
+    if scout_result['success']:
+        log.info(
+            "Successfully PGScouted a {:.1f}% lvl {} {} with {} CP (scout "
+            "level {}).".format(
+                scout_result['iv_percent'], scout_result['level'],
+                pokemon_name, scout_result['cp'], scout_result['scout_level']))
+    else:
+        log.warning("Failed PGScouting {}: {}".format(pokemon_name,
+                                                      scout_result['error']))
+    return scout_result
+
+
 # todo: this probably shouldn't _really_ be in "models" anymore, but w/e.
 def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
               key_scheduler, api, status, now_date, account, account_sets):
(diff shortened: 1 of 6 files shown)


static/js/map.js
```diff
@@ -424,41 +539,36 @@ function pokemonLabel(item) {
...
+    if (pMove1 !== 'gen/unknown') {
+        details += buildMovesDiv(encounterIdLong, pMove1, pMove2, ratingAttack, ratingDefense)
+    }
+    if (gender != null) {
+        details += buildGenderDiv(encounterIdLong, weight, height, gender)
     }
+    if (prob1 != null) {
+        details += buildProbsDiv(encounterIdLong, prob1, prob2, prob3)
+    }
+    var scoutLink = cp == null ? `<a href='javascript:void(0);' onclick='javascript:scout("${encounterId}");' title='Scout CP'>Scout</a>` : ''
```
sLoPPydrive: How do I make it that lint doesn't complain about function `scout` never being called? Doesn't lint know the template syntax? Can I attach the event handler somehow differently? Problem is that I don't have any DOM objects but only a string. :-/

Any help appreciated!

pogom/models.py
```diff
@@ -2886,5 +2950,21 @@ def database_migrate(db, old_ver):
                                 SmallIntegerField(null=True))
         )
 
+    if old_ver < 19:
+        migrate(
+            migrator.add_column('pokemon', 'level',
```
urinella: Why not using cp_multiplier here? Thats the thing you get from the API and should be stored.

static/js/map.js
Arengo: Just make it a normal if check.
```js
var scoutLink
if (cp === null) {
    scoutLink = `<a href='javascript:void(0);' onclick='javascript:scout("${encounterId}");' title='Scout CP'>Scout</a>`
} else {
    scoutLinK = ``
}
```

pogom/models.py
Arengo: Fortify.

static/js/map.js
sLoPPydrive: It may be easier to read and fit the existing code style, but I keep getting errors from `grunt lint`:
```
  406:10  error  'scout' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```
I don't actually call the `scout` method. I only define an HTML string in which the function is called via JavaScript. On the other hand I really, really wonder why this works with for example the JS function `openMapDirections` which is called the same way as `scout`.

sLoPPydrive: Oh, Rubber Duck Debugging to the rescue! By mentioning the function `openMapDirections` I just noticed the comment `// eslint-disable-line no-unused-vars`. ;-)

Finally:
```
Done, without errors.
```

pogom/models.py
sLoPPydrive: This will require a change to PGScout as well. But okay. ;)

sLoPPydrive: On second though: I'd rater **add** `cp_multiplier` instead of replacing the `level` field. Otherwise I would have to add several little code snippets here and there to keep the current functionality. Having the one information that the user wants (level) ready to use on the Pokemon entity cannot be a bad thing. :)

Arengo: I can only say that it does not fit in the current DB scheme.

mspitzbarth:
@sLoPPydrive can pgscout determine ditto? because in your screenshot there is ditto on the left

sLoPPydrive:
@Neayto No, PGScout does not catch Pokemon, so it cannot detect Dittos. This is a custom modification to my personal RocketMap fork.

sebastienvercammen:
This PR and PGScout were discussed on Discord. The consensus is that PGScout isn't an addition to RocketMap, rather it strips out part of its functionality to be re-integrated in yet another separate project, and even this part it doesn't optimize (it uses the same old bad practices: flask/werkzeug as webserver in a production environment, unnecessary threading and infinite loops).

Where something like PokeAlarm is a framework that can stand by itself, PGScout is not, so we've decided not to support it or this PR.

However, the moveset rating for example was a good idea, something that can definitely be added to RocketMap. For the future we recommend looking at adding things *to* RocketMap rather than taking away from it.

sLoPPydrive:
TL;DR: Too bad. ;-) Just wanted to give you guys a chance of how I am using PGScout with RocketMap. 

There are still some things I'd like to clarify:

What was discussed on Discord was the advertising of PGScout on Discord which was against the rules. I am totally fine with that. You asked me to tell you who my tool tries to help (so you can decide if it's something you can support) and I told you. I got no answer from you (these things happen) so I decided to open up this PR. I thought of it as an alternative way to get Pokémon stats. Like the alternative schedulers, only as an external service. I didn't want to take functionality from RocketMap. 

Alright, it has been decided not to support PGScout. I accept this decision but I still like the idea of smaller microsorvices (resilience, pluggability) instead of software monoliths.

PR #2010, opened 2017-05-02
Remove SpawnScan scheduler (-ss)

FrostTheFox:
Working on getting rid of -ss.
## Description

Removing the scheduler and associated configuration options, in favor of speedscan.
## Motivation and Context

The speed scan scheduler is a lot more advanced, works properly (for the most part) and is being maintained. The spawnscan scheduler is old and has not been touched in a while, and I'm pretty sure it's completely broken at this point.

## How Has This Been Tested?

Have not yet tested myself, though everything should work in theory. Passes flake8 in Atom.

## Types of changes

- [x] Removes old functionality
## Checklist:


SiteEffect:
While I see ``-ss`` as broken and unneeded at the moment, I do see use for an easy output of spawns without a direct database access. So ``--dump-spawnpoints`` into a file is not bad and removing it would slim functionality which we created in before.

FrostTheFox:
@Arengo I actually couldn't even get --dump-spawnpoints to work on my instance originally which was why I removed it too, took a look at what I removed and realized it was depending on -ss having defined a file, readded it and patched it so it's not done stupidly :)

runserver.py
```diff
@@ -291,15 +291,13 @@ def main():
 
         # Attempt to dump the spawn points (do this before starting threads of
         # endure the woe).
-        if (args.spawnpoint_scanning and
-                args.spawnpoint_scanning != 'nofile' and
-                args.dump_spawnpoints):
-            with open(args.spawnpoint_scanning, 'w+') as file:
-                log.info('Saving spawn points to %s', args.spawnpoint_scanning)
+        if args.dump_spawnpoints:
+            with open('spawnpoints.json', 'w+') as file:
```
Arengo: Thanks for re-adding. When we already have ``--dump-spawnpoints`` flag, why not making it a string flag to specify the file name to dump to. So it's basically just moving the ``-ss`` arg type to that one.

FrostTheFox: https://github.com/RocketMap/RocketMap/pull/2010/commits/6fd77c8c813e49bf64a76621622ec4be3aba8a47

myntath:
I still use SpawnScan for very quickly scanning all gyms over very large areas (100s of kilometers) using information dumped from the database. If SpwanScan is to go (and it probably is ready to be cut) I think we should prioritise merging #1957 first. Otherwise I think there will be a genuine loss of functionality for a small group of users.

FrostTheFox:
Well, that's functionality that SpawnScan was not intended for :(

Obihoernchen:
Regarding --dump-spawnpoints please have a look at https://github.com/RocketMap/RocketMap/pull/1948.
We should replace the existing method anyways imo.

So keeping it is :ok_hand:

SkOODaT:
im still using -ss on my main map works great as long as you have a way to geofence... and with Arengos new PR -ss could be useful again (only downside is no 60 min support)

FrostTheFox:
@SkOODaT You're not really getting the true usage of -ss, SpawnScan has not actually worked (I mean, it'll run, but the algo is wrong) for a while now because of the timer change from Niantic.

Why else would we bother with -speed? :D

SkOODaT:
yea -speed is better in process of converting over now but generally i haven't had any issues running -ss timers pop up right at 30 mins good enough for ppl to catch things lol it was the fact i needed -ss for a geofence tool i had, (odd scanning area)..... unless theirs been changes since the version i run and the newest stuff  that messed up -ss , with Arengos new PR for geofence tho i guess -ss is really getting obsolete

FrostTheFox:
The timers might show 30 minutes, that doesn't mean they're really 30 minutes.

SkOODaT:
they are lol ive timed multiple ones to make sure map is live now 30 plus ppl on it if thier were timer issues id know lol, oh im useing the old timer PR tho not -speed with my current public map -ss works great tho been running since nov
not sure about new rocket map and -ss tho if stuff got changed maby -ss is horribly broken now 
but i agree -ss is obsolete with new RM

FrostTheFox:
Did you export spawnpoint data with timers? Because if so, that can make -ss work sort of, but that's not because it's -ss, that's because you already had the timers.

FrostTheFox:
Either way, -ss wont have much use even with future PRs, -speed was designed to replace -ss and it has done so.

minus the slight chance we refactor the shit out of -ss

SkOODaT:
yea i exported and compressed + fenced off areas .....the good old way like back in the day lol :)

bluemode:
Plz don't remove -ss! It is outdated and pretty much useless! That is exactly what we need to keep finding the pokemon. Geez

richardfoltin:
I'm running a map  that finds 50.000 pokemon/hour and using a customized -ss.

**Customizations:**
- ss gets spawnpoints from spawnpoint table that was found with -speed before
- spawnpoints get clustered on the fly (at start of each instance)
- the timetreshold of clustering is added as an argument
- added another speedlimiter (to work with -ss)

**Benefits:**
- much less processor/memory/hard disk usage compared to -speed
- needs less RMP because of clustering
- less workers needed -> less proxies needed
- I can cover a much bigger area because I can use -st 4 beehives (The pokemon are very dense at some areas. Without clustering -st 4 could not be scanned without skipping under the speed limit)

PR #2011, opened 2017-05-02
make hash-key a requirement

h0us3cat:
## Description
Make hash-key a requirement.

## Motivation and Context
It fixes scrubs asking why their map is not working.

## How Has This Been Tested?
/

/

## Types of changes

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -412,7 +412,8 @@ def get_args():
     parser.add_argument('-spp', '--status-page-password', default=None,
                         help='Set the status page password.')
     parser.add_argument('-hk', '--hash-key', default=None, action='append',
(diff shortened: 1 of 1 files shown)


jagauthier:
This makes sense to me. I considered it as well.

Obihoernchen:
Nope sorry. There is -os as well. hash-key isn't required!

SiteEffect:
That's of no use, where and how you placed it. However, in ``runserver.py`` there might be use under 
```python
if not args.only_server:
    if not args.hash_key:
        log.error("Hashkey is required for scanning. Aborting.")
        sys.exit()
```

FrostTheFox:
good idea, implemented as @Arengo has it :)

PR #2015, opened 2017-05-04
Alter tables collation and charset

### LINKED ISSUE #1740: incorrect string value 1366 error using gym details with cyrillic names
so i am getting this every time i change my DB. (it mostly affects Easter Europe and other Cyrillic using locations) 

the solution is this: 

ALTER TABLE `gymdetails` COLLATE='utf8_general_ci', CONVERT TO CHARSET utf8;

if someone adds it to the guide it will save time for others.(ot devs change the gymdetails charset to utf8 by default :)

### LINKED ISSUE #2045: Error inserting data to database
**## Expected Behavior**
After enabling gym information, gym data should be stored correctly in the db.

**## Current Behavior**
One of my instances get stuck while inserting the info, without enabling gym info everything works fine. I'm running 5 more instances which are working fine. 
Here is the error log:

```
2017-05-16 01:07:03,243 [    db-updater-0][        models][   DEBUG] Inserting items 0 to 2.
2017-05-16 01:07:03,243 [    db-updater-0][        models][ WARNING] InternalError(1366, u"Incorrect string value: '\\xF0\\x9F\\x8E\\xAD O...' for column 'name' at row 1")... Retrying...
2017-05-16 01:07:04,259 [    db-updater-0][        models][   DEBUG] Inserting items 0 to 2.
2017-05-16 01:07:04,259 [    db-updater-0][        models][ WARNING] InternalError(1366, u"Incorrect string value: '\\xF0\\x9F\\x8E\\xAD O...' for column 'name' at row 1")... Retrying...
```

**## Your Environment**
* Version used: Latest dev
* Environment name and version: Python 2.7
* Operating System and version (desktop or mobile): Windows 10

Thetacz:
This PR alters all tables according to #1740 to fix OperationalError 1366 

## Description

I altered collation of tables from 'latin1_swedish_ci' to 'utf8_general_ci' and charsets to 'utf-8'.

Fixes #2045.
Fixes #1740.

## Motivation and Context

mariaDB has default collation 'latin1_swedish_ci' and default charset 'latin1' after installation, which causes warning when parsing non-default characters from gym names:
```
2017-05-04 11:59:51,832 [   search-worker-2][        models][    INFO] Upserted gyms: 3, gym members: 25.
2017-05-04 11:59:51,832 [   search-worker-2][        search][    INFO] Processing details of 3 gyms for location 49.155774,16.505332... Sleeping 6.37s until 12:00:06.
2017-05-04 11:59:51,834 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xC5\\x99\\xC3\\xAD\\xC5\\xBE...' for column 'name' at row 2")... Retrying...
2017-05-04 11:59:52,835 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xC5\\x99\\xC3\\xAD\\xC5\\xBE...' for column 'name' at row 2")... Retrying...
2017-05-04 11:59:53,835 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xC5\\x99\\xC3\\xAD\\xC5\\xBE...' for column 'name' at row 2")... Retrying...
2017-05-04 11:59:54,836 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xC5\\x99\\xC3\\xAD\\xC5\\xBE...' for column 'name' at row 2")... Retrying...
2017-05-04 11:59:55,838 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xC5\\x99\\xC3\\xAD\\xC5\\xBE...' for column 'name' at row 2")... Retrying...
```
This warning never goes away, keeps it keeps spamming.

Fixes issue #1740 .
Changing only GymDetails table is enough to fix this issue, but for consistency this PR changes all tables at once.

## How Has This Been Tested?

I accidentaly managed to reproduce #1740 on location 49.155774 16.505332. There are 3 gyms and one of them is causing OperationalError to happen, because it's name contains special czech characters.

Tested on clean pull, clean install, clean MariaDB install according to docs, empty database.
Warning starts to appear immidiately the first time when processing details of gyms.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2641,12 +2641,14 @@ def create_tables(db):
             SELECT table_name FROM information_schema.tables WHERE
             table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
             ''' % args.db_name
-        tables = db.execute_sql(cmd_sql)
-        if tables.rowcount > 0:
+        change_tables = db.execute_sql(cmd_sql)
+        if change_tables.rowcount > 0:
             log.info('Changing collation and charset on %s tables.',
-                     tables.rowcount)
+                     change_tables.rowcount)
+            if change_tables.rowcount == len(tables) + 1:
+                log.info('Changing whole database, this might take while.')
             db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
-            for table in tables:
+            for table in change_tables:
                 log.debug('Changing collation and charset on table %s.',
                           table[0])
                 cmd_sql = '''ALTER TABLE %s CONVERT TO CHARACTER SET utf8mb4
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -42,7 +42,7 @@ args = get_args()
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 19
+db_schema_version = 18
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -63,7 +63,8 @@ def init_database(app):
             host=args.db_host,
             port=args.db_port,
             max_connections=connections,
-            stale_timeout=300)
+            stale_timeout=300,
+            charset='utf8mb4')
     else:
         log.info('Connecting to local SQLite database')
         db = SqliteExtDatabase(args.db,
@@ -2626,19 +2627,24 @@ def create_tables(db):
         else:
             log.debug('Skipping table %s, it already exists.', table.__name__)
 
-    # fixing encoding on future tables
+    # fixing encoding on present and future tables
     if args.db_type == 'mysql':
-        db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
         cmd_sql = '''
             SELECT table_name FROM information_schema.tables WHERE
-            table_collation != "utf8mb4_general_ci" AND table_schema = "%s";
+            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
             ''' % args.db_name
         tables = db.execute_sql(cmd_sql)
-        for table in tables:
-            cmd_sql = '''ALTER TABLE %s COLLATE=utf8mb4_general_ci,
-                        CONVERT TO CHARSET utf8mb4;''' % str(table)
-            db.execute_sql(cmd_sql)
-        db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
+        if tables.rowcount > 0:
+            log.info('Changing collation and charset on %s tables.',
+                     tables.rowcount)
+            db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+            for table in tables:
+                log.debug('Changing collation and charset on table %s.',
+                          table[0])
+                cmd_sql = '''ALTER TABLE %s COLLATE=utf8mb4_unicode_ci,
+                            CONVERT TO CHARSET utf8mb4, FORCE;''' % str(table[0])
+                db.execute_sql(cmd_sql)
+            db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
 
     db.close()
 
@@ -2899,19 +2905,6 @@ def database_migrate(db, old_ver):
             migrator.add_column('pokemon', 'cp',
                                 SmallIntegerField(null=True))
         )
-
-    if old_ver < 19:
-        # Change charset of gym details table to utf-8 to fix gym
-        # name warning on special characters
-        # Change all tables for consistency
-        if args.db_type == 'mysql':
-            db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
-            tables = [str(x) for x in db.get_tables()]
-            for table in tables:
-                cmd_sql = '''ALTER TABLE %s COLLATE=utf8mb4_general_ci,
-                            CONVERT TO CHARSET utf8mb4;''' % table
-                db.execute_sql(cmd_sql)
-            db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
 
     # Always log that we're done.
     log.info('Schema upgrade complete.')
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2623,7 +2623,6 @@ def bulk_upsert(cls, data, db):
 
 def create_tables(db):
     db.connect()
-    verify_database_schema(db)
     tables = [Pokemon, Pokestop, Gym, ScannedLocation, GymDetails,
               GymMember, GymPokemon, Trainer, MainWorker, WorkerStatus,
               SpawnPoint, ScanSpawnPoint, SpawnpointDetectionData,
@@ -2634,32 +2633,6 @@ def create_tables(db):
             db.create_tables([table], safe=True)
         else:
             log.debug('Skipping table %s, it already exists.', table.__name__)
-
-    # fixing encoding on present and future tables
-    if args.db_type == 'mysql':
-        cmd_sql = '''
-            SELECT table_name FROM information_schema.tables WHERE
-            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
-            ''' % args.db_name
-        change_tables = db.execute_sql(cmd_sql)
-
-        if change_tables.rowcount > 0:
-            log.info('Changing collation and charset on %s tables.',
-                     change_tables.rowcount)
-
-            if change_tables.rowcount == len(tables) + 1:
-                log.info('Changing whole database, this might a take while.')
-
-            with db.atomic():
-                db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
-                for table in change_tables:
-                    log.debug('Changing collation and charset on table %s.',
-                              table[0])
-                    cmd_sql = '''ALTER TABLE %s CONVERT TO CHARACTER SET utf8mb4
-                                COLLATE utf8mb4_unicode_ci;''' % str(table[0])
-                    db.execute_sql(cmd_sql)
-                db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
-
     db.close()
 
 
@@ -2680,7 +2653,40 @@ def drop_tables(db):
     db.close()
 
 
+def verify_table_encoding(db):
+    if args.db_type == 'mysql':
+        db.connect()
+
+        cmd_sql = '''
+            SELECT table_name FROM information_schema.tables WHERE
+            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
+            ''' % args.db_name
+        change_tables = db.execute_sql(cmd_sql)
+
+        cmd_sql = "SHOW tables;"
+        tables = db.execute_sql(cmd_sql)
+
+        if change_tables.rowcount > 0:
+            log.info('Changing collation and charset on %s tables.',
+                     change_tables.rowcount)
+
+            if change_tables.rowcount == tables.rowcount:
+                log.info('Changing whole database, this might a take while.')
+
+            with db.atomic():
+                db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+                for table in change_tables:
+                    log.debug('Changing collation and charset on table %s.',
+                              table[0])
+                    cmd_sql = '''ALTER TABLE %s CONVERT TO CHARACTER SET utf8mb4
+                                COLLATE utf8mb4_unicode_ci;''' % str(table[0])
+                    db.execute_sql(cmd_sql)
+                db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
+        db.close()
+
+
 def verify_database_schema(db):
+    db.connect()
     if not Versions.table_exists():
         db.create_tables([Versions])
 
@@ -2707,6 +2713,7 @@ def verify_database_schema(db):
             log.error('Please upgrade your code base or drop all tables in '
                       'your database.')
             sys.exit(1)
+    db.close()
 
 
 def database_migrate(db, old_ver):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -96,7 +96,7 @@ class Pokemon(BaseModel):
     # We are base64 encoding the ids delivered by the api
     # because they are too big for sqlite to handle.
     encounter_id = CharField(primary_key=True, max_length=50)
-    spawnpoint_id = CharField(index=True)
+    spawnpoint_id = CharField(index=True, max_length=191)
     pokemon_id = SmallIntegerField(index=True)
     latitude = DoubleField()
     longitude = DoubleField()
@@ -1677,8 +1677,8 @@ class Versions(flaskDb.Model):
 
 
 class GymMember(BaseModel):
-    gym_id = CharField(index=True)
-    pokemon_uid = CharField(index=True)
+    gym_id = CharField(index=True, max_length=191)
+    pokemon_uid = CharField(index=True, max_length=191)
     last_scanned = DateTimeField(default=datetime.utcnow, index=True)
 
     class Meta:
@@ -1689,7 +1689,7 @@ class GymPokemon(BaseModel):
     pokemon_uid = CharField(primary_key=True, max_length=50)
     pokemon_id = SmallIntegerField()
     cp = SmallIntegerField()
-    trainer_name = CharField(index=True)
+    trainer_name = CharField(index=True, max_length=191)
     num_upgrades = SmallIntegerField(null=True)
     move_1 = SmallIntegerField(null=True)
     move_2 = SmallIntegerField(null=True)
@@ -2629,10 +2629,14 @@ def create_tables(db):
     # fixing encoding on future tables
     if args.db_type == 'mysql':
         db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+        cmd_sql = '''
+            SELECT table_name FROM information_schema.tables WHERE
+            table_collation != "utf8mb4_general_ci" AND table_schema = "%s";
+            ''' % args.db_name
+        tables = db.execute_sql(cmd_sql)
         for table in tables:
-            tables = [str(x) for x in db.get_tables()]
-            cmd_sql = '''ALTER TABLE %s COLLATE=`utf8_general_ci`,
-                        CONVERT TO CHARSET utf8;''' % table
+            cmd_sql = '''ALTER TABLE %s COLLATE=utf8mb4_general_ci,
+                        CONVERT TO CHARSET utf8mb4;''' % str(table)
             db.execute_sql(cmd_sql)
         db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
 
@@ -2901,11 +2905,11 @@ def database_migrate(db, old_ver):
         # name warning on special characters
         # Change all tables for consistency
         if args.db_type == 'mysql':
-            tables = [str(x) for x in db.get_tables()]
             db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+            tables = [str(x) for x in db.get_tables()]            
             for table in tables:
-                cmd_sql = '''ALTER TABLE %s COLLATE=`utf8_general_ci`,
-                            CONVERT TO CHARSET utf8;''' % table
+                cmd_sql = '''ALTER TABLE %s COLLATE=utf8mb4_general_ci,
+                            CONVERT TO CHARSET utf8mb4;''' % table
                 db.execute_sql(cmd_sql)
             db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
 
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2641,8 +2641,8 @@ def create_tables(db):
             for table in tables:
                 log.debug('Changing collation and charset on table %s.',
                           table[0])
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -2626,6 +2626,15 @@ def create_tables(db):
...
 
+    # fixing encoding on future tables
+    if args.db_type == 'mysql':
+        db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+        for table in tables:
+            cmd_sql = '''ALTER TABLE %s COLLATE=`utf8_general_ci`,
+                        CONVERT TO CHARSET utf8;''' % table
+            db.execute_sql(cmd_sql)
+        db.execute_sql('SET FOREIGN_KEY_CHECKS=1;')
+
```
Thetacz: This probably could be avoided by setting  "character_set_database" variable to "utf8", but I have not found a way to do it that works ...yet.

friscoMad: That will only fix future table creation but not previous tables.

friscoMad:
I love this but the problem is that if for db 20 we add a new table it will use database collation instead of utf8 so at least you should also add how to set that value in the docs.

pogom/models.py
Thetacz: Thats's why there is db version upgrade, that does the same for all previous tables present in the database. Wish I could simply set "character_set_database" in the update, but as I said, found no way that works yet. So I had to add this piece of code to create_tables(), that changes the collation any time we create tables. Which is when someone adds new table or user clears his database (delete_tables -> create_tables)...in theory :)

jakubkrysl:
@friscoMad that's why the same code is in create_tables(), which takes care of new tables creation and also clearing database. (Hope I am not wrong on that one...) So this should work for future tables too. Wish I could make it easy by setting "character_set_database" to "utf8" in the update. Simply "set character_set_database = utf8;" works in sql console, but for some reason not from RM.

friscoMad:
@Thetacz
My bad, you are completely right, I thought that it was in a migration.

Not sure if changing the collation of a table with already that collation ends instantly or if it really tries to convert it, if it is the later maybe we should do a one time migration and only change tables when they are created.

Anyway check http://stackoverflow.com/questions/5906585/how-to-change-the-default-collation-of-a-database as it seems it is possible (with the correct permissions) afaik changing db charset does not change table charsets only sets the default charset for new tables, so in theory with this and a one time migration there would be no need to migrate every table every time, but maybe peewee does create the tables with a different charset and then database charset wouldn't be used.

And please switch to utf8mb4 and utf8mb4_unicode_ci  http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci

jakubkrysl:
@friscoMad 
Thanks for the feedback, afik peewee creates tables according to the db charset, so changing the default charset and one time migration will indeed be enough, will try approach you mentioned, I cannot see how could I missed that.
As for the switch to utf8mb4: It seems like a good idea, but I will have to check our tables if that does not break anything. From what I've red, the utf8mb4 uses 4 bytes per character, but the limit (bytewise) stays the same, so it could lead to some corruption. I will see if I can prevent that, probably have to check lenght of every single column to lenght of real data in it....

pogom/models.py
Arengo: You should also consider ``utf8mb4_general_ci``

Alderon86:
good work, looking forward to test this once finished!

pogom/models.py
```diff
@@ -2626,6 +2626,16 @@ def create_tables(db):
         else:
             log.debug('Skipping table %s, it already exists.', table.__name__)
 
+    # fixing encoding on future tables
+    if args.db_type == 'mysql':
+        db.execute_sql('SET FOREIGN_KEY_CHECKS=0;')
+        for table in tables:
+            tables = [str(x) for x in db.get_tables()]
+            cmd_sql = '''ALTER TABLE %s COLLATE=`utf8_general_ci`,
+                        CONVERT TO CHARSET utf8;''' % table
```
sebastienvercammen: MySQL's proper UTF8 support is actually utf8mb4, so if you change utf8 to utf8mb4 (utf8mb4_general_ci), I'm 👍

jakubkrysl:
Ok I found no way to change database collation on the run to utf8mb4 (utf8mb4_general_ci), so I modified create_tables() to check for collation and if it is different than utf8mb4_general_ci, changes the collation of the table to the right one.
I also modified the migrator to the right collation.
The thing is, the changed collation caused peewee to misbehave (I think peewee uses latin_swedish_ci intenally...), throwing `InternalError: (1071, u'Specified key was too long; max key length is 767 bytes') `on certain CharFields when doing "-cd", so I had to reduce max_length to 191. I got this value by experimenting, 192 throws the error. I modified only those CharFields that were throwing the error, but in the future it could be necessary to set max_length in some new CharFields. This might be fixed by creating custom CharField with default max_length of 191, but I need some assistance there if we want to do that. :)

friscoMad:
Some more thoughts, not tested:

https://github.com/coleifer/peewee/blob/master/docs/peewee/database.rst#vendor-specific-parameters see that you can set the connection charset, maybe that helps with max size.
The limit seems to be known: https://github.com/coleifer/peewee/issues/131 but it seems that is 190.

Please change collation to utf8mb4_unicode_ci, general does not comply with the standards.

Also not sure if maxlength gets translatated to columnsize in the DB or if it only used internally for pewee, if it is then that columns should be migrated in the v19.

jakubkrysl:
Altering collation on table creation (if it differs from utf8mb4_unicode_ci) is enough and there is no longer any need to migrate the database to newer version. :)
I checked the columns and their collation changes with the table, so no need to alter the columns too.

@friscoMad the limit is exactly 191 characters, because 191*4 = 764 bytes, which is 3 bytes short to the excact limit.

I changed the collation to utf8mb4_unicode_ci, but keeps changing to utf8mb4_general_ci, probably because MariaDB has default collation for utf8mb4 set to utf8mb4_general_ci. So far I was unable to set it to unicode, I will keep trying.

Adding `charset='utf8mb4'` to db initialization does not seem to do anything, but it feels better and more obvious to have it there. If nothing at least it is a reminder. :)

jakubkrysl:
Now it changes properly to utf8mb4_unicode_ci. Can someone please try it on large database? I tried it only on small one which was running for an hour or so with few workers. It changed charset and collation properly without errors, -cd works fine too.

SiteEffect:
``'utf8mb4_unicode_ci' != 'utf8mb4_general_ci'``

friscoMad:
Nice work @Thetacz I will try tomorrow, maybe we can even drop the 'SET FOREIGN_KEY_CHECKS=0' in bulk_update with this changes.

@Arengo unicode_ci > general_ci http://stackoverflow.com/questions/2344118/utf-8-general-bin-unicode

jakubkrysl:
@friscoMad Great, thanks for offering to test this..
Can you also test it without 'SET FOREIGN_KEY_CHECKS=0' in bulk_upsert() as I sadly don't have the means to test that. It seems like nice idea to put it in this PR as it seems to solve the reason it was there in the first place.

@Arengo general_ci is a bit faster on very large DB, but unicode_ci offers more support to worldwide languages, which seems like more important for RM. I can leave it to general_ci if the concensus is made that general_ci is the right way to take. :)

sebastienvercammen:
@Thetacz Go for unicode.

friscoMad:
Ok, I fucked up my test and upgraded live map instead of test instance but I can confirm:

- Code works and updates collation on all tables
- The updated tables work with not updated code
- There is still 1 warning during the migration (index size over 1000) with engine MyIsam (not sure about InnoDB)
- Code works with MyIsam and InnoDB
- There is no need to disable foreignkeys with MyIsam (that's expected as MyIsam does not validate FKs)

I am using MariaDB 10.1 but my server was migrated from older one so my default engine was MyIsam, peewee does not define engine when creating tables so all my tables were created as MyIsam, faq and everyone expect tables to be InnoDB, not sure if we can/should add that migration to this PR but that migration really takes a lot of time for a big table like pokemon.

After I finish migrating to InnoDB I will report about the need of FK checks in bulk insert.
I must add that I do not encounter pokemons so that part has not been tested.

friscoMad:
Ok after migrating to InnoDB I do not get the errors also but it seems that I lost all FKs (maybe never had them as I was using MyIsam) so I will need some more test for now I will run it without FKs I hope I can do some more test tomorrow.
Also pokemon table went from 600Mb to 1.2Gb changing from MyIsam to InnoDB I only store 1 month of data.

jakubkrysl:
@friscoMad Thanks for the testing.

I changed default CharField() to Utf8mb4CharField() (name can change if needed, hope I did it the right way...) and replaced it everywhere for consistency. I changed default max_length=250 to max_length=191. We only had 1 max_length=250 in RM code in WorkerStatus.message, but it should be fine as long as noone is using worker username longer than 172 characters, as I found only 1 place where we use it: line 1183.

I tested removing 'SET FOREIGN_KEY_CHECKS' in bulk_upsert(), but it fails, because we not altering parent rows at the same time. But 'SET FOREIGN_KEY_CHECKS=0' does not seem like the proper way to do it, as this might lead to db corruption. Probably we should alter the parent row if needed. For example when altering ScannedLocations we need to alter ScanSpawnPoint first. Fortunately we are not deleting rows with FK (yet...), so it is fine for now.

jakubkrysl:
Just a note if merged: There probably should be note on discord for @everyone, that they should run only 1 instance, because this is messing with the whole database. It could take some time as @friscoMad mentioned and I am not sure, what would it do if run from multiple instances at once....

friscoMad:
@Thetacz I have not tested the new code but seems right, I did some investigation and I don't see why we are using FKs at all, the commit seems to say that it increase performance but I think that it is just because a FK will create a index for it, so creating the index and removing FKs could be the same. But anyway I doubt there is any need of a scanspawnpoint table at all, and the real performance increase will be just removing it (with a new column in spanwpoint with the cellid).
So I think the scope of this PR is good as it is right now and I will do some test with the new commits tomorrow.

jakubkrysl:
I think, FKs and ScanSpawnPoint table are needed for now, because each spawnpoint could belong to more bands now. If snap to grid gets merged, it will be obsolete. But we would have to rewrite the logic to first calculate, where the spawnpoint belongs to, before we write it to DB. So there is now way it could belong to more bands. Sadly, this is a thing for the future, which this PR can't solve now... I will probably do new PR about this once snap to grid gets merged, as I am working on speed scan gym scanning now, which among others deals with exactly the same issue - I had to create extra table to link gyms to bands :/

jakubkrysl:
@friscoMad Thanks for the tests again, I can do only limited ones, as I can spare only 5 workers and on separate database from my main scanning. I have about 200 users, so I can't mess with live server too much :/ Had to make black box clean copy...at least I am sure, there is nothing messing with my small tests. :) Still it can't beat live tests.
I tested migrating the DB for some/all tables with/without data with/without -cd, but again, the DB was almost empty....

pogom/models.py
```diff
@@ -2626,6 +2635,27 @@ def create_tables(db):
...
+        cmd_sql = '''
+            SELECT table_name FROM information_schema.tables WHERE
+            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
+            ''' % args.db_name
+        change_tables = db.execute_sql(cmd_sql)
+        if change_tables.rowcount > 0:
+            log.info('Changing collation and charset on %s tables.',
+                     change_tables.rowcount)
+            if change_tables.rowcount == len(tables) + 1:
+                log.info('Changing whole database, this might take while.')
```
sebastienvercammen: > this might take **a** while

pogom/models.py
```diff
@@ -2626,6 +2635,27 @@ def create_tables(db):
...
             log.debug('Skipping table %s, it already exists.', table.__name__)
 
+    # fixing encoding on present and future tables
+    if args.db_type == 'mysql':
+        cmd_sql = '''
+            SELECT table_name FROM information_schema.tables WHERE
+            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
+            ''' % args.db_name
+        change_tables = db.execute_sql(cmd_sql)
+        if change_tables.rowcount > 0:
```
sebastienvercammen: Can you add some whitespace between blocks? For example right above this `if`, to separate variable definition/initialization and the conditional, for readability.

jakubkrysl:
@sebastienvercammen Done, hope I did it the right way. I was thinking about nesting transaction for every table, but then I realized, this PR aims to change whole DB for consistency, so I did not do that.

sebastienvercammen:
@Thetacz 👍 The transaction needs to be around the whole block: all passes, or nothing passes.

Which reminds me: if something goes wrong during migration, we still update the db version, so it never updates again. We really need to make some changes to our migration methods, but that's outside of the scope of this PR. I've added it to the Discord list of things to do.

bellowsj:
@Thetacz @sebastienvercammen - Using this pr, I'm still getting the following error. ```2017-05-16 07:40:41,183 [      db-updater-0][        peewee][   DEBUG] ('REPLACE INTO `gymdetails` (`gym_id`, `name`, `description`, `url`, `last_scanned`) VALUES (%s, %s, %s, %s, %s)', [u'fe8df783f41248fcba005fd321d20797.16', u'Angry Frog \U0001f438 At The Angry Crab', None, u'http://lh5.ggpht.com/BNmFfE0re2x6VZ-N4KYKeOrEO7jZTrVvZR_TJLVSGQb1-k9qbphg3ek3YyQdTGteHvlXfgju4NeRuxEJDM6w2Q', datetime.datetime(2017, 5, 16, 14, 40, 41, 183384)])
2017-05-16 07:40:41,184 [      db-updater-0][          pool][   DEBUG] Returning 38568144 to pool.
2017-05-16 07:40:41,184 [      db-updater-0][        models][ WARNING] OperationalError(1366, "Incorrect string value: '\\xF0\\x9F\\x90\\xB8 A...' for column 'name' at row 1")... Retrying...```

The gym that is giving the error is located at:
Latitude = 33.496232
Longitude = -112.021273

Could you try scanning that gym and seeing if you get similar errors?

jakubkrysl:
@bellowsj I don't have time now, but I will check it tomorrow. In the meantime, can you please confirm, your collation and charset is changed to utf8mb4 / utf8mb4_unicode_ci?

`SELECT table_name, table_collation FROM information_schema.tables WHERE table_schema = "DB_NAME"` (change DB_NAME to the name of your database, rocketmapdb is the default if you followed mariaDB instalation tutorial here)

pogom/models.py
```diff
@@ -2626,6 +2635,31 @@ def create_tables(db):
         else:
             log.debug('Skipping table %s, it already exists.', table.__name__)
 
+    # fixing encoding on present and future tables
+    if args.db_type == 'mysql':
+        cmd_sql = '''
+            SELECT table_name FROM information_schema.tables WHERE
+            table_collation != "utf8mb4_unicode_ci" AND table_schema = "%s";
+            ''' % args.db_name
+        change_tables = db.execute_sql(cmd_sql)
```
sebastienvercammen: This entire block of migration code shouldn't be in `create_tables`.

Make `verify_table_encoding`, and call it in `verify_database_schema`.

Thetacz: Afaik we call `create_tables(db)` in runserver.py, than we `verify_database_schema` and than we actually create the tables in `create_tables`. So calling `verify_table_encoding` inside `verify_database_schema` would mean calling it inside `create_tables` before we actually create the tables (if they are not created yet). This could lead to the wrong encoding, as the check would be done long before the tables are created.
So I will move `verify_table_encoding` to runserver.py after `create_tables(db)` to be sure, all tables are checked for the right encoding. Is that ok?

Last sentence: did you mean moving `verify_database_schema` (not `verify_table_encoding`) out of `create_tables`? I can move that one to runserver.py too (adding connection to db inside the function ofc)

friscoMad: I think he wanted to refactor this whole part as it is currently a mess (createTable function, migrates and then creates the tables).
Until a proper refactor is done, the best thing would be encapsulating the table encoding migration in its own function and calling it where it is the code now. 
Later we should create a DBManager or whatever with a better migration system and with all this code in one single place instead of the 3K LoC model file or the runserver that should be as lean as possible.

jakubkrysl:
Allright, created new function `verify_table_encoding` and moved it to runserver.py. I also moved `verify_database_schema` out of `create_tables` to runserver.py before `create_tables`, because I do not want to break logic inside `verify_database_schema`.
So now everything is called from runserver.py and should be working fine (did only quick test with fresh db)

sebastienvercammen:
Very, very good. Thanks for the effort 👍 This was definitely necessary.

walaoaaa1234:
drop ``pokemon`` table to fix ``Specified key was too long``

bbdoc:
Hi,

This is throwing an exception when using InnoDB : 

`InternalError: (1071, u'Specified key was too long; max key length is 767 bytes')`

I tried to convert character set manually on a table of my DB in order to check what happens and this resulted in the same error : 

```
mysql> ALTER TABLE account CONVERT TO CHARACTER SET utf8mb4  COLLATE utf8mb4_unicode_ci;
ERROR 1709 (HY000): Index column size too large. The maximum column size is 767 bytes.
```

jakubkrysl:
@bbdoc This is because utf8mb4 uses 4 bytes per character prefix, so the original 256 characters of CharField gets reduced to 191 characters of CharField. I created new Utf8Mb4CharField with default max_length=191.  I was running into error 1071, but this fixed the issue for MariaDB (which is the one from docs tutorial). I also reduced max_length from to 191 in single field we had more than that (worker name), which should not be the cause of this.
I think InnoDB solves this a bit differently, so you need to manualy alter column size to max 191. There were 3 columns across the tables causing this (as far as I remember), so fix one at a time. It will always tell you, which table is the causing it, than check the models.py and try to limit column that is not explicitly limited yet. Or you can try [this](http://stackoverflow.com/a/31474509), which seems much easier. :)

@walaoaaa1234 you don't have to loose your data if you alter the tables instead of droping them.

bbdoc:
@Thetacz I understand the idea and indeed did already come to that conclusion reading the same stuff on stackoverflow, but this PR has been commited. We should not have to ask users to alter their DB tables after a PR has been commited. Any required migration should be handled directly in the code.

Do we really need varchar(255) fields in our DB ? If yes, then we have an issue as it looks like some engines can't handle it with current charset. If not, why don't we simply update DB to varchar(191) so that all RM users are using the same DB schema ?

Manually updating DB field is a very Dangerous process imo, as if a future PR needs to use more than 191 chars for whatever reason in the future, those people who altered their DB will run into issues and noone will remember the root cause of it.

friscoMad:
There is no column right now that does need a 255 lenght, the one that had it was because it is the default length in peewee.
Afaik there is no difference between InnoDB and mysiam, both have limits that with 171 should be happy, there is a difference in InnoDB with Maria 10.2 and Msql 5.7 as with some configuration there is no limit in the indexes.

Anyway if I recall correctly when I did the test, the error it was not an error, it was just a warning, tables are migrated and there is no operational error after it.
If it is an error then increasing version number should be enough to change column sizes and change database collation, as a workaround you could try to fix the issue changing the value in your version table to 17 and let RM to do the migration.

friscoMad:
Forget my last comment a db migration should be made for existing tables to change the max length.

jakubkrysl:
http://docs.peewee-orm.com/en/latest/peewee/models.html :
> Each field type has a corresponding SQL storage class (i.e. varchar, int), and conversion between python data types and underlying storage is handled transparently.

in the beginning I did db migration, but than I found out, peewee CharField(max_length=50) is the same as varchar(50). So I changed default max_length to the max value of utf8mb4 and peewee does the rest. `verify_table_collation` actually changes collation for columns in MariaDB too. At least for me it did in MariaDB 10.1. 
Are you using the latest MariaDB? I found out, they changed default engine from XtraDB to InnoDB in MariaDB 10.2 very recently, because I was using 10.1 MariaDB (XtraDB) for the tests, which I installed right before it. So there is probably about 2 weeks gap max, where if you installed MariaDB 10.2 in that time and started using it, it could cause these issues. Probably...

So as I see it, there are 2 possible solutions:
A) If updating MariaDB to 10.2+ does **NOT** break it after changing collation: Just drop / alter the tables, because there are only few people affected by this. For the future there is Utf8mb4CharField with default max_length lowered.
B) If updating MariaDB to 10.2+ does break it after changing collation: We should do migration as @friscoMad said.
I can't test it myself, because my database is empty....

friscoMad:
The problem is for tables that are already created, the collation is changed but the columns are not resized, it seems that in some setups (like mine) the collation change does not throw an error, tables collation are modified but columns are not resized.
Using max_lenght does change the size when the table is created for the first time and maybe even peewee does check the length before trying to insert the value, but the issue will be there anyway.

I am not sure about XtraDB but AFAIK the only setup that does not have any problems with index length are new installs of MariaDB 10.2 and Mysql 5.7 (defaults to InnoDB DYNAMIC with innodb_large_prefix enabled).
But for example, even when MariaDB 10.1 defaults to XtraDB, my VPS inherits the config from Mysql so it defaults to MyIsam (in fact MyIsam is a lot better than InnoDB in 10.1).

Anyway the problem is that there is no consistency between what we are generating and what we leave after migration, so even if some/most databases do work (depends on a lot of things) it is better to create a migration altering column sizes to 171.

PR #2017, opened 2017-05-04
Remove SpawnScan scheduler (-ss)

FrostTheFox:
Working on getting rid of -ss.
## Description

Removing the scheduler and associated configuration options, in favor of speedscan.
## Motivation and Context

The speed scan scheduler is a lot more advanced, works properly (for the most part) and is being maintained. The spawnscan scheduler is old and has not been touched in a while, and I'm pretty sure it's completely broken at this point.

## How Has This Been Tested?

Have not yet tested myself, though everything should work in theory. Passes flake8 in Atom.

## Types of changes

- [x] Removes old functionality
## Checklist:

#2010 back again, new PR because I don't know how to properly rebase because fuck git :3


Acadia1337:
Let me start by saying I'm a long time user of this pokemongo-map and I'm a big fan of the work. I appreciate anything that will make this software run better. I do not want to see the spawn scan functionality go away. There are viable ways to utilize the function and to remove functionality from Rocketmap seems wrong to me. The quality of the scanning mechanism and the code it uses is a valuable asset and I would hate to see it lost. With all the respect I can give I ask that you please do not merge this PR.  Thanks for everything you all do.

Sk00DaT:
this is also why i mentioned in the other PR i have been using -ss on my current public map it seems to be way easier on everything more passive then -speed, resources stretch further, after this weekend ill know for sure im switching to -speed going to give it a try on my public map...well see if i need more or less resource lol :)

FrostTheFox:
So a couple things to address.

The only viable function of ss is if you already have spawnpoints exported with timers. Gym scanning (which was something I was told about) is not a valid argument, that is not a valid use of -ss. 

So, say you do have spawnpoints + timers exported, and -ss seems faster or more efficient for you. Great. If there's benefits in -ss, they should be imported somehow into -speed, but there is 0 use in having two schedulers that obtain the same goal. I'm completely open to modifications that can be made that make one or the other more viable, but we're not keeping both. Something needs to change. It either gets overhauled, or it gets removed.

Also, @Dcstegg228 removing -ss does not remove functionality... -ss was essentially absorbed into -speed. The functionality is still there, just operates a bit differently.

ghost:
It seems like a lot less work to get speed to adopt the supposed efficiency of ss than it is to get ss up to par with speed. Removing ss and just optimizing speed definitely looks like a better option overall.

bluemode:
plz remove -ss and bring back example.py!

PR #2019, opened 2017-05-05
Add german translations for moves

sLoPPydrive:
Just added german translations for moves to `static/locales/de.json`.

## How Has This Been Tested?
On my own map.

image
(Note that this screenshot is from a modified RocketMap but it works the same on the original RocketMap)

## Types of changes

- [X] Added translation

## Checklist:


static/locales/de.json
```diff
@@ -502,17 +614,27 @@
...
+  "Pound": "Pfund",
+  "Powder Snow": "Pulverschnee",
+  "Power Gem": "Juwelenkraft",
+  "Power Whip": "Blattgeißel",
   "Primeape": "Rasaff",
   "Prinplup": "Pliprin",
   "Probopass": "Voluminas",
+  "Psybeam": "Psystrahl",
   "Psychic": "Psycho",
+  "Psychic": "Psychokinese",
```
Obihoernchen: Duplicate key `"Psychic"`. Please remove L625.
Besides that :+1:

sLoPPydrive: I see the problem. According to https://pokemongo.gamepress.gg/pokemon-move/psychic "Psychic" is both an element and a move and as the translations only work from english to other languages we don't know which is meant. Hmmmm...

sLoPPydrive: One could rename the move to "Psychic (Move)" so that "Psychic" stays the element name.

Obihoernchen: mhhh yeah that's bad...
```json
    "108": {
        "name": "Psychic",
        "type": "Psychic",
        "damage": 100,
        "duration": 2800,
        "energy": 100,
        "dps": 35.71
    },
```

MARTINI1: Same problem as the Spanish translation I'm doing, but not with psychic (luckily the translation is the same in type and attack)
My problem is with dark (poke type and style map xD)

If you have any idea how to solve the problem, comment it here and I will be grateful ;)

sLoPPydrive: There is no way around renaming the english names to make them all unique.

MARTINI1:
@sLoPPydrive Be careful that now with the pr that I have done that corrects the attacks "Future Sight" and "X-Scissor", remember to change them here ;)

SiteEffect:
^ #2021

static/locales/de.json
sebastienvercammen: The solution to the duplicate problem is by using a proper translation map:
```
'UNIQUE-KEY' => english-name
```
rather than using the English words as keys.

But this means a rework of how we do translations. I'm not against it, but someone would have to pick it up as I currently don't have the time.

sebastienvercammen:
@MARTINI1 @Arengo I merged #2021, so if there's be an issue there would be a merge conflict. All is OK 👍 But I do want to get some more attention to [my other comment in this PR](https://github.com/RocketMap/RocketMap/pull/2019#discussion_r115146856).

static/locales/de.json
```diff
@@ -202,36 +250,52 @@
...
   "Fraxure": "Sharfax",
   "Frillish": "Quabbel",
   "Froakie": "Froxy",
   "Frogadier": "Amphizel",
   "Froslass": "Frosdedje",
+  "Frost Breath": "Eisesodem",
   "Furfrou": "Coiffwaff",
   "Furret": "Wiesenior",
+  "Fury Cutter": "Zornklinge",
+  "Futuresight": "Seher",
```
MARTINI1: `"Future Sight": "Seher",`

I said that I had corrected it and now it is different

static/locales/de.json
```diff
@@ -729,23 +897,31 @@
...
   "Wobbuffet": "Woingenau",
   "Woobat": "Fleknoil",
   "Wooper": "Felino",
   "Wormadam": "Burmadame",
+  "Wrap (Green)": "Wickel (Grün)",
+  "Wrap (Pink)": "Wickel (Lila)",
+  "Wrap": "Wickel",
   "Wurmple": "Waumpel",
   "Wynaut": "Isso",
+  "X Scissor": "Kreuzschere",
```
MARTINI1: `  "X-Scissor": "Kreuzschere",`

Same as Future Sight

static/data/moves.json
```diff
@@ -856,7 +856,7 @@
         "dps": 39.39
     },
     "108": {
-        "name": "Psychic",
+        "name": "Psychic (Move)",
```
MARTINI1: little odd job :S 

In English those who look at the map will see "Psychic (move)" and I did not like this solution, if I had not already done with the Spanish translation and the word "Dark" but it was a solution that did not convince stay something ugly for those who use the map in English

I do not know what others think about this

His thing would be what Seb said to make a translation system, something like the pokealarm that has unique id attack and word to all translations

static/locales/de.json
sLoPPydrive: Yea right, was a  sloppy job of me. ;) Will fix it.

static/data/moves.json
sLoPPydrive: Renaming is indeed a quick hack to solve the problem but it's an ugly one. I hope I will find the time to make a proper translation system with unique IDs.

AEtHeLsYn:
After running some test I think I've a fix for all translations, at least for mons and moves. We should change the pokemon and moves json data to something like that:

```
"1": {
    "name": "Pokemon_001",
    "rarity": "Uncommon",
    "types": [
      {
        "type": "Grass",
        "color": "#78c850"
      },
      {
        "type": "Poison",
        "color": "#a040a0"
      }
    ],
    "spawn_rate": "61"
  },
  "2": {
    "name": "Pokemon_002",
    "rarity": "Rare",
    "types": [
```

And then have inside the locales folder the translation (including english):

 ```
 "Pokemon_718": "Zygarde",
  "Pokemon_719": "Diancie",
  "Pokemon_720": "Hoopa",
  "Pokemon_721": "Volcanion",
  "Move_001": "Thunder Shock",
  "Move_002": "Quick Attack",
  "Move_003": "Scratch",
  "Move_004": "Ember",
  "Move_005": "Vine Whip",
  "Move_006": "Tackle",
```
  
Thoughts?

SiteEffect:
@AEtHeLsYn your approach is fine but in my opinion harder to maintain than fetching new ingame changes and just translate mapping english to languageX.

PR #2023, opened 2017-05-09
Fake Search: Get Pokemon data from code instead of Niantic servers

TomSirgedas:
Generate fake Dittoes in a basic pattern, as an example.

`runserver.py -cf config/config.ini.example -k NEED_REAL_GOOGLE_API_KEY -u Bob -p bob -l 42.7,-84.5 -H 0.0.0.0 -P 7777 -hk fake_hash_key`


runserver.py
```diff
@@ -306,12 +307,18 @@ def main():
...
                 log.info('Finished exporting spawn points')
 
-        argset = (args, new_location_queue, pause_bit,
-                  heartbeat, db_updates_queue, wh_updates_queue)
-
-        log.debug('Starting a %s search thread', args.scheduler)
-        search_thread = Thread(target=search_overseer_thread,
-                               name='search-overseer', args=argset)
+        fake_pokemon_mode = True
+        if fake_pokemon_mode:
```
sLoPPydrive: This always yields True, so you can practically throw away the else-block (and the whole if/else statement).

sLoPPydrive:
I don't really get what this PR does. So basically if this gets merged the map is completely useless because it will only run in fake search mode.

I assume that this PR wants to introduce some sort of debug/testing/demo mode to let RocketMap show Pokemon on a map without actual API requests (even fake ones). That might be in fact helpful for frontend developers but at least make it optional please. ;-)

Sk00DaT:
This is an AWESOME pr if it does what i need it to .... would be awesome to fake any data off pokemons to test new code ....... IE we could of used this for form and im sure in future thier will be other new things

runserver.py
friscoMad: I think that you need to add another command line switch for this as right now it only fakes data

friscoMad:
Nice idea but needs some love.
I don't now if it is needed to be part of the searcher code, just a script under the test folder could do the same.
Also adding proper support to push all pokemon data would be great add_fake_pokemon only takes some data.
The test data should be customizable, maybe passing the name of a script to import or something like that, ditto test as an example is great but I think that it is a need being able to create your custom test scripts without modifying the version controlled code.

pogom/fakeSearcher.py
```diff
@@ -0,0 +1,63 @@
...
+            'move_2': None,
+            'cp': None,
+            'height': None,
+            'weight': None,
+            'gender': None,
+            'form': None,
+            }
+
+    if pokemon:
+        db_updates_queue.put((Pokemon, pokemon))
```
Arengo: for a real fake, there needs to be the data for the webhook too. So users jump out of their chair when they see the 100% Mewtwo from PokeAlarm.

runserver.py
TomSirgedas: I agree, it's not integrated well right now. I just wanted a "self-documented" way to show how to switch the fake mode on/off.

pogom/fakeSearcher.py
TomSirgedas: Yeah, this code doesn't handle webhooks, pokestops, gyms, etc, since I didn't need those. But, I wanted to share my work in case it was useful for others.

TomSirgedas:
Thanks for the feedback guys!

This code was useful for myself, and I just wanted to share. It's not in a merge-able state for sure. I'm not sure how to do that, but @friscoMad has good ideas.

343334:
Well looks like it has great potential to be used as a standalone server for multiple remote mysql servers where this map could accept incoming webhooks - and then display that data.

SiteEffect:
@343334 how about replication? :-D

343334:
@Arengo Why should we depend on a mysql db any more than we need to?  Most users tend to use the webhook features already so if RM had a way to run as a standalone server that accepts incomings POSTS to display the markers that would allow many local neighborhood scanners to team up together without much extra work as replication requires a lot of extra knowledge and configurations.  With a standalone server to handle this on its own all it would require is users to add another webhook and than prob set the firewall to only allow incoming connections from other scanner ips on that port if they wish to team up. Much more user friendly and in theory should be less load than pulling data from mysql.

pogom/fakeSearcher.py
TomSirgedas: I updated the code to be merge-able, but there's no webhook support. I hope that's OK.

runserver.py
TomSirgedas: updated!

TomSirgedas:
Hey @sebastienvercammen , any chance this could be merged? Please let me know if there's no chance or if there's something that you want. I think it opens up some possibilities, like collaborative maps (great idea, @343334 !)

SkOODaT:
i really like this feature for testing... any possibility it could be adapted to send the pokemon to webhooks too, ie again unown formatting, im trying to set this up currently and it would be of great help to be able to fake a spawn and/or an alert too
Works perfect tho....
Thank you

TomSirgedas:
@SkOODaT 
an attempt at supporting webhooks, let me know if this works or not:
https://github.com/TomSirgedas/RocketMap/pull/2
for unowns, your script can provide `'form': 26` (in addition to the usual `'pokemon_id': 201`, etc)

sebastienvercammen:
@TomSirgedas Sorry for the late reply, somehow the notification for your ping got lost and I've been swamped with shadowban/API work.

Yes, this definitely has a use case and will be merged if implemented properly. However, as it is, we can't merge yet.

This needs to be a completely stand-alone mode, separate from the main codebase, similar to how `-mock` does its thing. I'm not sure if `-fss` has any use for any other "fake search script" implementation than a single fake implementation that covers all use cases.

I understand the client/server example that you've added, but what are the advantages of this (more complex) implementation versus a single mock data mode that randomly adds Pokémon (customizable amount) of all types (start at Pokémon 1, increment, modulo to reset) with a single toggle rather than separate scripts?

Just curious about your thoughts. :)

TomSirgedas:
Some use cases:

- Want to test front end, just need some random pokemon
- Want to test front end, lots of pokemon to stress test
- Want to test front end, special scenarios, like showing unown with the proper letter, or doing something special for ultra-rares, etc.
- Draw a pattern on the map using pokemon (silly)
- Collaborative map -- your script would grab pokemon from multiple inputs
- Make a map from a Discord feed. This was my project (with blessing from the admin). I needed the Discord scraper to be in a different process than RocketMap, so the `client.py`/`listener.py` was useful for me.
- I'm sure there are more ideas!

Like frisco said:
> The test data should be customizable, maybe passing the name of a script to import or something like that, ditto test as an example is great but I think that it is a need being able to create your custom test scripts without modifying the version controlled code.

So, re: "I'm not sure if -fss has any use for any other "fake search script" implementation than a single fake implementation that covers all use cases.", I think it's valuable to specify a script!

I'm not sure what "completely stand-alone mode, separate from the main codebase" means exactly. I want to use the existing database, classes, etc. Changes to the main code path are minimal, just checking if "fake_search_script" mode is on. (Though it seems that `fake_pokemon_mode = True` `if fake_pokemon_mode:` is still in the code, instead of checking the config variable :/)

TomSirgedas:
@sebastienvercammen (bump)

SiteEffect:
I wouldn't have thought it, but this had use now, with broken hash servers.

PR #2025, opened 2017-05-09
SpawnPoint History and Nest Detection

BrownSlaughter:
gets 24 hour history for spawn points, if a pokemon spawns 25% of the time or more it is classed as frequent/nesting 

## Description
Spawnpoint history gets the last 24 of spawns for that single point.
Nest detection gets 24h history for all points on visible on screen, it then collects the Pokemon that spawn over 25% and lists them

## Motivation and Context
It is nice to see what spawns at a certain point, easy to identify nests

## How Has This Been Tested?
Tested for months on my personal map

sph
nest

## Types of changes

## Checklist:


MARTINI1:
I like it, I missed this function in rocketmap 👍

SiteEffect:
I really like it as a feature! We should give it a last beauty boost, though. The green for example is a bit too much "eye candy" :-P. Where does the ``grass1`` marker show up, btw?

BrownSlaughter:
if people have a better colour scheme please let me know. white was just terrible, especially since both green bars do different things when you click on them

pogom/models.py
```diff
@@ -433,6 +433,49 @@ def get_spawnpoints_in_hex(cls, center, steps):
...
+
+        return spawn_history
+
+    @classmethod
+    def get_pointhistory(cls, swLat, swLng, neLat, neLng):
+        if swLat is None or swLng is None or neLat is None or neLng is None:
+            query = (Pokemon
+                     .select()
+                     .where()
+                     .dicts())
```
Thetacz: If I am not mistaken, you propose to send whole database of pokemons when no boundaries are set. This seems like major security breach, as there should be no way to get any of those boundaries as None if used properly.
https://developers.google.com/maps/documentation/javascript/reference#Map
We are using getBounds(), which returns null or undefined only when the map is not initilized, but the map object gets initialized at the very beginning, so this should not be the case. So only way to get those is if someone is deliberately parsing them instead of normally viewing the map through the browser or there was some error during init...
So I think, we should send empty when no boundaries are recieved. :)

BrownSlaughter: I am pretty new at this but I thought that select nothing not the whole DB as you suggest

Thetacz: Alright, I gave it a try and `.select()` selects everything, `dicts() `changes it to dicts and `where() `causes error: :D
So if you get no boundaries, you send peewee error.
I think, the best approach here is to set `.where(Pokemon.pokemon_id == None)`, as this does not return error but empty response.

pogom/models.py
```diff
@@ -457,10 +457,7 @@ def get_spawn_history(cls, spawnpoint_id):
     @classmethod
     def get_pointhistory(cls, swLat, swLng, neLat, neLng):
         if swLat is None or swLng is None or neLat is None or neLng is None:
```
Arengo: For proper code-style change it to the following, by that also no ``else`` is needed.
```py
query = []
if swLat and swLng and neLat and neLng:
    do something
```

BrownSlaughter: good point, there is no need for the else statement anymore as it only needs to check if they are set and do stuff. To answer your earlier question I missed, the grass1 marker shows up when you click "show on map" and will show all points where that pokemon spawned 25%+ of the time. I really need to find a way to remove markers but have not managed it thus far.

BrownSlaughter:
I also changed the colours to be more eye friendly, still wanting ideas for colours or alternative

pogom/models.py
```diff
@@ -456,9 +456,7 @@ def get_spawn_history(cls, spawnpoint_id):
 
     @classmethod
     def get_pointhistory(cls, swLat, swLng, neLat, neLng):
-        if swLat is None or swLng is None or neLat is None or neLng is None:
-            query = []
-        else:
+        if swLat and swLng and neLat and neLng:
```
Arengo: The issue is, that you do not have a default return value now, so you will get an Exception for no initialization since you missed ``query = []`` before the ``if``-Statement if it doesn't get triggered. I also do not know if ``list(query)`` directly transforms a ``query.dict()`` into a standard python list of dicts. Have a look at the example I PM'd you some days ago.

SiteEffect:
@BrownSlaughter here is a nice [eye friendly color reference](https://s3.amazonaws.com/viget.com/legacy/uploads/image/blog/cc-link-1.png).

boby361:
I love the idea and just going to implement into fork. I was thinking with the current event however this would collide with events where a whole lot of spawnpoints spawn event pokemon. an exclusion list or just have it notice if there are a crazy amount of "nests" to simply not show (due to high probability of event).

(edited) 
Now that iv implemented it (assuming i did correctly) it seems that you A) cant hide the grass icon popup (the X doesn't work) B) it seems a very inconsistent as after playing with it (and having it reload a bunch of times) that the Pokemon that had 3 spawns now just didn't generate on the list (despite waiting awhile) C) if you look away from the window it seems to completely regenerate the list (which is more taxing and slower) then if there were to be some method to cache it (honestly no idea how things work) D) if there is a time out feature on which it pulls information it would be the reason why its inconsistent due to slower DBs. E) placing markers can be infinite. but removing them is only 1. F) Better detection method? I have a small park with 4 Girafarig spawned. This doesn't show up on the list at all despite it being obvious. I know the code is looking for a spawnpoint being repetitive although if a bunch of rare pokemon are in a small radius it would suggest a nest also. (obviously more demanding to calculate and hence using the DB would probably be wise.

LudwikPR:
+1

BrownSlaughter:
This is now closing due to a rewrite and reshuffle

PR #2026, opened 2017-05-09
[Optimization] Use protobuf objects directly instead of dicts

friscoMad:
## Description
This PR does avoid converting all response data from host, ~~to work needs a pgoapi PR (https://github.com/pogodevorg/pgoapi/pull/200) to be merged but after 2 weeks waiting without reply there I think that maybe posting this here could speed up review there.~~
API changes have been merged now so to test it you need to checkout this PR and then:
```
pip uninstall pgoapi
pip install --upgrade -r requirements.txt
```

## Motivation and Context
There is no gain in converting all objects to dicts and there are some issues:

- Wasted CPU time: Responses are converted to protobuf objects and then to dicts, also accessing dicts it is slower than accesing regular object attributes (even if O(1) there is no need of hashing and comparing the key before retrieval)
- Increase memory usage:  Dicts use more memory than regular objects as there is no need to store keys.
- Protobuf2dict has problems with falsy values: There are a lot of code that needs to check against this issue like gym checks in map parsing or empty lists, in the object the value is there the problem is within the conversion.

## How Has This Been Tested?
This code has been working for 3 weeks in my server, but I did not test -tut or encounters so I am almost sure that there is some error in that part as a lot of code needs to be converted and I have probable missed some lines.

Memory consumption of a -st 41 280 worker instance with develop (it is test instance):
pokemap tlp - pokemad neoseo es_22 - bitvise xterm_2017-05-09_14-03-00

Memory consumption of a -st 41 280 worker instance with this branch (it is test instance):
pokemap tlp - pokemad neoseo es_22 - bitvise xterm_2017-05-09_18-19-17

I timed the time spent parsing the response of every request with develop and with this pr against the pgoapi PR (it is backwards compatible). Tested with an instance of -st 5 and 1 thread to reduce cpu load noise and context switching problems:

Develop:
pokemap tlp - pokemad neoseo es_22 - bitvise xterm_2017-05-09_22-41-18

Protobuf:
pokemap tlp - pokemad neoseo es_22 - bitvise xterm_2017-05-09_22-33-19

## Types of changes

- [X] Improvement

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -11,7 +11,7 @@ from pgoapi import PGoApi
 from pgoapi.exceptions import AuthException
 
 from .fakePogoApi import FakePogoApi
-from .utils import in_radius, generate_device_info, equi_rect_distance
+from .utils import (in_radius, generate_device_info, equi_rect_distance)
 from .proxy import get_new_proxy
 
 log = logging.getLogger(__name__)
@@ -120,7 +120,7 @@ def rpc_login_sequence(args, api, account):
 
     try:
         request = api.create_request()
-        request.call()
+        request.call(False)
 
         total_req += 1
         time.sleep(random.uniform(.43, .97))
@@ -139,7 +139,7 @@ def rpc_login_sequence(args, api, account):
     try:
         req = api.create_request()
         req.get_player(player_locale=args.player_locale)
-        req.call()
+        req.call(False)
 
         total_req += 1
         time.sleep(random.uniform(.53, 1.1))
@@ -165,7 +165,7 @@ def rpc_login_sequence(args, api, account):
         request.get_inventory(last_timestamp_ms=0)
         request.check_awarded_badges()
         request.download_settings()
-        response = request.call()
+        response = request.call(False)
 
         parse_download_settings(account, response)
         parse_new_timestamp_ms(account, response)
@@ -206,7 +206,7 @@ def rpc_login_sequence(args, api, account):
             request.check_awarded_badges()
             request.download_settings(hash=account[
                 'remote_config']['hash'])
-            response = request.call()
+            response = request.call(False)
 
             parse_new_timestamp_ms(account, response)
 
@@ -226,9 +226,9 @@ def rpc_login_sequence(args, api, account):
             except KeyError:
                 break
 
-            result = response.get('result', 0)
-            page_offset = response.get('page_offset', 0)
-            page_timestamp = response.get('timestamp_ms', 0)
+            result = response.result
+            page_offset = response.page_offset
+            page_timestamp = response.timestamp_ms
             log.debug('Completed %d requests to get asset digest.',
                       req_count)
 
@@ -255,7 +255,7 @@ def rpc_login_sequence(args, api, account):
             request.check_awarded_badges()
             request.download_settings(
                 hash=account['remote_config']['hash'])
-            response = request.call()
+            response = request.call(False)
 
             parse_new_timestamp_ms(account, response)
 
@@ -275,9 +275,9 @@ def rpc_login_sequence(args, api, account):
             except KeyError:
                 break
 
-            result = response.get('result', 0)
-            page_offset = response.get('page_offset', 0)
-            page_timestamp = response.get('timestamp_ms', 0)
+            result = response.result
+            page_offset = response.page_offset
+            page_timestamp = response.timestamp_ms
             log.debug('Completed %d requests to download'
                       + ' item templates.', req_count)
 
@@ -293,7 +293,7 @@ def rpc_login_sequence(args, api, account):
         request.check_awarded_badges()
         request.download_settings(hash=account['remote_config']['hash'])
         request.get_buddy_walked()
-        response = request.call()
+        response = request.call(False)
 
         parse_new_timestamp_ms(account, response)
 
@@ -335,7 +335,7 @@ def rpc_login_sequence(args, api, account):
         request.download_settings(hash=account['remote_config']['hash'])
         request.get_buddy_walked()
         request.get_inbox(is_history=True)
-        response = request.call()
+        response = request.call(False)
 
         parse_new_timestamp_ms(account, response)
 
@@ -363,11 +363,10 @@ def get_tutorial_state(args, api, account):
     request = api.create_request()
     request.get_player(
         player_locale=args.player_locale)
-    response = request.call().get('responses', {})
-
-    get_player = response.get('GET_PLAYER', {})
-    tutorial_state = get_player.get(
-        'player_data', {}).get('tutorial_state', [])
+    response = request.call(False).get('responses', {})
+    if 'GET_PLAYER' not in response:
+        return []
+    tutorial_state = response['GET_PLAYER'].player_data.tutorial_state
     time.sleep(random.uniform(2, 4))
     return tutorial_state
 
@@ -381,7 +380,7 @@ def complete_tutorial(args, api, account, tutorial_state):
         request = api.create_request()
         request.mark_tutorial_complete(tutorials_completed=0)
         log.debug('Sending 0 tutorials_completed for %s.', account['username'])
-        request.call()
+        request.call(False)
 
     if 1 not in tutorial_state:
         time.sleep(random.uniform(5, 12))
@@ -397,20 +396,20 @@ def complete_tutorial(args, api, account, tutorial_state):
         })
         log.debug('Sending set random player character request for %s.',
                   account['username'])
-        request.call()
+        request.call(False)
 
         time.sleep(random.uniform(0.3, 0.5))
 
         request = api.create_request()
         request.mark_tutorial_complete(tutorials_completed=1)
         log.debug('Sending 1 tutorials_completed for %s.', account['username'])
-        request.call()
+        request.call(False)
 
     time.sleep(random.uniform(0.5, 0.6))
     request = api.create_request()
     request.get_player_profile()
     log.debug('Fetching player profile for %s...', account['username'])
-    request.call()
+    request.call(False)
 
     starter_id = None
     if 3 not in tutorial_state:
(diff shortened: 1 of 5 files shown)


MARTINI1:
This pr if that is an authentic zoom zoom to optimize the map, you make amazings PRs for optimizing map and not the wild trolls that appear from time to time making PRs useless xD

Almost 200 MB is a great improvement, above all I will note that I use a map in an rpi3, good job ;)

@friscoMad Hmm, I wanted to test it but nothing else to boot and while the workers connect I get this error when connecting each one:

[ERROR] Exception in search_worker under account xxxx Exception message: TypeError ('call () takes exactly 1 argument (2 given)',).

SiteEffect:
@MARTINI1 there is only one real troll @bluemode ;-)
I do agree, that this is a viable thing to consider. It does not harm development since the objects are still readable.

sebastienvercammen:
I like it, but we need a reply from pgoapi. I'm on the API maintainers list in pogodev, so I'll poke them for feedback.

michikrug:
@friscoMad you probably missed three spots

access to fort properties in:

https://github.com/RocketMap/RocketMap/pull/2026/files#diff-c08aa0b614366984e8695649c57eded7R260
and
https://github.com/RocketMap/RocketMap/pull/2026/files#diff-c08aa0b614366984e8695649c57eded7R283
and
https://github.com/RocketMap/RocketMap/pull/2026/files#diff-c08aa0b614366984e8695649c57eded7R285

friscoMad:
@michikrug Nice found, I fixed them along with some more in the spin poke request.
Also rebased and working with latest api

michikrug:
Encounter also doesn't seem to work, since the get_player_level for the encounter_result gives 0

friscoMad:
@michikrug It should be fixed by now at least the get_player_level, thanks for the report

michikrug:
Thanks. 
The last remaining errors I currently get are:

```
pogom/account.py", line 292, in spin_pokestop
captcha_url = spin_response['responses'][
TypeError: 'bool' object has no attribute '__getitem__'
```

```
pogom/models.py", line 2083, in parse_map
    if encounter_result is not None and (encounter_result['responses'][
TypeError: 'bool' object has no attribute '__getitem__'
```

I seems that the gym webhook messages are somehow not working anymore.
I think this is because you now copy the dict, which unfortunately has different field names.


EDIT: could be related to not up2date pgoapi on my site, will test further
EDIT2: No. Exceptions still remain.

Obihoernchen:
https://github.com/pogodevorg/pgoapi/pull/200 is merged now :thumbsup:

friscoMad:
@michikrug  I fixed those exceptions but they are only cascade exceptions, when the requests fails the response is a false object and does not work as a dict, so they shouldn't be related with this PR, anyway I am fixing them as it is pointless to add a new PR for just 2 lines.

michikrug:
Ah, I see. Thanks for the quick update.

What about the gym webhook message?

Obihoernchen:
Tested:
- [X] pokemon scanning
- [X] pokemon encounter
- [X] pokestops
- [X] gym + gym-info
- [X] pokemon webhook
- [ ] tutorial
- [ ] pokestop webhook
- [ ] gym webhook

Some early test results:
- ~15% less memory usage
- slightly lower CPU usage (5% maybe)

:+1:

friscoMad:
@michikrug I don't see any problem with gym webhooks, there are 2 wh for gyms:

- The hook with the info from parsemap that can be disabled with 'webhook-updates-only' it sends all the visible info but not the full pokemon rooster and defending players.
-   The hook with the full gim info, it will only get this info is --gym-info is enabled.

Please tell me which one is missing or the kind of problem you have.

michikrug:
I have set "webhook-updates-only" and therefore it should relate to the "gym_details" webhook messages. At least before the merge PokeAlarm received the messages.

AEtHeLsYn:
@friscoMad @michikrug According to command line documentation:
```
--webhook-updates-only
                    Only send updates (Pokemon & lured pokestops). [env
                    var: POGOMAP_WEBHOOK_UPDATES_ONLY]
```

If the switch is enabled no gym update info is sent to the webhook

michikrug:
@AEtHeLsYn This can't be true, since it was working for months. Also the code, afaik, tells other.

pogom/models.py
```diff
@@ -2354,91 +2347,69 @@ def parse_gyms(args, gym_responses, wh_update_queue, db_update_queue):
...
-                        'additional_cp_multiplier', 0),
-                    'iv_defense': member['pokemon_data'].get(
-                        'individual_defense', 0),
-                    'iv_stamina': member['pokemon_data'].get(
-                        'individual_stamina', 0),
-                    'iv_attack': member['pokemon_data'].get(
-                        'individual_attack', 0),
-                    'trainer_name': member['trainer_public_profile']['name'],
-                    'trainer_level': member['trainer_public_profile']['level'],
+                wh_gym_pokemon = gym_pokemon[i].copy()
```
michikrug: This part somehow stops correct gym_detail webhook messages to be sent. I removed it for testing and it worked again.

friscoMad:
Fixed issue with webhooks and gym detail, there was a silent error when converting gym details to json due to the last_seen field that was added when changed to code to copy.

pogom/account.py
```diff
@@ -268,39 +269,35 @@ def tutorial_pokestop_spin(api, player_level, forts, step_location, account):
...
     return 0
 
 
 def spin_pokestop(api, fort, step_location):
     spinning_radius = 0.04
-    if in_radius((fort['latitude'], fort['longitude']), step_location,
+    if in_radius((fort.latitude, fort.longitude), step_location,
                  spinning_radius):
-        log.debug('Attempt to spin Pokestop (ID %s)', fort['id'])
+        log.debug('Attempt to spin Pokestop (ID %s)', fort.id)
```
sebastienvercammen: Missing dot in log message.

pogom/models.py
```diff
@@ -1789,8 +1789,8 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
 
     # Consolidate the individual lists in each cell into two lists of Pokemon
     # and a list of forts.
-    cells = map_dict['responses']['GET_MAP_OBJECTS']['map_cells']
-    # Get the level for the pokestop spin, and to send to webhook.
+    cells = map_dict['responses']['GET_MAP_OBJECTS'].map_cells
+    # Get the level for the pokestop spin in any case delete inventory
```
sebastienvercammen: Punctuation.

pogom/models.py
sebastienvercammen: @friscoMad @Alderon86 Has someone reviewed this feedback?

pogom/models.py
```diff
@@ -2354,91 +2347,71 @@ def parse_gyms(args, gym_responses, wh_update_queue, db_update_queue):
...
-                    'iv_attack': member['pokemon_data'].get(
-                        'individual_attack', 0),
-                    'trainer_name': member['trainer_public_profile']['name'],
-                    'trainer_level': member['trainer_public_profile']['level'],
+                wh_gym_pokemon = gym_pokemon[i].copy()
+                wh_gym_pokemon.update({
+                    'trainer_name': member.trainer_public_profile.name,
+                    'trainer_level': member.trainer_public_profile.level
                 })
+                # Otherwise json conversion will fail and nothing will be sent
```
sebastienvercammen: Punctuation.

pogom/models.py
friscoMad: Yup this was addressed at 72b40e8

friscoMad:
Rebased and fixed punctuation as requested, feedback about gym info problem was addressed some time ago there was a silent exception when converting last seen date to json, so I just removed it as it was not being sent before.

urinella:
<s>`Exception while downloading map: TypeError('call() takes exactly 1 argument (2 given)',)`

I'm getting this error from latest version @friscoMad </s>

friscoMad:
At what line? I think you did merge this pr into something else as all calls are passing only False in every call. 
Also I am using this branch clean for my map and is running smoothly.

urinella:
<s>Using clean branch. Only this PR for testing.
I cannot tell you the line because I only get this error and nothing else.
Maybe I can try to look at verbose later.
```
git fetch RM pull/2026/head:ram
From https://github.com/RocketMap/RocketMap
 * [new ref]         refs/pull/2026/head -> ram
$ git checkout ram
Previous HEAD position was 57e45f6... Update pgoapi to fix json login errors (#2090)
Switched to branch 'ram'

...

$ git reset --hard
HEAD is now at 9eb4a8d Fix punctuation
``` 

Edit: -v & -vv doesn't give any information here

More info:

``` 
$python -V
Python 2.7.12
``` 

Made a fresh clone into a new folder: same issue</s>

friscoMad:
Ok, the "2 given" part misguided me, you are using an non up to date pgoapi, that's why it is complaining.
Please:
pip uninstall pgoapi

urinella:
<s>I'm on latest pgoapi... 78e79bda8b78f970d2b3692bd1a5bd17cefa5fdd
Downgrade to 1a77611c19f46d76681c515ddb0c9ab15c002d11 doesn't work, still same error. 
Sorry.
Maybe someone else can test the latest version of the PR. Could be a problem with my server, I don't know.

I'll do some more tests:
- using -tut:

```
  File "/home/RocketMap2/test/RocketMap/pogom/search.py", line 944, in search_worker_thread
    tutorial_state = get_tutorial_state(api, account)
  File "/home/RocketMap2/test/RocketMap/pogom/account.py", line 116, in get_tutorial_state
    response = request.call(False).get('responses', {})
TypeError: call() takes exactly 1 argument (2 given)
``` 
</s>

**Edit: Solved. Was my system with 2 versions of pgoapi. Sorry for bother you**

friscoMad:
@urinella That's really strange, I am at  1a77611c19f46d76681c515ddb0c9ab15c002d11 and working will try with latest version.
Please if you can contact me on discord (friscoMad) so we can debug this issue, if you can't please post your full ```pip list```, then do ```pip show pgoapi``` go to the location shown there + "/pgopai" and do a ```cat pgoapi.py | grep use_``` if should show something like ```def call(self, use_dict = True):``` if it does not show anything then for some reason you are on an older pgoapi

friscoMad:
Just to make it clear, @urinella was having issues with a second old pgoapi installed in a different directory, when he did fix that the errors were gone. So it was not a problem of this PR.

PR #2028, opened 2017-05-11
Fix previous issue as webhooks was overlooked

captainmedic:
## Description
This change prevents the workers from getting Key Errors when scanning an area that has Pokestops that have timers on them.

## Motivation and Context
It solves the Key Error for maps running webhooks. This change was previously committed in #1972 but did not cover those running webhooks

## How Has This Been Tested?
This is has been tested on my server with a known hive containing Pokestops with timers. Prior to this change it would receive Key Error, and now it is running smoothly

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -2136,14 +2136,14 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+            stop_ids = [f.get('id', 0) for f in forts if f.get('type') == 1]
             if stop_ids:
                 query = (Pokestop
                          .select(Pokestop.pokestop_id, Pokestop.last_modified)
                          .where((Pokestop.pokestop_id << stop_ids))
                          .dicts())
-                encountered_pokestops = [(f['pokestop_id'], int(
-                    (f['last_modified'] -
+                encountered_pokestops = [(f.get('pokestop_id', 0), int(
+                    (f.get('last_modified', 0) -
```
sebastienvercammen: In cases where we are 100% expecting and require the proper ID, we can't softly fall back to a default value. If the pokéstop ID is missing, something is wrong, and we should stop instead of falling back to `0`.

TheCrabity:
Bump.

friscoMad:
I think the issue was with enabled and that was already fixed on #2101 so this can be closed.
Could you @captainmedic or @PKMTracker confirm?

PR #2029, opened 2017-05-11
Add Cries for better notifications

BrownSlaughter:
This PR adds the ability to play cries (or other sounds) for notifications rather than the usual `ding`.
## Description
A switch for cries is added to the side bar to switch it on or off.
If you turn on cries but fail to put any in the folder it will revert to the `ding`
If a sound for a pokemon is not found it will play the normal `Ding`

Since cries are probably copywrite, I can not provide them but if you ask me on discord I can probably help you search.

The cries should be placed in static/sounds/cries/
they should be `pokemon id.wav` eg 1.wav for bulbasaur

## Motivation and Context
The map feels alive with the sound of anime voices

## How Has This Been Tested?
Tested on my map with anime voices

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -363,6 +363,7 @@ function initSidebar() {
     $('#spawnpoints-switch').prop('checked', Store.get('showSpawnpoints'))
     $('#ranges-switch').prop('checked', Store.get('showRanges'))
     $('#sound-switch').prop('checked', Store.get('playSound'))
+    $('#cries-switch').prop('checked', Store.get('playCries'))
     var searchBox = new google.maps.places.Autocomplete(document.getElementById('next-location'))
     $('#next-location').css('background-color', $('#geoloc-switch').prop('checked') ? '#e0e0e0' : '#ffffff')
 
@@ -764,8 +765,18 @@ function customizePokemonMarker(marker, item, skipNotification) {
 
     if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
         if (!skipNotification) {
-            if (Store.get('playSound')) {
+            if (Store.get('playSound') && !Store.get('playCries')) {
+                var audio = new Audio('static/sounds/ding.mp3')
                 audio.play()
+            } else if (Store.get('playSound') && Store.get('playCries')) {
+                var audio = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                audio.play().catch(function (err) {
+                    if (err) {
+                        console.log('sound for this pokemon is missing, using ding instead')
+                        var audio = new Audio('static/sounds/ding.mp3')
+                        audio.play()
+                    }
+                })
             }
             sendNotification(getNotifyText(item).fav_title, getNotifyText(item).fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
         }
@@ -778,8 +789,18 @@ function customizePokemonMarker(marker, item, skipNotification) {
         var perfection = getIv(item['individual_attack'], item['individual_defense'], item['individual_stamina'])
         if (notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection) {
             if (!skipNotification) {
-                if (Store.get('playSound')) {
+                if (Store.get('playSound') && !Store.get('playCries')) {
+                    var audio = new Audio('static/sounds/ding.mp3')
                     audio.play()
+                } else if (Store.get('playSound') && Store.get('playCries')) {
+                    var audio = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                    audio.play().catch(function (err) {
+                        if (err) {
+                            console.log('sound for this pokemon is missing, using ding instead')
+                            var audio = new Audio('static/sounds/ding.mp3')
+                            audio.play()
+                        }
+                    })
                 }
                 sendNotification(getNotifyText(item).fav_title, getNotifyText(item).fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
             }
@@ -2357,6 +2378,10 @@ $(function () {
 
     $('#sound-switch').change(function () {
         Store.set('playSound', this.checked)
+    })
+
+    $('#cries-switch').change(function () {
+        Store.set('playCries', this.checked)
     })
 
     $('#geoloc-switch').change(function () {
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -766,11 +766,10 @@ function customizePokemonMarker(marker, item, skipNotification) {
     if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
         if (!skipNotification) {
             if (Store.get('playSound') && !Store.get('playCries')) {
-                var audio = new Audio('static/sounds/ding.mp3')
                 audio.play()
             } else if (Store.get('playSound') && Store.get('playCries')) {
-                var audio = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
-                audio.play().catch(function (err) {
+                var audiocry = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                audiocry.play().catch(function (err) {
                     if (err) {
                         console.log('sound for this pokemon is missing, using ding instead')
                         var audio = new Audio('static/sounds/ding.mp3')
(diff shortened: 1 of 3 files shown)


SiteEffect:
This does the needful! Very welcomed 👍 
[Ding!](https://www.youtube.com/watch?v=uVJLFfRYOj0)

templates/map.html
```diff
@@ -323,6 +323,14 @@
                 <span class="switch-label" data-on="On" data-off="Off"></span>
                 <span class="switch-handle"></span>
                 </label>
+              </div><br><br>
+              <h3>Use Cries</h3>
+              <div class="onoffswitch">
```
Arengo: Hide the cry option if no sound is requested. You can use a wrapper around ``Notify with sound`` like here: 
https://github.com/RocketMap/RocketMap/pull/1852/files#diff-6c79d9e636b5b03d3d6deab07d9e5ebaR314

static/js/map.js
```diff
@@ -764,8 +765,17 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
     if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
         if (!skipNotification) {
-            if (Store.get('playSound')) {
+            if (Store.get('playSound') && !Store.get('playCries')) {
                 audio.play()
+            } else if (Store.get('playSound') && Store.get('playCries')) {
+                var audiocry = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                audiocry.play().catch(function (err) {
+                    if (err) {
+                        console.log('sound for this pokemon is missing, using ding instead')
```
sebastienvercammen: Change to:
```
console.log('Sound for Pokémon ' + item['pokemon_id'] + ' is missing, using generic sound instead.')
```

static/js/map.js
```diff
@@ -764,8 +765,17 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
+                        var audio = new Audio('static/sounds/ding.mp3')
```
sebastienvercammen: `audio` is already a global variable set to an Audio object with this value. Don't unnecessarily make new objects. Optionally, if we want it to play a second time (by resetting to start) if it's already playing, we can add this before `.play()`:
```
// Reset the current time so it plays for the new notification.
audio.currentTime = 0
```

static/js/map.js
```diff
@@ -778,8 +788,16 @@ function customizePokemonMarker(marker, item, skipNotification) {
         var perfection = getIv(item['individual_attack'], item['individual_defense'], item['individual_stamina'])
         if (notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection) {
             if (!skipNotification) {
-                if (Store.get('playSound')) {
+                if (Store.get('playSound') && !Store.get('playCries')) {
                     audio.play()
+                } else if (Store.get('playSound') && Store.get('playCries')) {
+                    var audiocryiv = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                    audiocryiv.play().catch(function (err) {
```
sebastienvercammen: `lowerCamelCase` for variables: `audioCryIV`.

static/js/map.js
```diff
@@ -778,8 +788,16 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
+                        if (err) {
+                            console.log('sound for this pokemon is missing, using ding instead')
+                            audio.play()
+                        }
+                    })
                 }
                 sendNotification(getNotifyText(item).fav_title, getNotifyText(item).fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
```
sebastienvercammen: Unrelated to the PR, but this should be reworked, we shouldn't be calling `getNotifyText` twice in a row. You can add the small fix in this PR:
```
var notifyText = getNotifyText(item);
sendNotification(notifyText.fav_title, notifyText.fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
```

templates/map.html
```diff
@@ -323,6 +323,14 @@
                 <span class="switch-label" data-on="On" data-off="Off"></span>
                 <span class="switch-handle"></span>
                 </label>
+              </div><br><br>
+              <h3>Use Cries</h3>
```
sebastienvercammen: I would clarify `Use Cries` as `Use Pokémon Cries`.

templates/map.html
sebastienvercammen: Agreed with @Arengo, the "Use Pokémon Cries" slider should only be shown if `Notify with sound` is turned on.

static/js/map.js
```diff
@@ -764,10 +766,19 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
     if (notifiedPokemon.indexOf(item['pokemon_id']) > -1 || notifiedRarity.indexOf(item['pokemon_rarity']) > -1) {
         if (!skipNotification) {
-            if (Store.get('playSound')) {
+            if (Store.get('playSound') && !Store.get('playCries')) {
                 audio.play()
+            } else if (Store.get('playSound') && Store.get('playCries')) {
+                var audioCry = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
+                audioCry.play().catch(function (err) {
+                    if (err) {
+                        console.log('Sound for Pokémon ' + item['pokemon_id'] + ' is missing, using generic sound instead')
```
sebastienvercammen: End sentences with a dot `.`.

static/js/map.js
```diff
@@ -778,10 +789,17 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
+                } else if (Store.get('playSound') && Store.get('playCries')) {
+                    audioCry.play().catch(function (err) {
+                        if (err) {
+                            console.log('Sound for Pokémon ' + item['pokemon_id'] + ' is missing, using generic sound instead')
+                            audio.play()
+                        }
+                    })
                 }
-                sendNotification(getNotifyText(item).fav_title, getNotifyText(item).fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
+                sendNotification(notifyText.fav_title, notifyText.fav_text, 'static/icons/' + item['pokemon_id'] + '.png', item['latitude'], item['longitude'])
```
sebastienvercammen: You changed it to use `notifyText` just like you did above, but down here `var notifyText = getNotifyText(item)` isn't set.

static/js/map.js
```diff
@@ -778,10 +789,17 @@ function customizePokemonMarker(marker, item, skipNotification) {
...
-                if (Store.get('playSound')) {
+                if (Store.get('playSound') && !Store.get('playCries')) {
                     audio.play()
+                } else if (Store.get('playSound') && Store.get('playCries')) {
+                    audioCry.play().catch(function (err) {
+                        if (err) {
+                            console.log('Sound for Pokémon ' + item['pokemon_id'] + ' is missing, using generic sound instead')
+                            audio.play()
+                        }
+                    })
```
sebastienvercammen: This entire block of code is duplicate, it's 100% the same above. Move this to its own function, `playPokemonSound(pokemon_id)` and call it in both places instead.

This will also fix the problem that you are trying to use `audioCry` without it being set, because you forgot this part of the code above:
```
var audioCry = new Audio('static/sounds/cries/' + item['pokemon_id'] + '.wav')
```

static/js/map.js
Arengo: ... ``.``.

templates/map.html
```diff
@@ -323,8 +323,18 @@
...
                 </label>
+              </div><br><br>
+              <div id="pokemoncries" style="display:none">
+              <h3>Use Pokémon cries</h3>
+                <div class="onoffswitch">
+                <input id="cries-switch" type="checkbox" name="cries-switch" class="onoffswitch-checkbox" checked>
+                <label class="onoffswitch-label" for="cries-switch">
+                <span class="switch-label" data-on="On" data-off="Off"></span>
+                <span class="switch-handle"></span>
+                </label>
```
sebastienvercammen: Indentation inside the label is off.

static/js/map.js
```diff
@@ -747,7 +749,22 @@ function getNotifyText(item) {
     }
 }
 
+function playPokemonSound(pokemonID) {
+    var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.wav')
+    if (Store.get('playSound') && !Store.get('playCries')) {
+        audio.play()
+    } else if (Store.get('playSound') && Store.get('playCries')) {
```
sebastienvercammen: Can you rewrite these conditionals to start with:
```
// Don't do anything if sound is disabled.
if (!Store.get('playSound')) {
    return;
}
```
for clarity? This also simplifies the following conditionals.

Make sure to use that check *first*, above `var audioCry = ...`, to save some resources.

static/js/map.js
```diff
@@ -750,16 +750,20 @@ function getNotifyText(item) {
...
-    } else if (Store.get('playSound') && Store.get('playCries')) {
-        audioCry.play().catch(function (err) {
-            if (err) {
-                console.log('Sound for Pokémon ' + pokemonID + ' is missing, using generic sound instead.')
-                audio.play()
-            }
-        })
+    if (!Store.get('playSound')) {
+        return
+    } else {
```
sebastienvercammen: Don't `else` it, to avoid some ugly conditional nesting. :P

Just:
```
if (somethingBad) {
    // Stop.
    return
}

// Otherwise, all is good!
doStuff()
```

PR #2034, opened 2017-05-14
Set Interval before Hashing Keys are updated to prevent deadlocks

Alderon86:
## Description

Add an Interval before hashing keys are updated.
## Motivation and Context

There might be an issue for big instances with the HashKey report on /status.
Apparently running instances with a total of 10k+ workers makes it so that they're all competing to update the HashKey table, most likely resulting in locks.

## How Has This Been Tested?

Just on my local map, but i cant reach a worker count that high!

## Types of changes

## Checklist:


pogom/search.py
```diff
@@ -1126,14 +1127,10 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
-                    # Prepare hashing keys to be sent to the db. But only
-                    # sent latest updates of the 'peak' value per key.
-                    hashkeys = {}
-                    hashkeys[key] = key_instance
-                    hashkeys[key]['key'] = key
-                    hashkeys[key]['peak'] = max(key_instance['peak'],
-                                                HashKeys.getStoredPeak(key))
-                    dbq.put((HashKeys, hashkeys))
+                    # Delay hashing key db update to prevent deadlocks.
+                    t = threading.Timer(20.0, upsertKeys, [key, key_instance,
```
Arengo: If any upload is delayed, by thousands of queries because of the workers, why should it then not happen when all try to upload it 20 s later, at the same time?

Alderon86: you are welcome to post better suggestions on how to use intervals to delay it

Obihoernchen:
move it to search_overseer_thread :+1: (see discord)

pogom/search.py
Arengo: I don't know of a best solution. The one I think that works as temp fix is a delay by random 1-10 seconds

pogom/search.py
```diff
@@ -491,6 +490,9 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
 
     # The real work starts here but will halt on pause_bit.set().
     while True:
+        if args.hash_key is not None:
+            for key in args.hash_key:
+                upsertKeys(key, key_scheduler.keys[key], db_updates_queue)
```
Obihoernchen: This should be `upsertKey(...)` without an s ;)

pogom/search.py
```diff
@@ -1162,6 +1163,17 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
 
+def upsertKey(key, key_instance, db_updates_queue):
+    # Prepare hashing keys to be sent to the db. But only
+    # sent latest updates of the 'peak' value per key.
+    hashkeys = {}
+    hashkeys[key] = key_instance
+    hashkeys[key]['key'] = key
+    hashkeys[key]['peak'] = max(key_instance['peak'],
+                                HashKeys.getStoredPeak(key))
+    db_updates_queue.put((HashKeys, hashkeys))
```
sebastienvercammen: Instead of sending each key to the db update queue one by one, we should prepare all of the hashkeys and run them all in the same query. This'll use one db queue item rather than one per hashing key.

Obihoernchen:
```py
        start = timeit.default_timer()
        if (args.hash_key is not None and
                (start - hashkeys_last_upsert) > hashkeys_upsert_min_delay):
                    upsertKeys(
                        args.hash_key, key_scheduler, db_updates_queue)

def upsertKeys(keys, key_scheduler, db_updates_queue):
    for key in keys:
        key_instance = key_scheduler.keys[key]
    # Prepare hashing keys to be sent to the db. But only
```

PR #2035, opened 2017-05-14
Fix empty queue issue + set timeout for all requests

Obihoernchen:
## Description
This PR adds a 5s timeout to the API version check to fix the empty queue issue.
In some rare cases with some bad proxies the API version check might hang there forever and blocks all search queue updates. Therefore after some time the queue runs out of items and your instance stops scanning.

Furthermore it adds a 5s to all GET and POST requests.

## Motivation and Context
running without a version check wasn't the best solution for me ;)

## How Has This Been Tested?
my map + a map from another discord user

## Types of changes

## Checklist:

### DIFF
--- a/pogom/altitude.py
+++ b/pogom/altitude.py
@@ -17,7 +17,8 @@ def get_gmaps_altitude(lat, lng, gmaps_key):
         r_session = requests.Session()
         response = r_session.get((
             'https://maps.googleapis.com/maps/api/elevation/json?' +
-            'locations={},{}&key={}').format(lat, lng, gmaps_key))
+            'locations={},{}&key={}').format(lat, lng, gmaps_key),
+            timeout=5)
         response = response.json()
         status = response['status']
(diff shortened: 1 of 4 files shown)


Obihoernchen:
@sebastienvercammen yep good idea. Will do so!

Sk00DaT:
im experiencing this issue as well added the timeout but issue still persists :( hopefully when you add the other timeouts it fixes the problem

Obihoernchen:
@sebastienvercammen done. I've added 5s to all requests I found. 5s is default proxy timeout as well.

sebastienvercammen:
@Obihoernchen Ah. Perhaps it makes more sense to use `args.proxy_timeout` for the request timeouts as well, instead of a hardcoded `5` seconds?

sebastienvercammen:
RE Discord: We'll leave it at 5s for now. Proxy timeout would be confusing when someone isn't using proxies.

PR #2036, opened 2017-05-14
Allow newer api version than forced api version

Obihoernchen:
## Motivation and Context
save the fox

## How Has This Been Tested?
my map with some different api_version strings

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] My code follows the code style of this project.

PR #2037, opened 2017-05-14
API watchdog improvements

Obihoernchen:
## Description
Improving and fixing quick fix PR #2036
- [fix] exception if version check failed and returned int(0)
- [new] StrictVersion will handle the version format verification now
- [new] pause scanner if returned api version format is invalid
- [new] pause scanner if api version request failed
- [new] clear pause_bit (continue scanning) if the version check is successful again (api_version >= forced_version)
- [new] log messages are warning now instead of info
- improve log message syntax (don't use .format())

## Motivation and Context
- last PR was rushed and not 100% perfect
- found some "issues" with current watchdog logic

## How Has This Been Tested?
yes with various fake version numbers, see comments below
and with some bad proxies this time ;)

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1265,6 +1265,15 @@ def check_forced_version(args, api_version, api_check_time, pause_bit):
         api_check_time = int(time.time()) + args.version_check_interval
         forced_api = get_api_version(args)
 
+        if not forced_api:
+            # Couldn't retrieve API version. Stop scanning.
+            pause_bit.set()
+            log.warning('Forced API check got no or invalid response. ' +
+                        'Possible bad proxy.')
+            log.warning('Scanner paused due to failed API check.')
+            return api_check_time
+
+        # got a response let's compare version numbers.
         try:
             if StrictVersion(api_version) < StrictVersion(forced_api):
                 # Installed api version is lower. Stop scanning.
@@ -1285,12 +1294,23 @@ def check_forced_version(args, api_version, api_check_time, pause_bit):
             log.warning('Stop the scanner process until RocketMap ' +
                         'has updated.')
         except Exception as e:
+            # Something else happened. Stop scanning as well.
+            pause_bit.set()
             log.warning('Unknown error on API version comparison: %s', repr(e))
+            log.warning('Scanner paused due to unknown API check error.')
 
     return api_check_time
 
 
 def get_api_version(args):
+    """Retrieve forced API version by Niantic
+
+    Args:
+        args: Command line arguments
+
+    Returns:
+        API version string. None if request failed.
+    """
     proxies = {}
 
     if args.proxy:
@@ -1311,8 +1331,7 @@ def get_api_version(args):
             'https://pgorelease.nianticlabs.com/plfe/version',
             proxies=proxies,
             verify=False)
-        return r.text[2:] if r.status_code == requests.codes.ok else '0.0'
+        return r.text[2:] if r.status_code == requests.codes.ok else None
     except Exception as e:
         log.warning('error on API check: %s', repr(e))
-        # Do not stop scanning because check failed.
-        return '0.0'
+        return None
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1304,7 +1304,7 @@ def get_api_version(args):
         s = requests.Session()
         s.mount('https://',
                 HTTPAdapter(max_retries=Retry(total=3,
-                                              backoff_factor=0.1,
+                                              backoff_factor=0.4,
                                               status_forcelist=[500, 502,
                                                                 503, 504])))
         r = s.get(
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -493,7 +493,7 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
     while True:
 
         odt_triggered = (args.on_demand_timeout > 0 and
-                        (now() - args.on_demand_timeout) > heartb[0])
+                         (now() - args.on_demand_timeout) > heartb[0])
         if odt_triggered:
             pause_bit.set()
             log.info('Searching paused due to inactivity...')
@@ -1269,7 +1269,7 @@ def check_forced_version(args, api_version, api_check_time, pause_bit,
         forced_api = get_api_version(args)
 
         if not forced_api:
-            # Couldn't retrieve API version. Stop scanning.
+            # Couldn't retrieve API version. Pause scanning.
             pause_bit.set()
             log.warning('Forced API check got no or invalid response. ' +
                         'Possible bad proxy.')
@@ -1279,15 +1279,13 @@ def check_forced_version(args, api_version, api_check_time, pause_bit,
         # Got a response let's compare version numbers.
         try:
             if StrictVersion(api_version) < StrictVersion(forced_api):
-                # Installed API version is lower. Stop scanning.
+                # Installed API version is lower. Pause scanning.
                 pause_bit.set()
                 log.warning('Started with API: %s, ' +
                             'Niantic forced to API: %s',
                             api_version,
                             forced_api)
                 log.warning('Scanner paused due to forced Niantic API update.')
-                log.warning('Stop the scanner process until RocketMap ' +
-                            'has updated.')
             else:
                 # API check was successful and
                 # installed API version is newer or equal forced API.
@@ -1296,15 +1294,13 @@ def check_forced_version(args, api_version, api_check_time, pause_bit,
                     pause_bit.clear()
 
         except ValueError as e:
-            # Unknown version format. Stop scanning as well.
+            # Unknown version format. Pause scanning as well.
             pause_bit.set()
             log.warning('Niantic forced unknown API version format: %s',
                         forced_api)
             log.warning('Scanner paused due to unknown API version format.')
-            log.warning('Stop the scanner process until RocketMap ' +
-                        'has updated.')
         except Exception as e:
-            # Something else happened. Stop scanning as well.
+            # Something else happened. Pause scanning as well.
             pause_bit.set()
             log.warning('Unknown error on API version comparison: %s', repr(e))
             log.warning('Scanner paused due to unknown API check error.')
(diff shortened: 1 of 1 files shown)


FrostTheFox:
He tested with some fake unknown versions in Discord:
```
2017-05-15 00:33:05,852 [   search-overseer][        search][ WARNING] Started with API: 0.63.1, Niantic forced to API: 6.6.6
2017-05-15 00:33:05,852 [   search-overseer][        search][ WARNING] Scanner paused due to forced Niantic API update.
2017-05-15 00:33:05,852 [   search-overseer][        search][ WARNING] Stop the scanner process until RocketMap has updated.

2017-05-15 00:33:46,464 [   search-overseer][        search][ WARNING] Niantic forced unknown API version format: 0.666.6.hankedition
2017-05-15 00:33:46,464 [   search-overseer][        search][ WARNING] Scanner paused due to unknown API version format.
2017-05-15 00:33:46,465 [   search-overseer][        search][ WARNING] Stop the scanner process until RocketMap has updated.

2017-05-15 00:34:24,400 [   search-overseer][        search][ WARNING] Niantic forced unknown API version format: hanke.is.god.v1
2017-05-15 00:34:24,400 [   search-overseer][        search][ WARNING] Scanner paused due to unknown API version format.
2017-05-15 00:34:24,401 [   search-overseer][        search][ WARNING] Stop the scanner process until RocketMap has updated.

2017-05-15 00:37:06,328 [   search-overseer][        search][ WARNING] Niantic forced unknown API version format: 0.0.63.1
2017-05-15 00:37:06,328 [   search-overseer][        search][ WARNING] Scanner paused due to unknown API version format.
2017-05-15 00:37:06,328 [   search-overseer][        search][ WARNING] Stop the scanner process until RocketMap has updated.
```
Looks good to me 👍

pogom/search.py
```diff
@@ -1300,8 +1311,8 @@ def get_api_version(args):
             'https://pgorelease.nianticlabs.com/plfe/version',
             proxies=proxies,
             verify=False)
-        return r.text[2:] if (r.status_code == requests.codes.ok and
-                              r.text[2:].count('.') == 2) else 0
+        return r.text[2:] if r.status_code == requests.codes.ok else '0.0'
     except Exception as e:
         log.warning('error on API check: %s', repr(e))
-        return 0
+        # Do not stop scanning because check failed.
```
sebastienvercammen:
```
# Do not stop scanning because check failed.
```
Except you're wrong, the intended behavior **is** to stop if we can't get the version.

It won't give any problems with a one-time connection problem because we retry the version check several times. And if we can't get the version in several queries, something is wrong with either our connection or Niantic's connection, and scanning most likely won't be stable.

Obihoernchen: Old behavior was to continue scanning as well. I didn't change the behavior.
```python
        if (api_version != forced_api and forced_api != 0):
            pause_bit.set()
...

def get_api_version(args):
        log.warning('error on API check: %s', repr(e))
        return 0
```

See:
https://github.com/RocketMap/RocketMap/blob/7d854bf3dac066ff90acb1fce2aec97a57d91930/pogom/search.py#L1267
https://github.com/RocketMap/RocketMap/blob/7d854bf3dac066ff90acb1fce2aec97a57d91930/pogom/search.py#L1305

FrostTheFox:
@sebastienvercammen I was under the impression that he had tested the original merge, considering he posted the code earlier before I merged the API update, which is why I merged it, because I could understand the code enough and I figured he tested it. The pasted message was said *after* I merged.

Also, I don't believe that stopping just because we're above the merged version is a bad thing. You're free to feel different, but let's be realistic, they're not going to just decide to reset or change versioning in the near future. Merging the latest API before it's forced *is* a good thing, we shouldn't wait until it's forced before adding support for the latest API. Even if they do, Information is practically instant via Discord, and that's 1 ***possible patch*** we would have to make to the watchdog, while allowing us to simply update the API version preemptively. I think with this the watchdog still sufficiently does it's job, while making it easier for us in the future. 

I really doubt they're going to try and lay some sort of trap with versioning.

friscoMad:
That's the problem when there are not tests or formal definition, you are never sure what's a feature or a bug. it's clear that the parsing code could and should be improved. So let's define what should be the behavior and then implement that.

sebastienvercammen:
@FrostTheFox I agree with @friscoMad. It's not just about this one small feature (the watchdog) that could or could not be broken by *a possible* update by Niantic just to mess with the map.

It's about having to come online to a series of commits of:

* update/fix X
* whoops
* fix
* wait, now is the real fix
* PR #X: properly implement last merges

With behavior of code sometimes being changed without ever having been discussed. Or, well, it was discussed, and every single time I explained why there are things to discuss about the Watchdog that are important enough that we can't just change the behavior of the old code.

And if there's anything we've learned from the past months, it's that *"I was under the impression of ..."* has happened several times before, and that's why we have testing and the rule of 2x approvals in place. I mean, I'm not blaming you, it's happened to me as well. But if we want people to stop doing it, they need to start following procedure.

@Obihoernchen I know the old code continued if it couldn't get the version, but the correct behavior is to  stop. If you implement a feature to strictly follow the API version to protect your users/accounts, then it doesn't make any sense to continue if you can't get the version for some reason. Something *might* be wrong, and that's reason enough to stop and investigate.

We'll probably want to increase the backoff factor for the retries (those that are in now, are copied from the webhooks) though, to make sure we give it enough time to retry properly.

Obihoernchen:
Yep I'll change it to:
- pause scanning if version format is invalid, but do another check really soon (don't wait another 60-300s)
- pause scanning if request failed, but do another check really soon (don't wait another 60-300s)
- clear the pause_bit if the version check is successful again (but respect args.on_demand_timeout!)

stays the same:
- pause scanning if forced version is greater
- continue scanning if pgoapi version is greater or equal forced version

This should be super safe and unreliable requests shouldn't cause too much trouble because of the decreased recheck delay.

What do you think @sebastienvercammen and others?

neskk:
@Obihoernchen Your idea looks good.
I would make a cycle with N attempts, briefly spaced between (2/3 secs) to try and make the request. If it fails after N attempts, either because request failed or format is invalid, scanning is paused. After a certain amount of time the check should be performed again and unpause/keep scanning if it is successful.
Both parameters, number of attempts and API version check interval, could be configurable or not, I don't think it's really a requirement...

sebastienvercammen:
@Obihoernchen Agreed with @neskk: the cycle of N attemps with spaced timing is already being done with the retries and [backoff factor](https://github.com/kennethreitz/requests/blob/master/requests/packages/urllib3/util/retry.py#L112).

The watchdog should indeed keep running in the background and unpause the scanner if everything is 👍 

I suggest increasing the backoff factor to handle the retries rather than retrying the request ourselves like @Obihoernchen suggested. The continued retries every `--version-check-interval` seconds will handle the rest.

FrostTheFox:
Mkay, I don't agree that an API update needs a pull request and 3x approval. I'll admit, it was poor judgement on my part at the time to change watchdog as opposed to reverting the API commit. However, if you consider the fact my brain was set on getting the new API version working for everyone, it's easy to see my reasons in merging the PR.

I do typically adhere to the procedure where it is necessary. I don't consider docs updates, or other non code updates as items that should require 3x approval and a PR (including API version updates). There has been some deviations in the past, probably the most recent to mind being the change that made hash key required to run. Anything I'm not sure about, you know I open a PR or otherwise ask #pr / @sebastienvercammen.

sebastienvercammen:
@FrostTheFox The Watchdog update is what I meant. With 2x approvals, we'd have followed procedure, and we'd either have avoided a merge without testing or we'd know who approved without having tested it.

Just asking in #pr doesn't always have the same effect as requiring 2x approvals. With an approval, at least people have been taking responsibility and have been reviewing fairly well as opposed to a quick "yeah it *looks* okay" in #pr.

FrostTheFox:
I understand that, in fact I just told you I'll admit that under the circumstances the right move would not have been to move forward with the api update at all; the API update should have been reverted.

@sebastienvercammen Also, to my knowledge it was and always had been 3x approval, when did we discuss it being 2?

sebastienvercammen:
@FrostTheFox According to Discord logs, at least since 12/30/2016.

Obihoernchen:
I think I've implemented all requested changes. Please have a look.

pogom/search.py
```diff
@@ -1260,26 +1262,61 @@ def stat_delta(current_status, last_status, stat_name):
...
+                # API check was successful and
+                # installed API version is newer or equal forced API.
+                # Continue scanning if on_demand_timout isn't triggered.
+                if not odt_triggered:
+                    pause_bit.clear()
+
+        except ValueError as e:
+            # Unknown version format. Pause scanning as well.
+            pause_bit.set()
+            log.warning('Niantic forced unknown API version format: %s',
```
sebastienvercammen: End sentence with a dot.

pogom/search.py
```diff
@@ -1260,26 +1262,61 @@ def stat_delta(current_status, last_status, stat_name):
...
+                        forced_api)
+            log.warning('Scanner paused due to unknown API version format.')
+        except Exception as e:
+            # Something else happened. Pause scanning as well.
+            pause_bit.set()
+            log.warning('Unknown error on API version comparison: %s', repr(e))
```
sebastienvercammen: End sentence with a dot.

pogom/search.py
```diff
@@ -1293,15 +1330,14 @@ def get_api_version(args):
         s = requests.Session()
         s.mount('https://',
                 HTTPAdapter(max_retries=Retry(total=3,
-                                              backoff_factor=0.1,
+                                              backoff_factor=0.4,
```
sebastienvercammen: Instead of an arbitrary `0.4`, we can just use `0.5`.

pogom/search.py
```diff
@@ -1293,15 +1330,14 @@ def get_api_version(args):
...
             'https://pgorelease.nianticlabs.com/plfe/version',
             proxies=proxies,
             verify=False)
-        return r.text[2:] if (r.status_code == requests.codes.ok and
-                              r.text[2:].count('.') == 2) else 0
+        return r.text[2:] if r.status_code == requests.codes.ok else None
     except Exception as e:
         log.warning('error on API check: %s', repr(e))
-        return 0
+        return None
```
sebastienvercammen: Instead of returning `None` on failure, we can return `False`. The rest of the logic would stay the same, but I find `False` to be clearer. Thoughts?

PR #2041, opened 2017-05-15
Fix file path errors and remove useless code.

neskk:
## Description

When utils.py is imported `config` object isn't completely initialized, so code (`get_args()`) that used `config['ROOT_PATH']` wasn't working properly, specifically the code that handles the parsing of webhooks whitelist file. `extract_sprites` suffered from the same problem, but this time it required a different fix since this method is called before Flask app is initialized so, app.root_path isn't available and `config['ROOT_PATH']` can't be initialized in time.

## Motivation and Context

Fixes:
https://github.com/RocketMap/RocketMap/issues/2039
and
https://github.com/RocketMap/RocketMap/issues/2030

Better handling of root_path to ensure that resources are found when required by some functionality (i.e. static01.zip and pokemon.min.json).

## How Has This Been Tested?
Local machine, Win7 x64, python 2.7.12.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1947,7 +1947,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     hlvl_account = account
                     hlvl_api = api
                 else:
-                    # Get account to use for IV or CP scanning.
+                    # Get account to use for IV and CP scanning.
                     hlvl_account = account_sets.next('30', scan_location)
 
                 # If we don't have an API object yet, it means we didn't re-use
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1948,8 +1948,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     hlvl_api = api
                 else:
                     # Get account to use for IV or CP scanning.
-                    if pokemon_id in args.enc_whitelist:
-                        hlvl_account = account_sets.next('30', scan_location)
+                    hlvl_account = account_sets.next('30', scan_location)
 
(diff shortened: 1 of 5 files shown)


pogom/models.py
```diff
@@ -1948,8 +1948,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     hlvl_api = api
                 else:
                     # Get account to use for IV or CP scanning.
```
sebastienvercammen: 👍 The conditional was left over from when someone changed the L25/L30 account pool code to be L30 only (after Niantic's second IV/CP update).

This is absolutely trivial and unimportant, but could you change `for IV or CP` to `for IV and CP`?

runserver.py
```diff
@@ -325,8 +328,19 @@ def main():
...
+                                      f.read().splitlines()]
+    elif args.webhook_blacklist_file:
+        with open(args.webhook_blacklist_file) as f:
+            args.webhook_blacklist = [get_pokemon_id(name) for name in
+                                      f.read().splitlines()]
+    else:
+        args.webhook_blacklist = [int(i) for i in
+                                  args.webhook_blacklist]
+        args.webhook_whitelist = [int(i) for i in
+                                  args.webhook_whitelist]
```
sebastienvercammen: Can you add `frozenset()` around the lists? `O(n)` to `O(1)` for `in` membership check.

pogom/models.py
neskk: Sure ;)

runserver.py
neskk: Ok, changing this and will push in a minute...

PR #2042, opened 2017-05-15
Adjust Pokemon default details

Arengo:
Making available information available without encountering.

## Description

The API currently sends ``gender`` as well as Unown's ``form`` inside the ``pokemon_display`` list even when the Pokémon is not encountered. (Even when just seen in ``nearby`` and not in ``wild_pokemon`` but we do not parse ``nearby``, yet)
This PR parses the information, which is already available, instead of limiting it to encountered Pokémon. The front-end was changed accordingly so that Gender is displayed even when no height/weight is known, which contrarily needs an encounter.
Additional small fix by using ``pokemon_id`` variable instead accessing the dict again for it. 

## Motivation and Context

Show what is retrievable, why not.

## How Has This Been Tested?

On my local setup.

## Screenshots
*Location has been cropped out by intent.*

bildschirmfoto 2017-05-15 um 18 31 13

bildschirmfoto 2017-05-15 um 18 31 07

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1925,13 +1925,12 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
             disappear_time = now_date + \
                 timedelta(seconds=seconds_until_despawn)
 
-            printPokemon(p['pokemon_data']['pokemon_id'], p[
-                         'latitude'], p['longitude'], disappear_time)
+            pokemon_id = p['pokemon_data']['pokemon_id']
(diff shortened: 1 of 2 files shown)


BrownSlaughter:
yes gender and form were the same when I tested with some lvl 2, lvl 25 and level 30 accounts a couple of weeks ago after the lvl30+IV change and there have been no changes that I know of since

SiteEffect:
Ready for approval review.

Sk00DaT:
ive been running this for a while have had no issues at all both gender and form are present without encounter

PR #2043, opened 2017-05-15
[Hotfix] Missing gym-sidebar

Arengo:
Various reports today, after weeeks, that gym-sidebar is missing. Did your cache timer expire today?

## Description

~~Showing gyms is enough reason to let the user decide to additionally show the gym-sidebar. Not having ``-gi`` enabled shouldn't be a reason to prevent users accessing the sidebar while there already is a default disclaimer that additional gym-info might be missing.~~
To display the gym-info sidebar you need to add the ``-gi`` or ``--gym-info`` flag to your ``-os`` instance. Otherwise you can't select to display or not displaying the sidebar while those who have an old cache will show the behaviour selected before. With an documentation update, this hopefully shall now be visible and known for the future.

## Motivation and Context

#help 

You know how the gym-sidebar looks like - or have you forgotten?

## Types of changes

- [x] Documentation

## Checklist:

### DIFF
--- a/docs/extras/gyminfo.md
+++ b/docs/extras/gyminfo.md
@@ -25,8 +25,12 @@ python runserver.py -gi
 Or update your config.ini:
 
 ```
-gym-info:true            # enables detailed gym info collection (default false)
+gym-info	            # enables detailed gym info collection (default false)
 ```
+
+## Displaying the gym details sidebar in the front-end
+
+To show the gym details sidebar on the front-end, ``-gi`` needs to be enabled on your webserver instance. To scan gyms for the gym details, ``-gi`` must also be enabled on your scanner instances. Hence, if have your scanners separated from the webserver instance, make sure to explicitly specify the gym details flag in the ``--only-server`` (or ``-os``) webserver instance of your setup.
 
 ## New Webhook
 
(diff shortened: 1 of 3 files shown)


docs/extras/gyminfo.md
```diff
@@ -25,9 +25,13 @@ python runserver.py -gi
 Or update your config.ini:
 
 ```
-gym-info:true            # enables detailed gym info collection (default false)
+gym-info			# enables detailed gym info collection (default false)
 ```
 
+## Displaying the gym details sidebar in the front-end
+
+To show the gym details sidebar on the front-end, ``-gi`` needs to be enabled on your webserver instance. To scan gyms for the gym details, ``-gi`` must also be enabled on your scanner instances. Hence, if have your scanners separated from the webserver instance, make sure to explicitly specify the gym details flag in the ``--only-server`` (or ``-os``) webserver instance of your setup.
```
sebastienvercammen: Remove this part, it's repetition:
```
Hence, if have your scanners separated from the webserver instance, make sure to explicitly specify the gym details flag in the ``--only-server`` (or ``-os``) webserver instance of your setup.
```

PR #2048, opened 2017-05-17
add mp3 support for cries

BrownSlaughter:
check for mp3 format cries

first it checks for wav, then mp3 and then default ding
## Motivation and Context
Due to wav file sizes of voices it was better to convert to mp3 to save bandwith.
This also gives more choice to the user without having to convert sounds.
Also explorer can't play wav files

## How Has This Been Tested?
on a map playing mp3 sounds and wav for notifications

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -63,7 +63,7 @@ var lastUpdateTime
 const gymTypes = ['Uncontested', 'Mystic', 'Valor', 'Instinct']
 
 const audio = new Audio('static/sounds/ding.mp3')
-const cryFileTypes = ['mp3', 'wav']
+const cryFileTypes = ['wav', 'mp3']
 
 const genderType = ['♂', '♀', '⚲']
 const unownForm = ['unset', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '!', '?']
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -18,7 +18,7 @@ var $selectSearchIconMarker
 var $selectLocationIconMarker
 var $switchGymSidebar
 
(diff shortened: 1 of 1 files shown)


static/js/map.js
```diff
@@ -777,11 +777,17 @@ function playPokemonSound(pokemonID) {
...
     } else {
-        var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.wav')
+        var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.mp3')
         audioCry.play().catch(function (err) {
             if (err) {
-                console.log('Sound for Pokémon ' + pokemonID + ' is missing, using generic sound instead.')
-                audio.play()
+                console.log('MP3 Sound for Pokémon ' + pokemonID + ' is missing, checking for Wav.')
+                var audioCryWav = new Audio('static/sounds/cries/' + pokemonID + '.wav')
+                audioCryWav.play().catch(function (err) {
```
sebastienvercammen: Don't keep nesting these catches. Instead, use a list (array) of items you'd like to try, and try them from first to last until you've found one that works. This will simplify your code a lot and it will avoid nested code.

static/js/map.js
```diff
@@ -777,11 +777,17 @@ function playPokemonSound(pokemonID) {
     if (!Store.get('playCries')) {
         audio.play()
     } else {
-        var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.wav')
+        var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.mp3')
         audioCry.play().catch(function (err) {
             if (err) {
-                console.log('Sound for Pokémon ' + pokemonID + ' is missing, using generic sound instead.')
-                audio.play()
+                console.log('MP3 Sound for Pokémon ' + pokemonID + ' is missing, checking for Wav.')
```
sebastienvercammen: `wav`, not `Wav`.

static/js/map.js
BrownSlaughter: oh I didn't know I could do it like that. I shall look into it

PR #2049, opened 2017-05-17
Continue to check API version if scanning is paused

### LINKED ISSUE #2047: API Watchdog check fails due to proxy -> pauses all scans
capture

Watchdog check failed due to Proxy timeout. Scan now paused. No attempt to retry the check was made before I closed the scan 20 minutes later.

Now running with -novc to prevent this occurring.

Have 10 proxies in use for the scan, with the following config settings:

proxy-display: full
proxy-rotation: random
proxy-refresh: 3600

I have restarted having updated the refresh to 600 and added proxy-timeout: 15, but given that I am running with -novc I do not know if this fixes my issue.

Perhaps the watchdog needs to be retried x number of times using all available proxies in the instance of a failure, to prevent false negatives and pausing the scan.

Obihoernchen:
## Description
- continue to check API version if scanning is paused (fix #2047)
- Don't check API version if on-demand-timeout is triggered and scanning is paused anyways (cosmetic change)

## Motivation and Context
All good things come by threes.

## How Has This Been Tested?
with a bunch of bad proxies.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -503,11 +503,10 @@ def search_overseer_thread(args, new_location_queue, pause_bit, heartb,
             for i in range(0, len(scheduler_array)):
                 scheduler_array[i].scanning_paused()
             # API Watchdog - Continue to check API version.
-            if not args.no_version_check:
+            if not args.no_version_check and not odt_triggered:
                 api_check_time = check_forced_version(args, api_version,
                                                       api_check_time,
-                                                      pause_bit,
-                                                      odt_triggered)
+                                                      pause_bit)
             time.sleep(1)
 
         # If a new location has been passed to us, get the most recent one.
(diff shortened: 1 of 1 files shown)


shamy13:
How to change API?
"Started with API: 0.61.0, Niantic forced to API: 0.63.1"

Obihoernchen:
@shamy13 `sudo pip install -r requirements.txt --upgrade`

shamy13:
@Obihoernchen still com: "Started with API: 0.61.0, Niantic forced to API: 0.63.1"

```
 git pull
 npm run build
```

Obihoernchen:
Update RocketMap you definitely use a old pull

hansendc-st:
I pause my scanner when it is not in use, and I noticed it recently un-pausing by itself.  I believe it's due to this change so that every time the API check is successfully performed, it un-pauses, no matter the original reason for the pause.

Could we have it record if the original pause was due to the API checking, and only un-pause when API checks are successful if the original pause was due to API checking?

Obihoernchen:
How do you pause your scanner?

hansendc:
I've tried a couple of different ways.  I have a curl script (for doing it via cron), a standalone web page outside of RocketMap with a button, and I've also tried just using the normal old RocketMap web interface slider button.  I haven't seen any difference in behavior.

Are my hacks doing something wrong, perhaps?

hansendc-st:
I have verified that this behavior occurs when just using the "Search" slider in the normal RocketMap web interface alone.  It seems like the pause-from-the-web-interface behavior should be the same as the odt_triggered behavior.  We *only* want to resume scanning because of an API check if scanning was paused due to API check issues.

PR #2050, opened 2017-05-18
Improved encounter- and webhook whitelist file parsing.

sLoPPydrive:
## Description
This allows Pokemon names or IDs to be present in whitelist files. Lines starting with “#” or “-“ are being ignored, so basically you have a combined black- and whitelist. I also removed separate blacklists and the possibility to give lists directly via commandline, so `-wwht`, `-wblk` and `-wblkf` are gone!

The file you edit looks something that (note that Pokemon names have to be english, like defined in the [pokemon.json](https://github.com/RocketMap/RocketMap/blob/develop/static/data/pokemon.json) file):
```
# ================
# = Generation 1
# ================
Bulbasaur
Ivysaur
Venusaur
Charmander
Charmeleon
Charizard
Squirtle
Wartortle
Blastoise
-Caterpie
-Metapod
Butterfree
-Weedle
-Kakuna
Beedrill
-Pidgey
-Pidgeotto
...
# ================
# = Generation 2
# ================
Chikorita
Bayleef
Meganium
Cyndaquil
Quilava
Typhlosion
Totodile
Croconaw
Feraligatr
-Sentret
Furret
...
```
Each line starting with "#" or "-" is being ignored which you can use for either **comments** or to **exclude Pokemon**. The example above excludes some common Pokemon while others get included. **This PR includes an example file containing all 251 Pokemon** from both active generations, so there is no need to create such a file by hand.

**This PR is backwards compatible, so if you already have a whitelist containing numeric Pokemon IDs you can still use your existing whitelist file!** Note that blacklist files are gone though.

These lists can be used for encounter- and webhook whitelists. A common setup might be to actually **use the exact same file**, so every Pokemon you encounter will get sent to the webhooks and others won't.

Using this file format makes blacklists redundant, so I removed all traces of the webhook blacklists. Also I don't think providing a list of Pokemon IDs on the commandline isn't really comfortable, so I also removed `-wwht` (webhook whitelist). Now encounters and webhooks each have exactly one option to provide a whitelist file. **Less is more!**

The **behavior of empty whitelists** is a little bit different between encounters and webhooks. An **empty encounter whitelist** means: **NO, do not encounter**. An **empty webhook whitelist** means: **YES, send to webhook**. If you think about it this is a _reasonable_ decision, since you really want to control your encounters (so you have to provide a list) on the one hand and on the other hand there is not much harm done if you don't limit your webhook sends.

## Motivation and Context
I believe it's just more comfortable to edit a file and check/uncheck the Pokemon you want instead of looking up Pokemon IDs and having a bazillion command line options to control what you want and what not.

## How Has This Been Tested?
On my local map.

## Types of changes

## Checklist:


sebastienvercammen:
It was voted by our PR members that all Pokémon lists would be by Pokémon ID only rather than Pokémon names. You could say it takes a small extra effort on the host's part, but it avoids a constant stream of these issues in our help channel on Discord:

- Non-English people trying to use Pokémon names in their own language.
- Misspelled Pokémon names (pebkac) leads people to believe the map is broken.
- Issues with Pokémon names such as `Nidoran♂`.
- Inconsistency with how other files are using Pokémon IDs rather than names.

With Pokémon IDs, merging the white- and blacklist doesn't make as much sense; if you want to encounter only a few (whitelist) you don't have to look up many Pokémon IDs, and if you want to encounter most and disable a few (blacklist) you also have less work.

However, if you want to continue work on this PR, I do agree that we can improve the handling. So perhaps you can introduce comments on the same line as the Pokémon IDs, and add the Pokémon's name behind the ID, like this:
```
1 # Bulbasaur
```
which would let us add a clear default whitelist example file without people having to look up IDs.

And if you make the in-line comments work, we can still keep the combined white-/blacklist, because that would solve the issue of having to look up more Pokémon IDs.

SiteEffect:
I agree to basically everything, @sebastienvercammen wrote above. Not always the case 😃 
Also you could then provide one example file ``filters/pokemon.txt.example`` with
```txt
1  # Bulbasaur
2  # Ivysaur
...
```
So people actually see which number is what instead creating a potentially huge list by hand what's really uncomfortable.

friscoMad:
Even if I think that the current number of flags is a problem, this will always be a whitelist, and for the ones that only want to block a small number of pokes it is a problem, and not future proof, as you are adding some parsing to the file maybe we can ask some more and add in the file if it is a white or black list with some extra parameter defaulting to white list if not present, that way we would be able to keep all the features with half the command line options.

sLoPPydrive:
I see your point. If by default every Pokemon is included it will be a lot of work to exclude most and just include a few (if you don't have a text editor that can work in column mode). The other way round is easier: Just search and replace all minus signs by empty string. ;-)

One could use a format like this:
```
+1  # Bulbasaur
-2  # Ivysaur
...
```
So all lines with + get included and those with - excluded. Now it's quite easy to switch from white- to blacklist by using search-and-replace (which every editor offers). What do you think?

friscoMad:
I like the proposal but I am not sure if mixing include and exclude in the same file is a good idea, if see no point in mixing them and could lead to more user problems.
What I was thinking was something like a YAML file with 2 properties the first one the type (white or blacklist defaulting to white) and a second one a list with all ids.
That is pretty clear, human readable, allows comments and it is imposible to mix white and blacklist elements in the same file, and just need 1 cl option for both types.

SiteEffect:
@sLoPPydrive this is, imo, a nice idea but what is now the default? If so, we only need one either + or -. Thus, we are back to a black/whitelist. A new proposal: Use directly a ``.json`` which is the proper way today and there area already filters around for PokeAlarm.

sebastienvercammen:
@friscoMad YAML would add another dependency, we currently don't use YAML.

@SiteEffect JSON would add complexity, not make it simpler. I disagree with PokeAlarm's format.

@sLoPPydrive Agreed with the others, we need both a black- and a whitelist.

PR #2052, opened 2017-05-18
Add the ability to specify zoom level

captainmedic:
These changes add onto the already existing ability to append to ?lat={{lat}}&lon={{lon}} a &zoom={{zoom}}

## Description
Changes were made to the files that are associated with the ability to call map specifics from the address bar of the browser.

## Motivation and Context
I thought about adding this pull request for those who utilize third party alarm systems such as pokealarm. Pokealarm has the ability to link users straight to google maps at GPS coordinates, however, having the ability to channel users to the map creators website could have a couple of benefits. One the website owner would be able to track hits from the pokealarm for statistical purposes. It would also allow for the most accurate time remaining and any additional info about Pokestops in the area and even other Pokemon in the nearby vicinity.

## How Has This Been Tested?
This was tested on a Windows testing environment, to ensure that it would not run into any errors.

example

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -363,6 +363,8 @@ function initSidebar() {
     $('#spawnpoints-switch').prop('checked', Store.get('showSpawnpoints'))
     $('#ranges-switch').prop('checked', Store.get('showRanges'))
     $('#sound-switch').prop('checked', Store.get('playSound'))
+    $('#pokemoncries').toggle(Store.get('playSound'))
+    $('#cries-switch').prop('checked', Store.get('playCries'))
     var searchBox = new google.maps.places.Autocomplete(document.getElementById('next-location'))
     $('#next-location').css('background-color', $('#geoloc-switch').prop('checked') ? '#e0e0e0' : '#ffffff')
 
@@ -768,7 +770,25 @@ function getNotifyText(item) {
     }
 }
 
+function playPokemonSound(pokemonID) {
+    if (!Store.get('playSound')) {
+        return
+    }
+    if (!Store.get('playCries')) {
+        audio.play()
+    } else {
+        var audioCry = new Audio('static/sounds/cries/' + pokemonID + '.wav')
+        audioCry.play().catch(function (err) {
+            if (err) {
+                console.log('Sound for Pokémon ' + pokemonID + ' is missing, using generic sound instead.')
+                audio.play()
+            }
+        })
+    }
+}
+
 function customizePokemonMarker(marker, item, skipNotification) {
+    var notifyText = getNotifyText(item)
     marker.addListener('click', function () {
         this.setAnimation(null)
         this.animationDisabled = true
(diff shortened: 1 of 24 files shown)


cptmedic:
I do welcome comments from more experienced coders on the implementation of this. I am sure it is botchy as my python skills are very little. Also to have it successfully work in the map.js file I changed `zoom:Store.get('zoomLevel')` to `zoom: centerZoom` the variable that is now assigned in the map.html. With this change it appears that if you resize the page and refresh it goes back to the default zoom specified which is 16. Would there be a better way to implement this?

peterchabs:
Awesome!
I've been wanting this feature for a bit, hope it works out!

friscoMad:
There is no need to pass the value from falsk to html, just parse it from the UI.
It is more code but cleaner, you could also use the same for lat and long.
See: http://stackoverflow.com/questions/901115/how-can-i-get-query-string-values-in-javascript

Also that way you can check if the user has some zoom stored and use it when it is not url defined, as right now you are removing the feature of storing the zoom between visits.

cptmedic:
Thanks for the suggestion. I have implemented this method of calling the resource instead of  making the changes to app.py. I have removed those commits and pushed the other method. Currently it is working well on my windows test build. Now user saved zoom levels will remain in effect unless specified by `&zoom=`

static/js/map.js
```diff
@@ -114,6 +114,10 @@ function removePokemonMarker(encounterId) { // eslint-disable-line no-unused-var
 }
 
 function initMap() { // eslint-disable-line no-unused-vars
+    if (centerZoom) { // checks to see if specified in url
```
friscoMad: You shouldn't store it, just use it when centering the map.
Otherwise we are overriding his stored zoom when opening a zoomed map.

templates/map.html
```diff
@@ -420,8 +420,18 @@
     <script src="{{ url_for('static', filename='dist/js/app.min.js').lstrip('/') }}"></script>
     <script src="{{ url_for('static', filename='js/vendor/classie.js').lstrip('/') }}"></script>
     <script>
-      var centerLat = {{lat}};
-      var centerLng = {{lng}};
+        function getParameterByName(name, url) {
```
friscoMad: Add the function directly to map.js
Drop lat/long parsing in app.py as we will be getting it from js

captainmedic: Thanks again for the suggestions. I have moved the function itself to map.js. However, when in the map.js file we lose functionality of it, in that it will no longer use the zoom that is define by url. I have tried several different methods to no avail, at least on my windows environment. I will try again tomorrow when I am not so tired to figure out why it doesn't work when moving the function, if you have any ideas I'd be open to it.

templates/map.html
```diff
@@ -420,8 +420,18 @@
...
+            if (!url) url = window.location.href;
+            name = name.replace(/[\[\]]/g, "\\$&");
+            var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
+                results = regex.exec(url);
```
sebastienvercammen: Oh dios, mis ojos. Take a look at `window.location.search`.

friscoMad:
@sebastienvercammen this is ready to be merged.

Issue #2051, opened 2017-05-18
API Watchdog check using bad proxy??


evilmoses:
Using latest version to test watchdog fix, and it works. The check is being performed again after a failure. Log shows something weird now though, maybe an error, maybe coincidence....

![capture](https://cloud.githubusercontent.com/assets/23256850/26191968/e485722a-3ba7-11e7-8c77-ab4b606125f9.PNG)

It looks in the screenshot that the Watchdog check is using a proxy that is known as being bad? The Proxy refresh is performed, and then one of the timed out proxies is being used for the check? Is that just because it happened so soon after the refresh,. or is there an issue?

At this point, the scans were paused to the watchdog check failure.

sebastienvercammen:
The connection to the Niantic server to check the version has timed out. This is most likely due to a slow-ish proxy.

I'll add a config item to customize the timeout on requests rather than having it hardcoded as it is now. I've assigned myself to remember.

evilmoses:
So setting the proxy-timeout: 15 makes no difference in this case currently? That only applies to the refresh?

The issue I raised, was that the Proxy refresh is done (first few lines), and then the API watchdog check is done afterwards, but using a proxy that had just been determined to be bad, instead of using one of the 3 known to be good. I blacked out all my proxies but the one that is bad to show you. Just seemed weird to use a known bad proxy to do the watchdog check?

friscoMad:
I am not sure but I can see this happening if you have proxy-file and proxy-refresh enabled, due to threads contexts switching overseer thread could start the version check before the proxy check was ended (it has a 5 sec timeout and the error is shown just 1.1 sec after the refresh end message, probably the proxy was selected before the proxy refresh start.

tobias86aa:
I have problem related to this thing to.

21:59:44,492 [   search-overseer][        search][ WARNING] error on API check: IOError(24, 'Too many open files')

I have been forced to shutdown the API watchdog on most of my maps.

FrostTheFox:
@tokar86a That's a different error, relating to ulimit. That's not an RM code issue. Google "too many open files".

FrostTheFox:
Also, I didn't dig in too far, but just from peeking at the watchdog code, it is implementing the same mechanism for fetching a new proxy as the scan threads do (calling get_new_proxy from proxy.py).

I would be inclined to agree with @friscoMad, based on the times your logs are showing.

sebastienvercammen:
[...]

PR #2054, opened 2017-05-19
Improved help and removed unused options

friscoMad:
## Description
This pr tries to help with CL options clutter by removing 3 optiosn:

- Encryption Library: No longer in use
- Altitude: Used as a default of a fallback, refactored to improve make it fixed
- Disable version check: Changed code to disable the checks using version-check-interval with 0

Also I have added groups to options so the help is clearer and related options are grouped
And several other improvements:

-  Removed step default and changed to required
- Added required to hash key
- Added required to workers
- Added exclusion to -ss and -speed
- Added exclusion to -ns and -os
- Detailed db type in mysql only options
- Fixed max db connections description
- Added group headers in docs
...

## Motivation and Context
I just wanted to remove -el ENCRYPT_LIB and ended up like this 🤷‍ 3 hours later

## How Has This Been Tested?
Just small instance runs, will test more later

## Types of changes

## Checklist:

- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.


Alderon86:
Can you also address Issue #2051 ? would be good to have all watchdog changes made together.

friscoMad:
@Alderon86 there is no such error and but I do not see how it is related to this PR anyways

friscoMad:
Rebased and added docs for --api-version

SchroedingersKater:
I like the general clean-up but limiting scans to practically speed is a step back. General hex scan is still viable and important i. e. for tutorials. So imo, leave out the PR part of ss and speed.

friscoMad:
Rebased and improved.
Doc file now shows horizontal scrollbars due to the width of lines, it can be fixed manually but it is not nice to be changing between printers just to update docs.

Anyway we should start instructing ppl to use -h to see this instead of going to wiki.

friscoMad:
I am closing this as it seems there was no interest on merging this for almost a year and it takes a lot to keep up with flag changes.

PR #2057, opened 2017-05-22
Stop using Denver as default location for accounts

Arengo:
With a proper API simulation, we want to submit a valid player location for our scanned locations.
Side-change: Encoded the ``config.ini.example`` to proper line-endings via unix LF UTF-8

## Description

Instead of passing a constant location for get_player API-Request, we now define a soft-coded one by defaulting it to current default and altering it afterwards according to geopy GoogleV3 reverse geolocation.
While ``country`` and ``timezone`` is somewhat specific to the scan location, ``language`` is not. So this remains at the default ``en`` if not the flag ``--locale`` is adjusted in the config.
Make sure you enable the Google timezone API on your already used API-Key.

## Motivation and Context

With the aim of a better app emulation of requests, this provides at least a valid get_player request.

## How Has This Been Tested?

Local test setup with my own API-Key
This might need testing in not-obvious locations, since we do not know if any possible combination is regarded as valid from PoGO.

```
{'country': 'US', 'language': 'en', 'timezone': 'America/Denver'}
```
```
{'country': u'DE', 'language': 'de', 'timezone': 'Europe/Berlin'}
```
## Types of changes

- [x] Improvement

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -307,7 +307,7 @@ def main():
         args.player_locale = gmaps_reverse_geolocate(
             args.gmaps_key,
             args.locale,
-            position[1] + ', ' + position[2])
+            position[0] + ', ' + position[1])
 
         # Gather the Pokemon!
 
--- a/runserver.py
+++ b/runserver.py
@@ -307,7 +307,7 @@ def main():
         args.player_locale = gmaps_reverse_geolocate(
             args.gmaps_key,
             args.locale,
-            position[0] + ', ' + position[1])
+            str(position[0]) + ', ' + str(position[1]))
 
         # Gather the Pokemon!
 
--- a/runserver.py
+++ b/runserver.py
@@ -15,11 +15,10 @@ from threading import Thread, Event
 from queue import Queue
 from flask_cors import CORS
 from flask_cache_bust import init_cache_busting
-from geopy.geocoders import GoogleV3
 
 from pogom import config
 from pogom.app import Pogom
-from pogom.utils import get_args, now, extract_sprites
+from pogom.utils import get_args, now, extract_sprites, gmaps_reverse_geolocate
 from pogom.altitude import get_gmaps_altitude
 
 from pogom.search import search_overseer_thread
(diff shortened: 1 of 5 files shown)


runserver.py
```diff
@@ -303,6 +304,31 @@ def main():
...
+                    'timezone': str(timezone)
+                })
+            except Exception as e:
+                log.warning('Exception while obtaining Google Timezone. ' +
+                            'Key probably not enabled: %s.', repr(e))
+                pass
+        except Exception as e:
+            log.warning('Exception while obtaining player locale: %s.',
+                        repr(e))
+            pass
```
sebastienvercammen: `pass` does nothing. You probably meant to use `return`.

runserver.py
```diff
@@ -303,6 +304,31 @@ def main():
...
+            country = location[-1].raw['address_components'][-1]['short_name']
+            try:
+                timezone = geolocator.timezone(args.location)
+                args.player_locale.update({
+                    'country': country,
+                    'timezone': str(timezone)
+                })
+            except Exception as e:
+                log.warning('Exception while obtaining Google Timezone. ' +
+                            'Key probably not enabled: %s.', repr(e))
```
sebastienvercammen: Catch and handle the specific exception for when Timezone API isn't enabled, and show a separate message.

All exceptions should also be logged with `log.exception` rather than `log.warning`.

runserver.py
```diff
@@ -303,6 +304,31 @@ def main():
...
+        # Find the reverse geolocation
+        geolocator = GoogleV3(api_key=args.gmaps_key)
+        args.player_locale = {
+            'country': 'US',
+            'language': args.locale,
+            'timezone': 'America/Denver'
+        }
+        try:
+            location = geolocator.reverse(args.location)
+            country = location[-1].raw['address_components'][-1]['short_name']
```
sebastienvercammen: What's the return value of `geolocator.reverse()` when it fails or gets invalid input? We should handle it properly.

SiteEffect:
Help wanted

Issue #836, opened 2016-08-19
Done two little extensions ... You are interest in?


elvquant:
1. Additional to webhook I have implement an push-notification based on prowl (iOS) and NotifyMyAndroid. Defining type and apikey, I can define rarity and/or Pokemon number, which should make an push notification. The notification includes notification type (Rarity/Pokemon), Name (language based), hide time and date, distance (metric) and direction (wind rose and degree). It is made very compact so it perfect displayed on my pebble. If western longitude and southern latitude are negative, the distance/direction should by calculated right. Only at -180-/180-longitude a problem exists.
1a. Added pokemon name and rarity to json data for the webhook call ...

2. Additional search definition (I called it step-orientation). On horizontal type it made additional quarter circle left and right to the hexagon. On vertiacal on the upper and lower side.

I made it in the version from 2016-08-12, I think. You are interest in theses features and code review, so I can a context-diff.

angrymoose:
It sounds interesting, do you have any screenshots ?

ghvanderweg:
Why not open a pull request?

elvquant:
@ghvanderweg I want to do it, but I'm failed. Possible, I'm mobile ... Back home in about 6 hours.

@angrymoose 
[https://picload.org/image/rridwolo/image.jpg](url)
[https://picload.org/image/rridwoop/image.jpg](url)
[https://picload.org/image/rridwolg/image.jpg](url)
[https://picload.org/image/rridwold/image.jpg](url)

roflinger:
pokealarm uses a nice config file - why don´t u also try this instead of a hell of parameter

justsomedudeonthenet:
I think it would be better to leave notifications to a separate webhook program.

If prowl gets added, why not Slack, email, irc, discord, msn messenger, etc.  Leaving that all separate makes more sense to me.  Plus I can change alert settings without restarting the scanner.

elvquant:
@justsomedudeonthenet you are right, but I was much to lazy to made a separate webhook service. Additional there are not all information given in json-data to the webhook. Name and Rarity are easy to add, because the information are available directly at create the data structure. Not as easy is it to get a reference location (you position, marker sign) to calculate distance and direction ...

I only ask for interest ... ;-)

elvquant:
I think about to make an webhook 

@roflinger all Parameters can also put into config.ini ;-)

PR #2059, opened 2017-05-23
Check if account is shadowbanned on the run

Thetacz:
This PR aims to identify if accounts are shadowbanned / get shadowban on the run.

## Description

One can identify shadowbanned accounts by missing pokemons in their scan. So if current account sees only "common" pokemons and no "rares", it pulls nearby active pokemons from database and compares them to what the current account sees. If there are any rare pokemons around and the current account can't see them, it sends a warning with pokemon IDs saying, the account is shadowbanned. It also writes the account to "acc_shadowbanned.csv" (only if not there yet) and sends it to recycler for -ari + 24h.
It also counts consecutive non-rare scans and if the number exceeds 20, it sends the account to recycle for -ari. It does not write to "acc_shadowbanned.csv" in this case, because this is not 100% reliable, as some rural areas can have only common pokemons. The number of consecutive counts is so high on purpose, because we have to give the previous method some time to work. This way fresh accounts get into rotation and we can use the 1st method.

## Motivation and Context

Who would not want to know, if their account gets blinded? And sort them out...

## How Has This Been Tested?

I was running -st 20 with blinded accounts few hours to identify common pokemons. Than I took few of these accounts and let them run the same area with not yet blinded accounts. Here is a log of found blinded account:
image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -46,6 +46,12 @@ cache = TTLCache(maxsize=100, ttl=60 * 5)
 
 db_schema_version = 20
 
+class Shadowbanned(Exception):
+    def __init__(self, account, missed_ids=[], final=False):
+        self.account = account
+        self.missed_ids = missed_ids
+        self.final = final
+
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
     pass
@@ -328,6 +334,46 @@ class Pokemon(BaseModel):
                  )
 
         return list(itertools.chain(*query))
+
+    @classmethod
+    def get_pokemons_nearby(cls, center, timestamp):
+
+        # Maximum distance, where we see pokemons nearby
+        visible_distance = 200
+
+        # Get box of visible distance
+        start = geopy.distance.distance(meters=visible_distance)
+        sw = start.destination(center, 225).format_decimal()
+        sw = [float(s) for s in sw.split(',')]
+        ne = start.destination(center, 45).format_decimal()
+        ne = [float(s) for s in ne.split(',')]
+
+        swLat = sw[0]
+        swLng = sw[1]
+        neLat = ne[0]
+        neLng = ne[1]
+
+        # Get active pokemons on those borders
+        query = (Pokemon
+                 .select(Pokemon.latitude.alias('lat'),
+                         Pokemon.longitude.alias('lng'),
+                         Pokemon.pokemon_id)
+                 .where((Pokemon.disappear_time >= timestamp) &
+                        ((Pokemon.latitude >= swLat) &
+                         (Pokemon.longitude >= swLng) &
+                         (Pokemon.latitude <= neLat) &
+                         (Pokemon.longitude <= neLng)))
+                 .dicts())
+        p = list(query)
+
+        # Remove pokemons outside visible circle
+        pokemons = []
+        for idx, sp in enumerate(p):
+            if geopy.distance.distance(
+                    center, (sp['lat'], sp['lng'])).meters <= visible_distance:
+                pokemons.append(p[idx]['pokemon_id'])
+
+        return pokemons
 
 
 class Pokestop(BaseModel):
@@ -1852,14 +1898,22 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     stopsskipped = 0
     forts = []
     forts_count = 0
+    encountered_pokemon = []
+    encountered_pokemon_ids = []
     wild_pokemon = []
     wild_pokemon_count = 0
-    nearby_pokemon = 0
+    nearby_pokemon = []
+    nearby_pokemon_ids = []
+    nearby_pokemon_count = 0
     spawn_points = {}
     scan_spawn_points = {}
     sightings = {}
     new_spawn_points = []
     sp_id_list = []
+    missed = []
+    common_ids = [16, 19, 23, 27, 29, 32, 41, 43, 46, 52, 54, 60, 69,
+                  72, 74, 77, 81, 98, 118, 120, 129, 161, 165, 167,
+                  177, 183, 187, 191, 194, 198, 209, 218]
 
     # Consolidate the individual lists in each cell into two lists of Pokemon
     # and a list of forts.
@@ -1875,13 +1929,14 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
             now_date = datetime.utcfromtimestamp(
                 cell.current_timestamp_ms / 1000)
 
-        nearby_pokemon += len(cell.nearby_pokemons)
+        nearby_pokemon_count += len(cell.nearby_pokemons)
         # Parse everything for stats (counts).  Future enhancement -- we don't
         # necessarily need to know *how many* forts/wild/nearby were found but
         # we'd like to know whether or not *any* were found to help determine
         # if a scan was actually bad.
         if config['parse_pokemon']:
             wild_pokemon += cell.wild_pokemons
+            nearby_pokemon += cell.nearby_pokemons
 
         if config['parse_pokestops'] or config['parse_gyms']:
             forts += cell.forts
@@ -1894,7 +1949,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     del map_dict['responses']['GET_MAP_OBJECTS']
 
     # If there are no wild or nearby Pokemon . . .
-    if not wild_pokemon and not nearby_pokemon:
+    if not wild_pokemon and not nearby_pokemon_count:
         # . . . and there are no gyms/pokestops then it's unusable/bad.
         if not forts:
             log.warning('Bad scan. Parsing found absolutely nothing.')
@@ -1909,6 +1964,38 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     done_already = scan_loc['done']
     ScannedLocation.update_band(scan_loc, now_date)
     just_completed = not done_already and scan_loc['done']
+
+    # Checking if account is blinded
+    if (nearby_pokemon or wild_pokemon) and config['parse_pokemon']:
+
+        # Create list of present pokemon IDs for blinded check
+        for p in nearby_pokemon:
+            nearby_pokemon_ids.append(p.pokemon_id)
+        for p in wild_pokemon:
+            nearby_pokemon_ids.append(p.pokemon_data.pokemon_id)
+
+            # Remove common pokemons from seen
+        rare_finds = [p for p in nearby_pokemon_ids if p not in common_ids]
+
+        # Checking if found only common pokemons
+        if len(rare_finds) == 0:
+            # Get nearby active pokemons from database
+            encountered_pokemon_ids = Pokemon.get_pokemons_nearby(
+                step_location, now_date)
+
+            for p in encountered_pokemon_ids:
+                if ((p not in nearby_pokemon_ids) and (p not in common_ids)):
+                    missed.append(p)
+
+            if missed:
+                raise Shadowbanned(account, missed, True)
+
+            status['nonrares'] += 1
+            if status['nonrares'] >= 20:
+                raise Shadowbanned(account)
+                # reset counter if rares are found
+        else:
+            status['nonrares'] = 0
 
     if wild_pokemon and config['parse_pokemon']:
         encounter_ids = [b64encode(str(p.encounter_id))
@@ -2258,7 +2345,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
              'pokestops: %d, gyms: %d, raids: %d.',
              len(pokemon) + skipped,
              filtered,
-             nearby_pokemon,
+             nearby_pokemon_count,
              len(pokestops) + stopsskipped,
              len(gyms),
              len(raids))
@@ -2289,7 +2376,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 log.info('Possible causes: Still doing initial scan, super'
                          ' rare double spawnpoint during')
                 log.info('hidden period, or Niantic has removed '
-                         'spawnpoint.')
+                         'spawnpoint or shadowbanned account.')
 
         if (not SpawnPoint.tth_found(sp) and scan_loc['done'] and
                 (now_secs - sp['latest_seen'] -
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -1777,15 +1817,40 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
     scan_spawn_points = {}
     sightings = {}
     new_spawn_points = []
     sp_id_list = []
     captcha_url = ''
+    missed = []
+    # there must be a better way...
+    # Common: 16, 19, 27, 29, 32, 41, 43, 46, 52, 54, 60, 69, 72, 74, 98, 118,
+    #         120, 129, 161, 165, 167, 177, 183, 187, 194, 198, 209, 218
+    rare_ids = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 17, 18, 20,
```
blueberry0101: Yes, there is a better way.

```
common = [16, 19, 27, 29, 32, 41, 43, 46, 52, 54, 60, 69, 72, 74, 98, 118,
          120, 129, 161, 165, 167, 177, 183, 187, 194, 198, 209, 218]

rare_ids = [x for x in range(0, 250) if x not in common]
```

Thetacz: I meant better way finding common pokemons than using this kind of list of IDs.Tthis is mostly cosmetic change, but looks better, you are right :)

Arengo: ``range(1,252)``

Thetacz: @Arengo the rare_ids list is constructed from range, but IDs of common pokemons are substracted from the list. So the range is not complete, that's the whole point :D

jakubkrysl:
I just added the rest of confirmed common pokemons, [source reddit](https://www.reddit.com/r/pokemongodev/comments/6cng1n/current_guess_at_shadowbanned_pokemon/).
I am looking into possiblity of confirming shadowbanned accounts by checking the store for items, it seems shadowbanned accounts can't see any items buyable by coins, they only see real money exchanges. To clarify: blindness is a symptom of shadowban.
So far I traced the request to GET_STORE_ITEMS protobuf object, but I have no idea how to access it as it is in platform of pgoapi and not with the rest of protobufs. Could definitely use some help there... 

EDIT: Checking store for missing items does not seem to be a reliable way to check for shadowban. The account has to be flagged too for this to happen, but flag != shadowban. So no point in doing another check this way, if it can't check for shadowban alone...

sebastienvercammen:
@Thetacz 

2: Preferably not. The idea of using the shop to identify blind accounts (because they could only see coins) was a good one, but it's apparently not true for everyone. As an alternative, I suggest looking at scan data: it seems blind accounts don't get all details the regular accounts are getting, as the `'enabled'` field was missing and triggered a `KeyError`. `'enabled'` is a field that's usually **always** received.

But we've also started working on a PGoClient wrapper around PGoAPI, which could include a status such as banned/flagged/blind, and which could automatically update the account status after every request. @Arengo 

3: Look at the flow for banned accounts. Iirc, we remove them from queue, output to logs, and never retry that account.

jakubkrysl:
Now it properly recycles the accounts (in case the shadowbans get lifted after some time - stil not confirmed). I can either leave it as it is and those accounts will get re-tried after user specified time, set some fixed time for shadowbanned (needs more reaserch, being done at #pogodev) or completely discard the account (or leave it there and add if that will skip shadowbanned accounts)

image

jakubkrysl:
@sebastienvercammen Just compared response from shadowbanned and not. Those responses only differ in visible pokemons (aside from account specific - types of pokemon it seen, stats etc...). They both have the same amount of `'enabled'` fields (same location). `'enabled'` field only occurs in forts, could not see it anywhere else. All `'enabled'` fields were set to true in both responses.

EDIT: missclick on close, sry :/

pogom/search.py
```diff
@@ -1011,6 +1011,20 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                         step_location[0], step_location[1],
                         parsed['count'])
                     log.debug(status['message'])
+
+                except Shadowbanned as e:
+                    status['message'] = (
+                        'Account {} could not find pokemon ID: {}. ' +
+                        'Possibly shadowbanned, switching accounts... '
+                        ).format(e.account['username'], e.missed_ids)
```
Alderon86: Please use the logger (%s) instead of `string with {} some stuff {} in it.format(foo,bar)`

Thetacz: @Alderon86 I used %s before, but changed it for consistency with line #836. If you insist so much, I will change it in a heatbeat. :)

Thetacz: Or should I split the mesage to more lines?

pogom/search.py
```diff
@@ -1014,9 +1014,9 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
 
                 except Shadowbanned as e:
                     status['message'] = (
-                        'Account {} could not find pokemon ID: {}. ' +
+                        'Account %s could not find pokemon ID: %s. ' +
                         'Possibly shadowbanned, switching accounts... '
-                        ).format(e.account['username'], e.missed_ids)
+                        ) % (e.account['username'], e.missed_ids)
```
Alderon86: status['message'] = (
                        'Account %s could not find pokemon ID: %s. ' +
                        'Possibly shadowbanned, switching accounts... ' ,e.account['username'], e.missed_ids)

pogom/search.py
```diff
@@ -1011,6 +1011,20 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                         step_location[0], step_location[1],
                         parsed['count'])
                     log.debug(status['message'])
+
+                except Shadowbanned as e:
+                    status['message'] = (
+                        'Account %s could not find pokemon ID: %s. ' +
+                        'Possibly shadowbanned, switching accounts... '
+                        ) % (e.account['username'], e.missed_ids)
```
Alderon86: status['message'] = (
                        'Account %s could not find pokemon ID: %s. ' +
                        'Possibly shadowbanned, switching accounts... ', e.account['username'], e.missed_ids)

Thetacz: https://stackoverflow.com/a/11589591
But you are right, looks nicer.

Alderon86:
`1.list of rare pokemons constains some common ones that I have not encountered yet in my area. For example: Have not seen bulba in ages, any nest to test if he is common?
2.Do we want to identify blinded accounts based on their inability to find rare pokemons? (Warn after X consecutive rareless finds...)
3.What do we want to do with identified blinded accounts?`

1.Bulba and any other of the starter pokemons are not common.
2.I like the idea of giving a warning after 5 times fails to scan a rare, after 10 fails mark it as shadowbanned
3.remove from rotation and give a log/status warning that the account is blind/shadowbanned and is removed from rotation.

Ofc just my two cents

Borges-Wellington:
Does anyone know how to implement mapbox leaflet in place of gmap? For Internet browsing and others. I have other instances that do a scan and wanted an interface booklet

Buncertainc:
It seems that No.77 Ponyta should also be included in the common_ids.

jakubkrysl:
@Buncertainc Do you have some source for that? Or any location of Ponyta nest, so I can test it?

Buncertainc:
@Thetacz No, Sorry.  It's only from my accounts' experience. I have several blinded accounts which are able to find Ponyta.

tomballgithub:
@Thetacz Ponyta nest @ 30.168453, -95.7407196

sebastienvercammen:
@wellington1092 This is not the place for that.

@Thetacz Awaiting changes, or ready for review?

jakubkrysl:
@Buncertainc Thanks for the tip, it seems to be true.
@tomballgithub Thanks, using that nest I confirmed, ponyta is among common pokemons. Adding it to the list.
@sebastienvercammen Just added ponyta to common list, so it is ready for review. Sorry for the delay, too much new stuff going on IRL :)

pki0:
What about a new column "blind" in workerstatus? If an account hits the shadowban set blind to '1', else '0'.
This would make scripts for  automatic  restart/account changing more easy and comfortable. 
Just a suggestion .

Edit: Or a column named "Status" with OK/Blind/Banned as options

jakubkrysl:
@pokemonkiel That seems like out of scope of this PR as that should be solved by proper account manager. Generally I like the idea, but accounts get unshadowbanned, so there would have to be a way to check and reset the flag in DB.  And that is definitely out of scope and work for proper account manager...Now we just send accounts to rest.
I implemented saving to file, so there is room for custom management. Without the need for future massive code change on RM side.

Kimuji:
I'm currently trying your PR.
From what i saw it seems that event related pokemons are still displayed for shadowbanned accounts.
So it means during events this PR isn't detecting those blind accounts.

I understand the logic behind your PR, but why is the warning "spawnpoint xxx has no Pokemon N times in a row." not used in a way or an other in the detection ?

SiteEffect:
An option would be to make the list a flag with current list as default and let map maintainers adjust them during an event.

SiteEffect:
This is the current list common_list for the live event:

```py
    common_ids = [16, 19, 23, 27, 29, 32, 37, 41, 43, 46, 52, 54, 58, 60,
                  69, 72, 74, 77, 81, 90, 98, 118, 120, 129, 155, 161, 165,
                  167, 177, 183, 187, 191, 194, 198, 209, 218, 220, 228]
```

If you find anything else being common, let us know. I am not sure about Lapras # 131.

friscoMad:
I am not sure about this system I do see what **-redacted-** does a lot simpler just check wild + nearby for rare pokemons, nearby list is large and is uncommon to don't have any rare pokemon in the list.
This system seem more exact except when the whole map start blinding at the same time, I would see this being better if we do pause blind accounts asap

SiteEffect:
Re-check list and rebase, please :-) 
Looks like the shadowbans are back.

jakubkrysl:
@Arengo Will do when I get to PC, in the middle of woods now till monday. Did the list change after event, remained event-like or returned back to pre-event?

AEtHeLsYn:
As @friscoMad said, **-redacted-** approach seems way simpler and works pretty well, it's uncommon to don't find a rare Pokemon in wild+nearby list

sebastienvercammen:
@friscoMad @AEtHeLsYn I've redacted your comments not to include the name of a project related to botting that isn't supported by RocketMap.

However, the point of using wild+nearby is one that should be discussed, but the system we decide to implement needs 100% success rate on determining whether an account is shadowbanned or not. We can't have "works 50/50", and "pretty uncommon" != "always".

tomballgithub:
Also need a plan to handle events which I believe has previously changed what constitutes 'common' per comments above

friscoMad:
There is not way of being 100% sure, even moving an account to check the account not seeing rares is not 100% sure, this PR tries to use older pokes already scanned that is a bit safer but still not 100% and increase db usage.
I think that just counting scans without rares (rares being user configurable to handle events) and let the user discard accounts after X scans without rares as we are doing with empty scans. The bad part is not telling the user what accounts are shadowbanned. but we are not doing that for banned or whatever right now so I don't see why that would be a problem.

dongemus:
Wouldn't it be 100% sure if the current encounters system was duplicated...
I.E. A user makes a "test-accounts: testing.csv" like currently is done with high-lvl-accounts: and when a rare is seen as specified in "rare-whitelist-file: rares-whitelist.txt" one or more test accounts goes to that spawn to see if they can see it.

Just an idea, wish my coding was good enough to help implement

friscoMad:
Except that the test accounts can be shadowbanned too and there is no way
to know as there is no flag for it.

2017-07-09 17:28 GMT+02:00 dongemus <notifications@github.com>:

dongemus:
But the test accounts would only be triggered if a rare was seen from
~rares-whitelist.txt
so the test accounts can only be tested for shadowban when a not
shadowbanned account from accounts.csv sees a rare. You are right it won't
work if all the accounts in accounts.csv are shadowbanned, but if any of
them are not it can process the accounts in ~testing.csv

Just a thought, I shouldn't bother everyone too much as I can't fix it
myself. Just thought that whoever came up with the brilliant encounters
system could apply the same technique to shadowban testing

sebastienvercammen:
@dongemus @friscoMad Don't worry too much, thanks about the feedback.

There **is** a way that works 100% of the time to detect shadowbans, but it'll take more time to implement. Until then, something that works the largest majority of the time will be _good enough_.

boby361:
Needs rebase. Models.py conflict. Appreciate the "Fix the rebase" however there is still a conflict.

SiteEffect:
There is a conflict with ``models.py`` left. You need to rebase to the new API objects instead of dicts.

jakubkrysl:
Sadly I don't have any free time now. Soonest date I can get to this is 4 days from now, but more probable is next week. Sorry guys for keeping you waiting with broken code.

SiteEffect:
No hurry 👍 enjoy!

jakubkrysl:
Missclick on wrong branch, reopening and working again :)

tobias86aa:
Nice :)

boby361:
So happy to see this. Have been so patient.

boby361:
So i have a weird bug. I restarted my server and i got this error:
  File "runserver.py", line 20, in <module>
    from pogom.app import Pogom
  File "C:\Servers\RocketMap-Boby361\pogom\app.py", line 19, in <module>
    from .models import (Pokemon, Gym, Pokestop, ScannedLocation,
  File "C:\Servers\RocketMap-Boby361\pogom\models.py", line 35, in <module>
    from .customLog import printPokemon
  File "C:\Servers\RocketMap-Boby361\pogom\customLog.py", line 5, in <module>
    args = get_args()
  File "C:\Servers\RocketMap-Boby361\pogom\utils.py", line 41, in wrapper
    rv = function(*args)
  File "C:\Servers\RocketMap-Boby361\pogom\utils.py", line 679, in get_args
    args.username.pop(i)
IndexError: pop index out of range

If i rename my shadow ban file to somethhing else. it works. The file hasn't been touched in awhile. The formatting seems all proper (UTF-8). there is nothing missing. no extra space. nothing. it fixes itself if i remove accounts out. It happened with 17 accounts in it. Thoughts? **Edit** Its happened with less accounts now.

jakubkrysl:
@boby361 This is still PR so things change from time to time and some are not backwards compatible. I added 4th field to the file, timestamp. Your old file has probably only 3 fields. So either check the code and add proper timestamp or delete it and let it create a new one.
Also check if you have the same account more than once in the file. Could not check the possibility, but would be very interested if it happens.

boby361:
Yeah the file only contains lines with the 4 fields. Everything seems to work properly. It's just weird how it randomly throws this issue. I have other people having the same issue with the same fix (using the same fork. Although it's just a up to date RM develop branch + this PR) it happened to me on windows server 2016 although one of my scanning contributors had same issue on Ubuntu.

jorgitoh:
Hi. Any progress with this PR?. Now we have Temp Ban and shadow bans again. I think is time to put an eye in this PR.

sebastienvercammen:
@jorgitoh Then go ahead and finish it yourself, it's open source. Otherwise stay silent if you have nothing to contribute.

PR #2063, opened 2017-05-24
[Debug] Add "Missed" to worker status stats.

neskk:
This PR is for debug purposes and it updates database schema version - just one tiny modification to `workerstatus` table - if you want to reverse the update, drop it and revert schema version to 19 and you're good.

It adds "Missed" column to status page and status printer (console) to track missed spawnpoints. The idea came from @Arengo and I decided to do a branch isolating this change.

## Description

Adds a new statistic to WorkerStatus: `missed` - that tracks whenever a spawnpoint is missed by the worker when map request is parsed. This can happen because account is "shadowed/blind" or if spawnpoint data is old and outdated.

## Motivation and Context

Get more information on "shadowed/blind" accounts and how to identify them.

## How Has This Been Tested?

Tested in local environment and server with several instances.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -42,7 +42,7 @@ args = get_args()
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 19
+db_schema_version = 20
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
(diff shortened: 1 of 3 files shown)


SiteEffect:
Good one :-)

bbdoc:
What about adding a message to log when a LVL 30 account is sent for an encounter but doesn't see the pokemon it should encounter ? That would mean a low level account could see it but your lvl 30 couldn't see it, so it's shadowbanned... Would make it very easy to identify shadowbans at least for high level accounts. Even better with some increment "L30 account xxx could not encounter a pokemon for xx times in a row".

Would be even better to add this info in workerstatus table but I don't think Lvl 30 accounts are in there...

Commenting on this PR as the objective is the same, help identify "blind" accounts.

PR #2072, opened 2017-05-30
device_info

rocketrobot99:
New table to store devices associated with an account

## Description
Added a new table to store the generated device info, rather than regenerating it every login.

## Motivation and Context
I am hoping this may help a bit with shadow bans. I don't think normal gamers switch phones everytime they log in.

## How Has This Been Tested?
I have tested the code to ensure it functions and stores device details along with the account.
As for results, too early to really tell if it helps with shadow bans.

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -810,7 +810,8 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
             # one each time we login
             device_info = AccountDevice.get_device_info(account['username'])
             api, device_info = setup_api(args, status, device_info)
-            dbq.put((AccountDevice, {0: AccountDevice.db_format(device_info,account['username'])}))
+            dbq.put((AccountDevice, {0: AccountDevice.db_format(
+                                        device_info, account['username'])}))
 
             # The forever loop for the searches.
             while True:
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -44,7 +44,7 @@ from pgoapi import utilities as util
 from pgoapi.hash_server import (HashServer, BadHashRequestException,
                                 HashingOfflineException)
 from .models import (parse_map, GymDetails, parse_gyms, MainWorker,
-                     WorkerStatus, HashKeys)
+                     WorkerStatus, HashKeys, AccountDevice)
 from .utils import now, clear_dict_response
 from .transform import get_new_coords, jitter_location
 from .account import (setup_api, check_login, get_tutorial_state,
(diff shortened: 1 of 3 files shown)


pogom/search.py
```diff
@@ -805,7 +805,13 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
...
             consecutive_noitems = 0
 
-            api = setup_api(args, status)
+            # So the idea is to store the device that gets generated. This way
+            # we can keep using the same device rather than using a random
+            # one each time we login
+            device_info = AccountDevice.get_device_info(account['username'])
+            api, device_info = setup_api(args, status, device_info)
+            dbq.put((AccountDevice, {0: AccountDevice.db_format(
+                                        device_info, account['username'])}))
```
Thetacz: You only need to put device info to DB once. If it is there, there is no point in replacing the info with the same data, it only places useless item in DB queue, making it unnecessarily longer. So maybe check if you have device info and put it to queue only if not.

rocketrobot99: Good point, my whole reason to put it in a separate table is that it shouldn't be updated. New commit release, thanks.

friscoMad:
Sorry if this is a bit destructive but just minutes after seeing this PR I thought that a better solution is to have deterministic devices based on the account name, add a configured salt if we want to be paranoid and avoid Niantic checking for bots using the deterministic algo.

You can hash account+salt then take the lower-higher bytes and mod X where X is the number of options in every step.

That avoids needing to store anything in DB, will keep the device for each account even after cleaning the DB as long as you keep the salt (it would change only if new devices/os are added to the pool).

Thats said I don't think this is a bad solution, just that I am a bit unconfortable adding the accounts to the DB when we are far from having an account manager and while thinking in other options reached to this idea.

If implemented salt should be a generic salt as we may find other uses for it later.

343334:
Well I think in due time we are going to need our accts in the db anyways so we can keep track of how long each acct was logged in for , and when was the last time they logged out so we can make it a bit more random than having every acct always scan for 2 hrs etc and always logged out for same durations.

rocketrobot99:
@friscoMad Hi, I can see what you are saying. I did wonder if there was a way to avoid storing the details in the database. However, there may be some advantages:
1. The user doesn't have to generate a salt, or understand what one is or why they have to keep it.
2. Accounts are often acquired (lv 5 or lv30) or transferred from various other programs. It would be easier to enter these given details it the table (app not included) and near enough impossible to produce the same device info if it was originally produced randomly or entered manually from a different program.
3. We could also consider adding a function in a year's time to upgrade the device_info to a newer iphone model or upgrade the IOS every 6 months. 

I know (2) is not really a concern for most users. However, it may start to "become a thing" when you purchase accounts in the future, that you are also given the device info that was used to level it.

rocketrobot99:
@343334 Indeed. We need a design for all the account related entities. Account device is not a bad start through.

michikrug:
You missed to update the api setup for the high level accounts in models.py.

sLoPPydrive:
By the way... the device ID (UDID) has a bad format (always had). It's a UUID where it should be a SHA1 hash. Just saying...

sebastienvercammen:
PR closed in favor of #2074, which is a more elegant way of doing this. The only step up is to implement better account handling.

PR #2073, opened 2017-06-02
Introducing Pokestop spinning and Egg hatching.

Alderon86:
## Description

This PR will set an 50% chance to spin a Pokestop to act more human-like.
Items are trashed also.
Eggs will be incubated in order of 2>5>10.
Pokemons are released automatically + some other minor improvements.
## Motivation and Context

RocketMap needs to act more like a real Player!

## How Has This Been Tested?

On my own setup.
## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -134,7 +134,7 @@ def rpc_login_sequence(args, api, account):
 
     try:
         req = api.create_request()
-        req.call()
+        req.call(False)
 
         total_req += 1
         time.sleep(random.uniform(.43, .97))
@@ -153,7 +153,7 @@ def rpc_login_sequence(args, api, account):
     try:
         req = api.create_request()
         req.get_player(player_locale=args.player_locale)
-        response = req.call()
+        response = req.call(False)
 
         parse_get_player(account, response)
 
@@ -184,7 +184,7 @@ def rpc_login_sequence(args, api, account):
         req.get_inventory(last_timestamp_ms=0)
         req.check_awarded_badges()
         req.download_settings()
-        response = req.call()
+        response = req.call(False)
 
         parse_download_settings(account, response)
         parse_new_timestamp_ms(account, response)
@@ -225,7 +225,7 @@ def rpc_login_sequence(args, api, account):
             req.check_awarded_badges()
             req.download_settings(hash=account[
                 'remote_config']['hash'])
-            response = req.call()
+            response = req.call(False)
 
             parse_new_timestamp_ms(account, response)
 
@@ -274,7 +274,7 @@ def rpc_login_sequence(args, api, account):
             req.check_awarded_badges()
             req.download_settings(
                 hash=account['remote_config']['hash'])
-            response = req.call()
+            response = req.call(False)
 
             parse_new_timestamp_ms(account, response)
 
@@ -319,7 +319,7 @@ def rpc_login_sequence(args, api, account):
         req.check_awarded_badges()
         req.download_settings(hash=account['remote_config']['hash'])
         req.get_buddy_walked()
-        response = req.call()
+        response = req.call(False)
 
         parse_new_timestamp_ms(account, response)
 
@@ -338,7 +338,7 @@ def rpc_login_sequence(args, api, account):
     try:  # 7 - Make an empty request to retrieve store items.
         req = api.create_request()
         req.get_store_items()
-        response = req.call()
+        response = req.call(False)
 
         total_req += 1
         time.sleep(random.uniform(.6, 1.1))
@@ -361,7 +361,7 @@ def rpc_login_sequence(args, api, account):
         req.download_settings(hash=account['remote_config']['hash'])
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
 
         parse_new_timestamp_ms(account, response)
 
@@ -392,7 +392,7 @@ def complete_tutorial(args, api, account):
         req = api.create_request()
         req.mark_tutorial_complete(tutorials_completed=0)
         log.debug('Sending 0 tutorials_completed for %s.', account['username'])
-        req.call()
+        req.call(False)
 
     if 1 not in tutorial_state:
         time.sleep(random.uniform(5, 12))
@@ -408,20 +408,20 @@ def complete_tutorial(args, api, account):
         })
         log.debug('Sending set random player character request for %s.',
                   account['username'])
-        req.call()
+        req.call(False)
 
         time.sleep(random.uniform(0.3, 0.5))
 
         req = api.create_request()
         req.mark_tutorial_complete(tutorials_completed=1)
         log.debug('Sending 1 tutorials_completed for %s.', account['username'])
-        req.call()
+        req.call(False)
 
     time.sleep(random.uniform(0.5, 0.6))
     req = api.create_request()
     req.get_player_profile()
     log.debug('Fetching player profile for %s...', account['username'])
-    req.call()
+    req.call(False)
 
     starter_id = None
     if 3 not in tutorial_state:
@@ -432,24 +432,24 @@ def complete_tutorial(args, api, account):
             'aa8f7687-a022-4773-b900-3a8c170e9aea/1477084794890000',
             'e89109b0-9a54-40fe-8431-12f7826c8194/1477084802881000'])
         log.debug('Grabbing some game assets.')
-        req.call()
+        req.call(False)
 
         time.sleep(random.uniform(1, 1.6))
         req = api.create_request()
-        req.call()
+        req.call(False)
 
         time.sleep(random.uniform(6, 13))
         req = api.create_request()
         starter = random.choice((1, 4, 7))
         req.encounter_tutorial_complete(pokemon_id=starter)
         log.debug('Catching the starter for %s.', account['username'])
-        req.call()
+        req.call(False)
 
         time.sleep(random.uniform(0.5, 0.6))
         req = api.create_request()
         req.get_player(
             player_locale=args.player_locale)
-        responses = req.call().get('responses', {})
+        responses = req.call(False).get('responses', {})
 
         if 'GET_INVENTORY' in responses:
             for item in (responses['GET_INVENTORY'].inventory_delta
@@ -463,33 +463,33 @@ def complete_tutorial(args, api, account):
         req = api.create_request()
         req.claim_codename(codename=account['username'])
         log.debug('Claiming codename for %s.', account['username'])
-        req.call()
+        req.call(False)
 
         time.sleep(random.uniform(1, 1.3))
         req = api.create_request()
         req.mark_tutorial_complete(tutorials_completed=4)
         log.debug('Sending 4 tutorials_completed for %s.', account['username'])
-        req.call()
+        req.call(False)
 
         time.sleep(0.1)
         req = api.create_request()
         req.get_player(
             player_locale=args.player_locale)
-        req.call()
+        req.call(False)
 
     if 7 not in tutorial_state:
         time.sleep(random.uniform(4, 10))
         req = api.create_request()
         req.mark_tutorial_complete(tutorials_completed=7)
         log.debug('Sending 7 tutorials_completed for %s.', account['username'])
-        req.call()
+        req.call(False)
 
     if starter_id:
         time.sleep(random.uniform(3, 5))
         req = api.create_request()
         req.set_buddy_pokemon(pokemon_id=starter_id)
         log.debug('Setting buddy pokemon for %s.', account['username'])
-        req.call()
+        req.call(False)
         time.sleep(random.uniform(0.8, 1.8))
 
     # Sleeping before we start scanning to avoid Niantic throttling.
@@ -819,7 +819,7 @@ def spin_pokestop_request(api, account, fort, step_location):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
         parse_new_timestamp_ms(account, response)
         return response
 
@@ -841,7 +841,7 @@ def fort_details_request(api, account, fort):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
         parse_new_timestamp_ms(account, response)
         return response
     except Exception as e:
@@ -865,7 +865,7 @@ def encounter_pokemon_request(api, account, encounter_id, spawnpoint_id,
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
         parse_new_timestamp_ms(account, response)
         return response
     except Exception as e:
@@ -883,7 +883,7 @@ def clear_inventory_request(api, account, item_id, drop_count):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        clear_inventory_response = req.call()
+        clear_inventory_response = req.call(False)
 
         parse_new_timestamp_ms(account, clear_inventory_response)
         parse_inventory(api, account, clear_inventory_response)
@@ -908,7 +908,7 @@ def request_use_item_egg_incubator(api, account, incubator_id, egg_id):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
 
         parse_new_timestamp_ms(account, response)
 
@@ -930,7 +930,7 @@ def request_release_pokemon(api, account, pokemon_id, release_ids=[]):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        release_p_response = req.call()
+        release_p_response = req.call(False)
 
         parse_new_timestamp_ms(account, release_p_response)
 
@@ -952,7 +952,7 @@ def parse_level_up_rewards(api, account):
         req.check_awarded_badges()
         req.get_buddy_walked()
         req.get_inbox(is_history=True)
-        response = req.call()
+        response = req.call(False)
 
         parse_new_timestamp_ms(account, response)
 
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -200,7 +200,7 @@ def rpc_login_sequence(args, api, account):
     log.debug('Fetching asset digest...')
     config = account.get('remote_config', {})
 
-    if config.get('asset_time') > old_config.get('asset_time', 0):
+    if config.get('asset_time', 0) > old_config.get('asset_time', 0):
         i = random.randint(0, 3)
         req_count = 0
         result = 2
@@ -253,7 +253,7 @@ def rpc_login_sequence(args, api, account):
     # 5 - Get item templates.
     log.debug('Fetching item templates...')
 
-    if config.get('template_time') > old_config.get('template_time', 0):
+    if config.get('template_time', 0) > old_config.get('template_time', 0):
         i = random.randint(0, 3)
         req_count = 0
         result = 2
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -449,7 +449,7 @@ def complete_tutorial(args, api, account):
         req = api.create_request()
         req.get_player(
             player_locale=args.player_locale)
-        response = req.call(False).get('responses', {})
+        responses = req.call(False).get('responses', {})
 
         if 'GET_INVENTORY' in responses:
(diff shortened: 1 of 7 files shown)


GamingMaster2000:
Forgive me if I'm misreading the code but does clear_inventory literally get rid of everything after spinning a stop? A human player wouldn't get rid of 100% of items every time.

Alderon86:
@GamingMaster2000 right now it is yes, we could consider to keep maybe 5 of each to be more close to real players.

pogom/humanize.py
```diff
@@ -0,0 +1,147 @@
...
+                         spinning_radius)
+    now = time.time()
+    pause_needed = 'cooldown_complete_timestamp_ms' in fort and fort[
+        'cooldown_complete_timestamp_ms'] / 1000 > now
+    return in_range and not pause_needed
+
+
+# 50% Chance to spin a Pokestop.
+def spinning_try(api, fort, step_location, account):
+    time.sleep(random.uniform(0.8, 1.8))  # Do not let Niantic throttle.
```
friscoMad: Add the sleep only if we will spin, otherwise is just wasting time.

pogom/humanize.py
```diff
@@ -0,0 +1,147 @@
...
+    # Set 50% Chance to spin a Pokestop.
+    if random.randint(0, 100) < 50:
+        spin_response = spin_pokestop_request(api, fort, step_location)
+        time.sleep(random.uniform(2, 4))  # Do not let Niantic throttle.
```
friscoMad: The same as before, add the sleep after an Ok spin

friscoMad:
Improvements:

- Please, base this PR on top of PGOClient, there is no reason to continue adding more requests when that PR should be merged soon.
- Humanize does not seem a great module name, if they are stops related function stops is better, inventory part should be in account or inventory module.
- Do not spin every stop (with 50% chance) spin max a stop per map request
 - I would prefer spin a stop every X min (for example base probability 10% then increasing with time and at X min it will be 100%).
 - Clear inventory randomly instead of after every pokespin.

This changes could cause speed violations as schedulers does take into account step location and time since last search, we are adding some more sleeps and requests without changing the stored location so it could lead to trying to scan a step that is near enough since the last GMO but not since the last request (threads and context switching is a real PITA). But I am not sure how the speed check is done if it is only using GMOs then everything is fine.

Alderon86:
@friscoMad I think theres a replacement coming for PGoClient, if so i will rebase it afterwards. But i wont rebase it on something that is not confirmed and marked as [WIP].
Also i thought humanize would be nice, because the are coming more things to improve behaviour. So in the end we dont want to create new files for every action.
Also Seb requested to drop items recieved from Stops automatically. So for now im keeping 5 each to not have an empty inventory in the end again.

pogom/humanize.py
Alderon86: hm its the same code, rocketmap uses for -tut. But i can move the sleep.

weizh888:
Maybe clear inventory after it is full, or clear some specific items first, instead of dropping them all

Alderon86:
Clearing the inventory will stay as it is. 5 Items of each will be kept, rest is going into the trash can.

pogom/models.py
```diff
@@ -2162,7 +2159,8 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                      datetime(1970, 1, 1)).total_seconds())) for f in query]
 
         # Complete tutorial with a Pokestop spin
-        if args.complete_tutorial and not (len(captcha_url) > 1):
+        if (args.complete_tutorial and not (len(captcha_url) > 1) and not
```
Obihoernchen: Please remove `and not (len(captcha_url) > 1)` cause it's never set for the scan account. It's only set for the high level account.

pogom/humanize.py
```diff
@@ -0,0 +1,152 @@
...
+            item_id = item['item_id']
+            count = item['count']
+            # Keep 5 Items in Inventory
+            total_items = item.get('count', 0)
+            items_to_drop = total_items - 5
+            if item_id in ITEMS:
+                item_name = ITEMS[item_id]
+            if total_items > 5:
+                # Do not let Niantic throttle
+                time.sleep(random.uniform(0.8, 1.8))
```
Obihoernchen: I'm pretty sure this will cause throttle warnings. Use something >= 1.75s

pogom/humanize.py
```diff
@@ -0,0 +1,152 @@
...
+    pause_needed = 'cooldown_complete_timestamp_ms' in fort and fort[
+        'cooldown_complete_timestamp_ms'] / 1000 > now
+    return in_range and not pause_needed
+
+
+# 50% Chance to spin a Pokestop.
+def spinning_try(api, fort, step_location, account):
+    # Set 50% Chance to spin a Pokestop.
+    if random.randint(0, 100) < 50:
+        time.sleep(random.uniform(0.8, 1.8))  # Do not let Niantic throttle.
```
Obihoernchen: I'm pretty sure this will cause throttle warnings. Use something >= 1.75s

pogom/humanize.py
```diff
@@ -0,0 +1,152 @@
...
+        spin_response = spin_pokestop_request(api, fort, step_location)
+        time.sleep(random.uniform(2, 4))  # Do not let Niantic throttle.
```
Obihoernchen: Why sleep here again?

pogom/humanize.py
```diff
@@ -0,0 +1,152 @@
...
+def clear_inventory(api, account, map_dict):
+    inventory_items = get_inventory_items(map_dict)
+    clear_responses = []
+    for item in inventory_items:
+        if 'item_id' and 'count' in item:
+            item_id = item['item_id']
+            count = item['count']
+            # Keep 5 Items in Inventory
+            total_items = item.get('count', 0)
+            items_to_drop = total_items - 5
```
Obihoernchen: randomize this (5-10 for example)

pogom/humanize.py
```diff
@@ -0,0 +1,152 @@
...
+            return True
+        # Catch all other results.
+        elif spin_result is 2:
+            log.info('Pokestop %s was not in range to spin for account %s',
+                     fort['id'], account['username'])
+        elif spin_result is 3:
+            log.info('Failed to spin Pokestop %s. %s Has recently spun this' +
+                     'stop.', fort['id'], account['username'])
+        elif spin_result is 4:
+            log.info('Failed to spin Pokestop %s. %s Inventory is full.',
```
Obihoernchen: Run `clear_inventory()` here as well

pogom/models.py
```diff
@@ -2213,6 +2211,12 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         'active_fort_modifier': active_fort_modifier
                     }))
 
+                # Spin Pokestop with 50% chance.
+                if args.pokestop_spinning and pokestop_spinnable(
+                        f, step_location):
+                    spinning_try(api, f, step_location, account)
+                    clear_inventory(api, account, map_dict)
```
Obihoernchen: I wouldn't make another request every time.
1) it doesn't make sense to run clear_inventory() if no pokestop was spun
2) it's too many requests imo. I would clear inventory every ~20 requests or something like this

Kimuji:
wouldn't you consider include what pr #2080  is about before you try to spin ?

Alderon86:
@Kimuji i already did locally, thats why the Pr is still tagged as WIP.
Besides that we shouldnt make another request, we can use existing ones.
Just wait till im done ;)

Bart274:
how far is this from being merged? this would improve the odds of having accounts run longer than they do now

Alderon86:
@Bart274 im rewriting parts of the code atm. If this is done, all left until merge would be 2 approvals

Bart274:
@Alderon86 Thanks for the quick reaction and for the nice work on this PR :-)

I think the only thing left to humanize the workers is having the accounts not only encounter the pokemon but also to let them try to catch the pokemon and to have them discard pokemons

343334:
@Bart274 Theres a lot more to it than just that.  A start would be to randomize the ari / asi by account, and not accounts.csv so each individual acct is different instead of the whole batch acting together. But that's for a different pr than this.

boby361:
Its been brought to my attention that apparently the PR doesn't request level up items. Which is creating an exception. Obviously if we want to act as a human,  a client would make this request.

pogom/humanize.py
```diff
@@ -0,0 +1,309 @@
...
+                        if (ready_to_hatch is not None and egg_id is not None
+                                or needs_egg is not None and egg_id
+                                is not None):
+                            if ready_to_hatch is None:
+                                basic_incubator = needs_egg
+                            else:
+                                basic_incubator = ready_to_hatch
+                            if (basic_incubator_empty is True and
+                                    pokemon_data[
+                                        'egg_km_walked_target'] == 2.0):
```
Kimuji: Tried it, there is en exception thrown here and below because "egg_km_walked_target" may not be in pokemon_data

pogom/humanize.py
```diff
@@ -0,0 +1,309 @@
...
+                                time.sleep(random.uniform(2, 4))
+                                egg_hatching_response = egg_hatching_request(
+                                    api, egg_id, basic_incubator)
+                            elif (basic_incubator_empty is True and 2.0 not in
```
Kimuji: Tried it, there is en exception thrown here and below because 2.0 is a float and it cannot iterate float, same for 5.0 below

Alderon86: @Kimuji i have nearly rewriten it all, going to push when its finished!

boby361: Awesome! Glad to hear it. Appreciate the great work!

pogom/humanize.py
Arengo: It's already [here](https://github.com/RocketMap/RocketMap/pull/2073/files#diff-1c093ecd5a518965b676864da75b95eeR89) :-O

pogom/account.py
```diff
@@ -453,3 +451,283 @@ def next(self, set_name, coords_to_scan):
...
+            parsed_items += item_count
+        elif 'egg_incubators' in item_data:
+            incubators = item_data['egg_incubators']['egg_incubator']
+            for incubator in incubators:
+                account['incubators'][incubator['id']] = {
+                    'item_id': incubator['item_id'],
+                    'uses_remaining': incubator.get('uses_remaining', 0),
+                    'pokemon_id': incubator.get('pokemon_id', 0)
+                }
+        if 'pokemon_data' in item_data:
```
friscoMad: Shouldn't this be a elif?

pogom/account.py
```diff
@@ -453,3 +451,283 @@ def next(self, set_name, coords_to_scan):
...
+            p_data = item_data['pokemon_data']
+            p_id = p_data.get('id', 0)
+            pokemon_id = p_data.get('pokemon_id', 0)
+            is_egg = p_data.get('is_egg', False)
+            if p_id and pokemon_id:
```
friscoMad: probably not wrong but reading the code makes more sense to check if not p_id and continue and doing an if/else on pokemon_id or is_egg but maybe there is some other option.

pogom/account.py
```diff
@@ -453,3 +451,283 @@ def next(self, set_name, coords_to_scan):
...
+            egg_id = random.choice(egg_ids)
+            km_target = account['eggs'][egg_id]['km_target']
+            time.sleep(random.uniform(2.0, 4.0))
+            request_use_item_egg_incubator(api, account, incubator_id,
+                                           egg_id)
+            log.info('Egg #%s ({:.1f} km) is on incubator #%s.').format(
+                    egg_id, km_target, incubator_id)
+            del account['eggs'][egg_id]
+        elif incubator['pokemon_id'] != 0:
+            log.info('Already incubating egg!')
```
friscoMad: remove from incubators, we are only storing empty ones

pogom/account.py
```diff
@@ -545,8 +545,9 @@ def spin_pokestop(api, account, args, fort, step_location):
                     account['username'])
         return False
     # Set 50% Chance to spin a Pokestop.
+    if account['level'] == 1:
+        random.randint == 49
```
Obihoernchen: No pls not. How is this supposed to work?!

Do:
```python
if random.randint(0, 100) < 50 or account['level'] == 1:
    fort_details_request(api, account, fort)
```

Obihoernchen: And please use: `random.random() < 0.5` instead. It's way faster.

```bash
# python -m timeit -s "import random" "random.randint(0,100) < 0.5"
1000000 loops, best of 3: 1.2 usec per loop
# python -m timeit -s "import random" "random.random() < 0.5"
10000000 loops, best of 3: 0.0824 usec per loop
```

Arengo: ``if rand.choice([0 1]):``

sebastienvercammen:
Closed, at least temporarily.

Since this brings us into actual botting territory, which we're against, we're going to keep it closed until PGoClient (API request rework) and the GMO rework are finished and merged.

By doing this, we can test the effect of PGoClient/GMO and reopen 2073 only if it's absolutely necessary.

sebastienvercammen:
Due to current priorities, this PR is now back on the table.

Alphawarz:
No errors since start up.

pogom/account.py
```diff
@@ -143,59 +143,58 @@ def rpc_login_sequence(args, api, account):
...
+        req.check_awarded_badges()
+        req.download_settings()
+        response = req.call(False)
 
         parse_download_settings(account, response)
         parse_new_timestamp_ms(account, response)
+        parse_inventory(api, account, response)
 
         total_req += 1
-        time.sleep(random.uniform(.53, 1.1))
```
friscoMad: No sleep or nulltime exception check?

pogom/account.py
```diff
@@ -418,47 +407,48 @@ def complete_tutorial(args, api, account, tutorial_state):
...
         time.sleep(random.uniform(0.5, 0.6))
-        request = api.create_request()
-        request.get_player(player_locale=args.player_locale)
-        responses = request.call(False).get('responses', {})
+        req = api.create_request()
+        req.get_player(
+            player_locale=args.player_locale)
+        responses = req.call(False).get('responses', {})
 
         if 'GET_INVENTORY' in responses:
```
friscoMad: This needs tests, it seems there is a GET_INVENTORY response without asking for it, if true it should be parsed properly and select a poke from inv to buddy it later.

pogom/account.py
```diff
@@ -543,34 +508,90 @@ def get_player_level(map_dict):
...
+    spins_h = account['session_spins'] * 3600.0 / elapsed_time
+    account['hour_spins'] = spins_h
+
+    # Refresh visited pokestops that were on timeout.
+    used_pokestops = dict(account['used_pokestops'])
+    for pokestop_id in account['used_pokestops']:
+        last_attempt = account['used_pokestops'][pokestop_id]
+        if (last_attempt + pokestop_timeout) < time.time():
+            del used_pokestops[pokestop_id]
+            account['used_pokestops'] = used_pokestops
```
friscoMad: wrong indentation

pogom/account.py
```diff
@@ -543,34 +508,90 @@ def get_player_level(map_dict):
...
+
+
+# Check if Pokestop is spinnable and not on cooldown.
+def pokestop_spinnable(fort, step_location):
+    spinning_radius = 0.038
+    in_range = in_radius((fort.latitude, fort.longitude),
+                         step_location, spinning_radius)
+    now = time.time()
+    pause_needed = fort.cooldown_complete_timestamp_ms / 1000 > now
```
friscoMad: .cooldown_complete_timestamp_ms  could be 0? if it can then this is not right.

pogom/account.py
```diff
@@ -581,6 +602,212 @@ def spin_pokestop(api, account, fort, step_location):
...
+        player_level = get_player_level(api_response)
+        if player_level:
+            account['level'] = player_level
+
+
+def parse_get_player(account, api_response):
+    if 'GET_PLAYER' in api_response['responses']:
+        player_data = api_response['responses']['GET_PLAYER'].player_data
+
+        account['warning'] = api_response['responses']['GET_PLAYER'].warn
```
friscoMad: change to account['status'] and store warn or ban?, we can use it later for shadow also

pogom/account.py
```diff
@@ -581,6 +602,212 @@ def spin_pokestop(api, account, fort, step_location):
...
+            log.info('Egg #%s (%.0f km) is on incubator #%s.',
+                     egg['id'], egg['km_target'], incubator['id'])
+            account['incubators'].remove(incubator)
+        else:
+            log.warning('Failed to put egg on incubator #%s.', incubator['id'])
+
+    return
+
+
 def spin_pokestop_request(api, account, fort, step_location):
```
friscoMad: I would parse inventory here instead of in the caller, but parse_inventory will need to handle that it could be no inventory to parse if there is captcha or similar.

pogom/models.py
```diff
@@ -2110,6 +2101,17 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                 else:
                     lure_expiration, active_fort_modifier = None, None
 
+                # Spin Pokestop with 50% chance.
+                if args.pokestop_spinning and pokestop_spinnable(
+                        f, step_location):
+                    spin_pokestop(api, account, args, f, step_location)
+                if args.pokestop_spinning and not config[
```
friscoMad: Why not move this out of the parse pokestop if?, I don't see why this should be on for spinnin

pogom/models.py
```diff
@@ -2253,6 +2255,8 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                             wh_update_queue.put(('raid', wh_raid))
 
         # Helping out the GC.
+        if 'GET_INVENTORY' in map_dict['responses']:
```
friscoMad: This is not needed anymore it is cleared in clean_dict_response as inventory is parsed before all this.

pogom/utils.py
```diff
@@ -304,6 +304,15 @@ def get_args():
                         help=('Use speed scanning to identify spawn points ' +
                               'and then scan closest spawns.'),
                         action='store_true', default=False)
+    parser.add_argument('-spin', '--pokestop-spinning',
+                        help=('Spin Pokestops with "50%" Chance'),
+                        action='store_true', default=False)
+    parser.add_argument('-ams', '--account-max-spins',
+                        help='Maximum number of Pokestop spins per hour.',
+                        type=int, default=80)
+    parser.add_argument('-prt', '--pokestop-refresh-time',
```
friscoMad: We are checking ingame cooldowns, so this should not be needed

pogom/models.py
friscoMad: In fact we can spin also gyms so make sense to leave it at the top without checking the type (but we will need some changes as spining a gym needs a gym_get_info instead of a fort_details) up to you.

friscoMad:
Fully reviewed, there are some changes needed, but not a lot, the only big change I think we sould do is move all parse_inventory to request functions that way we will not forget to handle it later, if for some reason we skip some parse we will be leaking inventory items (eggs hatched mostly as regular items are not sent as delta) that could lead to full box but without the data to release the pokes.

Also there are a lot of format changes in exceptions that I like more as they were before, but personal preference only 😈 

In general great work and pretty neat code and handling, good work @Alderon86 !!

Alderon86:
It never ever can lead to a full box, try to think of how many eggs would be needed for that!
also if we somehow skip the parsing the next parsing will give you correct values and clear them if there are more then 5-10 accordingly.

pogom/account.py
friscoMad: This is not a problem of this PR I did open #2158 some time ago.

friscoMad:
Codewise I don't see any bug, I would make it possible to spin without parsing pokestops as both are totally unrelated. But can not approve without full testing, there are a lot situations covered in this pr and needs a long run test to at least be half sure nothing is broken.

pogom/models.py
```diff
@@ -2091,16 +2092,6 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     (f['last_modified'] -
                      datetime(1970, 1, 1)).total_seconds())) for f in query]
 
-        # Complete tutorial with a Pokestop spin
-        if args.complete_tutorial:
```
friscoMad: Another thing, this PR disables -tut as a flag, tutorial is done with it enabled or not, so we should remove it.

pogom/account.py
```diff
@@ -544,34 +510,89 @@ def get_player_level(map_dict):
...
 
 
-def spin_pokestop(api, account, fort, step_location):
-    spinning_radius = 0.04
-    if in_radius((fort.latitude, fort.longitude), step_location,
-                 spinning_radius):
-        log.debug('Attempt to spin Pokestop (ID %s).', fort.id)
-
+def reset_account(account):
+    account['start_time'] = time.time()
```
friscoMad: We need to reset also hlvl accounts

PR #2074, opened 2017-06-05
Deterministic Device Information

neskk:
Pretty much tries to do the same as  #2072 but without the need to save the data in the database.
This is accomplished by hashing the username an derive all used specs from that hash. This idea was talked about by @friscoMad and it seemed quite elegant so I decided to implement it.

## Description

Pretty simple, should make sure that every time username "X" logs in, it does so using the same device as before.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -27,7 +27,10 @@ def setup_api(args, status):
     if args.mock != '':
         api = FakePogoApi(args.mock)
     else:
-        device_info = generate_device_info(status['username'])
+        identifier = status['username']
+        if args.hashing_salt:
+            identifier += args.hashing_salt
+        device_info = generate_device_info(identifier)
         api = PGoApi(device_info=device_info)
 
     # New account - new proxy.
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -27,7 +27,7 @@ def setup_api(args, status):
     if args.mock != '':
         api = FakePogoApi(args.mock)
     else:
-        device_info = generate_device_info()
+        device_info = generate_device_info(status['username'])
         api = PGoApi(device_info=device_info)
 
     # New account - new proxy.
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -22,14 +22,12 @@ class TooManyLoginAttempts(Exception):
 
 
 # Create the API object that'll be used to scan.
-def setup_api(args, status):
+def setup_api(args, status, account):
     # Create the API instance this will use.
     if args.mock != '':
         api = FakePogoApi(args.mock)
     else:
-        identifier = status['username']
-        if args.hashing_salt:
-            identifier += args.hashing_salt
+        identifier = account['username'] + account['password']
         device_info = generate_device_info(identifier)
         api = PGoApi(device_info=device_info)
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -13,7 +13,8 @@ import socket
 import struct
 import zipfile
 import requests
-from uuid import uuid4
+import hashlib
+
 from s2sphere import CellId, LatLng
 
 from . import config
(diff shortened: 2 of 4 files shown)


rocketrobot99:
@neskk I do like this PR and @friscoMad's original idea. I just have a couple of concerns:
1. Would this yeild the same device_info on every PC? If so, then if Niantic wanted to find all the RM users they just need to md5 hash the account name and marks all the ones that are the same as the device ID. I think this only works if a Salt is applied.
2. Using the mod of the length of each list only works if the list doesn't get altered. If we add a new device or firmware then a particular account will decode into different device info.

neskk:
@rocketrobot99 yep your concerns are all valid.
This was just an experiment and I think it's better than having a different device info at each login.
1. I'm thinking how we could use a salt without asking users for one, like using hostname as a salt. But... I don't like this idea so I will probably add a config argument for the salt.
2. I'm not really concerned about this, because we probably won't update the device list that often. Comparing with what we currently have it's way better because this info changes at every login but accounts still seemed to work fine. I'm thinking if there's a way to prevent that all accounts change their device info if lists are updated but I'm not seeing anything simple to solve this problem.

pogom/account.py
```diff
@@ -27,7 +27,10 @@ def setup_api(args, status):
     if args.mock != '':
         api = FakePogoApi(args.mock)
     else:
-        device_info = generate_device_info()
+        identifier = status['username']
+        if args.hashing_salt:
```
friscoMad: I would prefer to do the salting inside generate device function, it makes more sense to have everything related in the same place.

pogom/utils.py
```diff
@@ -887,7 +891,11 @@ def get_blacklist():
            'iPhone9,4': 'D111AP'}
 
 
-def generate_device_info():
+def generate_device_info(identifier):
+    pick_hash = hash(identifier)
```
sLoPPydrive: Just read this comment on SO - can this be a problem?

> It is not guaranteed to be the same every time, but it is guaranteed to return the same value within the same process. If you want a non-changing hash, use hashlib functions instead.

Source: https://stackoverflow.com/questions/40298023/python-hash-function-on-strings#comment67854083_40298051

rocketrobot99:
@neskk I am just thinking, would it be possible to use the account password as a salt somehow? I don't think Niantic would / should have access to an unencrypted version. Just an idea.

friscoMad:
@rocketrobot99 I forgot that we are using delegated services for login, that solution is perfect!!
And we can add a comment on the device section to only update the devices once or twice a year, I don't think that such changes could be any problem with niantic.

friscoMad:
@rocketrobot99 I forgot that we are using delegated services for login, that solution is perfect!!
And we can add a comment on the device section to only update the devices once or twice a year, I don't think that such changes could be any problem with niantic.

pogom/utils.py
friscoMad: Hash() is defined by python implementation and not by the standard so it can change between versions, os, etc if we want to be cross server compatible then a proper hash function should be used, but I am not sure if it is really needed.

sebastienvercammen:
@neskk RE @rocketrobot99: Using the account's password for the salt is perfect. The only time it would yield a different result is when they've changed their password, but changing devices once in a while is not unrealistic.

RE @sLoPPydrive and @friscoMad: We should indeed be cross server compatible, so we need to use a hash function that stays constant between versions.

neskk:
I agree with suggestions but, not sure if using the password as a salt is that good:
1. setup_api doesn't have the account data available (it's not that difficult to change but...)
2. without a specified salt you have less control of whether you want to reset device information (by changing the salt)
I will move away from the hash() function and use some other in hashlibs. Probably will make the change to remove `--hashing-salt` aswell but I would like more feedback on this.
@friscoMad the "salting" happens in setup_api so that I don't have to pass more parameters to `generate_device_information`. If using password as a salt probably I will have to modify `setup_api` and `generate_device_information` to receive an account as parameter so what you're asking probably will happen.

PR #2077, opened 2017-06-07
Shifted api_version to args.

neskk:
Since most of the functions that require api_version already have access to `args` variable, it seemed like a good spot to store api_version.

## Description

Created a config argument for the API version currently in use. I know this parameter is not really suppose to be changed by the user but this way we'll have a clean and centralized variable with the api_version in use.
This will come in handy to make the `download_remote_config_version` request.

## Types of changes

## Checklist:


pogom/search.py
```diff
@@ -1081,10 +1079,8 @@ def search_worker_thread(args, api_version, account_queue, account_sets,
                                     current_gym, len(gyms_to_update),
                                     step_location[0], step_location[1])
                             time.sleep(random.random() + 2)
-                            response = gym_request(api,
-                                                   step_location,
-                                                   gym,
-                                                   api_version)
+                            response = gym_request(api, account, step_location,
```
neskk: Sorry, left behind `account` variable :|

PR #2088, opened 2017-06-13
[Optimization] Rework login flow for better App Simulation.

Alderon86:
## Motivation and Context

We need to behave more like the real app, including a correct login sequence.
## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -262,7 +262,19 @@ def check_login(args, account, api, position, proxy_url):
                       repr(e))
         raise LoginSequenceFail('Failed during login sequence.')
 
-    # TODO: # 7 - Make a request to get Shop items.
+    # Needs updated PGoApi to be used.
+    # log.info('Retrieving Store Items...')
+    # try:  # 7 - Make an empty request to retrieve store items.
+    #    request = api.create_request()
+    #    request.get_store_items()
+    #    response = request.call()
+    #    total_req += 1
+    #    time.sleep(random.uniform(.6, 1.1))
+    # except Exception as e:
+    #    log.exception('Login for account %s failed. Exception in ' +
+    #                  'retrieving Store Items: %s', account['username'],
+    #                  repr(e))
+    #    raise LoginSequenceFail('Failed during login sequence.')
 
     log.info('Check if Account has leveled up and collect rewards...')
     try:  # 8 - Check if there are level up rewards to claim.
@@ -282,7 +294,8 @@ def check_login(args, account, api, position, proxy_url):
 
     except Exception as e:
         log.exception('Login for account %s failed. Exception in ' +
-                      'level_up_rewards: %s', account['username'], repr(e))
+                      'Level up rewards request: %s', account['username'],
+                      repr(e))
         raise LoginSequenceFail('Failed during login sequence.')
 
     log.info('RPC Login Sequence for account %s with %s requests successful.',
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -21,6 +21,10 @@ class TooManyLoginAttempts(Exception):
     pass
 
 
+class LoginSequenceFail(Exception):
+    pass
+
+
 # Create the API object that'll be used to scan.
 def setup_api(args, status, account):
     # Create the API instance this will use.
@@ -56,7 +60,8 @@ def setup_api(args, status, account):
 
 # Use API to check the login status, and retry the login if possible.
 def check_login(args, account, api, position, proxy_url):
-
+    total_req = 0
+    app_version = int(args.api_version.replace('.', '0'))
     # Logged in? Enough time left? Cool!
     if api._auth_provider and api._auth_provider._ticket_expire:
         remaining_time = api._auth_provider._ticket_expire / 1000 - time.time()
@@ -98,8 +103,191 @@ def check_login(args, account, api, position, proxy_url):
             account['username'], num_tries)
         raise TooManyLoginAttempts('Exceeded login attempts.')
 
-    log.debug('Login for account %s successful.', account['username'])
-    time.sleep(20)
+    time.sleep(random.uniform(2, 4))
+
+    log.info('Starting RPC Login Sequence...')
+    try:  # 1 - Make an empty request to mimick real app behavior.
+        request = api.create_request()
+        request.call()
+        total_req += 1
+        time.sleep(random.uniform(.43, .97))
+    except Exception as e:
+        log.exception('Login for account %s failed.' +
+                      ' Exception in call request: %s', account['username'],
+                      repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    log.info('Trying to fetch Player...')
+    try:  # 2 - Get Player request.
+        req = api.create_request()
+        req.get_player(
+            player_locale={
+                'country': 'US',
+                'language': 'en',
+                'timezone': 'America/Denver'})
+        req.call()
+        total_req += 1
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Get Player request: %s', account['username'], repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    log.info('Trying to download Remote Config Version...')
+    try:  # 3 - Download Remote Config Version request.
+        old_config = account.get('remote_config', {})
+        request = api.create_request()
+        request.download_remote_config_version(platform=1,
+                                               app_version=app_version)
+        request.check_challenge()
+        request.get_hatched_eggs()
+        request.get_inventory(last_timestamp_ms=0)
+        request.check_awarded_badges()
+        request.download_settings()
+        response = request.call()
+        total_req += 1
+        parse_download_settings(account, response)
+        parse_new_timestamp_ms(account, response)
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Error while downloading remote config: %s.', repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    log.info('Trying to get Asset Diggest...')
+    # 4 - Get Asset Digest request.
+    config = account.get('remote_config', {})
+    if config.get('asset_time', 0) > old_config.get('asset_time', 0):
+        i = random.randint(0, 3)
+        req_count = 0
+        result = 2
+        page_offset = 0
+        page_timestamp = 0
+        time.sleep(random.uniform(.7, 1.2))
+        while result == 2:
+            try:
+                request = api.create_request()
+                request.get_asset_digest(
+                    platform=1,
+                    app_version=app_version,
+                    paginate=True,
+                    page_offset=page_offset,
+                    page_timestamp=page_timestamp)
+                request.check_challenge()
+                request.get_hatched_eggs()
+                request.get_inventory(last_timestamp_ms=account[
+                    'last_timestamp_ms'])
+                request.check_awarded_badges()
+                request.download_settings(hash=account[
+                    'remote_config']['hash'])
+                response = request.call()
+                parse_new_timestamp_ms(account, response)
+                req_count += 1
+                total_req += 1
+                if i > 2:
+                    time.sleep(random.uniform(1.4, 1.6))
+                    i = 0
+                else:
+                    i += 1
+                    time.sleep(random.uniform(.3, .5))
+                get_asset_digest = response['responses']['GET_ASSET_DIGEST']
+                result = get_asset_digest.get('result', 0)
+                page_offset = get_asset_digest.get('page_offset', 0)
+                page_timestamp = get_asset_digest.get('timestamp_ms', 0)
+                log.debug('Completed %d requests to get asset digest.',
+                          req_count)
+
+            except KeyError:
+                break
+
+    log.info('Trying to download Items Templates...')
+    # 5 - Download Item Templates request.
+    if config.get('template_time', 0) > old_config.get('template_time', 0):
+        i = random.randint(0, 3)
+        req_count = 0
+        result = 2
+        page_offset = 0
+        page_timestamp = 0
+        while result == 2:
+            try:
+                request = api.create_request()
+                request.download_item_templates(paginate=True,
+                                                page_offset=page_offset,
+                                                page_timestamp=page_timestamp)
+                request.check_challenge()
+                request.get_hatched_eggs()
+                request.get_inventory(last_timestamp_ms=account[
+                    'last_timestamp_ms'])
+                request.check_awarded_badges()
+                request.download_settings(hash=account[
+                    'remote_config']['hash'])
+                response = request.call()
+                parse_new_timestamp_ms(account, response)
+                req_count += 1
+                total_req += 1
+                if i > 2:
+                    time.sleep(random.uniform(1.4, 1.6))
+                    i = 0
+                else:
+                    i += 1
+                    time.sleep(random.uniform(.3, .5))
+
+                download_item_templates = response['responses'][
+                    'DOWNLOAD_ITEM_TEMPLATES']
+                result = download_item_templates.get('result', 0)
+                page_offset = download_item_templates.get('page_offset', 0)
+                page_timestamp = download_item_templates.get('timestamp_ms', 0)
+                log.debug('Completed %d requests to download' +
+                          ' Item Templates.', req_count)
+            except KeyError:
+                break
+
+    log.info('Trying to get Player Profile...')
+    try:  # 6 - Get Player Profile request.
+        request = api.create_request()
+        request.get_player_profile()
+        request.check_challenge()
+        request.get_hatched_eggs()
+        request.get_inventory(last_timestamp_ms=account['last_timestamp_ms'])
+        request.check_awarded_badges()
+        request.download_settings(hash=account['remote_config']['hash'])
+        request.get_buddy_walked()
+        response = request.call()
+        total_req += 1
+        parse_new_timestamp_ms(account, response)
+        time.sleep(random.uniform(.2, .3))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Get Player Profile request: %s', account['username'],
+                      repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    # TODO: # 7 - Make a request to get Shop items.
+
+    log.info('Check if Account has leveled up and collect rewards...')
+    try:  # 8 - Check if there are level up rewards to claim.
+        request = api.create_request()
+        request.level_up_rewards(level=account['level'])
+        request.check_challenge()
+        request.get_hatched_eggs()
+        request.get_inventory(last_timestamp_ms=account['last_timestamp_ms'])
+        request.check_awarded_badges()
+        request.download_settings(hash=account['remote_config']['hash'])
+        request.get_buddy_walked()
+        request.get_inbox(is_history=True)
+        response = request.call()
+        total_req += 1
+        parse_new_timestamp_ms(account, response)
+        time.sleep(random.uniform(.45, .7))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'level_up_rewards: %s', account['username'], repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    log.info('RPC Login Sequence for account %s with %s requests successful.',
+             account['username'], total_req)
+    time.sleep(random.uniform(10, 20))
 
 
 # Check if all important tutorial steps have been completed.
@@ -249,7 +437,7 @@ def tutorial_pokestop_spin(api, player_level, forts, step_location, account):
             account['username'])
         for fort in forts:
             if fort.get('type') == 1:
-                if spin_pokestop(api, fort, step_location):
+                if spin_pokestop(api, account, fort, step_location):
                     log.debug(
                         'Account %s successfully spun a Pokestop ' +
                         'after completed tutorial.',
@@ -275,24 +463,23 @@ def get_player_level(map_dict):
     return 0
 
 
-def spin_pokestop(api, fort, step_location):
+def spin_pokestop(api, account, fort, step_location):
     spinning_radius = 0.04
     if in_radius((fort['latitude'], fort['longitude']), step_location,
                  spinning_radius):
         log.debug('Attempt to spin Pokestop (ID %s)', fort['id'])
-
         time.sleep(random.uniform(0.8, 1.8))  # Do not let Niantic throttle
-        spin_response = spin_pokestop_request(api, fort, step_location)
+        response = spin_pokestop_request(api, account, fort, step_location)
         time.sleep(random.uniform(2, 4))  # Do not let Niantic throttle
 
         # Check for reCaptcha
-        captcha_url = spin_response['responses'][
+        captcha_url = response['responses'][
             'CHECK_CHALLENGE']['challenge_url']
         if len(captcha_url) > 1:
             log.debug('Account encountered a reCaptcha.')
             return False
 
-        spin_result = spin_response['responses']['FORT_SEARCH']['result']
+        spin_result = response['responses']['FORT_SEARCH']['result']
         if spin_result is 1:
             log.debug('Successful Pokestop spin.')
             return True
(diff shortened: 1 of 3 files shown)


pogom/account.py
```diff
@@ -98,8 +98,133 @@ def check_login(args, account, api, position, proxy_url):
...
+        request.download_settings()
+        request.get_buddy_walked()
+        request.call()
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Background Device: %s', account['username'], repr(e))
+
+    # TODO: # 9 Make a request to get Shop items.
+    # Also we need to parse last_timestamp_ms for each Account with
```
IcognitoUnown: Have a look how it was handled in [PGoClient](https://github.com/RocketMap/RocketMap/pull/2060/files#diff-196fc76d2004b90f04582f0a09fc1f3cR142).

pogom/account.py
```diff
@@ -98,8 +98,133 @@ def check_login(args, account, api, position, proxy_url):
...
+                      repr(e))
+
+    try:  # 2 - Get Player request.
+        req = api.create_request()
+        req.get_player(
+            player_locale={
+                'country': 'US',
+                'language': 'en',
+                'timezone': 'America/Los_Angeles'})
+        req.check_challenge()
```
IcognitoUnown: Are you really sure about the following lines? MITM or Monocle tells different.

pogom/account.py
```diff
@@ -98,8 +98,133 @@ def check_login(args, account, api, position, proxy_url):
...
+        request.download_settings()
+        request.get_buddy_walked()
+        request.call()
+        time.sleep(random.uniform(.45, .7))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'level_up_rewards: %s', account['username'], repr(e))
+
+    try:  # 8 - Register Background Device.
```
IcognitoUnown: To include or not include, that is here the question. MITM from the past included it but we don't know if it still exists since it got harder/impossible to inspect.

pogom/account.py
```diff
@@ -98,8 +98,133 @@ def check_login(args, account, api, position, proxy_url):
...
+                      ' Exception in call request: %s', account['username'],
+                      repr(e))
+
+    try:  # 2 - Get Player request.
+        req = api.create_request()
+        req.get_player(
+            player_locale={
+                'country': 'US',
+                'language': 'en',
+                'timezone': 'America/Los_Angeles'})
```
IcognitoUnown: Why LA?

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+        response = request.call()
+
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Background Device: %s', account['username'], repr(e))
+
+    # TODO: # 9 Make a request to get Shop items.
+
+        log.debug('Login for account %s successful.', account['username'])
```
Arengo: One indent less here and in the following line.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+        request = api.create_request()
+        request.call()
+        time.sleep(random.uniform(.43, .97))
+    except Exception as e:
+        log.exception('Login for account %s failed.' +
+                      ' Exception in call request: %s', account['username'],
+                      repr(e))
+
+    try:  # 2 Get Player request.
+        req = api.create_request()
```
Arengo: Consistency "request".

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+
+                    time.sleep(random.uniform(.53, 1.1))
+                    log.debug('Completed %d requests to get asset digest.',
+                              req_count)
+
+            except Exception as e:
+                log.exception('Error while downloading Asset Digest: %s.',
+                              repr(e))
+
+    # 5 Download Item Templates request.
```
Arengo: Dash.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+        request.download_settings(hash=account['remote_config']['hash'])
+        request.get_buddy_walked()
+        response = request.call()
+        time.sleep(random.uniform(.45, .7))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'level_up_rewards: %s', account['username'], repr(e))
+
+    try:  # 8 Register Background Device request.
```
Arengo: Dash.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+        request.download_settings(hash=account['remote_config']['hash'])
+        request.get_buddy_walked()
+        response = request.call()
+
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Background Device: %s', account['username'], repr(e))
+
+    # TODO: # 9 Make a request to get Shop items.
```
Arengo: Dash.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+    try:  # 1 - Make an empty request to mimick real app behavior.
+        request = api.create_request()
+        request.call()
+        time.sleep(random.uniform(.43, .97))
+    except Exception as e:
+        log.exception('Login for account %s failed.' +
+                      ' Exception in call request: %s', account['username'],
+                      repr(e))
+
+    try:  # 2 Get Player request.
```
Arengo: Dash.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+        req = api.create_request()
+        req.get_player(
+            player_locale={
+                'country': 'US',
+                'language': 'en',
+                'timezone': 'America/Denver'})
+        req.call()
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.error('Exception getting player information: %s', repr(e))
```
Arengo: Consistency "log.exception" and adjust the actual text. Keep it consistent with the other exception texts.

pogom/account.py
```diff
@@ -98,8 +98,183 @@ def check_login(args, account, api, position, proxy_url):
...
+    try:  # 3 - Download Remote Config Version request.
+        old_config = account['remote_config']
+        request = api.create_request()
+        request.download_remote_config_version(platform=1,
+                                               app_version=int(
+                                                args.api_version.replace(
+                                                    '.', '0')))
+        request.check_challenge()
+        request.get_hatched_eggs()
+        request.get_inventory(last_timestamp_ms=0)
```
Arengo: When does it get updated besides an initialisation ``reset_timestam``? Make sure to update from API response and track provided information of ``get_inventory`` since when using a timestamp != 0 you don't get any inventory content with the most requests anymore. Compare with [PGoClient](https://github.com/RocketMap/RocketMap/pull/2060/files#diff-196fc76d2004b90f04582f0a09fc1f3cR143).

friscoMad: I think it is good as it is now as each pr should work when merged as is, whichever PR is merged later should adapt/rebase it's code to the other one.

pogom/account.py
```diff
@@ -98,8 +98,190 @@ def check_login(args, account, api, position, proxy_url):
...
+        result = 2
+        page_offset = 0
+        page_timestamp = 0
+        time.sleep(random.uniform(.7, 1.2))
+        while result == 2:
+            try:
+                request = api.create_request()
+                request.get_asset_digest(
+                    platform=1,
+                    app_version=int(args.api_version.replace('.', '0')),
```
friscoMad: save api version replaced and converted to int somewhere instead of doing each time.

pogom/account.py
```diff
@@ -98,8 +98,190 @@ def check_login(args, account, api, position, proxy_url):
...
+        request.get_buddy_walked()
+        response = request.call()
+        parse_new_timestamp_ms(account, response)
+        time.sleep(random.uniform(.45, .7))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'level_up_rewards: %s', account['username'], repr(e))
+
+    try:  # 8 - Register Background Device request.
```
friscoMad: I know it is as Monocle is doing and probably this is not a problem right but regular users does not do this request right now, at least that's what Chrales said, so I think removing it is better.

pogom/account.py
```diff
@@ -98,8 +98,190 @@ def check_login(args, account, api, position, proxy_url):
...
+
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'Background Device: %s', account['username'], repr(e))
+
+    # TODO: # 9 - Make a request to get Shop items.
+
+        log.debug('Login for account %s successful.', account['username'])
+        time.sleep(random.uniform(10, 20))
```
friscoMad: All exceptions are catched and does not return an error, it just continue the flow, they should return an error or rethrow.

friscoMad:
As level is tracked in account, you must also modify parse_map function to use level from account, else most of the time it won't be present (maybe it is always present even with timestamps, not sure).

pogom/account.py
Arengo: Correct.

pogom/account.py
Arengo: Afaik the last MITM data said that it was always registered, even when there was no Apple Watch at the iPhone. Currently this can't checked back anymore. So it's a vague guess, kind of 50:50.

pogom/account.py
Arengo: Not completely right afaik, When an exception occurs, it get's catched but it exits the function as one, too - resulting in an exception of login handling in search.py resulting in an account switch.

SiteEffect:
@friscoMad already being addressed.

pogom/account.py
Alderon86: @friscoMad exactly what Arengo said. Exception is catched resulting in an login failure and account switch

pogom/search.py
```diff
@@ -1223,28 +1226,27 @@ def map_request(api, position, no_jitter=False):
...
+        log.debug('Getting details for gym @ %f/%f (%fkm away)',
                   gym['latitude'], gym['longitude'],
                   calc_distance(position, [gym['latitude'], gym['longitude']]))
         req = api.create_request()
         req.get_gym_details(gym_id=gym['gym_id'],
                             player_latitude=f2i(position[0]),
                             player_longitude=f2i(position[1]),
                             gym_latitude=gym['latitude'],
-                            gym_longitude=gym['longitude'],
-                            client_version=api_version)
```
friscoMad: api_version is sent by the client even if monocle do not send it

pogom/account.py
friscoMad: Maybe I did not explain myself properly, what I wanted to say is that all exceptions in each request are catched, that's nice, we have a log but as they are catched in each request the flow do continue to the next one unless it is an uncatched exception that will buble up to search.py.

For example an exception in line 105 will get catched on 108 and flow will continue with the login process at some point it will crash due to missing info or something else but we should stop the login process as soon as we now there is a problem and report back.

pogom/search.py
Arengo: Is this only the case for ``get_gym_details``?

pogom/account.py
Arengo: So you mean we should like ``return False`` in any exception?

pogom/models.py
```diff
@@ -2031,8 +2031,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                         else:
                             # Update level indicator before we clear the
                             # response.
-                            encounter_level = get_player_level(
-                                encounter_result)
+                            encounter_level = account['level']
```
Arengo: ``hlvl_account['level']``

pogom/account.py
```diff
@@ -56,7 +56,6 @@ def setup_api(args, status, account):
```
Arengo: Commit message: ``adding missing call``. Is it missing?

SiteEffect:
This develops really good and is almost out of WIP and ready for a general review.

pogom/account.py
Alderon86: no its not, maybe a i have choosen the wrong commit description,
theres indeed one missing to retrieve store items. Lets figure it out

pogom/account.py
friscoMad: yeah that's what I think it is the best, as I have seen that in 2.7 when raising a new exception it is not possible to keep the stack trace, so imho the best thing would be catching, logging and returning false.

pogom/account.py
```diff
@@ -98,8 +99,206 @@ def check_login(args, account, api, position, proxy_url):
...
+                parse_new_timestamp_ms(account, response)
+                req_count += 1
+                total_req += 1
+                if i > 2:
+                    time.sleep(random.uniform(1.4, 1.6))
+                    i = 0
+                else:
+                    i += 1
+                    time.sleep(random.uniform(.3, .5))
+                    result = response['responses']['GET_ASSET_DIGEST']
```
michikrug: Wrong indentation and assignment of result

pogom/account.py
```diff
@@ -98,8 +99,206 @@ def check_login(args, account, api, position, proxy_url):
...
+                req_count += 1
+                total_req += 1
+                if i > 2:
+                    time.sleep(random.uniform(1.4, 1.6))
+                    i = 0
+                else:
+                    i += 1
+                    time.sleep(random.uniform(.3, .5))
+
+                    result = response['responses']['DOWNLOAD_ITEM_TEMPLATES']
```
michikrug: Wrong indentation and assignment of result

pogom/account.py
```diff
@@ -181,13 +181,12 @@ def check_login(args, account, api, position, proxy_url):
...
                     i += 1
                     time.sleep(random.uniform(.3, .5))
-                    result = response['responses']['GET_ASSET_DIGEST']
-                    page_offset = result['page_offset']
-                    page_timestamp = result['timestamp_ms']
-
-                    time.sleep(random.uniform(.53, 1.1))
-                    log.debug('Completed %d requests to get asset digest.',
-                              req_count)
+                result = response['responses']['GET_ASSET_DIGEST']
```
michikrug: Still wrong assignment of result. Should be like:
`res = response['responses']['GET_ASSET_DIGEST']`
`result = res['result']`
`page_offset = res['page_offset']`
`page_timestamp = res['timestamp_ms']`

pogom/account.py
```diff
@@ -225,12 +224,12 @@ def check_login(args, account, api, position, proxy_url):
                     i += 1
                     time.sleep(random.uniform(.3, .5))
 
-                    result = response['responses']['DOWNLOAD_ITEM_TEMPLATES']
-                    page_offset = result['page_offset']
-                    page_timestamp = result['timestamp_ms']
-                    log.debug('Completed %d requests to download' +
-                              ' item templates.', req_count)
-                    time.sleep(random.uniform(.53, 1.1))
+                result = response['responses']['DOWNLOAD_ITEM_TEMPLATES']
```
michikrug: Still wrong assignment of result. Should be like:
`res = response['responses']['DOWNLOAD_ITEM_TEMPLATES']`
`result = res['result']`
`page_offset = res['page_offset']`
`page_timestamp = res['timestamp_ms']`

Arengo: res, response, result get mixed up often. We make a request, get a response which carries a result that carries the single call responses we specified in the call sequence. I vote for completely removing all ``res`` and using ``response`` only, not extra storing ``result``.

michikrug: the result field is required for the while condition and is currently wrongly assigned (compared to other implementations)

sebastienvercammen:
@michikrug @Arengo @friscoMad Thoughts on last commit? Any additions to your previous comments, any changes still required from your point of view?

SiteEffect:
@sebastienvercammen looks good for final review (I mean actual testing, not code-review), imo.
Please test:
- login+exceptions
- GMO
- ``--gym-info``
- tutorial on new accounts (I know this was not scope, just addressing this that it won't spit any exceptions)
- encountering

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+                    time.sleep(random.uniform(.3, .5))
+
+                download_item_templates = response['responses'][
+                    'DOWNLOAD_ITEM_TEMPLATES']
+                result = download_item_templates.get('result', 0)
+                page_offset = download_item_templates.get('page_offset', 0)
+                page_timestamp = download_item_templates.get('timestamp_ms', 0)
+                log.debug('Completed %d requests to download' +
+                          ' item templates.', req_count)
+                time.sleep(random.uniform(.53, 1.1))
```
michikrug: I would say no need to sleep again

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+                log.debug('Completed %d requests to get asset digest.',
+                          req_count)
+
+            except Exception as e:
+                log.exception('Error while downloading Asset Digest: %s.',
+                              repr(e))
+                raise LoginSequenceFail('Failed during login sequence.')
+
+    # 5 - Download Item Templates request.
+    if config.get('template_time', 0) > old_config.get('template_time', 0):
```
michikrug: As I have understood, this can be left out to safe some requests, since it is only done when a new app version is installed or updated.

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+        parse_download_settings(account, response)
+        parse_new_timestamp_ms(account, response)
+        time.sleep(random.uniform(.53, 1.1))
+    except Exception as e:
+        log.exception('Error while downloading remote config: %s.', repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    # 4 - Get Asset Digest request.
+    config = account.get('remote_config', {})
+    if config.get('asset_time', 0) > old_config.get('asset_time', 0):
```
michikrug: As I have understood, this can be left out to safe some requests, since it is only done when a new app version is installed or updated.

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+        time.sleep(random.uniform(.45, .7))
+
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'level_up_rewards: %s', account['username'], repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    log.debug('Login for account %s with %s requests successful.',
+              account['username'], total_req)
+    time.sleep(random.uniform(10, 20))
```
michikrug: Why the long sleep?

pogom/account.py
```diff
@@ -321,31 +509,35 @@ def spin_pokestop(api, fort, step_location):
...
             player_longitude=step_location[1])
         req.check_challenge()
         req.get_hatched_eggs()
-        req.get_inventory()
+        req.get_inventory(last_timestamp_ms=account['last_timestamp_ms'])
         req.check_awarded_badges()
-        req.download_settings()
         req.get_buddy_walked()
-        spin_pokestop_response = req.call()
+        req.get_inbox(is_history=True,
```
michikrug: is_reverse=False,
not_before_ms=0

can be left out, since its default

pogom/account.py
```diff
@@ -356,18 +548,56 @@ def encounter_pokemon_request(api, encounter_id, spawnpoint_id, scan_location):
...
+                         .get('DOWNLOAD_REMOTE_CONFIG_VERSION', 0))
+        if 'asset_digest_timestamp_ms' in remote_config:
+            asset_time = remote_config['asset_digest_timestamp_ms'] / 1000000
+        if 'item_templates_timestamp_ms' in remote_config:
+            template_time = remote_config['item_templates_timestamp_ms'] / 1000
+
+        download_settings = {}
+        download_settings['hash'] = api_response[
+            'responses']['DOWNLOAD_SETTINGS']['hash']
+        download_settings['asset_time'] = asset_time
```
michikrug: Maybe add some default value if "if clause" before does not match

pogom/account.py
michikrug: But for new accounts, this probably should be called.

pogom/account.py
michikrug: But for new accounts, this probably should be called.

pogom/account.py
Arengo: Doesn't hurt, though

pogom/account.py
Arengo: Originally it even was strict 20s, we wan't at least some time before the first GMO, maybe 3-5 would also be enough, to emulate a phone which has to load graphics.

pogom/account.py
Arengo: Copy Pasta from Monocle.

pogom/account.py
Arengo: We could check that after the get_player request and parsing the tutorial completion steps.

jagauthier:
I'm getting this an awful lot:
```
2017-06-27 07:46:05,480 [   search-worker-6][       account][   ERROR] Error while downloading remote config: KeyError('DOWNLOAD_SETTINGS',).
  File "/home/jgauthier/RocketMap/pogom/account.py", line 146, in check_login
    parse_download_settings(account, response)
  File "/home/jgauthier/RocketMap/pogom/account.py", line 577, in parse_download_settings
    'responses']['DOWNLOAD_SETTINGS']['hash']
KeyError: 'DOWNLOAD_SETTINGS'
2017-06-27 07:46:05,480 [   search-worker-6][        search][   ERROR] Exception in search_worker under account XXXXX Exception message: LoginSequenceFail('Failed during login sequence.',).

```

Small instance, about 10 accounts, with a login delay of 15.
Even after it smooths out, I see it sporadically as accounts are rotated in.

pogom/account.py
Alderon86: if its called on updates also we shouldnt move it to the tut imo, maybe check if api version changed and then call it during login.

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+                    time.sleep(random.uniform(1.4, 1.6))
+                    i = 0
+                else:
+                    i += 1
+                    time.sleep(random.uniform(.3, .5))
+                get_asset_digest = response['responses']['GET_ASSET_DIGEST']
+                result = get_asset_digest.get('result', 0)
+                page_offset = get_asset_digest.get('page_offset', 0)
+                page_timestamp = get_asset_digest.get('timestamp_ms', 0)
+                time.sleep(random.uniform(.53, 1.1))
```
friscoMad: ^you are doing the i thing to emulate random sleeps, there is no need to add another one here.

pogom/account.py
friscoMad: Monocle is not sleeping here 😄

pogom/account.py
```diff
@@ -98,8 +103,192 @@ def check_login(args, account, api, position, proxy_url):
...
+    except Exception as e:
+        log.exception('Login for account %s failed. Exception in ' +
+                      'get_player_profile: %s', account['username'], repr(e))
+        raise LoginSequenceFail('Failed during login sequence.')
+
+    # TODO: # 7 - Make a request to get Shop items.
+
+    try:  # 8 - Check if there are level up rewards to claim.
+        request = api.create_request()
+        request.level_up_rewards()
```
friscoMad: Level up rewards is missing the level param

pogom/account.py
```diff
@@ -356,18 +548,56 @@ def encounter_pokemon_request(api, encounter_id, spawnpoint_id, scan_location):
...
         req.check_challenge()
         req.get_hatched_eggs()
-        req.get_inventory()
+        req.get_inventory(last_timestamp_ms=account['last_timestamp_ms'])
         req.check_awarded_badges()
-        req.download_settings()
         req.get_buddy_walked()
-        encounter_result = req.call()
+        req.get_inbox(is_history=True,
+                      is_reverse=False,
```
friscoMad: Only set is_history in all get_inbox

friscoMad:
`Login for account pokeFrasco287 with 31 requests successful.`

Templates and assets do have a huge impact in the number of requests and time to login, as this is done only once per version (and using a stored asset_time) and device, I do think it is not probable that they are using this for flagging otherwise it will flag when you use a new account in your phone without redownloading.
Also doing it everytime can cause flagging, I know that Monocle do this, that's the only thing holding me back, but I would say that we should keep that part at random (1 ot of 10 times?), or setting the defaults timestamps to the values from a week ago (instead of 0) until we have some account persistency to avoid doing it every single time.

Aside from that it seems really nice.

Alderon86:
@jagauthier Accounts are banned :( sorry

Alderon86:
Changes i did not address where intentionally skipped, as every major bot/map is doing exactly this

PR #2093, opened 2017-06-23
[WIP] Player locale based on user's geolocated IP address

ibuckshot5:
This PR allows custom Player Locale based on the user's geolocated IP.

**This is very WIP and feel free to tweak any part of this.**

## Description

I added a new method to pogom/utils, `get_player_locale`. This returns locale for the user's IP, as stated before.

## Motivation and Context

Having a set locale can flag accounts?

## How Has This Been Tested?

On my personal installation

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -11,7 +11,7 @@ from pgoapi import PGoApi
 from pgoapi.exceptions import AuthException
 
 from .fakePogoApi import FakePogoApi
-from .utils import in_radius, generate_device_info, equi_rect_distance
+from .utils import in_radius, generate_device_info, equi_rect_distance, get_player_locale
 from .proxy import get_new_proxy
 
 log = logging.getLogger(__name__)
@@ -20,6 +20,8 @@ log = logging.getLogger(__name__)
 class TooManyLoginAttempts(Exception):
     pass
 
+
+locale = get_player_locale()
 
 # Create the API object that'll be used to scan.
 def setup_api(args, status, account):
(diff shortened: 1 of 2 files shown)


SiteEffect:
Much better than the geopy try of #2057. The problem is that people using proxies or running on a server do not want to use the IP as reference but the search location.

pogom/account.py
```diff
@@ -21,6 +21,8 @@ class TooManyLoginAttempts(Exception):
     pass
 
 
+locale = get_player_locale()
```
sebastienvercammen: Don't make this a global. Memoize `get_player_locale` instead.

pogom/utils.py
```diff
@@ -720,6 +720,19 @@ def get_args():
...
+    locale = {'country': 'US',
+                       'language': 'en',
+                       'timezone': 'America/Denver'}
+    try:
+        res = requests.get(url='http://freegeoip.net/json')
+        dict = json.loads(res.text)
+        locale['country'] = dict['country_code']
+        locale['timezone'] = dict['time_zone']
+    except Exception:
+        pass
```
sebastienvercammen: Don't `pass`: at least log the exception, e.g.:
```
except Exception as e:
    log.exception('Exception while getting player locale: %s.', e)
```
before returning the default value.

pogom/utils.py
```diff
@@ -720,6 +720,19 @@ def get_args():
 
     return args
 
+def get_player_locale():
+    locale = {'country': 'US',
+                       'language': 'en',
+                       'timezone': 'America/Denver'}
+    try:
+        res = requests.get(url='http://freegeoip.net/json')
```
sebastienvercammen: Your request **needs** a timeout, otherwise it can hang indefinitely.

friscoMad:
I do think that geolocation from map coords seem a better solution, it is more complex due to gmaps api but maybe we can use another api to do the same, using the server ip it is not better than using just english.

SiteEffect:
@sebastienvercammen Can be closed due to merged #2057

sebastienvercammen:
Closed because we'll be using #2057 + #2106.

PR #2095, opened 2017-06-26
Fixed Gym Info and added Check Inbox

friscoMad:
## Description
This PR tries to recover the features lost with the api update and add the new check Inbox 

- Added the new Check Inbox method to all requests needding it.
- Recovered gym info feature, adjusting map for the differences

This PR do not add Raids as I would like to merge this fast and then work on raids.

To make it compatible with the new system I adjusted gym level to number of pokes deployed, also slots are correctly calculated based on pokes.

Show open gyms select changed to a switch.
Reduced min and max level filter for gyms to 0-6
Removed prestige bar

## Motivation and Context
API change 

## How Has This Been Tested?
Local instance, not tested gym WH even if I tried to make them work.

rocketmap_2017-06-26_20-10-58

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -336,7 +336,7 @@ def spin_pokestop_request(api, fort, step_location):
         req.check_awarded_badges()
         req.download_settings()
         req.get_buddy_walked()
-        req.request.get_inbox(is_history=True)
+        req.get_inbox(is_history=True)
         spin_pokestop_response = req.call()
 
         return spin_pokestop_response
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -336,6 +336,7 @@ def spin_pokestop_request(api, fort, step_location):
         req.check_awarded_badges()
         req.download_settings()
         req.get_buddy_walked()
+        req.request.get_inbox(is_history=True)
         spin_pokestop_response = req.call()
 
         return spin_pokestop_response
@@ -360,6 +361,7 @@ def encounter_pokemon_request(api, encounter_id, spawnpoint_id, scan_location):
         req.check_awarded_badges()
         req.download_settings()
         req.get_buddy_walked()
+        req.get_inbox(is_history=True)
         encounter_result = req.call()
 
         return encounter_result
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1086,7 +1086,7 @@ def search_worker_thread(args, account_queue, account_sets,
                             # API gets cranky about gyms that are ALMOST 1km
                             # away.)
                             if response['responses'][
-                                    'GET_GYM_DETAILS']['result'] == 2:
+                                    'GYM_GET_INFO']['result'] == 2:
                                 log.warning(
                                     ('Gym @ %f/%f is out of range (%dkm), ' +
                                      'skipping.'),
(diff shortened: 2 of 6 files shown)


pogom/account.py
```diff
@@ -336,6 +336,7 @@ def spin_pokestop_request(api, fort, step_location):
         req.check_awarded_badges()
         req.download_settings()
         req.get_buddy_walked()
+        req.request.get_inbox(is_history=True)
```
sebastienvercammen: A few lines down, you have
```
req.get_inbox(is_history=True)
```

sirlindqvist:
I'll update pr2085 tonight to match your changes.

pogom/search.py
```diff
@@ -1229,25 +1230,26 @@ def gym_request(api, position, gym, api_version):
...
+            player_lng_degrees=f2i(position[1]),
+            gym_lat_degrees=gym['latitude'],
+            gym_lng_degrees=gym['longitude'])
         req.check_challenge()
         req.get_hatched_eggs()
         req.get_inventory()
         req.check_awarded_badges()
         req.download_settings()
         req.get_buddy_walked()
-        x = req.call()
```
Arengo: THANKS!

static/js/map.js
```diff
@@ -1350,32 +1336,8 @@ function processGyms(i, item) {
...
-                closePrestige = 2500
-                break
-            case 4:
-                closePrestige = 5000
-                break
-        }
-
-        if (!gymHasOpenSpot(gymLevel, item.pokemon.length) && (gymPrestige[gymLevel - 1] > closePrestige + item.gym_points || gymLevel === 10)) {
+    if (Store.get('showOpenGymsOnly')) {
+        if (item.pokemon.length === 6) {
```
Arengo: Clean. I like it :-)

pogom/account.py
Arengo: ``req.get_inbox(is_history=True)`` is the correct one. Friscos line was probably a copy-pasta-typo.

friscoMad:
Fixed spin request problem and WH related issue, thanks to Duke for reporting, it should be good to go.

Still not tested in my main instance as I had all my accounts banned.

SiteEffect:
I'll give it a shot and ban all my accounts again, with that login flow :-P

pki0:
- What about the motivation percentage? 
- I'd love to store also the original cp (cp_when_deployed) This would be needful for statistics.
- Good value could be the duration in gym or time of deploy, too.

Or even we could store all infos from API in the database. Then user can decide which values they can show on the map.

I know that this will be more then a "fix", but why not doing it in this push completely?

friscoMad:
@pokemonkiel I have changed to store real cp as this PR wanted to just recover the broken features, Arengo is working on a PR with decaying cp and Raids, I will add deployment time to that also. 
I wanted to keep it simple to get it merged as soon as possible, adding new info means db changes which are more complex to test and we would need UI changes too so that could take a while.

We can add more data to the DB but it comes with a lot of data (such as a log of berry actions) and our mysql are already struggling so if we can cherry pick data I think it is wiser

FrostTheFox:
The issue that Sebby pointed out has been fixed, everything else looks OK to me, we have two approvals from repo maintainers and 3 non-repo maintainers, I'm considering this good to go.

PR #2097, opened 2017-06-26
Do the anti-silly.

sebastienvercammen:
## Description
Handle the jQuery cache-busting timestamp.

## Motivation and Context
Let [these guys](https://github.com/smoo7h/Pokemon_Go_Map_PWN/blob/master/PogoMapPwn/Program.cs#L36) go _rururu_ for a little bit.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -3,6 +3,7 @@
 
 import calendar
 import logging
+import time
 
 from flask import Flask, abort, jsonify, render_template, request,\
     make_response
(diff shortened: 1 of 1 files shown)


FrostTheFox:
👍

pogom/app.py
```diff
@@ -193,6 +194,13 @@ def raw_data(self):
             self.search_control.clear()
         d = {}
 
+        # No, no, no.
+        now_long = int(time.time() * 1000)
+        jquery_cache = int(request.args.get('_', now_long))
+
+        if jquery_cache < now_long - 1000:
```
friscoMad: Increase to 2 secs just in case?

sebastienvercammen:
@friscoMad I agree, this is a trivial implementation that doesn't even rely on anything in RocketMap, it relies on a timestamp added because of jQuery's `.cache`. And even then, it's nowhere near real functionality to stop scrapers.

It was just a fun small thing to do :) I'm closing my own PR now!

PR #2105, opened 2017-07-01
Sanity fixes

Thetacz:
## Description

Some small fixes that prevent errors and future headache.
## Motivation and Context

With all the merges I decided to check code for warnings, this is the result.
## Types of changes

## Checklist:


pogom/account.py
```diff
@@ -670,7 +672,9 @@ def __init__(self, kph):
         self.next_lock = Lock()
 
     # Set manipulation.
-    def create_set(self, name, values=[]):
+    def create_set(self, name, values=None):
```
Arengo: I don't get that. You default to ``None`` to change the default to ``[]`` if it's ``None``, so you could keep ``[]`` in the first place.

Thetacz: Security immutability. If that is none in the first place and than set to list, you cannot change it in the middle.

Unhelpful: Default arguments are evaluated when the function is defined, not each time it is called. In this case, every to create set that doesn't provide the values argument will share, and possibly modify, the same list instance.

Arengo: That actually was helpful. Kappa.

pogom/account.py
```diff
@@ -617,6 +617,8 @@ def parse_download_settings(account, api_response):
     if 'DOWNLOAD_REMOTE_CONFIG_VERSION' in api_response['responses']:
         remote_config = (api_response['responses']
                          .get('DOWNLOAD_REMOTE_CONFIG_VERSION', 0))
+        asset_time = 0
+        template_time = 0
```
Alderon86: Still not needed in my opinion, if it fails to download remote config your account is 95% banned.
so in any other case it will go through and return asset and template time, so we shouldnt set it to 0 if we cant get it.
you will create the next error with that, resulting in failed to get player profile with KeyError last_timestamp_ms

Thetacz: But if we can't get it and it will go throu, you have get an error. So the best way is to throw exception if it remains 0 I guess. I am opened to suggestions...

Alderon86: imho just leave it as it is right now in develop.
Theres no case we *cant* get it and it will go through.
If it goes through we have values, if not account is banned/hashing offline/niantic offline. there are no other options

sebastienvercammen: @Thetacz @Alderon86 Raising an exception ourselves is the best way to go in this case.

pogom/account.py
sebastienvercammen: 👍 Good catch

Alderon86:
`Raising an exception ourselves is the best way to go in this case.`

Im fine with that one.

pogom/account.py
```diff
@@ -172,6 +177,10 @@ def rpc_login_sequence(args, api, account):
 
         total_req += 1
         time.sleep(random.uniform(.53, 1.1))
+    except NullTimeException as e:
+        log.exception('Could not get %s time for Account %s,'
+                      + 'probably banned or Hashing error. Exception: %s.',
+                      e.type, account['username'], e)
```
jagauthier: I'm not seeing a string for exception.

2017-07-11 16:29:43,447 [   search-worker-0][       account][   ERROR] Could not get asset time for Account xyz,probably banned or Hashing error. Exception: .

friscoMad: Yup the same for me and it needs a space after ','

Alderon86: imo a ValueError would fit better.

PR #2106, opened 2017-07-01
Cache Google Geolocator API results in DB

Arengo:
Adding a Google API caching via DB, since the limit is reached quickly.

## Description

Introducing PlayerLocale table to cache the player locales and prevent that the Google API key is loaded too much. This **only** happens in certain cases but we do not want ``America/Denver`` for them, though.

## Motivation and Context

Google API resources are a ****.

## How Has This Been Tested?

On two setups in CLI, config and without anything just using Google API.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -763,20 +763,18 @@ class PlayerLocale(BaseModel):
 
     @staticmethod
     def get_locale(location):
-        query = (PlayerLocale
-                 .select()
-                 .where(PlayerLocale.location == location)
-                 .dicts())
-
         locale = None
-        for l in query:
+        try:
+            query = PlayerLocale.get(PlayerLocale.location == location)
             locale = {
-                'country': l['country'],
-                'language': l['language'],
-                'timezone': l['timezone']
+                'country': query.country,
+                'language': query.language,
+                'timezone': query.timezone
             }
-
-        return locale
+        except PlayerLocale.DoesNotExist:
+            log.debug('This location is not yet in PlayerLocale DB table.')
+        finally:
+            return locale
 
 
 class ScannedLocation(BaseModel):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -755,6 +755,30 @@ class LocationAltitude(BaseModel):
         InsertQuery(cls, rows=[cls.new_loc(loc, altitude)]).upsert().execute()
 
 
+class PlayerLocale(BaseModel):
+    location = Utf8mb4CharField(primary_key=True, max_length=50, index=True)
+    country = Utf8mb4CharField(max_length=2)
+    language = Utf8mb4CharField(max_length=2)
+    timezone = Utf8mb4CharField(max_length=50)
+
+    @staticmethod
+    def get_locale(location):
+        query = (PlayerLocale
+                 .select()
+                 .where(PlayerLocale.location == location)
+                 .dicts())
+
+        locale = None
+        for l in query:
+            locale = {
+                'country': l['country'],
+                'language': l['language'],
+                'timezone': l['timezone']
+            }
+
+        return locale
+
+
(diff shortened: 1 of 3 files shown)


sebastienvercammen:
Rather than setting a manual locale, you can cache the result and re-use it if a location falls within a certain distance.

TheCrabity:
Problem is sometimes the key wont even return a result to get cached. They rate limit and lock you out pretty quickly on these keys, just like the Altitude setting we have. Our workaround for that, was the same as this method.

SiteEffect:
We can save it though in the DB like we do with altitude, scan ``location`` as primary key.

sebastienvercammen:
@Arengo Sounds good. Using SQL to query by distance to a location is also pretty efficient, so that's an added bonus 👍 Just remember not to cache the value when it wasn't retrieved from the Google API.

friscoMad:
Some ideas, move this and version to a new table named settings, we don't want to use denver for all maps but even maps with different instances if using the same db would be using the same player locale.
So why not just storing them as key:value in a settings table (version, locale, country, timezone), that way we avoid having a single entry table that will store a lot of entries if the map just move around, and move towards more interactive configuration if needed in the future.

If we don't like the settings idea then for me the logic should be the other way around, just use the cache if the google request fails then use the last valid value, as it is right now heavily moving the map between different countries will use the old locale forever.

SiteEffect:
@friscoMad why would they use the same player locale? It totally is depended of that instance ``--location``. So instance one is on the one side of a timezone border while instance two is on the other sides. The location differs in a few kilometres and since it's not the same lat/lon pair it has another DB row with the correctly corresponding timezones which are used independently. You can only move the map between different countries via ``--location`` scan on demand which is rarely used. In case of scan on demand in none of the cases it would fetch different timezones currently, since only the scheduler is affected.

Regarding a settings/options table,  that's not a bad idea
```sql
setting   |  value
------------------
version   |  19
country   |  US
```

pogom/models.py
```diff
@@ -755,6 +755,30 @@ def save_altitude(cls, loc, altitude):
...
+    @staticmethod
+    def get_locale(location):
+        query = (PlayerLocale
+                 .select()
+                 .where(PlayerLocale.location == location)
+                 .dicts())
+
+        locale = None
+        for l in query:
+            locale = {
```
sebastienvercammen: Instead of looping, why not limit your query to only one result?

jagauthier:
Pulled this PR in, and gave it a test.  I confirmed that my locale was stored in the DB after a successful google query:
```
MariaDB [rocketmapdb]> select * from playerlocale;
+----------------------+---------+----------+------------------+
| location             | country | language | timezone         |
+----------------------+---------+----------+------------------+
| 39.xxxxxx,-84.xxxxxx | US      | US       | America/New_York |
+----------------------+---------+----------+------------------+
```

I also confirmed that the data was retrieved from the DB by adding a line in models.py to confirm  it:
`2017-07-03 13:10:41,444 [        MainThread][        models][    INFO] Locale retrieved from database.`

I don't have specific review comments, but confirmed that this seems to work as intended on my small test.

pogom/models.py
```diff
@@ -755,6 +755,28 @@ def save_altitude(cls, loc, altitude):
...
+    @staticmethod
+    def get_locale(location):
+        locale = None
+        try:
+            query = PlayerLocale.get(PlayerLocale.location == location)
+            locale = {
+                'country': query.country,
+                'language': query.language,
+                'timezone': query.timezone
+            }
```
sebastienvercammen: If we retrieved a locale from DB, log it to `.debug()` for debugging purposes.

SiteEffect:
@sebastienvercammen Debug logging addressed in ``ac34089``.

PR #2108, opened 2017-07-03
Implement hash key checks

ibuckshot5:
Check if hash keys are working before they are used

## Description

Before hash keys are added to the hash scheduler, check if they are working with a HTTP request to Bossland.

## Motivation and Context

RocketMap can no longer run without a hash key, so it's important to make sure the key provided works first.

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -1174,49 +1174,57 @@ class KeyScheduler(object):
                 }
                 log.debug("Using proxy: {}".format(pr['http']))
 
-            status_code = requests.get('https://pokehash.buddyauth.com', proxies=pr).status_code
+            status_code = requests.get('https://pokehash.buddyauth.com', 
+                                       proxies=pr).status_code
             log.debug("Bossland returned: {}".format(status_code))
             if status_code == 200:
-                log.debug("Bossland returned 200. Check OK. Going to key check")
+                log.debug("Bossland returned 200. Check OK. Start key check")
                 for key in keys:
-                    # TODO: If you find a better way to do the hash key check, please
-                    # say it. It's better for everyone that way.
+                    # TODO: If you find a better way to do the hash key check,
+                    # please say it. It's better for everyone that way.
                     try:
                         # TODO: Translate api-version flag into hash URL
                         r = requests.post(
-                            'https://pokehash.buddyauth.com/api/v137_1/hash', data={
-                            'Timestamp': now(),
-                            'Latitude': 0,
-                            'Longitude': 0,
-                            'Altitude': 0,
-                            'AuthTicket': 'dG90bw==',
-                            'SessionData': 'dG90bw==',
-                            'Requests': []
-                        }, proxies=pr, headers={
-                            'Content-Type': 'application/json',
-                            'X-AuthToken': key
-                        }, timeout=5)
+                            'https://pokehash.buddyauth.com/api/v137_1/hash', 
+                            data={
+                                'Timestamp': now(),
+                                'Latitude': 0,
+                                'Longitude': 0,
+                                'Altitude': 0,
+                                'AuthTicket': 'dG90bw==',
+                                'SessionData': 'dG90bw==',
+                                'Requests': []
+                            }, 
+                            proxies=pr, 
+                            headers={
+                                'Content-Type': 'application/json',
+                                'X-AuthToken': key
+                            }, 
+                            timeout=5)
 
                         if r.status_code == 200:
                             self.keys[key] = {
                                 'remaining': 0,
-                                'maximum': r.headers.get('x-maxrequestcount'),
+                                'maximum': r.headers
+                                                .get('x-maxrequestcount'),
                                 'peak': 0,
-                                'expires': r.headers.get('x-authtokenexpiration')
+                                'expires': r.headers
+                                                .get('x-authtokenexpiration')
                             }
                         elif r.status_code == 401:
-                            log.warning('Hash key "{}" appears invalid or expired,' +
+                            log.warning('Hash key "{}" appears invalid,' +
                                         'not adding into queue.'.format(key))
                         else:
-                            log.error('Invalid HTTP status code received from ' +
+                            log.error('Invalid HTTP status code received on ' +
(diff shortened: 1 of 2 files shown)


SiteEffect:
Fiddled something together to prevent over-stress of hash-servers while constantly checking them.
This is not a HashKey check itself but a availability check of their servers, however I thought it fits.
```py
    retry_time = 0
    while True:
        log.debug("Using proxy: {}".format(status['proxy_url']) )
        status_code = requests.get('https://pokehash.buddyauth.com', proxies=status['proxy_url']).status_code
        log.debug("Bossland returned: {}".format(status_code))
        if status_code == 200:
            log.debug("Bossland returned 200. Check OK.")
            break

        if args.proxy:
            proxy_num, status['proxy_url'] = get_new_proxy(args)
            if args.proxy_display.upper() != 'FULL':
                status['proxy_display'] = proxy_num
            else:
                status['proxy_display'] = status['proxy_url']

        retry_time += 1
        log.warning("Bossland check failed, switching proxies, retrying in {} s.".format(retry_time))
        time.sleep(retry_time)
```

ibuckshot5:
@Arengo makes sense

ibuckshot5:
Alright, implemented that in

pogom/schedulers.py
```diff
@@ -1152,20 +1154,68 @@ def get_scheduler(name, *args, **kwargs):
...
+    def __init__(self, keys, args, db_updates_queue):
         self.keys = {}
-        for key in keys:
-            self.keys[key] = {
-                'remaining': 0,
-                'maximum': 0,
-                'peak': 0,
-                'expires': None
-            }
+        # Thanks Arengo
```
Arengo: Please - don't.

pogom/schedulers.py
```diff
@@ -1152,20 +1154,68 @@ def get_scheduler(name, *args, **kwargs):
...
+        # Check if bossland is available
+        retry_time = 0
+        while True:
+            log.debug("Using proxy: {}".format(status['proxy_url']) )
```
Arengo: You don't have access to ``status`` here.

pogom/schedulers.py
```diff
@@ -1152,20 +1154,68 @@ def get_scheduler(name, *args, **kwargs):
...
+                        else:
+                            log.error('Invalid HTTP status code received from ' +
+                                      'key check: {}. Check if hashing is down.'
+                                      .format(r.status_code))
+                    except requests.Timeout:
+                        log.warning('Hashing check request timed out, adding key ' +
+                                    'to queue anyways.')
+
+            if args.proxy:
+                proxy_num, status['proxy_url'] = get_new_proxy(args)
```
Arengo: Also no ``status`` access here and in the following lines.

SiteEffect:
Check Travis CI for bad lines.

SiteEffect:
@ibuckshot5 Seeing some struggle with Travis CI, think about using Atom.io "IDE" and manually use ``flake8`` from console with ``flake8 pogom`` and ``flake8 runserver.py`` and also ``npm run lint`` or use the Atom extensions. It saves a ton of nerves, especially yours.

ibuckshot5:
@Arengo thanks for the recommendation. Fixed the linting errors

pogom/schedulers.py
```diff
@@ -1152,20 +1154,78 @@ def get_scheduler(name, *args, **kwargs):
         raise NotImplementedError(
             "The requested scheduler has not been implemented")
 
-
```
Arengo: Re-introduce this line and remove the spacing between the follow-up comment and ``class KeyScheduler``. That's also Travis conform :-)

ibuckshot5: Done

Alderon86:
@ibuckshot5  another indent problem, line 1999 has a whitespace left ;)

ibuckshot5:
@Alderon86 fixed

Alderon86:
@ibuckshot5 always make sure to run
```
flake8 --statistics --show-source --disable-noqa
npm run build
npm run lint
```
locally to check for errors, before pushing commits.
Also you may want to read https://www.atlassian.com/git/tutorials/merging-vs-rebasing
just get familiar with how things are done here and you have an easy life

pogom/schedulers.py
```diff
@@ -1156,16 +1158,75 @@ def get_scheduler(name, *args, **kwargs):
...
+                                        'not adding into queue.'.format(key))
+                        else:
+                            log.error('Invalid HTTP status code received:' +
+                                      ' {}. Check if hashing is down.'
+                                      .format(r.status_code))
+                    except requests.Timeout:
+                        log.warning('Hashing check request timed out, ' +
+                                    'adding to queue anyways.')
+
+                    break
```
Arengo: For me it seems that the ``break`` will result in a loop break after the first iteration, always. Is this intended?

friscoMad:
I am not sure if this whole PR make sense, Hash Errors should be checked on client/2060+ or whatever is called and just retry if it is offline/timeout/throttling etc if we also have hash proxy bans then we should add to proxy checking, currenty hashing is quite random so adding a check during init doesn't fix anything

sebastienvercammen:
I agree with @friscoMad: even if I try to consider using it as some kind of general automatic checking, it will use requests of your hashing keys just for the sake of checking and will reduce your available RPM.

Keys and their headers (RPM count) should be used and updated on-the-go instead of on a separate scheduler like we do with proxies.

SiteEffect:
I vote for having a retry cycle for failed hashing requests, while checking for access to the server in before by ping.

Alderon86:
```
I vote for having a retry cycle for failed hashing requests, while checking for access to the server in before by ping.
```
imo this should be done by the api, not rocketmap

sebastienvercammen:
@Arengo @Alderon86 Any long-term solution is outside of the scope of this PR, so for now I'm closing this. It's already on the general roadmap so we'll get there if we keep working on our priority list.

With BossLand's fixed (read: reduced) RPM, this PR is even less of an option.

PR #2110, opened 2017-07-03
Updated and improved api version check

friscoMad:
## Description
This PR needs the latest version of pgoapi, requirements.txt have been changed to reflect that.
(but you can test how it fails to start if you don't update to latest pogoapi).
There was an old api version check that compared to package version 1.1.7 when we have been on 1.2.0 for ages.
Also a new method get_api_version was added to pgoapi to enable this kind of checks, it will stop the instance if it is a scanner instance and the api it is running it is not exactly the same as we are asking for.
This is made on purpouse as new apis are usually not backwards compatible without risk of flagging so I think it is better to tie an api version with an RM version.

This pr has nothing to do with version-check that checks Niantic forced version with RM api version.

## Motivation and Context
Help to identify users running on old apis without noticing.

## How Has This Been Tested?
Local instance, checked against current configured api (fails with a proper error) and latest api (works).
Also used --api-version to do tests with future and previous RM api versions.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -29,9 +29,6 @@ from pogom.webhook import wh_updater
 
 from pogom.proxy import check_proxies, proxies_refresher
 
-# Currently supported pgoapi.
-pgoapi_version = "1.1.7"
-
 # Moved here so logger is configured at load time.
 logging.basicConfig(
     format='%(asctime)s [%(threadName)18s][%(module)14s][%(levelname)8s] ' +
@@ -41,19 +38,11 @@ log = logging.getLogger()
 # Assert pgoapi is installed.
 try:
     import pgoapi
-    from pgoapi import utilities as util
+    from pgoapi import PGoApi, utilities as util
 except ImportError:
     log.critical(
         "It seems `pgoapi` is not installed. Try running " +
         "pip install --upgrade -r requirements.txt.")
-    sys.exit(1)
-
-# Assert pgoapi >= pgoapi_version.
-if (not hasattr(pgoapi, "__version__") or
-        StrictVersion(pgoapi.__version__) < StrictVersion(pgoapi_version)):
-    log.critical(
-        "It seems `pgoapi` is not up-to-date. Try running " +
-        "pip install --upgrade -r requirements.txt again.")
     sys.exit(1)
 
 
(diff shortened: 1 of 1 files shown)


h0us3cat:
Does this fix search control on/off?
https://github.com/RocketMap/RocketMap/issues/2089

runserver.py
```diff
@@ -89,6 +78,35 @@ def handle_exception(exc_type, exc_value, exc_traceback):
         exc_type, exc_value, exc_traceback))
 
 
+def scan_checks(args):
+    # Currently supported pgoapi.
+    pgoapi_version = "1.2.0"
+    api_version_error = ('It seems `pgoapi` is not up-to-date. Try running ' +
+                         '"pip uninstall pgopapi" and then ' +
+                         '"pip install --upgrade -r requirements.txt" again.')
```
sebastienvercammen: Wording:

> The installed `pgoapi` is out of date. Please refer to http://DOCS-LINK/updating-pgoapi to update.

runserver.py
```diff
@@ -89,6 +78,35 @@ def handle_exception(exc_type, exc_value, exc_traceback):
         exc_type, exc_value, exc_traceback))
 
 
+def scan_checks(args):
```
sebastienvercammen: Function name is too generic.

runserver.py
```diff
@@ -89,6 +78,35 @@ def handle_exception(exc_type, exc_value, exc_traceback):
...
+            StrictVersion(pgoapi.__version__) < StrictVersion(pgoapi_version)):
+        log.critical(api_version_error)
+        sys.exit(1)
+
+    # Abort if we don't have a hash key set.
+    if not args.hash_key:
+        log.critical('Hash key is required for scanning. Exiting.')
+        sys.exit()
+
+    # Check pgoapi version against implemented api version.
```
sebastienvercammen: This doesn't check pgoapi version (which is `1.2.0`), it checks what PoGo version the API supports. Reword the comment.

runserver.py
```diff
@@ -89,6 +78,35 @@ def handle_exception(exc_type, exc_value, exc_traceback):
...
+    api = PGoApi()
+    try:
+        if api.get_api_version() != int(args.api_version.replace('.', '0')):
```
sebastienvercammen: `get_api_version` should be a static method in pgoapi so we don't need to create an instance.

runserver.py
```diff
@@ -293,11 +311,8 @@ def main():
     config['GMAPS_KEY'] = args.gmaps_key
 
     if not args.only_server:
-
-        # Abort if we don't have a hash key set
-        if not args.hash_key:
-            log.critical('Hash key is required for scanning. Exiting.')
-            sys.exit()
+        # Check if we are able to scan.
+        scan_checks(args)
```
sebastienvercammen: RE: function name too generic: in general, you would do
```
if can_start_scanning():
    ... do stuff
else:
    ... no bueno.
```
not `if_can_start_scan_do_stuff()`.

sebastienvercammen:
@h0us3cat No, it doesn't fix that.

runserver.py
friscoMad: I found that when I was developing this but nullpixel was really fast with the merge.
 I can PR again to change that but I am not sure when it will be merged.
Also it doesn't seem like a big issue, as the instance would be deleted as soon as we leave the function that's why I just left it as it is.

runserver.py
```diff
@@ -89,6 +78,38 @@ def handle_exception(exc_type, exc_value, exc_traceback):
...
+
+    # Abort if we don't have a hash key set.
+    if not args.hash_key:
+        log.critical('Hash key is required for scanning. Exiting.')
+        return False
+
+    # Check the PoGo api pgoapi implements against what RM is expecting
+    api = PGoApi()
+    try:
+        if api.get_api_version() != int(args.api_version.replace('.', '0')):
```
friscoMad: Will update to static version as soon as: https://github.com/pogodevorg/pgoapi/pull/208 is tested and merged

sebastienvercammen: pgoapi's PR is now merged.

runserver.py
```diff
@@ -89,6 +78,38 @@ def handle_exception(exc_type, exc_value, exc_traceback):
...
+    # Abort if we don't have a hash key set.
+    if not args.hash_key:
+        log.critical('Hash key is required for scanning. Exiting.')
+        return False
+
+    # Check the PoGo api pgoapi implements against what RM is expecting
+    try:
+        if PGoApi.get_api_version() != int(args.api_version.replace('.', '0')):
+            log.critical(api_version_error)
+            sys.exit(1)
```
sebastienvercammen: Missed a `sys.exit(1)` here.

PR #2118, opened 2017-07-06
Reworked proxy tests to use requests_futures instead of custom threads.

sebastienvercammen:
## Description
Consistent with how we send webhook data, proxy tests now use requests_futures' FutureSession for asynchronous requests and a proper worker pool.

FutureSession with urllib3's `Retry` allows for async requests that can automatically be retried, although retrying is disabled for now until we properly implement automatically retrying API requests to allow unstable proxies.

Also cleaned up `proxy.py` and separated proxy loading from proxy checking.

Added config items:
* `-pxre, --proxy-test-retries`: Number of times to retry sending proxy test requests on failure. Default: 0.
* `-pxbf, --proxy-test-backoff-factor`: Factor (in seconds) by which the delay until next retry will increase. Default: 0.25.
* `-pxc, --proxy-test-concurrency`: Async requests pool size. Default: 0 (= automatically set pool size to the number of proxies we have to test).

Tested performance difference is between 10% and 30%.

## Motivation and Context
Readability, consistency, performance.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -127,7 +127,10 @@
 #proxy:                         # Proxy URL e.g. socks5://127.0.0.1:9050 or a list of proxies
                                 # e.g. [socks5://127.0.0.1:9050,socks5://127.0.0.1:9050]
 #proxy-skip-check               # Disable checking of proxies before start.
-#proxy-timeout:                 # Timeout before proceeding with next proxy while checking if the proxy works. (default=5)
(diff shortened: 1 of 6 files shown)


sebastienvercammen:
Received more test reports from Alderon, Arengo, and some other people on Discord. All test results are positive, with all of them showing improvement (on minimum, median and average timing) between 10% and 30%.

I've also renamed all config items to start with `--proxy-test-` rather than `--proxy-` so they are clearer and can't be confused with general proxy settings not related to proxy testing.

PR #2119, opened 2017-07-07
Remove lowest_pokemon_motivation from major webhook updates field

jagauthier:
It is causing a substantial amount of gym webhook updates, that we really do not need.


sebastienvercammen:
Closed, because this is already being fixed in the actual raid PR, and merging this PR would only create conflicts.

PR #2120, opened 2017-07-08
Stop external app by default. Make it opt-in.

sebastienvercammen:
## Description
Someone built a mobile app (a web browser with changed interface) called [ThunderRadar](https://thunderstrike.software/thunderradar/) (unaffiliated to Thunderfox on Discord) that can be used to browse to any RocketMap website.

This change disables the app from visiting/scraping your website (regular browsers will still work) by default. Renaming or removing the file `static/app/settings.json` will re-enable it.

This changes it to act as an opt-in rather than an opt-out, which gives more control to server owners and it's much more in line with the EU's regulations on privacy of data.

## Motivation and Context
Privacy, choice, regulations.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ static/css/custom.css
 !contrib/supervisord/install-reinstall.sh
 !static/data/*
 !static/icons/hsl*.png
+!static/app/settings.json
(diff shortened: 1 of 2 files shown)


123FLO321:
If you want to allow ThunderRadar you should change appAllowed inside the settings.json to true (`"appAllowed": true`) instead of renaming or removing the file.

If the settings.json file is not found, the App will display a popup informing the user that he/she should only continue if he/she is allowed to do so.

PR #2126, opened 2017-07-11
Reworked the logger and added /logs.

Alderon86:
## Description

Removed --very-verbose because loggers argparse has `count`.
Verbosity is now defined with `-v/-vv/-vvv`.
Also added a `logs` directory in which logs are saved in format `log_year_month_day.txt`
Log.txt is created even if no -v flag is set.
`-dfl --disable-file-logs` added for those who dont want logs at all.
This also fixes users error like `where i can find my logs`

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -448,10 +448,14 @@ def get_args():
                               'through these trusted proxies.'))
     parser.add_argument('--api-version', default='0.67.2',
                         help=('API version currently in use.'))
-    parser.add_argument('-v', '--verbose',
+    parser.add_argument('-v',
                         help=('Show debug messages from RocketMap ' +
                               'and pgoapi. Can be repeated up to 3 times.'),
                         action='count', default=0)
+    parser.add_argument('--verbose',
+                        help=('Show debug messages from RocketMap ' +
+                              'and pgoapi.'),
+                        type=int, default=0)
     parser.add_argument('--no-file-logs',
                         help=('Disable logging to files. ' +
                               'Does not disable --access-logs.'),
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -413,7 +413,7 @@ def get_args():
     parser.add_argument('-slt', '--stats-log-timer',
                         help='In log view, list per hr stats every X seconds',
                         type=int, default=0)
-    parser.add_argument('-sn', '--status-name', default=None,
+    parser.add_argument('-sn', '--status-name', default=os.getpid(),
                         help=('Enable status page database update using ' +
                               'STATUS_NAME as main worker name.'))
     parser.add_argument('-spp', '--status-page-password', default=None,
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -453,7 +453,7 @@ def get_args():
                               'and pgoapi. Can be repeated up to 3 times.'),
                         action='count', default=0)
     parser.add_argument('--no-file-logs',
-                        help=('Disable logging to files.' +
+                        help=('Disable logging to files. ' +
                               'Does not disable --access-logs.'),
                         action='store_true', default=False)
     parser.add_argument('--log-path',
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -452,7 +452,7 @@ def get_args():
                         help=('Show debug messages from RocketMap ' +
                               'and pgoapi. Can be repeated up to 3 times.'),
                         action='count', default=0)
-    parser.add_argument('--verbose',
+    parser.add_argument('--verbosity',
                         help=('Show debug messages from RocketMap ' +
                               'and pgoapi.'),
                         type=int, default=0)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -459,6 +459,7 @@ def get_args():
     parser.add_argument('--log-path',
                         help=('Defines directory to save log files to.'),
                         default='logs/')
+    parser.set_defaults(DEBUG=False)
 
     args = parser.parse_args()
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -448,20 +448,13 @@ def get_args():
                               'through these trusted proxies.'))
     parser.add_argument('--api-version', default='0.67.2',
                         help=('API version currently in use.'))
-    verbosity = parser.add_mutually_exclusive_group()
-    verbosity.add_argument('-v', '--verbose',
-                           help=('Show debug messages from RocketMap ' +
-                                 'and pgoapi. Optionally specify file ' +
-                                 'to log to.'),
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    verbosity.add_argument('-vv', '--very-verbose',
-                           help=('Like verbose, but show debug messages ' +
-                                 'from all modules as well.  Optionally ' +
-                                 'specify file to log to.'),
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    parser.set_defaults(DEBUG=False)
+    parser.add_argument('-v', '--verbose',
+                        help=('Show debug messages from RocketMap ' +
+                              'and pgoapi. Can be repeated up to 3 times.'),
+                        action='count', default=0)
+    parser.add_argument('-dfl', '--disable-file-logs',
+                        help=('Disables file logging completley.'),
+                        action='store_true', default=False)
 
     args = parser.parse_args()
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -452,9 +452,13 @@ def get_args():
                         help=('Show debug messages from RocketMap ' +
                               'and pgoapi. Can be repeated up to 3 times.'),
                         action='count', default=0)
-    parser.add_argument('-dfl', '--disable-file-logs',
-                        help=('Disables file logging completley.'),
+    parser.add_argument('--no-file-logs',
+                        help=('Disable logging to files.' +
+                              'Does not disable --access-logs.'),
                         action='store_true', default=False)
+    parser.add_argument('--log-path',
+                        help=('Defines Directory to save Debug log files to.'),
+                        action='store_true', default='logs/')
 
     args = parser.parse_args()
 
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -452,7 +452,7 @@ def get_args():
     verbose.add_argument('-v',
                          help=('Show debug messages from RocketMap ' +
                                'and pgoapi. Can be repeated up to 3 times.'),
-                         action='count', default=0)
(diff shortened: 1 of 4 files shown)


pogom/utils.py
```diff
@@ -434,20 +434,13 @@ def get_args():
...
-                                 'from all modules as well.  Optionally ' +
-                                 'specify file to log to.'),
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    parser.set_defaults(DEBUG=False)
+    parser.add_argument('-v', '--verbose',
+                        help=('Show debug messages from RocketMap ' +
+                              'and pgoapi. Can be repeated up to 3 times.'),
+                        action='count', default=0)
+    parser.add_argument('-dfl', '--disable-file-logs',
```
jagauthier: Would change this to --no-file-logs to be consistent with the other negation options.

runserver.py
```diff
@@ -29,11 +29,28 @@
...
+if args.verbose:
+    log.setLevel(logging.DEBUG)
+else:
+    log.setLevel(logging.INFO)
+
+# Always write to log file.
+if not os.path.exists('logs'):
+    os.mkdir('logs')
+if not args.disable_file_logs:
+    filelog = logging.FileHandler(strftime('logs/log_%Y_%m_%d.txt'))
```
jagauthier: Sorry, I don't like the forced log file name.  I set each of my instances to specific log files so I can monitor them specifically.

pogom/utils.py
Alderon86: conflicts with -no-fixed-location

runserver.py
Alderon86: we could maybe add the instances name?

pogom/utils.py
sebastienvercammen: @Alderon86 Just remove the shorthand version and only keep `--no-file-logs`. Not everything needs a short version.

pogom/utils.py
```diff
@@ -434,20 +434,13 @@ def get_args():
...
+                        help=('Disables file logging completley.'),
```
sebastienvercammen: Typo: completely.

Also not entirely true, considering `--access-logs` is an individual flag and will still work even if `--no-file-logs` is enabled. It would be more correct to say "Disable logging to files. Does not disable --access-logs.".

runserver.py
jagauthier: What about letting the user specify, and if not default to the proposed log name above?
--log-file whatever.txt

pogom/utils.py
jagauthier: --no-logs
--no-output-logs
--no-debug-logs
or something :)

runserver.py
sebastienvercammen: @jagauthier @Alderon86 To fix, add `--log-filename-format`, which is something like `%Y%m%d%s_INSTANCE-NAME.log` by default. Also get all of this stuff out of the global scope... I'm not sure why you moved it there, but never pollute the global scope.

Notice that I added `%s` for seconds since multiple instances would try to write to the same log file if you open them on the same day with your format.

runserver.py
```diff
@@ -160,50 +177,32 @@ def main():
...
-    if args.very_verbose:
-        logging.getLogger('peewee').setLevel(logging.DEBUG)
-        logging.getLogger('requests').setLevel(logging.DEBUG)
+    elif args.verbose == 2:
         logging.getLogger('pgoapi.pgoapi').setLevel(logging.DEBUG)
-        logging.getLogger('pgoapi.rpc_api').setLevel(logging.DEBUG)
         logging.getLogger('rpc_api').setLevel(logging.DEBUG)
+    elif args.verbose >= 3:
+        logging.getLogger('pgoapi.rpc_api').setLevel(logging.DEBUG)
+        logging.getLogger('requests').setLevel(logging.DEBUG)
```
sebastienvercammen: You think requests is not important? What about webhooks or proxy testing? Imo, something like this makes more sense:
```
1: General logging.
2: pgoapi, requests, peewee
3+: werkzeug, rpc_api, pgoapi.rpc_api
```

friscoMad: I vote for moving peewee to 3, it is really noisy.

pogom/utils.py
```diff
@@ -438,9 +438,13 @@ def get_args():
...
                         action='count', default=0)
-    parser.add_argument('-dfl', '--disable-file-logs',
-                        help=('Disables file logging completley.'),
+    parser.add_argument('--no-file-logs',
+                        help=('Disable logging to files.' +
+                              'Does not disable --access-logs.'),
                         action='store_true', default=False)
+    parser.add_argument('--log-path',
+                        help=('Defines Directory to save Debug log files to.'),
+                        action='store_true', default='logs/')
```
sebastienvercammen: This is not an `action='store_true'`.

pogom/utils.py
```diff
@@ -438,9 +438,13 @@ def get_args():
...
                               'and pgoapi. Can be repeated up to 3 times.'),
                         action='count', default=0)
-    parser.add_argument('-dfl', '--disable-file-logs',
-                        help=('Disables file logging completley.'),
+    parser.add_argument('--no-file-logs',
+                        help=('Disable logging to files.' +
+                              'Does not disable --access-logs.'),
                         action='store_true', default=False)
+    parser.add_argument('--log-path',
+                        help=('Defines Directory to save Debug log files to.'),
```
sebastienvercammen: Lowercase: directory. Remove the word "Debug".

runserver.py
```diff
@@ -186,6 +171,24 @@ def main():
     config['parse_gyms'] = not args.no_gyms
     config['parse_raids'] = not args.no_raids
 
+    # Always write to log file.
+    if not args.no_file_logs:
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
+        date = strftime('%Y_%m_%d_%H_%M_')
+        filename = os.path.join(
+            args.log_path, '{}_{}.log'.format(date, args.status_name))
```
sebastienvercammen: You have an underscore at the end of date and you're appending it to another underscore:
```
"%Y_%m_%d_%H_%M_" + "_" + status_name + ".log"
```

runserver.py
```diff
@@ -186,6 +171,24 @@ def main():
     config['parse_gyms'] = not args.no_gyms
     config['parse_raids'] = not args.no_raids
 
+    # Always write to log file.
+    if not args.no_file_logs:
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
+        date = strftime('%Y_%m_%d_%H_%M_')
```
sebastienvercammen: I personally wouldn't add any underscores at all, but at the very least I would reduce the amount:
```
date = strftime('%Y%m%d_%H%M')
```

runserver.py
sebastienvercammen: @friscoMad peewee to 3 is good by me.

pogom/utils.py
```diff
@@ -434,20 +434,17 @@ def get_args():
...
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    parser.set_defaults(DEBUG=False)
+    parser.add_argument('-v', '--verbose',
+                        help=('Show debug messages from RocketMap ' +
+                              'and pgoapi. Can be repeated up to 3 times.'),
+                        action='count', default=0)
+    parser.add_argument('--no-file-logs',
+                        help=('Disable logging to files.' +
+                              'Does not disable --access-logs.'),
```
sebastienvercammen: Missing space.

runserver.py
```diff
@@ -160,50 +162,50 @@ def main():
...
 
-    config['parse_pokemon'] = not args.no_pokemon
-    config['parse_pokestops'] = not args.no_pokestops
-    config['parse_gyms'] = not args.no_gyms
-    config['parse_raids'] = not args.no_raids
-
     # Turn these back up if debugging.
-    if args.verbose or args.very_verbose:
+    if args.verbose == 1:
         logging.getLogger('pgoapi').setLevel(logging.DEBUG)
```
sebastienvercammen: Just a thought, but maybe log level 1 should be RocketMap logging only? Since we only have `pgoapi` left at level one, we might as well move it to level two and use `-v` for RocketMap logging, and anything more for additional verbosity.

jagauthier:
You should have a different -sn per instance. If you do not, then IMO, you are doing it wrong :)

Alderon86:
@jaake77 what @jagauthier said :D

sebastienvercammen:
@Alderon86 What about people who don't use `-sn` at all? We should have a fallback: no name = randomly generated ID. Feel free to go with something like `random.getrandbits(32)`.

friscoMad:
If you don't have -sn wouldn't be OK to just skip it?

sebastienvercammen:
@friscoMad No, because starting multiple instances in the same minute would make them try to get a file lock on the same file (which won't work, and will crash).

As suggested on Discord, using the pid instead of a random number is an even better way.

pogom/utils.py
```diff
@@ -448,20 +448,17 @@ def get_args():
...
-                                 'to log to.'),
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    verbosity.add_argument('-vv', '--very-verbose',
-                           help=('Like verbose, but show debug messages ' +
-                                 'from all modules as well.  Optionally ' +
-                                 'specify file to log to.'),
-                           nargs='?', const='nofile', default=False,
-                           metavar='filename.log')
-    parser.set_defaults(DEBUG=False)
```
friscoMad: not sure if removing this line was intentional

runserver.py
```diff
@@ -184,50 +185,49 @@ def main():
...
-    if args.verbose and args.verbose != 'nofile':
-        filelog = logging.FileHandler(args.verbose)
-        filelog.setFormatter(logging.Formatter(
-            '%(asctime)s [%(threadName)16s][%(module)14s][%(levelname)8s] ' +
-            '%(message)s'))
-        logging.getLogger('').addHandler(filelog)
-    if args.very_verbose and args.very_verbose != 'nofile':
-        filelog = logging.FileHandler(args.very_verbose)
+    # Let's not forget to run Grunt / Only needed when running with webserver.
+    if not args.no_server and not validate_assets(args):
```
friscoMad: This should go after logging is properly configured as it needs to log errors.

friscoMad:
And just one general comment, I am ok as it is now but it will be better if all logging setup is refactored as a new function, we should start breaking down those > 100  lines long functions.

sebastienvercammen:
@friscoMad Agreed: a mutually exclusive group that stores their values in the same attribute name (with `dest`). That way you also don't need to write any manual parsing code.

PR #2133, opened 2017-07-12
Raid Filters

ClarkyKent:
## Description
This PR adds two filter for Raid alerts on map

- Option to show raids or not

- Option to choose min level of raid to show on Map

## Motivation and Context
Because we need it?

## How Has This Been Tested?
On my local Map setup.

screenshot_20170718-140305

screenshot_20170718-140337 screenshot_20170718-140341

screenshot_20170718-140346 screenshot_20170718-140454

## Types of changes

## Checklist:


elincognito:
Can you add screenshots

SiteEffect:
How about an own raid-toggle? Combining with gym info windows when gyms are enabled, too. Otherwise showing just raids or just gyms when no raids are enabled but gyms are?

ClarkyKent:
@Arengo Ah.. yes.. did not think about no-raid argument in config.. i will modify code accordingly.

sebastienvercammen:
Needs a rebase since the last merge wave. Think you can rebase, @ClarkyKent?

343334:
once this gets merged, a spinoff pr that might be useful would be adding in --gdelay option for those who are not updating gyms super fast, as raids give 2 hour notification so really dont need to scan under 2 hours in time, but after this pr, when you enable active raids only - it wont be 100% accurate as it really depends on last_scanned time vs current time so a --gdelay option could let the user tell their frontend map everything is delayed by this amount of time so whats truely active is correct. And maybe add a bounce option like pokemon notification, so active raids are bouncing on map if viewing all.

sebastienvercammen:
@343334 If I understood correctly, you want to delay gym rescanning and show a message on the front-end. This PR addresses front-end fixes unrelated to the back-end's scheduler.

The idea has been noted, but this isn't the place for it.

343334:
not delay. but like if i look at my map now and were to just choose an option of active raids, there would be some eggs t hat should be showing as active but wont since they werent recently scanned to display it, so i guess active raids should take into effect the scanned date vs current time vs how long it takes for the user to update all gyms etc.  but yes thanks for the notation not wanting to change this pr a simple select last_scanned from gym order by last_scanned desc;  if the user has no stale gyms in db, it should display the oldest scan compare to current time and can determine that. so shouldnt need to touch a scheduler or im thinking too much and just do a start time from raid vs currenttime and if start time is past currentime make it active.

houkop:
can we hide gyms without raids?

ClarkyKent:
@houkop no, there is no option for that.
For the moment it can hide Raids to show only gyms (even with Raids, ti will simply replace raids icons to normal gym) and you can set min level of raid to show them on map.

richardfoltin:
the purpose of this feature should be hiding every gym (+ possible low level raids) that the user is not interested in.

ClarkyKent:
@richardfoltin  and @houkop 
So you want that when user enables "Show Raids" toggle button, it will show only raids but not other gyms?
Or showing normal gyms with raid gyms, is something needed in this feature, as well (which this PR does already)?

z0n:
It would also be nice to have an option to only display raids that are already running. Not sure if this is in the scope of this PR though.

ClarkyKent:
I will add the features suggested by @richardfoltin and @z0n by tomorrow.. I can't test them today as raids will end soon here.

ClarkyKent:
I will add the features suggested by @richardfoltin and @z0n by tomorrow.. I can't test them today as raids will end soon here.

sebastienvercammen:
@ClarkyKent Perfect, I look forward to those changes. :) You can expect us to pick up your PR immediately after you've pushed the changes.

richardfoltin:
@ClarkyKent To clarify, I could imagine two different approaches:

1.
We have two separate toggle buttons: Show Gyms and Show Raids. (both of them are always visible) If I switch 'show gyms' on then I will only see gyms (without raids). If I switch 'show raids' on  then I will see only those gyms that have raid.

2.
There is a toggle button that only appears if Show Gyms is on, but it's named something like 'Hide gyms without Raid'. If I switch it on then I will see only those gyms that have raid.

Also, (whatever the approach will be) there could be another raid filter (added to the level filter) that has 3 options: ongoing raids, eggs, both

tomballgithub:
Why not have 2 buttons:  
1) SHOW GYMS:  Turns on gyms just like the normal feature has always worked
2) SHOW RAIDS:  Turns on RAIDS/EGGS regardless of SHOW GYMS settings. 

If both are on, you get GYMS and RAIDS/EGGS.  If you only have SHOW RAIDS then you don't see normal gyms.  If you only have SHOW GYMS then you don't see RAIDS/EGGs.

Seems simple to me...

SiteEffect:
This PR is pretty important right now, so I am looking forward to review it asap you push the desired changes to it. If you need any help feel free to ask in Discord, we are happy to help.

ClarkyKent:
@Arengo  I did some changes, according to comments i read here yesterday. 
I really would like if you can review it and give your feedback.

templates/map.html
```diff
@@ -87,8 +87,40 @@
                 </label>
               </div>
             </div>
-              {% endif %}
+              {% endif %}
```
Arengo: Trailing whitespaces, also below. Think about using Atom.io for example as easy to use IDE.

static/js/map.common.js
```diff
@@ -874,6 +874,18 @@ var StoreOptions = {
         default: false,
         type: StoreTypes.Boolean
     },
+    'showRaidGyms': {
```
Arengo: Rename it to make it more clear what it does: enable Raid by default?!

static/js/map.common.js
```diff
@@ -874,6 +874,18 @@ var StoreOptions = {
...
+        default: true,
+        type: StoreTypes.Boolean
+    },
+    'showActiveRaidsOnly': {
+        default: true,
+        type: StoreTypes.Boolean
+    },
+    'showRaidMinLevel': {
+        default: 0,
```
Arengo: Change default to ``1`` as logical value, this might prevent case issues in ``map.js`` in the future.

static/js/map.js
```diff
@@ -9,6 +9,9 @@ var $textPerfectionNotify
 var $selectStyle
 var $selectIconSize
 var $switchOpenGymsOnly
+var $switchRaidGymsOnly
```
Arengo: ``RaidOnly`` would be more clear, compare with my ``map.common.js`` comment.

static/js/map.js
```diff
@@ -472,7 +479,7 @@ function pokemonLabel(item) {
                 <span class='pokemon'>Level: </span><span class='pokemon'>${pokemonLevel}</span>
                 <span class='pokemon links exclude'><a href='javascript:excludePokemon(${id})'>Exclude</a></span>
                 <span class='pokemon links notify'><a href='javascript:notifyAboutPokemon(${id})'>Notify</a></span>
-                <span class='pokemon links remove'><a href='javascript:notifyAboutPokemon(${id})'>Remove</a></span>
+                <span class='pokemon links remove'><a href='javascript:removePokemonMarker("${encounterId}")'>Remove</a></span>
```
Arengo: First, rebase this PR to reflect newest fix on develop to this line. Also, you are correctly fixing another issue: It should only remove one marker and not all Pokémon 👍

static/js/map.js
```diff
@@ -504,7 +511,7 @@ function pokemonLabel(item) {
             <span class='pokemon'>Level: </span><span class='pokemon no-encounter'>n/a</span>
             <span class='pokemon links exclude'><a href='javascript:excludePokemon(${id})'>Exclude</a></span>
             <span class='pokemon links notify'><a href='javascript:notifyAboutPokemon(${id})'>Notify</a></span>
-            <span class='pokemon links remove'><a href='javascript:notifyAboutPokemon(${id})'>Remove</a></span>
+            <span class='pokemon links remove'><a href='javascript:removePokemonMarker("${encounterId}")'>Remove</a></span>
```
Arengo: See Comment to L#482.

static/js/map.js
```diff
@@ -817,6 +824,7 @@ function isRangeActive(map) {
     return Store.get('showRanges')
 }
 
+
```
Arengo: Remove additional line break.

static/js/map.js
```diff
@@ -1490,6 +1499,26 @@ function processGyms(i, item) {
...
+    if (Store.get('showRaidGyms')) {
+        if (raidLevel === 0) {
+            removeGymFromMap(item['gym_id'])
+            return true
+        }
+    }
+
+    if (Store.get('showRaidGyms') && Store.get('showActiveRaidsOnly')) {
+        if (item.raid.pokemon_id === null) {
```
Arengo: An old raid also has a ``pokemon_id`` but won't get removed from Map by this line. So you have to check against the ``end`` time, too.

templates/map.html
```diff
@@ -87,8 +87,40 @@
                 </label>
               </div>
             </div>
-              {% endif %}
+              {% endif %} 
             <div id="gyms-filter-wrapper" style="display:none">
+              {% if show.raids %}
+              <div class="form-control switch-container" id="raid-gym-wrapper">
+                <h3>Show Raids Only</h3>
```
Arengo: Remove the ``Show``, also for the active raids string.

templates/map.html
```diff
@@ -87,8 +87,40 @@
...
+                <div class="onoffswitch">
+                  <input id="raid-gym-switch" type="checkbox" name="raid-gym-switch" class="onoffswitch-checkbox" checked>
+                  <label class="onoffswitch-label" for="raid-gym-switch">
+                  <span class="switch-label" data-on="On" data-off="Off"></span>
+                  <span class="switch-handle"></span>
+                  </label>
+                </div>
+              </div>
+              <div class="form-control switch-container" id="raid-active-gym-wrapper">
+                <h3>Show Active Raids</h3>
```
Arengo: Can be placed under a wrapper, only accessible when ``Raids Only`` is set, since with enabling this option all normal Gyms will vanish too - so the ``Raids Only`` is implicitly true either way.

static/js/map.js
ClarkyKent: This is not my modification, i think i got them after reabse.

static/js/map.js
```diff
@@ -1470,6 +1477,7 @@ function updatePokestops() {
 
 function processGyms(i, item) {
     var gymLevel = getGymLevel(item)
+    var raidLevel = item.raid.level
```
friscoMad: There is no guaranteed raid in each gym, with time it would be but a clean map will have no raids stored and this will fail.

static/js/map.js
```diff
@@ -1491,13 +1499,33 @@ function processGyms(i, item) {
...
+    if (Store.get('showRaidsOnly')) {
+        if (raidLevel === 0) {
+            removeGymFromMap(item['gym_id'])
+            return true
+        }
+    }
+
+    if (Store.get('showRaidsOnly') && Store.get('showActiveRaidsOnly')) {
+        var now = new Date()
+        if (item.raid.pokemon_id === null || item.raid.end < now.getTime()) {
```
friscoMad: There is a function called isOngoingRaid that you shold be using here

static/js/map.js
```diff
@@ -1491,13 +1499,33 @@ function processGyms(i, item) {
...
+        return true
+    }
+
     if (Store.get('showTeamGymsOnly') && Store.get('showTeamGymsOnly') !== item.team_id) {
         removeGymFromMap(item['gym_id'])
         return true
     }
 
     if (Store.get('showLastUpdatedGymsOnly')) {
-        var now = new Date()
```
friscoMad: why removing this?

static/js/map.js
ClarkyKent: Ok, didn't see that function.. I will replace it.

static/js/map.js
ClarkyKent: Because as you can see I used this variable in L#1511. I had lint error otherwise.

friscoMad: Change var for const and leave the 2 lines, var adds it to the global scope that's why it is complaining, but we need to get the current date when the function is called.

sebastienvercammen: @friscoMad `var` is local scope, it would be in the function's scope, not the global scope.

The proper fix for block scoping is `const` or `let`, which will limit the scope to the nearest block (in this case, the `if`).

`const` should be used when the identifier shouldn't be re-assigned, and `let` when you do want to re-assign.

ClarkyKent: As I will put `isOngoingRaid` function at L#1511. I don't need now variable anymore.
So I can put this one as it was `var now` or should I change it to `let`?

sirlindqvist:
Tried this now and it's good you can toggle active only when showing raids only. However I think it should have an option to show gyms that's not raided in combination with above parameters. Like:

Show raids (will show both gyms and raids)
Show raids only
Active only

Now it's not possible for example to show gyms and/or level 4 raids.

sirlindqvist:
Forget about it.. it works just like that :)

Good job!

urinella:
Your first screenshot:
- Only the Gyms-Switch active and I can see Eggs? 

image

Your last screenshot:
- Gyms-Switch is active but no gyms are shown? 

image

Seem buggy and not logical for me, sorry.
Also: Dropdown menu on wrong place? Gym-switch is over Raid-switch but Gym-dropdown below Raid-switch?

Why not make it more simple?:

Gym-switch are gyms:
- No eggs
- No bosses
- just gyms like it ever was

Raid-switch are raids:
- No gyms
- Show eggs
- Show bosses
- Open menu with: minimum raidlevel and show boss only

If both are switched on:
- Show gyms
- Show raids
- Show both drop-down options in correct order

sirlindqvist:
Maybe I don't get this but I think theres one "level" missed? When raid clicked it shouldn't exclude normal gyms.

In my opinion it should look like this:

raid1

This will show raids only and exclude normal gyms.

raid2

This will show raids AND normal gyms. If lv3 chosen as minimum it will show all normal gyms and all lv3 and 4 upcoming and ongoing raids.

Could also be an extra option with "active only" which would show normal gyms and ongoing raids and NOT any upcoming.

sirlindqvist:
I think we want same thing @urinella? Just that we describe it in different ways?

sirlindqvist:
Maybe better to create a new submenu in nav called "Gyms" ?

And move the toggle "Gym" in there? On=show gyms, off=dont show gyms.

Under here "Raids"

Show raids only
Show active only
Level filter

And those toggles won't affect "Gyms" at all since if you don't want standard gyms turn that off.

urinella:
@sirlindqvist I don't think exactly the same.

I just want to suggest that it would be better to separate **gyms** and **raids**. This is what @tomballgithub already said days ago.

In the author's screenshots the shown things on map are incorrect and not logical:
- Switches are on but not showing the content (Gyms in this case)
- dropdown-menus are in wrong order

Hard to describe it. I made an edit on my previous post. Maybe now more understandable.

friscoMad:
pokemad_2017-07-17_11-54-24
overwolf quick launcher_2017-07-17_11-55-09

I think that somethign like this is ideal, but what I think we should do is that if gyms is marked then all gyms should be shown always, if raid (or active only active) is disabled just showing the default icon with level.

sirlindqvist:
raid_1
Will show all standard gyms and all raids, both upcoming and ongoing.

raid_2
Will show all standard gyms and ongoing but exclude upcoming.

raid_3
Will not show standard gyms but all raids, both upcoming and ongoing.

raid_4
Will not show standard gyms but ongoing and exclude upcoming.

This can be combined with minimum level. Either with one dropdown chasing minimum level for both upcoming and ongoing or two dropdown to be able to choose different level for upcoming and ongoing.

friscoMad:
Yup except for the second case, for me that should show all gyms with default icon except for active raids that would show the active raid icon.

sirlindqvist:
Don't get it @friscoMad :)

Case 1 will show all raids both ongoing and upcoming and Case 2 will show only ongoing.

ClarkyKent:
For me raids were the part of gyms. That is why they were under Gyms switch.

Example of @friscoMad looks more clear and I think more or less everyone describing same thing. 
If everyone agree, i can make changes according to that.

urinella:
@sirlindqvist You are missing one option: Gyms on, Raid off
-> Show only all gyms like it was before the raid update (No eggs, No raids)

sirlindqvist:
@urinella if using that gyms that has an upcoming raid won't show at all. Those gyms are still deplyable and considered "standard" in one way. Let's say you go out and play for taking down gyms. Gyms with upcoming raid won't show then.

sirlindqvist:
Well could be possible to that is way more complicated and needs extra work to map.js because markers for upcoming raids is fetched. Also there's timer till raid and you will need to rework map.js to display a standard gym marker instead and remove the timer if those upcoming raid gyms should be treated as standard instead.

sirlindqvist:
@ClarkyKent got time to fix and push new commit? If so I can test today.

ClarkyKent:
@sirlindqvist I will try to commit today, but can't promise.

sirlindqvist:
Would be awesome! Looking forward to it.

sirlindqvist:
Hmmm now it works totally different and not at all in the way we've been discussing today :(

urinella:
Tested latest update and I have a few gyms with wrong marker now. Some are fine and some are incorrect.

bildschirmfoto 2017-07-17 um 22 03 07

sirlindqvist:
Coz all gyms no matter if upcoming raid, ongoing raid are treated as standard gyms.

ClarkyKent:
@sirlindqvist let's discuss on discord.. to finalize it. What's your discord pseudo?

sebastienvercammen:
@ClarkyKent It's "sirlindqvist" on Discord as well. He's in #pr.

static/js/map.js
```diff
@@ -1490,6 +1508,26 @@ function processGyms(i, item) {
...
+            removeGymFromMap(item['gym_id'])
+            return true
+        }
+    }
+
+    if (!Store.get('showGyms') && Store.get('showRaidMinLevel') && raidLevel < Store.get('showRaidMinLevel')) {
+        removeGymFromMap(item['gym_id'])
+        return true
+    }
+
```
friscoMad: I would group all those 3 inside a big if of !Store.get('showGyms')

templates/map.html
```diff
@@ -87,7 +131,7 @@
               </div>
             </div>
               {% endif %}
-            <div id="gyms-filter-wrapper" style="display:none">
+            <div id="gyms-filter-wrapper" style="display:none">
```
friscoMad: Delete empty spaces

templates/map.html
```diff
@@ -143,7 +187,7 @@
                 </select>
               </div>
             </div>
-              {% endif %}
+              {% endif %}
```
friscoMad: Delete empty

holapex:
This seems to be the cleanest set up to filter raids and gyms. 
fiklter

sirlindqvist:
Much better! I'll help you later with what we talked about. This will fix issue #2157 as well.

343334:
as how we have item.raid.end > Date.now()  is there a way to also add in a clause of item.raid.start < (Date.now() + 60min) so any gyms that werent recently scanned would be removed if their raid start time was over an hour from now as it would already be ended due to the delay in some maps updating gyms.

jagauthier:
Functionally, this works.  It's definitely needed.  My only concern is that when I have both gyms and raids enabled, then disable gyms all gyms disappear.  After 1-3 second pause, then raids show up as expected.

static/js/map.js
```diff
@@ -592,7 +610,7 @@ function gymLabel(gym, includeMembers = true) {
         </div>`
     }
 
-    if (isUpcomingRaid || isRaidStarted) {
+    if ((isUpcomingRaid || isRaidStarted) && isRaidFilterOn && isGymSatisfiesRaidMinMaxFilter(raid) && !isActiveRaidFilterOn) {
```
richardfoltin: This addition is unnecessary and causes a bug:
If only active raids are on, and an egg just turned into an active raid a few minutes ago, but the scanner hasn't scanned the gym again -> the label does not show the time left on the raid. (and if I turn active raids off, it shows)

static/js/map.js
```diff
@@ -666,7 +684,7 @@ function gymLabel(gym, includeMembers = true) {
         </div>`
 
 
-    if (!isRaidStarted && includeMembers) {
+    if (includeMembers) {
```
richardfoltin: I dont think this is necessary either

static/js/map.js
```diff
@@ -1010,13 +1036,14 @@ function setupGymMarker(item) {
 }
 
 function updateGymMarker(item, marker) {
-    if (item.raid !== null && item.raid.end > Date.now() && item.raid.pokemon_id !== null) {
+    let raidLevel = getRaidLevel(item.raid)
+    if (item.raid !== null && item.raid.end > Date.now() && item.raid.pokemon_id !== null && Store.get('showRaids') && raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
```
richardfoltin: unnecessary, because of same reasons

static/js/map.js
```diff
@@ -1010,13 +1036,14 @@ function setupGymMarker(item) {
...
         marker.setIcon({
-            url: 'static/images/raid/' + item.raid.pokemon_id + '.png',
+            url: 'static/images/raid/' + item['raid']['pokemon_id'] + '.png',
             scaledSize: new google.maps.Size(48, 48)
         })
         marker.setZIndex(google.maps.Marker.MAX_ZINDEX + 1)
-    } else if (item.raid !== null && item.raid.end > Date.now()) {
+    } else if (item.raid !== null && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') && raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
```
richardfoltin: unnecessary because of same reasons

static/js/map.js
ClarkyKent: This was necessary to fix issue #2157

static/js/map.js
ClarkyKent: Why? Can you explain the reason?

static/js/map.js
richardfoltin: OK. So this is a bugfix

holapex:
Legionaries are 2 days from now, this feature would be great to have before then.

static/js/map.js
richardfoltin: I suppose it does not cause bugs, it's just cleaner coding. The filtering happens before calling updateGymMarker function, so we don't need another check.

updateGymMarker function's purpuse is to decide which icon to use on the map. The icon should not depend on filtering settings.

ClarkyKent: Yea.. sorry if it was not clear. But we have two icons (for raids and for gyms). So in case raid filter is not enable, this line of code replace them with normal gym image and vice versa.

richardfoltin: I dont know about the others but I would not need that. More information is always better I think

343334:
just tested and if you turn off gyms, and leave raids enabled everything goes blank for about 3 secs than both come back so looks like in current status you cannot see just raids as enabling raids also shows gyms whether enabled or not.

tomballgithub:
It is the middle of the night locally as I type this so there are no RAIDs going on.  Just loaded this PR to test.  When I selected GYMS OFF but RAIDS ON I see the following which is a bunch of non-RAID gyms.  I expected to see no gyms at all since it is the middle of the night.  The pattern seems to be that these are gyms that have not been scanned in a while because I turned those areas off after the hash key RPM $ increases.   Why are they still showing since I have non-RAID gyms turned off?   

Note.  I have another map for which no areas are turned off for scanning and it worked as expected, no gyms are showing at this time.

image

That said, here is another that is not nearly as old although 18 hrs ago would have still put the last time scanned in at about 8am.

image

houkop:
Gyms with legends are not showing on tha map... ( I've modified db to check it)
upd.: latest version doesn't show gyms with ongoing raids at all, only eggs 
image

nhumber:
The addition of gym details during a raid looks great but the filters themselves are confusing for users and they aren't working well. Is it possible to keep the ui changes but go back to the standard raids on/off filter only?

friscoMad:
@nhumber What exactly do you find confusing?

nhumber:
Since implementing this yesterday, i've had almost all map users ask me how to use the raid filters because they don't understand it. either its because its confusing or maybe its because its just not working well with the current bugs.

nhumber:
these are problems i see.. its still too early to see an active raid here but so far this morning.

i can see upcoming raids with these settings

raid on
active off
gym on

i see no upcoming raids at all, just standard gyms with these settings
raids on
active on
gym on

i see nothing at all with 
raids on 
gym off

friscoMad:
After talking with @nhumber it became clear that Active Raids is not really clear on what's doing, maybe we can rename that option to "Show Active Raids only" or "Active Raids only" to make it clear that it is a way to hide eggs.

nhumber:
after talking with friscomad i understand the use cases better. ill update this when an active raid starts but currently the only problem is with this case and nothing shows.

i think another toggle would be the best fix

raids on/off
upcoming on/off
active on/off

this gives you the ability to see all raids, only upcoming, and only active.

Nothing is showing with raids on and gyms off, there are currently upcoming raids on the map.

ClarkyKent:
@nhumber  Did you copy new static01.zip and images in raids folder.. and then npm install && npm run build?

@friscoMad  Ok, i will rename that to "Show Active Raids only"

nhumber:
@ClarkyKent, yes. ill chime in again when some active raids start.

houkop:
@ClarkyKent, unziped new static01, than static04(btw, will it be updated too?)
I see nothing at all with:
raids on
active raids off
gyms off

I see nothing at all with:
raids on
active raids on
gyms off

I see ongoing and upcoming raids with:
raids on
active raids on
gyms off

I see upcoming raids with:
raids on
active raids off
gyms off

nhumber:
i have a few active in my area now, and all toggles are working correctly, except the one i posted photos of above, raids on, active on or off, and gyms off shows nothing where i think we should expect to see only gyms with upcoming/active raids.

still i vote for a new toggle which adds more options and leaves nothing to figure out.

raids on/off
upcoming on/off
active on/off

ClarkyKent:
@nhumber  last commit should fix the problem of not showing raids, while "gyms" off.

houkop:
@ClarkyKent it works, thanks

jeffrhap:
The active raid images are missing? It tries to load Team_PokeId.png, but only PokeId.png are available?

tomballgithub:
I agree with @jeffrhap that the boss raid icons from static04.zip don't work properly with this PR.  It works fine for me without this PR.

sirlindqvist:
Use the icons found in static01.zip

tomballgithub:
It doesn't work for me with this pr.  It does without pr

nhumber:
this pull specifically needs the new raid icons located in static01.zip until someone updates the unofficial static04.zip.

tomballgithub:
@nhumber What icons do you have working?  Since you say to use static01 that implies to me the colored circles with numbers and not raid boss icons.  Is that correct?

holapex:
Im also only seeing number for raid bosses, where are the actual pictures located at?

sirlindqvist:
@nhumber "this pull specifically needs the new raid icons located in static01.zip until someone updates the unofficial static04.zip."

tomballgithub:
I have no issue with functionality of this PR.  It works as designed.  It does need new static04 icons.  I have also received complaints that the start time has been removed, which is used to quickly plan raid visits.  Users are having to do unnecessary math now.

Alphawarz:
Working as described.
Amazing determination and work.

tomballgithub:
My issue from yesterday with old non-raid gyms still showing up with RAID enabled is still occurring in current code.  For instance, here is with RAID enabled and gyms disabled (does NOT work as expected since non-raid gyms are showing):
image

Here is with RAID disabled  and gyms disabled: (works as expected)
image

holapex:
I just tested everything on my server and it works perfectly fine, i tested with and without raids, tested all features and they worked as they should.

friscoMad:
@tomballgithub could you verify whats returning raw_data for that gyms and whats in the database?
Or if you want pm me and I will check the data coming from backend.
The code seems right and you db should not have raids for that gyms or old expired raids.

stefanvds1:
i pulled this PR in, and whilst the filters seem to work, an active raid is never shown.
i can filter the eggs, and when i select 'show active raids only' it just doesnt show any. something is wrong with the scanning i guess.

i did 
git pull origin pull/2133/head
npm install
npm run build

do you need -gi? i tried with and without. also, does the speedlimit matter? i thought not?

nhumber:
@stefanvds1 its been stated 20 times you need to add the new icons located in static01.zip for this to work.

ClarkyKent:
@tomballgithub  yea.. raw_data can help debug the problem/bug you have.

sebastienvercammen:
@richardfoltin @SiteEffect Are you up for updating your reviews?

tomballgithub:
I can't seem to be able to reproduce the problem with old gyms showing.  It was happening last night. 

I currently am seeing eggs with question marks that indicate they are active raids but they actually are not. 

http://pokemontomball.xyz:88/

tomballgithub:
That is a fresh develop build with raid filter applied last night that is serving as a front end only off existing active database.

tomballgithub:
screenshot_20170722-111938

343334:
Well just tested this and absolutely no raids show up on the map.  Swapped back to dev branch and now they all show up.   I think its something with the datetime lines vs the UTC timestamps in the db.

343334:
Ok after further evaluation.  This PR is working 100%.   I did discover a secondary issue though.  as this PR is a frontend change, I found no reason to ever run a -ns with this pr.  Well if you do not download sprites at all and you only run an -os line, it does NOT automatically uninstall the static01.zip that naturally occurs on the first run when doing a -ns line which when that is automatically done it auto goes into /static folder, if you manually extract static01.zip it goes right to /images or /raids.  Either way, if icons are installed properly manually this PR works perfectly or if you run a -ns line along with it.

tomballgithub:
Where should the files be manually unzipped?  I was putting the in /static/

343334:
@tomballgithub thats where they should go.  my comment was the static01.zip contents are images/ icons/ etc instead of static/images/ static/icons,  so runserver.py when executed with a -ns will extract those contents to static/  but since most of us here probably already have our db populated we tested with just an -os line which if we dont manually extract that static01.zip to static/  than it wont work. obviously npm install npm run build clear browser cache after.

PR #2138, opened 2017-07-14
Use push.js for notifications

### LINKED ISSUE #2061: Notifications not working in most mobile browsers
## Expected Behavior
Notification received when pokemon is in view
Pokemon jumping
Pokemon tapable to see its details

## Current Behavior
Notification is not received (notifications accepted when accessing map the first time)
Pokemon not jumping
Pokemon is not tapable to see its details

## Possible Solution
Is there a browser setting affecting this?

Enable notification for a pokemon type, reload page.
No notification.

## Your Environment
Android

Affected browsers:
- chrome
- opera
- samsung internet (previous version worked)

Still works in:
- firefox

pogo-excalibur:
The notification method 'sendNotification' has been modified to use push.js.

## Description
This switches notifications from the old [Notifications API](https://developer.mozilla.org/en/docs/Web/API/notification) to [push.js](https://github.com/Nickersoft/push.js/).
Push.js uses the Notifications API by default, but can upgrade to the new [Push API](https://developer.mozilla.org/en/docs/Web/API/Push_API) in browsers where the Notifications API has been deprecated.

## Motivation and Context
This is a response to #2061 which identified that Chrome for Android has now deprecated use of the Notifications API. This deprecation meant that Chrome for Android (and browsers based off it e.g. Samsung Internet) can no longer display notifications.

## How Has This Been Tested?
Tested on desktop:
* Firefox 54
* Chrome 59
* Edge 40

(All on Windows 10)

Tested on mobile:
* Firefox for Android 54
* Chrome for Android 59
* Samsung Internet 5.4
* Opera for Android 42

(All on a Galaxy S6 Edge)

## Types of changes

* Fixes #2061 `bug fix`
* This adds support for notifications on iOS `new feature`
* This removes notification sounds (as push.js doesn't support them). Sounds can/will still be played through the browser (just before the notification is sent), but will not be played on delivery of a notification. `breaking change`

## Checklist:


friscoMad:
Some comments, min and map files are not needed, grunt build them (even if map is not properly working due to flask cache buster).

I don't see why you need to add another route in app.py for service workier you should be able to fetch it like in the rest of the files, it seems that push.js has a settings to tell the path to the file, so you can use that with the cache buster (store path in html template -> setup pushjs with that).
Even better would be to use cdn for that, but I didn't found the latest version online :(

Also not sure if the setup of the service worker is needed as the docs doesn't say anything about that.

pogo-excalibur:
@friscoMad cheers for taking a look

Responses inline:

> min and map files are not needed, grunt build them (even if map is not properly working due to flask cache buster).

Good call. As you say using a cdn would be best but cdnjs doesn't seem to be up-to-date for push.js. I've opened Nickersoft/push.js#111 to try to resolve this.

In the meantime I'll add a commit adding push.js to static/js/vendor and adding it to the Gruntfile.

> I don't see why you need to add another route in app.py for service workier you should be able to fetch it like in the rest of the files

When using Service Workers, the path of the service worker script determines the scope of the worker. As map notifications are sent from '/', a worker registered under the default scripts directory ('/static/dist/js/') does not receive the notifications. I've tested this to confirm.

The service worker therefore needs to be available somewhere under '/', hence the route in app.py.

> Also not sure if the setup of the service worker is needed as the docs doesn't say anything about that.

Would you mind clarifying what you're referring to here? (Sorry :( )

friscoMad:
@pogo-excalibur 
The last part was about `createServiceWorkerReceiver` function, I have no experience with this lib and I couldn't find any example using something similar to that in the docs, that's why I was unsure if that is needed or not.

pogo-excalibur:
@friscoMad 
Ah - thank you.

Service workers operate outside the DOM context, so this was necessary to allow the service worker to direct the map to center on the notified Pokemon.

This is achieved by sending a message back from the service worker to the map. The code in createServiceWorkerReceiver adds a listener so that the map can receive these messages.

push.js doesn't seem to support responses to the sent messages.

pogo-excalibur:
This now uses push.js from cdnjs.

jagauthier:
Tested on Chrome for desktop.  Did not really see much difference.  Tested with Chrome for Android.  I received a notification through the Android notification system.

Alderon86:
Tested with IOS Safari/Firefox, pokemon are jumping on the map again, but i couldnt manage to get a sound or a notification.

pogo-excalibur:
@Alderon86 
no sounds is expected - push.js doesn't support them (I've updated the top comment to reflect this)
no notifications on iOS is also expected - iOS doesn't provide an API for web notifications (you have to have a native app)

Alderon86:
@pogo-excalibur Include Toastr messages for IOS support.

sebastienvercammen:
@pogo-excalibur As @Alderon86 said, fall back to [toastr](https://github.com/CodeSeven/toastr) (already in use in RM) on iOS for notifications.

pogo-excalibur:
@Alderon86 , @sebastienvercammen 
Good idea - I'll (hopefully) push something later today.

pogo-excalibur:
@Alderon86, @sebastienvercammen etc.
Question: what behaviour would you expect when a user has **denied** the (device) notifications permission?
a. No notification at all
b. A toastr notification is displayed

i.e. Should denying the notifications permission be interpreted as blocking any/all types of notification, or only blocking device notifications?

edit: Do we need another toggle? 🙄

pogo-excalibur:
iOS devices are now supported through toastr.

(@Alderon86 , @sebastienvercammen)

Alderon86:
I think blocking the device notifications should only affect push messages.
And if the user really wants no notifications at all he can still turn that off in his browser, so imo no need for another toggle

pogo-excalibur:
@Alderon86 
Looking back at my question, I didn't explain what I meant particularly well.
Better question: on a device which supports push.js, if a user disables push notifications should they get a toastr notification instead?

Alderon86:
as i wrote, yes it should, toastr message != push notification

pogo-excalibur:
@Alderon86 
Thank you for clarifying. I'll hopefully push that tonight.

pogo-excalibur:
Notifications will now fallback to toastr if the Notifications API/Push API etc. is unavailable (either because the the device doesn't support web notifications **or** if the permission to display device notifications has been denied).

sirlindqvist:
Works on latest iOS for me.

sirlindqvist:
Tested on iOS 10.10.3 with both Safari and Firefox.

templates/map.html
```diff
@@ -450,6 +450,7 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/js/select2.min.js"></script>
     <script src="https://cdn.datatables.net/1.10.12/js/jquery.dataTables.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment-with-locales.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/push.js/1.0.4/push.min.js"></script>
```
sebastienvercammen: We're including push.min.js from a CDN, so the local file needs to be removed (and removed from the build process).

friscoMad: Service worker needs to be served from our domain, see: https://github.com/RocketMap/RocketMap/pull/2138#issuecomment-315623866

pogo-excalibur:
@friscoMad chrome on android is working in my tests

Do note that chrome on android will only run serviceworkers if they are served over https.

In theory we could have a fallback to the Push API for non-https chrome on android; *but* the Push API will soon (likely later this month) also require https on chrome on android (RM-based source can be found in the screenshot in #2283). So falling back to an alternative that's about to be deprecated isn't (in my view) worth the effort.

templates/map.html
sebastienvercammen: 👍

PR #2078, opened 2017-06-07
Improved HLVL encounters - captcha handling - bad responses.

neskk:
## Description

If 2captcha automatic captcha mode is enabled (`--captcha-solving` + `--captcha-key`) and a HLVL account encounters a captcha, instead of putting the account aside RM now handles the Captcha.

Also detects bad status response from encounter and logs the errors accordingly.
Changed 'captcha' to 'failed' field in accounts that are put in the HLVL Account Set because it provides a broader semantic and acts a simple "on/off" switch for accounts.

## Motivation and Context

Seems obvious that HLVL accounts need captcha handling even if they don't get those very often.

## How Has This Been Tested?

I haven't actually tested this because I currently refuse to touch my LVL 30 accounts and even if I did, it's very unlikely that we get a captcha. Nevertheless I'm fairly confident that this should work without problems. One side effect I just remembered is that if captcha response takes too long, the SleepDelay won't be respected and will be nearly 0 because of previous changes to Schedulers.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2139, opened 2017-07-14
Refactor encountering with handling of all errors

friscoMad:
## Description
Supersedes #2078 and #2007 
I tried to handle all situations to avoid getting lv30 accounts blocked and adding encounter result handling.  Right now when there where exceptions the encounter will fail and the account will no be released properly, even the whole GMO parse will be discarded, this no longer happens.

I did not add auto captcha handling as that needs to be done globaly, currently it is only working on GMOs
I intentionally left hLvl accounts with lvl < 30 blocked forever after logging the error.
Increased the log levels for encouters to be in line with gmos and other requests.

## Motivation and Context
#2078 was abandoned and with #2026 there were some changes in the results, that lead to me to review that encounter code and trying to fix as many situations as possible.

## How Has This Been Tested?

Small instance with only 1 lvl30 account, tried to trigger a encounter error without luck as it seems to work without issues even if teleporting.
But tested a lot of hashing errors.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2456,7 +2456,7 @@ def encounter_pokemon(args, pokemon, account, api, account_sets, status,
 
     except Exception as e:
         log.exception('There was an error encountering Pokémon ID %s with ' +
-                      'account %s, pokemon_id', hlvl_account['username'], e)
+                      'account %s', pokemon_id, hlvl_account['username'], e)
 
     # We're done with the encounter. If it's from an
     # AccountSet, release account back to the pool.
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1855,7 +1855,6 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     sightings = {}
     new_spawn_points = []
     sp_id_list = []
-    captcha_url = ''
 
     # Consolidate the individual lists in each cell into two lists of Pokemon
     # and a list of forts.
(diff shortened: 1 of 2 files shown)


AEtHeLsYn:
While this PR is great, and probably solves #2006, it would be even better combined with PR #2059 to avoid encountering with a shadowbanned lvl30 account. Anyway great job, i'll give it a try.

RamiroWincor:
Yup #2006 should be fixed with this mostly, it is not retried but the account is not locked forever.
I am not sure what happens when you try to encounter a poke that is shadowbanned, if it doesn't work it will be logged but the account will go back to the pool, so you will be losing some encounters.
Anyway pls check and report back as I don't usually have -enc active and just got 1 account to do my tests.

friscoMad:
Yup #2006 should be fixed with this mostly, it is not retried but the account is not locked forever.
I am not sure what happens when you try to encounter a poke that is shadowbanned, if it doesn't work it will be logged but the account will go back to the pool, so you will be losing some encounters.
Anyway pls check and report back as I don't usually have -enc active and just got 1 account to do my tests.

SiteEffect:
@friscoMad It used to send response code 8 ``encounter_blocked_by_anticheat`` but I doubt, that this is still the case.

pogom/models.py
```diff
@@ -2485,6 +2345,127 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+        time.sleep(args.encounter_delay)
+
+        # If we didn't get an account, we can't encounter.
+        if not hlvl_account:
+            log.error('No L30 accounts are available, please' +
+                      ' consider adding more. Skipping encounter.')
+            return False
+
+        # Logging.
+        log.info('Encountering Pokémon ID %s with account %s at %s, %s.',
```
AEtHeLsYn: Pokemon instead Pokémon

friscoMad:
Added the missing bits of #2007

pogom/models.py
```diff
@@ -2485,6 +2342,127 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+        # If not args.no_api_store is enabled, we need to
+        # re-use an old API object if it's stored and we're
+        # using an account from the AccountSet.
+        if not args.no_api_store and using_accountset:
+            hlvl_api = hlvl_account.get('api', None)
+
+        # Make new API for this account if we're not using an
+        # API that's already logged in.
+        if not hlvl_api:
+            hlvl_api = setup_api(args, status, account)
```
AEtHeLsYn: Why do we setup the hlvl_api with regular account instead of using hlvl account?

friscoMad: good call fixing it

PR #2170, opened 2017-07-21
[bugfix] Make country lookup more robust

fosJoddie:
## Description
Code wrongfully make assumptions about the returned details from reverse geocoding. Instead of looking through the results it always assumed that the last result was the country. In some locations this was in fact the ZIP code.

## Motivation and Context
Noticed that RM tried to insert a string of more than 2 chars into the player locale table. 
Found out it was trying to insert a zip code because sloppy code assumed country was always the last result.

Obviously this would also cause the wrong country to be reported with requests.

## How Has This Been Tested?
Tested on local setup on a location that would produce a "zip code country" in current develop branch

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1008,7 +1008,23 @@ def gmaps_reverse_geolocate(gmaps_key, locale, location):
 
     try:
         reverse = geolocator.reverse(location)
-        country_code = reverse[-1].raw['address_components'][-1]['short_name']
+        address = reverse[-1].raw['address_components']
+        country_code = 'US'
+
+        # google returns an array of components, we need to find the one that is the country
+        for component in address:
+            component_is_country = False
+
+            # for good measure, each component has one or more types, look for country
+            for type in component['types']:
+                if type == 'country':
+                    component_is_country = True
+                    break
+
+            # Only when we found the country can we stop looking
+            if component_is_country:
+                country_code = component['short_name']
+                break
 
         try:
             timezone = geolocator.timezone(location)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1011,17 +1011,16 @@ def gmaps_reverse_geolocate(gmaps_key, locale, location):
         address = reverse[-1].raw['address_components']
         country_code = 'US'
 
-        # google returns an array of components, we need to find the one that is the country
+        # find country component
         for component in address:
             component_is_country = False
 
-            # for good measure, each component has one or more types, look for country
+            # look for country
             for type in component['types']:
                 if type == 'country':
                     component_is_country = True
                     break
 
-            # Only when we found the country can we stop looking
             if component_is_country:
                 country_code = component['short_name']
                 break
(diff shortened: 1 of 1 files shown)


fosJoddie:
not 100% sure how I managed to duck up my commit in such a way that it says it was made by someone else.. but there you go

sebastienvercammen:
@fosJoddie That's... interesting...?

pogom/utils.py
```diff
@@ -1021,7 +1036,7 @@ def gmaps_reverse_geolocate(gmaps_key, locale, location):
                           + 'Please check that you have Google Timezone API'
                           + ' enabled for your API key'
                           + ' (https://developers.google.com/maps/'
-                          + 'documentation/timezone/intro): %s.', e)
+                          + 'documentatiogit lon/timezone/intro): %s.', e)
```
sebastienvercammen: I think you did a whoopsy here.

pogom/utils.py
```diff
@@ -1008,7 +1008,22 @@ def gmaps_reverse_geolocate(gmaps_key, locale, location):
...
+        address = reverse[-1].raw['address_components']
+        country_code = 'US'
+
+        # find country component
+        for component in address:
+            component_is_country = False
+
+            # look for country
+            for type in component['types']:
+                if type == 'country':
```
sebastienvercammen: Don't use `type` as a variable name, `type()` is a built-in function. Feel free to use another one or shorten it as `t`.

pogom/utils.py
```diff
@@ -1008,7 +1008,22 @@ def gmaps_reverse_geolocate(gmaps_key, locale, location):
...
+                    component_is_country = True
+                    break
+
+            if component_is_country:
+                country_code = component['short_name']
+                break
```
sebastienvercammen: You can simply the above block of code as:
```
        # Find country component.
        for component in address:
            # Look for country.
            component_is_country = any([t == 'country' for t in component.get('types', [])])

            if component_is_country:
                country_code = component['short_name']
                break
```

sebastienvercammen:
@fosJoddie To rewrite your old commits with proper author info:
```
git filter-branch --env-filter '
WRONG_EMAIL="wrong@example.com"
NEW_NAME="New Name Value"
NEW_EMAIL="correct@example.com"

if [ "$GIT_COMMITTER_EMAIL" = "$WRONG_EMAIL" ]
then
    export GIT_COMMITTER_NAME="$NEW_NAME"
    export GIT_COMMITTER_EMAIL="$NEW_EMAIL"
fi
if [ "$GIT_AUTHOR_EMAIL" = "$WRONG_EMAIL" ]
then
    export GIT_AUTHOR_NAME="$NEW_NAME"
    export GIT_AUTHOR_EMAIL="$NEW_EMAIL"
fi
' --tag-name-filter cat -- --branches --tags
```

fosJoddie:
@sebastienvercammen  my beautiful visage is now on all my commits, thank you

PR #2171, opened 2017-07-21
Visualize geofences on front-end

IcognitoUnown:
PRing the visualization of the geofences after the initial functionality merge.

## Description

This PR adds a DB table ``geofence`` to retrieve data in ``app.py`` for the front-end. No additional flags have to be set. When one regular geofence flag is set, this option becomes automatically available as switch button in the front-end.
Make sure your geofences do not have duplicate names. The same geofence with the same name is fine, when used in multiple instances.

## Motivation and Context

Geofences are a great tool for RocketMap to outline areas of interest or restricted ones. To make this transparent, we want to show it to the users as well as "debug" our geofences itself if they are drawn correctly.

## How Has This Been Tested?

This was tested on my local machine.

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -100,6 +100,79 @@ def get_all(cls):
...
+                (DeleteQuery(Geofence)
+                 .where(Geofence.name == g['name'])
+                 .execute())
+
+    @staticmethod
+    def push_geofences(geofences):
+        Geofence.remove_duplicates(geofences)
+
+        db_geofences = []
+        id = 0
```
friscoMad: id is not needed, isn't it?

pogom/models.py
```diff
@@ -100,6 +100,79 @@ def get_all(cls):
...
+
+        # Make a DB save.
+        with flaskDb.database.transaction():
+            Geofence.insert_many(db_geofences).execute()
+
+        return db_geofences
+
+    @staticmethod
+    def get_geofences():
+        query = Geofence.select().dicts()
```
friscoMad: Control the situation when there are no geofences.
I know that this should be called in that situation but you never know.

friscoMad:
Not sure if I would delete all and insert every fence on load instead of replacing, the difference in timing should be neglible and you avoid problems when changing the fence and leaving the same name.

Also not sure if this handles properly fences without name (not sure if it is allowed by RM but think so).

pogom/models.py
SiteEffect: This is controlled by just returning an empty query which doesn't get looped due to ``for gf in query`` and returns am empty array ``geofences``.

friscoMad: @SiteEffect Peewee raises an exception if there are not data, not an empty iterable, not sure why

friscoMad:
Also it came to my mind recently that this does not handle different instances, not sure if we want to show all geofences from all instances in all maps, but if 2 instances have different fences with the same name, it will show the last one that was imported.

SiteEffect:
@friscoMad intended behaviour. Either you use the same name for two geofences or the same geofence on two instances, which then should contain the same content as well or you should alter the name of one when two geofences contain different content.

tomballgithub:
Is there a way to show geofences that are not active in Rocketmap?  For example, I am not using geofences in Rocketmap but would love to be able to load (even via database manipulation) my PokeAlarm geofences so the users can see them.

elincognito:
Can you fix the conflicts  @IcognitoUnown 😉

sirlindqvist:
Rebase??

PR #2173, opened 2017-07-22
Added marker clustering for Pokémon & improved front-end performance.

sebastienvercammen:
## Description
Small changes with major impact:

- Added automatic marker clustering with the updated [gmaps-marker-clusterer](https://github.com/gmaps-marker-clusterer/gmaps-marker-clusterer). Zoom level till which it clusters is set with `maxClusterZoomLevel` in `Store`, `14` by default.
- Separated globals in groups of type in `.eslintrc.json`, for readability.
- Reworked some code to be more readable (non-functional changes).
- Reworked updating functions not to force unnecessary recreation of Markers or unnecessarily calling redraw methods. These functions will now also batch remove/add Pokémon markers instead of handling them one by one (single redraw).
- Changed `Date` tricks (e.g. `+new Date` or `new Date().toLocaleString()`) to properly use `Date.now()` or moment.js.
- Moved code around so the redraw calls are done at better times.
- Set redraw calls after zooming on a global timer, so the user is free to zoom in/out as many times as they want (within 750ms) and we'll only redraw once.

While I've focused on performance, the following points still need to be properly tested/reviewed:

- Range indicators. I've disabled these for Pokémon for performance reasons, perhaps we should consider keeping them disabled? Is this a feature that's actively being used?
- Notifications for Pokémon (IV filter, rarity filter). Pokémon on the notify list will separated from the clusters so they're always visible.

## Motivation and Context
Zooming was near impossible with the old code: 3+ second delays for zooming or complete browser freezes.

## How Has This Been Tested?
Local setup and 3 others.

image
image
image
image

## Types of changes
- [x] Enhancement

## Checklist:


343334:
I see this is strictly for clustering pokemon at least from testing, any thoughts about doing gyms/pokestops as well?  Also never found a reason to use ranges.

sebastienvercammen:
@343334 For now, clustering only Pokémon lets us do some more practical testing. Pokémon are the heaviest part of the front-end, and with the small performance reworks I added in this PR, gyms/pokéstops will also work a lot better.

I'm going to push an update that separates Pokémon that match a notification filter, so they can be animated and visible outside of the clusters.

PR #2178, opened 2017-07-24
Docker and requirements.txt permissions

FrostTheFox:
Fix permissions on requirements.txt and Docker build
## Description

Docker build was failing due to the new unzip in npm, cross-unzip seems to call the linux unzip utility, so if we add this to docker it should build.

requirements.txt had +x for no reason, pip can read requirements fine without +x, also, it's a .txt with +x, wtf
## Motivation and Context

Docker is failing to build on every commit.

requirements.txt doesn't need +x.
## How Has This Been Tested?

I hate Docker, so not tested, but it should work.

And uh, a .txt should never need +x.

## Types of changes

## Checklist:


Toetje585:
I think it's also installing a older setuptools, I needed to add:

&& pip install --upgrade setuptools \

under:

&& pip install --no-cache-dir -r requirements.txt \

Otherwise I was getting some AttributeErrors while running the map :)

PR #2179, opened 2017-07-24
fix warning msg when running gym scanner

343334:
For those running a raid/gym scanner only there is no need to fill up the logfile with warnings about no wild/nearby pokemon when using -np

## Description

changed warning for no wild pokemon to only display if not using -np

## Motivation and Context

creates less logs when unnecessary

## How Has This Been Tested?

self map
## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1892,14 +1892,19 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
     # If there are no wild or nearby Pokemon . . .
     if not wild_pokemon and not nearby_pokemon:
         # . . . and there are no gyms/pokestops then it's unusable/bad.
-        if not forts:
-            log.warning('Bad scan. Parsing found absolutely nothing.')
-            log.info('Common causes: captchas or IP bans.')
+        if args.no_pokemon:
+            if not forts:
+                log.warning('Bad scan. Parsing found absolutely nothing.')
+                log.info('Common causes: captchas or IP bans.')
         else:
-            # No wild or nearby Pokemon but there are forts.  It's probably
-            # a speed violation.
-            log.warning('No nearby or wild Pokemon but there are visible gyms '
-                        'or pokestops. Possible speed violation.')
+            if not forts:
+                log.warning('Bad scan. Parsing found absolutely nothing.')
+                log.info('Common causes: captchas or IP bans.')
+            else:
+                # No wild or nearby Pokemon but there are forts.  It's probably
+                # a speed violation.
+                log.warning('No nearby or wild Pokemon but there are visible gyms '
+                            'or pokestops. Possible speed violation.')
 
     scan_loc = ScannedLocation.get_by_loc(step_location)
     done_already = scan_loc['done']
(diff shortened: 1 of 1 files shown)


friscoMad:
The logic makes little sense, just leave the if as it was and change the else to `elsif not args.no_pokemon`

pogom/models.py
```diff
@@ -1895,11 +1895,15 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         if not forts:
             log.warning('Bad scan. Parsing found absolutely nothing.')
             log.info('Common causes: captchas or IP bans.')
-        else:
-            # No wild or nearby Pokemon but there are forts.  It's probably
-            # a speed violation.
-            log.warning('No nearby or wild Pokemon but there are visible gyms '
-                        'or pokestops. Possible speed violation.')
+        elif not args.no_pokemon:
+            if not forts:
```
friscoMad: Maybe I did not explain it properly :( this if will never be true as it was already checked on 1895.

The second if is not needed at all, somwething like this is enough:
if no pokes: 
  if not forts:
    log 'bad scan, captchas, bans...'
  elif not args.no_pokemon:
    log 'no nearby, possible speed'

This way the first case is where the real error is no matter if no pokes is enabled or not, the second case is a warning when no pokes found a poke search is enabled. There are no other flows that needs to be checked.

Obihoernchen:
I think this is the 4th PR to suppress this message :D

pogom/models.py
```diff
@@ -1895,11 +1895,11 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         if not forts:
             log.warning('Bad scan. Parsing found absolutely nothing.')
             log.info('Common causes: captchas or IP bans.')
-        else:
+        elif not args.no_pokemon:
```
sebastienvercammen: `args.no_pokemon` disables parsing `wild_pokemon` and storing to DB, but it doesn't disable `nearby_pokemon`, which should almost always have Pokémon in it. So why should this warning be disabled entirely?

@friscoMad, thoughts?

friscoMad: Because we don't mind if we are violating speed limits for gym/pokestop scanning as it does not affect us.
If you use hex scan (the best one currently for gym only maps) it will warn you almost in every single gmo (unless you use -bh), that's the main reason for this PR.

sebastienvercammen: Removed "on hold", added "pending changes".

@friscoMad explained it well: we'll hit speed limits so Pokémon will be invisible, but we won't care about it because accounts that are flagged for speed limit can still scan gyms.

To get this PR merged, just add a comment above the conditional branch for clarification, e.g.:
```
# When gym scanning we'll go over the speed limit
# and Pokémon will be invisible, but we'll still be able
# to scan gyms so we disable the error logging.
elif not args.no_pokemon:
```

Obihoernchen: @sebastienvercammen we had exactly this PR already ;)
https://github.com/RocketMap/RocketMap/pull/1708
https://github.com/RocketMap/RocketMap/pull/1799

sebastienvercammen: @Obihoernchen None implemented the commenting for clarification, which is the most important part about it.

PR #2183, opened 2017-07-25
Bundle webhook messages per timed frames.

sebastienvercammen:
## Description
* Removed global "args", moved it for dependency injection.
* Replaced timing code with `timeit.default_timer()`.
* Bundled webhook messages in message frames which will send all messages received in the last `--wh-frame-interval` milliseconds (default 500) as a single message.

## Motivation and Context
Performance, reduces number of total requests, increases throughput, reduces load on receiving server, improves requests' thread use since a single thread will be able to spend more time on I/O.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -3,9 +3,9 @@
 
 import logging
 import requests
-from datetime import datetime
-from cachetools import LFUCache
 import threading
+from cachetools import LFUCache
+from timeit import default_timer
 from .utils import get_args, get_async_requests_session
 
 log = logging.getLogger(__name__)
@@ -44,7 +44,7 @@ def send_to_webhook(session, message_type, message):
 
 
 def wh_updater(args, queue, key_caches):
-    wh_threshold_timer = datetime.now()
+    wh_threshold_timer = default_timer()
     wh_over_threshold = False
 
     # Set up one session to use for all requests.
@@ -123,12 +123,12 @@ def wh_updater(args, queue, key_caches):
             if (not wh_over_threshold) and (
                     queue.qsize() > wh_warning_threshold):
                 wh_over_threshold = True
-                wh_threshold_timer = datetime.now()
+                wh_threshold_timer = default_timer()
             elif wh_over_threshold:
                 if queue.qsize() < wh_warning_threshold:
                     wh_over_threshold = False
                 else:
-                    timediff = datetime.now() - wh_threshold_timer
+                    timediff = default_timer() - wh_threshold_timer
 
                     if timediff.total_seconds() > wh_threshold_lifetime:
                         log.warning('Webhook queue has been > %d (@%d);'
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -133,7 +133,7 @@ def wh_updater(args, queue, key_caches):
 
             if num_messages > 0 and (time_passed_sec >
                                      frame_interval_sec):
-                log.debug('Sending %d items to %d webhooks.',
+                log.debug('Sending %d items to %d webhook(s).',
                           len(frame_messages),
                           len(args.webhooks))
                 send_to_webhooks(args, session, frame_messages)
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -121,7 +121,8 @@ def wh_updater(args, queue, key_caches):
             now = default_timer()
             time_passed_sec = now - frame_last_sent_sec
 
-            if (time_passed_sec > frame_interval_sec):
+            if len(frame_messages) > 0 and (time_passed_sec >
+                                            frame_interval_sec):
                 frame_last_sent_sec = now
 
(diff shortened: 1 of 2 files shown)


friscoMad:
This could be really weird with more than 1 wh-thread, also I think that if we are grouping then we should always empty the queue, as it is right now if the thread gets priority after more than wh-frame-interval msecs it will only send 1 request in the bundle.

Also to make things nicer for PA and others, it would be great to change the bundle message to be an object with custom type (like bundle) and the list inside, that way at least we would be sending always an object.

kvangent:
PA is updated to handle this change as of  https://github.com/kvangent/PokeAlarm/commit/2a9fa53268dc436baf6521f050cc8cdb3b487fed.

sebastienvercammen:
@friscoMad Queue is thread safe, so no unexpected weirdness there. Emptying queue would result in a massive dump of information on one request, but no data spread over other requests, which breaks the purpose of concurrent requests and would introduce new timing issues.

You're correct that thread switching *could* make a case happen where we happen to only send a single message, but Python threads are built to execute a minimum amount of instructions, so that is very unlikely. And even if it happens, it wouldn't happen on every single request.

Timing thread execution time is non-trivial without our own thread scheduling and the webhook threads are not CPU intensive. To add, a single webhook thread is more than enough for (almost?) all use cases, since it's only by increasing the request concurrency (`--wh-concurrency`) that you're adding room for more asynchronous I/O.

pogom/webhook.py
```diff
@@ -107,28 +110,38 @@ def wh_updater(args, queue, key_caches):
...
-                wh_threshold_timer = datetime.now()
+                wh_threshold_timer = default_timer()
             elif wh_over_threshold:
                 if queue.qsize() < wh_warning_threshold:
                     wh_over_threshold = False
                 else:
-                    timediff = datetime.now() - wh_threshold_timer
+                    timediff = default_timer() - wh_threshold_timer
 
                     if timediff.total_seconds() > wh_threshold_lifetime:
```
elincognito: Getting 
`if timediff.total_seconds() > wh_threshold_lifetime:
 AttributeError: 'float' object has no attribute 'total_seconds'
 2017-07-25 15:22:48,096 [      wh-updater-0][       webhook][   ERROR] Exception in wh_updater: 'float' object has no attribute 'total_seconds'.
   File "******", line 146, in wh_updater
     if timediff.total_seconds() > wh_threshold_lifetime:
 AttributeError: 'float' object has no attribute 'total_seconds'`

friscoMad:
By weirdness I wanted to express that messages could be kept waiting for long time and sent out of order, not a big issue just werid.

Gil in python 2.7 is released in 2 situations (according to http://www.dabeaz.com/python/UnderstandingGIL.pdf): 
 * max tick reached (default 100 ticks = 100 interpreter instructions so much less than 100 lines)
 * I/O operation

OS Scheduler has it's own algo to select the next thread that usually if the thread has run for just a little time it will select the same thread.

So my issue is that during CPU congestion the behaviour will be similar as it is currently as it will get 1 message, check that this thread has not sent anything in the last 500 ms and send it directly, that will release Gil, if we are lucky enough to get the priority again everything will be nice a couple of messages will be queued (in the best case) and Gil released again. If another thread get the lock we are back to first step.

If the thread gets frequent CPU time emptying the queue is fast and nice, if not I am not sure how spreading the queue over multiple requests can help, you have more context switching, network processing etc. Jumboframes has always be a solution for improving network throughput. I am not saying that it is a nice solution but I don't think spreading the request is really nicer in this situation.

I see this as we have to send X+Y+Z sending all at once will be always cheaper than sending them individually, I can be wrong, and probably I am at several levels, but I don't see how.

At the very least we should change the logic to wait for X secs since the first message was added to the list of messages to send, that way it is a lot more probable to send several messages together.

sebastienvercammen:
@friscoMad The biggest problem is the jump you make from saying that it switches threads every 100 instructions to saying that CPU congestion will make it so that we only send a single message, which assumes that our webhook thread will only run once in 500 milliseconds.

Assuming 100 instructions, our thread should run a lot more than just one time in 500 milliseconds, because the other thread that is currently busy has the same limitation.

Your point is not an invalid one: thread switching is complicated and since Python's threads run on a single core, we should be relying on them a lot less to do work and only rely on them for I/O, which is exactly the reason this PR is positive: with less total requests to send and bigger requests, more time is spent on the request threads that are handling I/O instead of on the threads that are picking up a single item and trying to send it instantly.

The PR results in less instructions per wh-thread and longer I/O handled in the request threads, so it reduces the problem you mentioned instead of increasing it. Anything I/O also doesn't take time away from our wh-threads: the GIL is only concerned with Python code being executed - until the data is moved into Python there is no reason to acquire the GIL.

I'll push an update to use the time since the first message was added instead of time that has passed since the start.

PR #2185, opened 2017-07-25
Create a filter for webhooks types

friscoMad:
## Description
Create a new argument that superseedes wh-scheduler and wh-updates-only, and also add support for sending or not every type of webhook.

You need to explicitly set what types of wh do you want with wh-types: [tth, lure], where valid types are: 'pokemon', 'gym', 'raid', 'egg', 'tth', 'gym-info', 'pokestop', 'lure'

Also I made some changes to pokestop wh, as it was being sent always when we are testing if it is already processed so moved to be inside that condition and made possible to send all pokestops or only lures. Keep in mind that due to the condition of not being previously processed even restarting the instance will not send all the pokestops again as it is doing for all other types.

Docs are not updated due to #2054 if this is going to merged before 2054 I will add docs.

## Motivation and Context
Currently wh-updates-only allows to send only pokes + lures or everything, there is no other option.
After Gym rework we have a lot more of gym updates and a lot of mappers only wants raids, so this would allow to only send what you want.

## How Has This Been Tested?
Tested in local instance, couldn't test 'lure' type and 'tth' is really strange, after reading the code and how it works there is something strange there.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -149,14 +149,14 @@
 
 #webhook:                       # Webhook URL e.g. http://127.0.0.1:12345 or a list for multiple webhooks
                                 # [http://127.0.0.1:1345,http://127.0.0.1:12346] (default=None)
+#wh-types:[]                    # List of events to be sent: gym, raid, egg, tth, gym-info, pokestop, lure. (default= nothing)
 #wh-threads:                    # Number of webhook threads; increase if the webhook queue falls behind. (default=1)
-#webhook-updates-only           # Only send updates to webhooks (excludes gyms & non-lured pokéstops). (default=False)
-#webhook-scheduler-updates      # Send webhook updates with scheduler status (use with -wh). (default=True)
 #wh-retries:                    # Number of times to retry sending webhook data on failure (default=5)
 #wh-timeout:                    # Timeout (in seconds) for webhook requests (default=2).
 #wh-concurrency:                # Async requests pool size. (default=25)
 #wh-backoff-factor:             # Factor (in seconds) by which the delay until next retry will increase. (default=0.25).
 #wh-lfu-size:                   # Webhook LFU cache max size (default=1000).
+
 
 
 # Status and logs
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2054,7 +2054,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     'cp_multiplier': pokemon_info.cp_multiplier
                 })
 
-            if args.webhooks:
+            if 'pokemon' in args.wh_types:
                 if (pokemon_id in args.webhook_whitelist or
                     (not args.webhook_whitelist and pokemon_id
                      not in args.webhook_blacklist)):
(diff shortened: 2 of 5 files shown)


pogom/models.py
```diff
@@ -2401,13 +2388,13 @@ def encounter_pokemon(args, pokemon, account, api, account_sets, status,
...
         elif (args.proxy and
               (hlvl_api._session.proxies['http'] not in args.proxy)):
-                proxy_idx, proxy_new = get_new_proxy(args)
-                hlvl_api.set_proxy({
-                    'http': proxy_new,
-                    'https': proxy_new})
-                hlvl_api._auth_provider.set_proxy({
-                    'http': proxy_new,
-                    'https': proxy_new})
+            proxy_idx, proxy_new = get_new_proxy(args)
```
friscoMad: Not intentional yapf should have changed this.

pogom/search.py
```diff
@@ -566,7 +566,7 @@ def search_overseer_thread(args, new_location_queue, control_flags, heartb,
         threadStatus['Overseer']['accounts_captcha'] = len(account_captchas)
 
         # Send webhook updates when scheduler status changes.
-        if args.webhook_scheduler_updates:
+        if args.speed_scan and 'tth' in args.wh_types:
```
friscoMad: Moved the speed_scan test as the function only sends wh for this scheduler.

pogom/models.py
```diff
@@ -2158,13 +2127,28 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
...
+                    wh_update_queue.put(('pokestop', wh_pokestop))
+
             # Currently, there are only stops and gyms.
             elif config['parse_gyms'] and f.type == 0:
                 b64_gym_id = b64encode(str(f.id))
                 gym_display = f.gym_display
                 raid_info = f.raid_info
                 # Send gyms to webhooks.
-                if args.webhooks and not args.webhook_updates_only:
+                if 'gym' in args.wh_types:
```
AEtHeLsYn: If we don't set gym in webhook type we won't receive eggs or raids webhooks even if we add them.

friscoMad: Here in GH seems that raids are below this if but it is under a different branch so it is not needed to have gym for raid or egg, in fact I tested it yesterday and was working properly.

AEtHeLsYn: Just tested and working fine 👍

config/config.ini.example
```diff
@@ -149,16 +149,16 @@
 
 #webhook:                       # Webhook URL e.g. http://127.0.0.1:12345 or a list for multiple webhooks
                                 # [http://127.0.0.1:1345,http://127.0.0.1:12346] (default=None)
+#wh-types:[]                    # List of events to be sent: gym, raid, egg, tth, gym-info, pokestop, lure. (default= nothing)
```
j16sdiz: Missing `pokemon` in example.

friscoMad: Damm, you can use pokemon but I missed it 😭

PR #2191, opened 2017-07-26
Get rid of config[] and use only args

friscoMad:
## Description
This PR removes config dict and uses args everywhere, as we where storing in config the same things we had on args.

Also removes an unused function in utils.

## Motivation and Context
When doing the flask pr #2189 I found out of this var and it made no sense as we have everything on args.

## How Has This Been Tested?
Local instance, with locale.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/customLog.py
+++ b/pogom/customLog.py
@@ -15,8 +15,8 @@ args = get_args()
 
 def printPokemon(id, lat, lng, itime):
     if args.display_in_console:
-        pokemon_name = get_pokemon_name(id).lower()
-        pokemon_rarity = get_pokemon_rarity(id).lower()
+        pokemon_name = get_pokemon_name(args, id).lower()
+        pokemon_rarity = get_pokemon_rarity(args, id).lower()
         pokemon_id = str(id)
         doPrint = True
         # if args.ignore:
--- a/pogom/customLog.py
+++ b/pogom/customLog.py
@@ -15,8 +15,8 @@ args = get_args()
 
 def printPokemon(id, lat, lng, itime):
     if args.display_in_console:
-        pokemon_name = get_pokemon_name(args, id).lower()
-        pokemon_rarity = get_pokemon_rarity(args, id).lower()
+        pokemon_name = get_pokemon_name(id).lower()
+        pokemon_rarity = get_pokemon_rarity(id).lower()
         pokemon_id = str(id)
         doPrint = True
         # if args.ignore:
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -824,7 +824,7 @@ def i8ln(word):
                 i8ln.dictionary = json.loads(f.read())
         else:
             # If locale file is not found we set an empty dict to avoid
-            # checking thefile every time, we skip the warning for english as
+            # checking the file every time, we skip the warning for English as
             # it is not expected to exist.
             if not args.locale == 'en':
                 log.warning(
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -20,8 +20,6 @@ from requests.packages.urllib3.util.retry import Retry
 from requests.adapters import HTTPAdapter
 from cHaversine import haversine
 
-from . import config
-
 log = logging.getLogger(__name__)
 
 
@@ -237,9 +235,8 @@ def get_args():
     parser.add_argument('-P', '--port', type=int,
                         help='Set web server listening port.', default=5000)
     parser.add_argument('-L', '--locale',
-                        help=('Locale for Pokemon names (default: {}, check ' +
-                              '{} for more).').format(config['LOCALE'],
-                                                      config['LOCALES_DIR']),
+                        help=('Locale for Pokemon names (check' +
+                              ' static/dist/locales for more).'),
                         default='en')
     parser.add_argument('-c', '--china',
                         help='Coordinates transformer for China.',
@@ -773,6 +770,8 @@ def get_args():
         else:
             args.wh_types = frozenset([i for i in args.wh_types])
 
+    args.locales_dir = 'static/dist/locales'
+    args.data_dir = 'static/dist/data'
     return args
 
 
@@ -813,14 +812,14 @@ def in_radius(loc1, loc2, radius):
     return distance(loc1, loc2) < radius
 
 
-def i8ln(word):
-    if config['LOCALE'] == "en":
+def i8ln(args, word):
+    if args.locale == 'en':
         return word
     if not hasattr(i8ln, 'dictionary'):
         file_path = os.path.join(
-            config['ROOT_PATH'],
-            config['LOCALES_DIR'],
-            '{}.min.json'.format(config['LOCALE']))
+            args.root_path,
+            args.locales_dir,
+            '{}.min.json'.format(args.locale))
         if os.path.isfile(file_path):
             with open(file_path, 'r') as f:
                 i8ln.dictionary = json.loads(f.read())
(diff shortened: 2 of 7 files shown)


pogom/utils.py
```diff
@@ -830,76 +829,63 @@ def i8ln(word):
...
+            args.root_path,
+            args.data_dir,
             'pokemon.min.json')
 
         with open(file_path, 'r') as f:
             get_pokemon_data.pokemon = json.loads(f.read())
     return get_pokemon_data.pokemon[str(pokemon_id)]
 
 
-def get_pokemon_id(pokemon_name):
```
friscoMad: It is not used anywhere, just on tests

SiteEffect:
In my opinion, the other way around would have been better, storing everything in ``config`` which is easier accessible.

friscoMad:
I can see the point for that, seems nicer that passing args all over the code. 
But I would suggest storing args as a package property not just mapping every entry in args to config as we were doing for some right now.

Also this mess with args could be solved moving to a more OOP code with classes that are initialized with the params in the constructor, but that is an even greater refactor.

Any solution seems fine to me but leaving both systems would just lead to more errors.

sebastienvercammen:
Since `get_args` is accessible everywhere and is memoized, I don't think there exists a "more accessible" option. However, for the sake of not polluting the global namespace and dependency injection, we pass it as an argument to functions.

I agree with @friscoMad that adding some more abstract data structures of our own that initialize themselves with the params is better and more future-proof. It's something we can work on gradually without a single major refactor.

This PR is good as is.

pogom/customLog.py
```diff
@@ -15,8 +15,8 @@
 
 def printPokemon(id, lat, lng, itime):
     if args.display_in_console:
-        pokemon_name = get_pokemon_name(id).lower()
-        pokemon_rarity = get_pokemon_rarity(id).lower()
+        pokemon_name = get_pokemon_name(args, id).lower()
+        pokemon_rarity = get_pokemon_rarity(args, id).lower()
```
sebastienvercammen: To all occurrences of `get_pokemon_data`, `get_pokemon_name`, `get_pokemon_rarity`, ... we should only pass the required data, not the entire `args` object.

pogom/utils.py
```diff
@@ -819,14 +818,14 @@ def in_radius(loc1, loc2, distance):
     return equi_rect_distance(loc1, loc2) < distance
 
 
-def i8ln(word):
-    if config['LOCALE'] == "en":
+def i8ln(args, word):
+    if args.locale == "en":
```
sebastienvercammen: Single quotes `'en'`.

pogom/utils.py
```diff
@@ -819,14 +818,14 @@ def in_radius(loc1, loc2, distance):
     return equi_rect_distance(loc1, loc2) < distance
 
 
-def i8ln(word):
-    if config['LOCALE'] == "en":
+def i8ln(args, word):
```
sebastienvercammen: Pass `locale` rather than `args`.

pogom/utils.py
```diff
@@ -819,14 +818,14 @@ def in_radius(loc1, loc2, distance):
...
+    if args.locale == "en":
         return word
     if not hasattr(i8ln, 'dictionary'):
         file_path = os.path.join(
-            config['ROOT_PATH'],
-            config['LOCALES_DIR'],
-            '{}.min.json'.format(config['LOCALE']))
+            args.root_path,
+            args.locales_dir,
+            '{}.min.json'.format(args.locale))
```
sebastienvercammen: Either this should be separated from this function (because it's not the translation function's job to load the data, and attaching it to the function itself is some python-esque yet incredibly ugly coding), or if we want to KISS you can call `get_args()` (it's memoized) in this conditional branch.

That last suggestion breaks writing functions for dependency injection, but it's more reasonable than passing the entire `args` object, which also ruins dependency injection because it's too vague.

pogom/utils.py
```diff
@@ -839,76 +838,63 @@ def i8ln(word):
...
+def get_move_energy(args, move_id):
+    return i8ln(args, get_moves_data(move_id)['energy'])
 
 
-def get_move_type(move_id):
-    move_type = get_moves_data(move_id)['type']
-    return {"type": i8ln(move_type), "type_en": move_type}
+def get_move_type(args, move_id):
+    move_type = get_moves_data(args, move_id)['type']
+    return {"type": i8ln(args, move_type), "type_en": move_type}
```
sebastienvercammen: Single quotes: `'type_en'`.

tests/test_utils.py
```diff
@@ -1,18 +1,22 @@
 import unittest
+import os
 from pogom import utils
 
 
+class Args:
+    locale = 'en'
+    root_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
+    data_dir = 'static/dist/data'
+    locales_dir = 'static/dist/locales'
```
sebastienvercammen: Dict makes more sense than a class.

pogom/utils.py
friscoMad: Umm there are 3 functions using this system
I can think of moving everything to a new class (or 2 l18n and pokedata) , do a proper init with the required args to load all files and use it like a singleton in all other code, it is clear but it will be more verbose when the functions are called.
Other than that I don't see a way to intialize a function which is what we are doing rigth now with that ugly code.
Anyway I don't think that change could belong to this PR so I can implement adding get_args() to the if and if you are ok with it create another PR to change it to an OOP version.

tests/test_utils.py
friscoMad: But it is mocking arg that it is a argparse.Namespace class so I think there is no way to use a dict.

pogom/utils.py
sebastienvercammen: One external function could initialize all Pokémon data functions, otherwise implementing `get_args()` and moving to something better later on is definitely good. 👍

friscoMad:
Ok all functions changed, in fact not even i18n needed a locale argument @sebastienvercammen.
It should be good to go.

pogom/utils.py
```diff
@@ -814,54 +813,43 @@ def in_radius(loc1, loc2, radius):
...
         if os.path.isfile(file_path):
             with open(file_path, 'r') as f:
                 i8ln.dictionary = json.loads(f.read())
         else:
-            log.warning(
-                'Skipping translations - unable to find locale file: %s',
-                file_path)
-            return word
+            # If locale file is not found we set an empty dict to avoid
+            # checking thefile every time, we skip the warning for english as
```
sebastienvercammen: > thefile

> English

pogom/utils.py
```diff
@@ -814,54 +813,43 @@ def in_radius(loc1, loc2, radius):
 
 
 def i8ln(word):
-    if config['LOCALE'] == "en":
-        return word
```
sebastienvercammen: @friscoMad Any reason this tiny optimization was removed? If it's English, language loading can be skipped and we can return the word as is.

friscoMad: I just moved it around to avoid needing to pass the locale or get_args every single call, in the first time the function is called it will check if the file exists like any other language and as it does not exist it just creates an empty dict that will return the word directly as it is not in the dict.

sebastienvercammen: @friscoMad I was first going to agree, but after thinking it through it makes a lot of sense to pass `locale` to `i8ln(locale, word)` on every call.

I would pass `locale` as an arg and add the conditional back. What do you think?

friscoMad: But locale can not be changed in runtime, it will make sense if we could use diferent locales but it has never been the case.
Adding the locale will force for us to be able to store different locale dicts and use the one needed properly, not a big deal but just not needed to the current behaviour.
In fact we should avoid doing that wok on server side but that's another issue.

PR #1722, opened 2017-01-16
Standardise distance calculations.

cydine:
## Description

Ditched geopy in favour of local functions to improve performance. Removed duplicate functions. Standardised earth radius for all functions.
## Motivation and Context

Several different formulas and constants were being used to perform the same distance calculations - not ideal. Using Vincenty is over the top for the distances we are dealing with.
## How Has This Been Tested?

Tested on my own machine. Pretty sure I've not broken anything......
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2192, opened 2017-07-26
[Optimization] Refactor distance calculations

friscoMad:
## Description
This PR supersedes #1722.

It changes al distance calculations from equi_rect to haversine algorithm which is more precise and not more complex.
It uses cHaversine package as it is almost 1 order of magnitude faster as it is done in Cython.
The only issue with cHaversine is that the Earth radius is not very precise but it is enough for our small scale calculations.

Some benchmarks were done for different algos and libraries, here is the code and results:
https://gist.github.com/friscoMad/f103d8a1facd8e33bd1fcc79f8e582a8

1.1Km distance measured with different functions, the error is below 30cm with cHaversine (location europe)
```
pokemad@pokemad.es [~/pokemon]# python geotest.py
#0 euclidean fun: 1097.45976537 (3.1403529644)
#1 haversine fun: 1097.45976518 (4.6673989296)
#2 great circle fun: 1097.45976518 (6.55744719505)
#3 geopy: 1096.16083708 (62.5600829124)
#3b geopy haversine: 1097.76745313 (33.4386270046)
#4 pyproj: 1096.16083747 (28.0757129192)
#5 geographiclib: 1096.16083747 (313.493932009)
#6 geodesy: 1097.4582493 (2.10232114792)
#7 cHaversine: 1096.84581781 (0.56662607193)
#7 Gryd: 1096.16083709 (17.6412119865)
```

I also used the get_new_cords function in #1722 for jittering as it is also 10x faster, but not for the rest of functions as other calls are for instance setup and need more precise values (like hives, or step calculation).

There are some changes over the code as fast_get_new_coords gets the distance in meters instead of kms and distance  (old equi_rect_distance) also returns meters instead of kms, our calculations are about meters so it is clearer.

## Motivation and Context

Unify the algorithms of distance improving precision and increasing speed.

## How Has This Been Tested?

Small local instance, needs serious reviews and tests and this messes with schedulers and a lot of critical things.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -1135,10 +1135,11 @@ class SpeedScan(HexSearch):
                     # Did we find the spawn?
                     if sp_id in parsed['sp_id_list']:
                         self.spawns_found += 1
-                    elif start_delay > 0:   # not sure why this could be
-                        # negative, but sometimes it is
+                    elif start_delay > 0:
+                        # Not sure why this could be negative,
+                        # but sometimes it is.
 
-                        # if not, record ID and put back in queue
+                        # If not, record ID and put back in queue.
                         self.spawns_missed_delay[
                             sp_id] = self.spawns_missed_delay.get(sp_id, [])
                         self.spawns_missed_delay[sp_id].append(start_delay)
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -47,7 +47,6 @@ add it to __scheduler_classes
 import itertools
 import logging
 import math
-import geopy
 import json
 import time
 import sys
@@ -62,7 +61,7 @@ from datetime import datetime, timedelta
 from .transform import get_new_coords
 from .models import (hex_bounds, SpawnPoint, ScannedLocation,
                      ScanSpawnPoint, HashKeys)
-from .utils import now, cur_sec, cellid, equi_rect_distance
+from .utils import now, cur_sec, cellid, distance
 from .altitude import get_altitude
 from .geofence import Geofences
 
@@ -315,7 +314,7 @@ class HexSearchSpawnpoint(HexSearch):
 
     def _any_spawnpoints_in_range(self, coords, spawnpoints):
         return any(
-            geopy.distance.distance(coords, x).meters <= 70
+            distance(coords, x) <= 70
             for x in spawnpoints)
 
     # Extend the generate_locations function to remove locations with no
@@ -956,8 +955,8 @@ class SpeedScan(HexSearch):
 
                 # If we are going to get there before it starts then ignore.
                 loc = item['loc']
-                distance = equi_rect_distance(loc, worker_loc)
-                secs_to_arrival = distance / self.args.kph * 3600
+                meters = distance(loc, worker_loc)
+                secs_to_arrival = meters / self.args.kph * 3.6
                 secs_waited = (now_date - last_action).total_seconds()
                 secs_to_arrival = max(secs_to_arrival - secs_waited, 0)
                 if ms + secs_to_arrival < item['start']:
@@ -976,7 +975,7 @@ class SpeedScan(HexSearch):
 
                 # For spawns, score is purely based on how close they are to
                 # last worker position
-                score = score / (distance + .01)
+                score = score / (meters + 10)
 
                 if score > best.get('score', 0):
                     best = {'score': score, 'i': i,
@@ -1039,10 +1038,9 @@ class SpeedScan(HexSearch):
                                         + ' under the speed limit.')
                 return -1, 0, 0, 0, messages, 0
 
-            distance = equi_rect_distance(loc, worker_loc)
-            if (distance >
-                    (now_date - last_action).total_seconds() *
-                    self.args.kph / 3600):
+            meters = distance(loc, worker_loc)
+            if (meters > (now_date - last_action).total_seconds() *
+                    self.args.kph / 3.6):
                 # Flag item as "parked" by a specific thread, because
                 # we're waiting for it. This will avoid all threads "walking"
                 # to the same item.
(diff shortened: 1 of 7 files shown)


pogom/schedulers.py
```diff
@@ -976,7 +975,7 @@ def next_item(self, status):
 
                 # For spawns, score is purely based on how close they are to
                 # last worker position
-                score = score / (distance + .01)
+                score = score / (meters + .01)
```
friscoMad: This needs to be 100

pogom/utils.py
```diff
@@ -794,20 +801,41 @@ def cellid(loc):
...
+
+    # convert decimal degrees to radians
+    lon1, lat1, lon2, lat2 = map(radians, [pos1[1], pos1[0], pos2[1], pos2[0]])
+
+    # haversine formula
+    dlon = lon2 - lon1
+    dlat = lat2 - lat1
+    a = sin(dlat/2)**2 + cos(lat1) * cos(lat2) * sin(dlon/2)**2
+    c = 2 * asin(sqrt(a))
+    r = 6371009  # IUGG mean earth radius in kilometers.
```
friscoMad: meters

Obihoernchen:
Just out of curiosity, I benched Noctem's pogeo cython lib as well.
As you can see it's about as fast as cHaversine. So cHaversine is the best choice 👍 
```
#0 euclidean fun: 676371.322420009 (1.7452428340911865)
#1 haversine fun: 675656.2994818708 (2.179184675216675)
#2 great circle fun: 675656.2994818708 (3.16937518119812)
#3 geopy: 677789.5312317796 (32.097493171691895)
#3b geopy haversine: 675845.7290243531 (15.252706050872803)
#4 pyproj: 677789.531232748 (11.268158197402954)
#5 geographiclib: 677789.5312327481 (196.10552430152893)
#6 geodesy: didn't work for me
#7 cHaversine: 675278.3208080824 (0.251234769821167)
#8 Gryd: 677789.5312334074 (7.870181560516357)
#9 pogeo: 675656.299481871 (0.26857995986938477)
```

pogom/utils.py
```diff
@@ -22,6 +21,13 @@
 
 from . import config
 
+# Trying to import cHaversine as it is faster for distances.
+try:
+    from cHaversine import haversine
+except ImportError as e:
+    # Pass as this is an optional requirement.
+    pass
```
Obihoernchen: hum it's used for distance(), so I don't think it is optional ;)

ghost:
How does our old equi_rect_distance stack up in terms of accuracy and time?

pogom/utils.py
friscoMad: ups, this is from a previous iteration where I was unsure if we could use cHaversine in all platforms.

friscoMad:
@mew905 it should be like "euclidean fun" in the test, it is the same algorithm. So a 5x improvement.

Obihoernchen:
```
2017-08-07 13:04:21,517 [   search-worker-0][        search][    INFO] Getting details for gym @ XX.XXX/XX.XXX (478.998759m away)
```
I think we can remove some decimal places ;)
Right now it's μm accuracy :D

Obihoernchen:
Testing this since ~2 weeks and it's working flawlessly. But I didn't test any single scheduler etc.

Obihoernchen:
Oh found two more log messages:
https://github.com/friscoMad/RocketMap/blob/1de5c5da3da97f73164eb7f3a6989701596f04aa/pogom/search.py#L1055
https://github.com/friscoMad/RocketMap/blob/1de5c5da3da97f73164eb7f3a6989701596f04aa/pogom/search.py#L1084

Please change to `%.0fm` as well.

pogom/schedulers.py
```diff
@@ -1139,7 +1136,7 @@ def task_done(self, status, parsed=False):
                     if sp_id in parsed['sp_id_list']:
                         self.spawns_found += 1
                     elif start_delay > 0:   # not sure why this could be
```
sebastienvercammen: Put this comment on the next line, so it aligns with the rest of the sentence. Also capitalize & add punctuation (I'll leave the proper length up to you, not in my IDE atm).
```
# Not sure why this could be
# negative, but sometimes it is.
```

pogom/search.py
```diff
@@ -1027,9 +1027,9 @@ def search_worker_thread(args, account_queue, account_sets,
                     gyms_to_update = {}
                     for gym in parsed['gyms'].values():
                         # Can only get gym details within 1km of our position.
-                        distance = calc_distance(
-                            step_location, [gym['latitude'], gym['longitude']])
-                        if distance < 1.0:
+                        if distance(step_location, [
+                                gym['latitude'], gym['longitude']
+                        ]) < 1000:
```
sebastienvercammen: This is a bit of an awkward notation. Can we clarify with a variable (e.g. `dist`, `gym_dist` or `gym_distance`), does that make it more readable?

pogom/utils.py
```diff
@@ -803,20 +803,14 @@ def cellid(loc):
...
-    R = 6371  # Radius of the earth in km.
-    lat1 = math.radians(loc1[0])
-    lat2 = math.radians(loc2[0])
-    x = (math.radians(loc2[1]) - math.radians(loc1[1])
-         ) * math.cos(0.5 * (lat2 + lat1))
-    y = lat2 - lat1
-    return R * math.sqrt(x * x + y * y)
+# Return approximate distance in meters.
+def distance(pos1, pos2):
+    return haversine((tuple(pos1))[0:2], (tuple(pos2))[0:2])
```
sebastienvercammen: Why the constant conversion to tuple and `[0:2]` indexing? Applying this to all calls seems unnecessary.

friscoMad: cHaversine is done in Cython so due to the mappings it needs a tuple of 2 elements (at least without it throws me errors).
Most of our pos are list and have altitude.
I will do some more tests but I think they are required.

friscoMad: Yup reducing the number of elements is needed as it  uses unpack instead of accessing by index and the tuple is required by cython, I think we could PR cHaversine to use indexes and be less restrictive but not sure how fast it will be merged.

sebastienvercammen: @friscoMad Very good, thanks for the update. 👍 Will merge this PR as is, but you're definitely welcome to PR cHaversine to improve the project.

PR #2205, opened 2017-07-27
Properly close db connections on -os

friscoMad:
## Description
As I was calling a private api I did not see that the next call was also needed, otherwise models will open their connections but will not close it properly, that's why we reached mas db connections opened so easily.

## Motivation and Context

Fixes db connection issues in -os and mixed instances.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -84,7 +84,7 @@ def init_database(app):
 
     # Using internal method as the other way would be using internal var, we
     # could use initializer but db is initialized later
-    flaskDb._load_database(app, db)    
+    flaskDb._load_database(app, db)
     if app is not None:
         flaskDb._register_handlers(app)
     return db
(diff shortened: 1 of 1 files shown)


friscoMad:
Currently testing in -ns instance to avoid more regressions, needs tests in -os and regular instances to see if it fixes the issue.

pogom/models.py
```diff
@@ -84,8 +84,9 @@ def init_database(app):
 
     # Using internal method as the other way would be using internal var, we
     # could use initializer but db is initialized later
-    flaskDb._load_database(app, db)
-
+    flaskDb._load_database(app, db)
```
sebastienvercammen: Ouch, trailing whitespace. :(

friscoMad: models py nuclide___neoseo es_home_pokemad_pokemon_rocketmap_pogom atom_2017-07-28_00-19-11

I wanted to pr the fix that much that I did not even haven't waited for Flask review

PR #2214, opened 2017-07-29
Add gym_data to raids and gym if possible

friscoMad:
## Description
Adds gym-info data to gym and raid webhooks using wh cache.

## Motivation and Context
Getting gim name and some extra info is a request we have got a lot of times so this tries to do it with the least impact possible.

It just looks for gym-info in wh cache and if it is, it is added as gym-data field to the content.

By design gym-info will be always outdated, as gym and raid wh are sent before the gym-info is retrieved, and due to that all the initial gym and raid whs will not have gym-info until the same gym needs to be sent for a second time.

## How Has This Been Tested?

Tested with small instance.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -37,6 +37,14 @@ def send_to_webhooks(args, session, message_frame):
             log.exception(e)
 
 
+def add_extra_info(message, key_caches):
+    msg = message['message']
+    if (message['type'] in ['raid', 'gym'] and
+            msg['gym_id'] in key_caches['gym_details']):
+        msg['gym_details'] = key_caches['gym_details'][msg['gym_id']]
+    return message
+
+
 def wh_updater(args, queue, key_caches):
     wh_threshold_timer = default_timer()
     wh_over_threshold = False
@@ -45,9 +53,7 @@ def wh_updater(args, queue, key_caches):
     # Requests to the same host will reuse the underlying TCP
     # connection, giving a performance increase.
     session = get_async_requests_session(
-        args.wh_retries,
-        args.wh_backoff_factor,
-        args.wh_concurrency)
+        args.wh_retries, args.wh_backoff_factor, args.wh_concurrency)
 
     # Extract the proper identifier. This list also controls which message
     # types are getting cached.
@@ -121,13 +127,15 @@ def wh_updater(args, queue, key_caches):
                         if __wh_object_changed(whtype, key_cache[ident],
                                                message):
                             key_cache[ident] = message
-                            frame_messages.append(frame_message)
+                            frame_messages.append(
+                                add_extra_info(frame_message, key_caches))
                             log.debug('Queued updated %s to webhook: %s.',
                                       whtype, ident)
                         else:
                             log.debug('Not queuing %s to webhook: %s.', whtype,
                                       ident)
                 queue.task_done()
+
             # Store the time when we added the first message instead of the
             # time when we last cleared the messages, so we more accurately
             # measure time spent getting messages from our queue.
@@ -141,8 +149,7 @@ def wh_updater(args, queue, key_caches):
             # If enough time has passed, send the message frame.
             time_passed_sec = now - frame_first_message_time_sec
 
-            if num_messages > 0 and (time_passed_sec >
-                                     frame_interval_sec):
+            if num_messages > 0 and (time_passed_sec > frame_interval_sec):
                 log.debug('Sending %d items to %d webhook(s).',
                           num_messages,
                           len(args.webhooks))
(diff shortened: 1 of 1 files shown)


friscoMad:
There are a lot of changes due to yapf8 (somehow I lost the option to format just a selection so the whole file was reformated) if that is a problem I can update with the least possible changes.

pogom/webhook.py
```diff
@@ -23,14 +23,25 @@ def send_to_webhooks(args, session, message_frame):
...
         except requests.exceptions.RequestException as e:
             log.exception(e)
 
 
+def add_extra_info(message, key_caches):
+    msg = message['message']
+    if (message['type'] in ['raid', 'gym'] and
+            msg['gym_id'] in key_caches['gym_details']):
+        msg['gym_data'] = key_caches['gym_details'][msg['gym_id']]
+    return message
```
sebastienvercammen: Are we sure this is enough? Webhook messages aren't assumed to always be sent or received in order - are we sure there is no case where we're receiving `raid` before receiving `gym_details`? 

One inevitable problem is when we haven't scanned the gym's details yet, are scanning `gym` for the first time since scanner startup, and try to send the `gym` message. In this case, we should have retrieved the gym's details from the database.

pogom/webhook.py
```diff
@@ -112,12 +118,13 @@ def wh_updater(args, queue, key_caches):
                     # data to webhooks.
                     if __wh_object_changed(whtype, key_cache[ident], message):
                         key_cache[ident] = message
-                        frame_messages.append(frame_message)
-                        log.debug('Queued updated %s to webhook: %s.',
-                                  whtype, ident)
+                        frame_messages.append(
+                            add_extra_info(frame_message, key_caches))
+                        log.debug('Queued updated %s to webhook: %s.', whtype,
+                                  ident)
```
sebastienvercammen: Is it yapf that has such inconsistent indentation?

friscoMad: Yup, it has a lot of options so maybe some adjusts are needed: 
https://github.com/google/yapf/blob/master/yapf/yapflib/style.py

pogom/webhook.py
friscoMad: we will receive raid always before gym_details that's why I said the info is outdated by design, getting gym_details before could be possible with more than 1 wh-thread and some strange context switches as we are also synching on cache access.

gym-info is retrieved after the whole GMO is parsed, and other WH are enqueued before that.

Using db will only improve the startup info missing but not the order issue in general, and I wanted to avoid hitting db as much as possible just for sending gym name, that it seems that is what most people want.

We could change this to the other way around, adding gym and raid to gym-info so that gym-info has all the gym related information.

Another thing that has come to my mind is that without enabling gym-info in wh filter this would not add anything :(
The other option I can think of is just caching all gym names and just adding that if we want.

sebastienvercammen: @friscoMad Gym name currently in DB should be sent with all gym messages. `gym_data` would cover all cases, but by including all the gym's data we're sending duplicate data over and over again.

Gym name has a use case for PokeAlarm, the other fields (currently) do not, so we'll keep it light by only including gym name, and we'll query DB to retrieve the gym's name if we don't currently have it stored in the cache.

friscoMad: I will rebase but will not change the logic, I really don't like any of the solutions to this problem, with any of the proposed solutions gym name can not guaranteed to be present at all times in the message even if you enable gym-info as during the first scan with new DB you will always get raid and gym before get_gym_info.

Going to the DB does fix the intial RM load with a populated DB and gi but that is already mostly covered in PA even without this PR. From my experience PA is restarted a lot less than RM.

But if we want to use DB we will be hitting it for every gym/raid/egg sent until 'gym-info' is sent (you need to enable 'gym-info' whs just to get gym name) and gym-info does not change that frequently. If we opt for using a cache then we will need to add another cache just for gym details or faking a gym-info wh just to insert it in the cache. Whatever we do is ugly.

Also sending about just the name while I see that is somewhat less traffic it will be just a concession from our vision on whs. We are sending whole entities and we always said that the info that comes from API goes to WH. So if we send the name we should send the whole details as related entity. 
We have had gym and gym-info whs for a long time and no one asked for this, nobody knows what will come tomorrow but I don't think some extra data with raids is such a problem and will help some use cases now ir in the future and it is more clean.

Anyway I do think a better solution (still don't like it too much) would be delaying gym WH until we have sent the gym-details (if gi is enabled) it does require some refactor in parse code but it is possible and I think it fixes all use cases.
Other possible solution is that PA just do get the info via RM webserver just when it is needed.

I general I think that it is best to move towards a system where every message that comes from Niantic goes to wh without additional parsing or logic is the other application the one that needs to do it's process as it needs: caching, filtering or whatever. Splitting pokes, stops, gym and raids from a single GMO while it make it easier to consume for PA is making it impossible to cover other use cases.
And related to that it is really simple to just serialize protobuf objects back to json to be sent as whs.

friscoMad:
Rebased

PR #2230, opened 2017-08-03
Filter raids by Pokemon (raid boss)

faith9:
Added a new option to filter the raids by pokemon (raid boss)

## Description

My first PR so please be gentle :)

## Motivation and Context

A lot of people this days only search for Moltres, Lugia or Tyranitar
It's easier to only show them on the map.

## How Has This Been Tested?

Tested on my local map

http://i.imgur.com/eEYV5IN.png

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -882,6 +882,10 @@ var StoreOptions = {
         default: 5,
         type: StoreTypes.Number
     },
+    'showRaidPokemon': {
+        default: 0,
+        type: StoreTypes.Number
+    },
     'showGyms': {
         default: false,
         type: StoreTypes.Boolean
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -12,6 +12,7 @@ var $switchOpenGymsOnly
 var $switchActiveRaidGymsOnly
 var $switchRaidMinLevel
 var $switchRaidMaxLevel
+var $switchRaidPokemon
 var $selectTeamGymsOnly
 var $selectLastUpdateGymsOnly
 var $selectMinGymLevel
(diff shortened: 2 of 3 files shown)


neskk:
Nice effort 👍 but I think that removing Gyms from map when they don't have the desired raid boss is kinda bad and the other raid filters try to avoid doing this. If you enable raid filtering you won't see Gyms being hidden, just their raids.

daniel-mitran:
Thank you for your feedback.
I tried to mirror @ClarkyKent 's raid filters #2133 , which were merged 12 days ago.
If I read the code correctly, this is how we currently filter by raid level, by removing the gym altogether.

`        if (raidLevel > Store.get('showRaidMaxLevel') || raidLevel < Store.get('showRaidMinLevel')) {
             removeGymFromMap(item['gym_id'])
             return true
         }`

friscoMad:
The remove is fine, it is under a if that check is gyms are disabled, the problem is that it will not work if gyms are enabled.

Also I think a multiselect like notifications or hide will be better instead of only choosing one.
And lastly if you don't want to do it multi select I think that you at least order them by id and use locale to get their names.

tomballgithub:
Is this dead, would be a nice feature to have?

tomballgithub:
I think the statement by @neskk that the other raids filters don't remove gyms is incorrect.  They all do the same call to filter out a gym that this PR does -> removeGymFromMap

Example:
```
        if (Store.get('showActiveRaidsOnly')) {
            if (!isOngoingRaid(item.raid)) {
                removeGymFromMap(item['gym_id'])
                return true
            }
        }
```

faith9:
I'm using it on my map, it can be applied  on develop (I'll sync the branch today).
Unfortunately I lack the abilities and the time to make the changes requested above (convert the dropdown to a multiselect and find how to use locales).

If someone else finds it useful and like to take over it's fine with me.

faith9:
Regarding the sorting: before today the sorting was by CP, the most powerful pokemons being the most interesting.
I have added today the beasts, that will not respect this order (but will be the most filtered pokemon, in the following months)

tomballgithub:
Why not try to copy the multi selection that was done for 'notify of pokemon' or 'notify of rarity' selections?

tomballgithub:
I can tackle if you don't want to

faith9:
Please tackle it, I have nothing against it :)

tomballgithub:
There's a bug somewhere, I just saw it get into a state where it would only show Raikou.  If I switched to anything else include 'show all' it would just be blank.  Switching back to Raikou worked again.  When I cleared the local storage in my browser it went back to normal.  Will try to figure out what caused it

billyjbryant:
Is this dead?

tomballgithub:
I still use it on my map.  It works.  I think it needs improvement which is
why I have not redone this PR.

tomballgithub:
The problem with this feature is it needs constant updating with the ever changing raid boss list.  Unless you support all pokemon as being possible raid bosses is there a way to support dynamic changes to current raid bosses?

PR #2232, opened 2017-08-03
Remove gym distance control

friscoMad:
## Description
Currently it seems that all gyms we get from GMO are fetchable so there is no reason to do an additional 
distance check.
We ask for cells in 500m radius but for some reason it seems that we get things a bit more far away but the most I have seen is aroun 720m that it is still nearer than the 1Km gym action radius.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1025,35 +1025,25 @@ def search_worker_thread(args, account_queue, account_sets,
                     # Build a list of gyms to update.
                     gyms_to_update = {}
                     for gym in parsed['gyms'].values():
-                        # Can only get gym details within 1km of our position.
-                        distance = calc_distance(
-                            step_location, [gym['latitude'], gym['longitude']])
-                        if distance < 1.0:
-                            # Check if we already have details on this gym.
-                            # Get them if not.
-                            try:
-                                record = GymDetails.get(gym_id=gym['gym_id'])
-                            except GymDetails.DoesNotExist:
-                                gyms_to_update[gym['gym_id']] = gym
-                                continue
+                        # Check if we already have details on this gym.
+                        # Get them if not.
+                        try:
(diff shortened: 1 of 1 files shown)


sebastienvercammen:
I'm not sure I agree with removing it. We added it for good reason: because we couldn't trust the data received. So perhaps we should keep it in as a safety measure?

One good reason to remove would be if it severely impacts performance, but it doesn't look like it would.

friscoMad:
I made this as some ppl asked about if it was possible to not being able to get info from a gym returned by GMO, it made sense and did some tests and I even run this change for several days without issues in my main instance but I agree that it is a safety measure and it is not that costly.

sebastienvercammen:
@friscoMad Agreed on closing the PR and keeping it as is for safety?

friscoMad:
I was about to close it, but then tought that while niantic does not change a thing leaving the check does not do anything, if the change it it could be 3 situations:
 
- They change the range limit to lower, then this code will do even less
- They change the range limite to be higher, then this code will be a bug as we had it for months limited at 450m when  1km was allowed
- They change it so that they return gyms that can not be retrieved, then the check will be great iif they keep the limit to 1km (in fact I have only see gym request for less than 800m), if they also change the limit we will do fail requests or request less gyms than possible.

So most of the time if they change something this check will not be ok anyways so I think it is better to remove it and let it fail if they change anything, we already have the gym request response logged for distance error in case something changes, and can fix it easily if needed.

sebastienvercammen:
@friscoMad We have many other lines of code that rely on allowable/scannable distance. If Niantic changes the distances, we would have to update the code anyway.

In the meantime, we should never assume to trust Niantic's data and fall back to a reasonable/safe handling instead of hoping for errors that make all scanners unusable.

Closing PR, we can reopen if this situation changes.

PR #2234, opened 2017-08-03
Add cache and lock to avoid getting the same gym twice

### LINKED ISSUE #2162: Gym info being requested multiple times.
This was always the case but hidden under a distinct in Gym.get_gyms() but also due to the new 1Km range we did in raids PR this will happen more usually as there is no control that avoids scanning the same gym twice, I did monitor and I constantly have 4-5 gyms in that situation and that's even discounting the gyms that where requested twice but cleared in between.

We should add some kind of shared control to avoid this situation.

friscoMad:
## Description
Just avoid asking details of the same gym twice setting a small ttl cache with 1 min eviction and capacity of 10K gym ids (it is a big number just in case it shouldn't never be such big).

~This PR is not production ready for some reason you can not kill the process anymore with Ctrl+C and flask stop working as soon as I add the line `with gym_cache_lock:`~ Seems to be working perfectly now 😅 

**Seb:** Fixes #2162.

## Motivation and Context
Currently we check for last_modified to see if we need to as gym info but between we decide to go for it and when we store the data back in the DB it could take some time, so other workers can be already doing the same, that ends with gym members duped and time wasted.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -1040,7 +1040,7 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                                 continue
                             else:
                                 # Set the gym as in progress it will just be
-                                # locked for 60 seconds due to TTL eviction
+                                # locked for 60 seconds due to TTL eviction.
                                 gym_cache[gym['gym_id']] = True
 
                         # Can only get gym details within 1km of our position.
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -342,6 +342,7 @@ def search_overseer_thread(args, new_location_queue, control_flags, heartb,
     hashkeys_last_upsert = timeit.default_timer()
     hashkeys_upsert_min_delay = 5.0
     gym_cache = None
+
     if args.gym_info:
         gym_cache = TTLCache(maxsize=10000, ttl=60)
 
@@ -1038,6 +1039,8 @@ def search_worker_thread(args, account_queue, account_sets, account_failures,
                                     gym['latitude'], gym['longitude'])
                                 continue
                             else:
+                                # Set the gym as in progress it will just be
+                                # locked for 60 seconds due to TTL eviction
                                 gym_cache[gym['gym_id']] = True
 
                         # Can only get gym details within 1km of our position.
--- a/pogom/search.py
+++ b/pogom/search.py
@@ -28,6 +28,7 @@ import requests
 import schedulers
 import terminalsize
 import timeit
+import threading
 
 from datetime import datetime
 from threading import Thread, Lock
@@ -37,6 +38,7 @@ from collections import deque
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 from distutils.version import StrictVersion
+from cachetools import TTLCache
 
 from pgoapi.utilities import f2i
 from pgoapi import utilities as util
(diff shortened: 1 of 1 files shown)


darkelement1987:
I've had 28 triggers for the same gym to pokealarm, does this fix that problem?

image

Edit: i tried this stuff and it gives me multiple errors like:

> 2017-08-12 00:59:06,388 [  search-worker-24][        search][   ERROR] Exception in search_worker under account glonkpochkuk Exception message: TypeError('check_login() takes exactly 5 arguments (4 given)',).

friscoMad:
@darkelement1987 if those events are gym-info events it will help with the issue, but anyway they should be handled by the wh cache, so unless you modified wh-lfu-size or have a really instance with thousands of gyms (either way increase wh-lfu-size to be bigger than your number of gyms) or you have done something wrong.

The errors seem to be a merge problem with the sanity fixes. I have rebased it recently so please tell me if it doesn't work.

darkelement1987:
I already fixed it by decreasing my wh-threads! Thanks for your reaction :)

tomballgithub:
Code looks good, I am going to test...

tomballgithub:
I have been running this PR on my public map since Aug 14 with no known issues.  I just verified that searching on my map I could not find any instances of the original problem.  I had previously reviewed the code and saw no issue.  I approve

pogom/search.py
```diff
@@ -338,6 +341,9 @@ def search_overseer_thread(args, new_location_queue, control_flags, heartb,
     api_check_time = 0
     hashkeys_last_upsert = timeit.default_timer()
     hashkeys_upsert_min_delay = 5.0
+    gym_cache = None
+    if args.gym_info:
```
sebastienvercammen: Add whitespace above this line.

pogom/search.py
```diff
@@ -54,6 +56,7 @@
 log = logging.getLogger(__name__)
 
 loginDelayLock = Lock()
+gym_cache_lock = threading.Lock()
```
sebastienvercammen: `gym_cache_lock` is only used in the infinite loop in the search worker thread, so it doesn't need to be defined globally.

pogom/search.py
```diff
@@ -1025,6 +1030,16 @@ def search_worker_thread(args, account_queue, account_sets,
...
                     for gym in parsed['gyms'].values():
+                        with gym_cache_lock:
+                            if gym['gym_id'] in gym_cache:
+                                log.debug(
+                                    ('Skipping update of gym @ %f/%f, ' +
+                                     'already in progress.'),
+                                    gym['latitude'], gym['longitude'])
+                                continue
+                            else:
+                                gym_cache[gym['gym_id']] = True
```
sebastienvercammen: It never gets removed from the gym cache, so a gym will only update once and never again afterwards.

pogom/search.py
friscoMad: Moving it to the function wouldn't make it local to every thread instead of being a global lock?

pogom/search.py
friscoMad: It is TTLCache with 60 seconds to avoid that

sebastienvercammen: @friscoMad 👍 Can you add a comment above this line to clarify it?

pogom/search.py
sebastienvercammen: @friscoMad 👍 Missed the threading, you're absolutely correct. The other option is to send it to the thread just like we do with `gym_cache` but those arguments are already getting massive.

PR #2238, opened 2017-08-05
Resurrect Spawn-Scanning

Alderon86:
Be aware -ss doesnt deliver a "Missed Scan" message, so missed Spawns / shadowbans are not tracked.
## Motivation and Context

We hate Shadowbans :D
## Types of changes

## Checklist:

Thanks @mew905 and @neskk


pogom/models.py
```diff
@@ -1292,6 +1335,10 @@ def get_spawnpoints_in_hex(cls, center, steps):
                              (SpawnPoint.longitude >= w) &
                              (SpawnPoint.longitude <= e)
                              ))
+        if args.spawnpoint_scanning and not args.no_server:
+            query = (query
+                     .where(SpawnPoint.disappear_time.is_null(False)))
```
neskk: `SpawnPoint.disappear_time` field doesn't exist. Consider removing this check/query condition and maybe use `Spawnpoint.start_end` to fetch disappear time and then filter the Spawnpoints based on that.

pogom/schedulers.py
```diff
@@ -358,25 +357,29 @@ def __init__(self, queues, status, args):
...
+                    # getting last scanned as secs in hour
+                    for gym in gyms:
+                        gym['time'] = gym['time'] % 3600
+            else:
+                log.debug('Loading spawn points from database.')
+                spawns += SpawnPoint.get_spawnpoints_in_hex(
+                    self.scan_location, self.args.step_limit)
+                log.debug('Loaded %s spawns from database.' % len(spawns))
+
+        self.locations.extend(gyms)
```
neskk: `self.locations` is initialized with False and when you try to extend a `bool` it goes 💣 
In Line 355 change `self.locations = False` to `self.locations = []`
Small side-note: I have an hard time thinking that this PR was tested before being submitted. I don't mean to be rude but at least try out your changes before pushing a PR 👍

pogom/schedulers.py
```diff
@@ -353,30 +352,33 @@ def __init__(self, queues, status, args):
             self.step_distance = 0.070
 
         self.step_limit = args.step_limit
-        self.locations = False
+        self.locations = []
```
neskk: We also need to change line 475 - `self.locations = None` -> `self.locations = []`

pogom/schedulers.py
```diff
@@ -353,30 +352,33 @@ def __init__(self, queues, status, args):
...
-
+        spawns = []
+        gyms = []
+        stops = []
         # No locations yet? Try the database!
         if not self.locations:
-            log.debug('Loading spawn points from database')
-            self.locations = SpawnPoint.get_spawnpoints_in_hex(
-                self.scan_location, self.args.step_limit)
+            if self.args.no_pokemon:
```
friscoMad: There is no reason for adding gyms and stops, if using -bh as we should those would be parsed anyway, and for stops and gym scanning there is no reason to use spawn scan, hex is better and a lot less counter intuitive.

pogom/schedulers.py
```diff
@@ -398,15 +400,13 @@ def _generate_locations(self):
         log.info('Total of %d spawns to track', len(self.locations))
 
         # locations.sort(key=itemgetter('time'))
-
```
friscoMad: Forgot to say that gym and stops does not have a time field

pogom/schedulers.py
```diff
@@ -353,30 +352,33 @@ def __init__(self, queues, status, args):
...
+                    gyms += Gym.get_gyms_in_hex(self.scan_location,
+                                                self.args.step_limit)
+                    log.debug('Loaded %s gyms from database.' % len(gyms))
+                    # getting last scanned as secs in hour
+                    for gym in gyms:
+                        gym['time'] = int(time.mktime(gym['time']
+                                                      .timetuple()) % 3600.0)
+            else:
+                log.debug('Loading spawn points from database.')
+                spawns += SpawnPoint.get_spawnpoints_in_hex(
```
friscoMad: SpawnPoint.get_spawnpoints_in_hex should call to select_in_hex_by_location(center, steps), it does the same but the filter it is a lot better as it only gets spawns inside the hex and not in a circle, that way there is no possibility of shared spawnpoints when using -bh

pogom/schedulers.py
neskk: @friscoMad `-bh` shouldn't affect scheduler, using `-bh -ss` or just `-ss` will be same in terms of scanning method, just different area configurations.
I can understand the changes in this PR because they save a lot of GMO requests for people that are just running raid/gym parsing.

pogom/schedulers.py
```diff
@@ -388,25 +377,22 @@ def _generate_locations(self):
...
         # {"lat": 37.53079079414139, "lng": -122.28811690874117,
         #  "spawnpoint_id": "808f9f1601d", "time": 511
 
         log.info('Total of %d spawns to track', len(self.locations))
 
-        # locations.sort(key=itemgetter('time'))
-
         if self.args.verbose:
             for i in self.locations:
                 sec = i['time'] % 60
```
neskk: So... this code is also bad? `self.locations` contains SpawnPoints and SpawnPoint currently doens't have a 'time' field. Am I crazy or failing to see something? There's a couple of places where `time` field is used and this must be addressed, no?

ghost:
Is it strictly spawnscanning that doesn't recieve shadowbans? I've been using it with clustering for months now without *too much* trouble, but I assumed the clustering was the real hero by delaying and offsetting when and where we check an area for a spawn in order to get more at once.

pogom/schedulers.py
michikrug: The 'time' field is calculated using the class method 'start_end' that is called in 'get_spawnpoints_in_hex'

neskk: Yeah after a bit I got that. I was trying to use friscoMad suggestion: `select_in_hex_by_location` but then we have to call start_end and do some "magic" with appears, leaves times because SS is expecting a timestamp not secs after the hour.

Alderon86:
@mew905 we are testing that, if it is clustering, i will add it to the pr.

pogom/schedulers.py
friscoMad: Maybe I did not explain correctly what I did was call select_in_hex_by_location from get_spawnpoints_in_hex remove the filter and leave the last part that adds time (I also needed to change a bit in spawn scheduler as instead of lat and long we have latitude and longitude)

Alderon86:
@mew905 if everything else gets the 🆗 , we will check that.

pogom/cluster.py
```diff
@@ -0,0 +1,136 @@
...
+    def __contains__(self, item):
+        return item in self._spawnpoints
+
+    def __len__(self):
+        return len(self._spawnpoints)
+
+    def append(self, spawnpoint):
+        # update centroid
+        f = len(self._spawnpoints) / (len(self._spawnpoints) + 1.0)
+        self.centroid = intermediate_point(
```
friscoMad: This shouldn't be a call to simulate_centroid (and rename it to calculate new centroid).

pogom/cluster.py
```diff
@@ -0,0 +1,136 @@
...
+        assert cluster.min_time <= p['time'] <= cluster.max_time
+
+
+def cluster_spawnpoints(spawns, radius=70, time_threshold=240):
+    # 4 minutes is alright to grab a pokemon since most times are 30m+
+    clusters = cluster(spawns, radius, time_threshold)
+
+    try:
+        for c in clusters:
+            test(c, radius, time_threshold)
```
friscoMad: I don't see if it is possible for test to fail if everything was done properly.

pogom/cluster.py
```diff
@@ -0,0 +1,136 @@
...
+    except AssertionError:
+        raise
+
+    clusters.sort(key=lambda x: len(x))
```
friscoMad: why sort them?

pogom/cluster.py
Alderon86: better save than sorry?

pogom/schedulers.py
```diff
@@ -388,77 +411,36 @@ def _generate_locations(self):
...
-                late_by = cur_sec() - location['time']
-                appears = now() + 3600 - late_by
+        log.info('Tracking a total of %d spawn points.', len(self.locations))
 
-            location['appears'] = appears
-            location['leaves'] = appears + 900
+        # Cluster spawnpoints.
+        if self.args.ss_cluster:
+            self.locations = cluster_spawnpoints(
+                self.locations, 70, self.args.ss_cluster_time)
```
friscoMad: I am not sure of this but I think that 70m is the step size but I think it is a diameter and not radius. As said needs confirmation I am not really sure.

pogom/search.py
```diff
@@ -911,15 +911,49 @@ def search_worker_thread(args, account_queue, account_sets,
...
+                if extra_delay > 0:
+                    status['message'] = (
+                        'Too fast for {:6f},{:6f}; waiting {}s...'
+                        .format(step_location[0], step_location[1],
+                                extra_delay))
+                    log.info(status['message'])
+                    continue_time = time.time() + extra_delay
+
+                    while time.time() < continue_time:
+                        time.sleep(1)
```
friscoMad: why not just sleep for continue_time?

pogom/utils.py
```diff
@@ -303,6 +303,12 @@ def get_args():
                               'grid). Scans in a circle based on step_limit ' +
                               'when on DB.'),
                         nargs='?', const='nofile', default=False)
+    parser.add_argument('-ssc', '--ss-cluster',
+                        help='Cluster spawnpoints before scanning',
+                        action='store_true', default=False)
+    parser.add_argument('-ssct', '--ss-cluster-time',
```
friscoMad: Personally I hate those features that needs 2 flags one for enabling and another for config, I would say that just use ss-cluster-time and if value > 0 then enable the feature otherwise disable it.

pogom/cluster.py
friscoMad: well if we don't did it properly we raise an exception and it breaks everything. I will prefer being sure that the clustering code works 😄

pogom/cluster.py
neskk: I just adapted the code from a previous PR #1586 .  I guess this puts bigger clusters first but you're right, we don't need to sort, specially because after the clustering is done we sort the list again using `appears` key.

pogom/utils.py
neskk: I understand your criticism but this way it's harde  for someone to mess things up. If @sebastienvercammen agrees with you we'll change this.

pogom/search.py
neskk: Good question, this comes from another old PR #1476. I will change this.

pogom/utils.py
Alderon86: default value is fine, just start with -ssc. 
You just need to up your cluster time if u have a huge amount of spawns

pogom/cluster.py
neskk: If no one gets an AssertionException in their tests then I guess it's ok to remove this overhead test.

pogom/utils.py
neskk: @Alderon86, what @friscoMad was suggesting is removing `-ssc` parameter and change `-ssct` default value to 0 - spawnpoint clustering disabled. If anyone wants to enable the feature they have to specify a value, i.e., `-ssct 240`.
I'm ok either way ;)

pogom/cluster.py
Alderon86: havent seen a single one so far, and im using it quite a lot

Alderon86: havent seen a single one so far and im using it quite a lot.

pogom/utils.py
Obihoernchen: Only `-ssct` please.

neskk:
`-ssc` removed. To use spawnpoint clustering you must specify the time threshold between spawns, e.g. `-ssct 240`.
Since people that tested this PR didn't get any AssertionError we think that is safe to remove the "double-check" before returning clusters.
Created a new var `cluster_range` that is set accordingly to `no-jitter` parameter so that we never get skip a spawn point due to jittering.

Toetje585:
I deff want to test this but I must admit that it has been a while that I used spawn scanning, does it matter how you collect spawnpoints for example can I just use the ones that are in the database already?

When I got a quick look into the code I noticed you removed loading spawns from jsons files. What is your approach getting the spawn points for SS scanning?

Edit: Got it running, testing with 7k spawnpoints

pogom/search.py
```diff
@@ -37,6 +37,7 @@
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 from distutils.version import StrictVersion
+from geopy.distance import vincenty
```
Obihoernchen: Please don't  use geopy for this. It's kinda slow. See: https://github.com/RocketMap/RocketMap/pull/2192
We already have equi_rect_distance in utils.py. Is there a reason we can't use it?

Alderon86: This comes from an older pr, i will switch to chaversine

sebastienvercammen:
Waiting for #2192 to get merged.

pogom/search.py
```diff
@@ -880,11 +883,28 @@ def search_worker_thread(args, account_queue, account_sets,
                         break
 
                 # Grab the next thing to search (when available).
-                step, step_location, appears, leaves, messages, wait = (
+                step, next_location, appears, leaves, messages, wait = (
                     scheduler.next_item(status))
+
+                # We dont need to check distance during login.
+                if not first_run:
```
friscoMad: Why not just use `if not step_location:` and set it to none inside the first while?

friscoMad: If we are using speed this will always yield an empty sleep (but it will do the calcs and call sleep, not sure in python but a empty sleep usually is a hint to yield thread priority and we dont want that) on the other hand if we are using hex it will make it follow speed limits (which is not desirable for gym scans).
So probably it is a much better idea to just move this code to spawnpoint scheduler and return it with wait as it is doing -speed.

neskk: I didn't developed this speed limiter. This version has several limitations but I think it's not that easy to move everything to the scheduler. I actually preferred the old implementation of the speed limiter that was in this PR, before the latest "speed limiter" commit from Alderon. I don't have much time right now but will look into this 👍

neskk:
💤 it would be nice that someone tested the latest changes. I feel this is ready and I've been testing it in my servers without any problems.

sebastienvercammen:
@neskk This PR is already scheduled to be tested, but not for this iteration of merges.

pogom/cluster.py
```diff
@@ -0,0 +1,115 @@
...
+    max_time = max(cluster.max_time, spawnpoint['time'])
+
+    if max_time - min_time > time_threshold:
+        return float('inf')
+    else:
+        sp_position = (spawnpoint['lat'], spawnpoint['lng'])
+        return distance(sp_position, cluster.centroid)
+
+
+def check_cluster(spawnpoint, cluster, radius, time_threshold):
```
sebastienvercammen: Rename to less generic function name, `check_cluster` doesn't say what it does.

pogom/cluster.py
```diff
@@ -0,0 +1,115 @@
...
+
+            if check_cluster(p, c, radius, time_threshold):
+                c.append(p)
+            else:
+                c = SpawnCluster(p)
+                clusters.append(c)
+    return clusters
+
+
+def cluster_spawnpoints(spawns, radius=70, time_threshold=240):
```
sebastienvercammen: Function `cluster` versus `cluster_spawnpoints`: by their function names, it's not clear what they do differently.

pogom/schedulers.py
```diff
@@ -352,30 +352,57 @@ def __init__(self, queues, status, args):
...
-                    repr(e))
-
         # No locations yet? Try the database!
-        if not self.locations:
-            log.debug('Loading spawn points from database')
-            self.locations = SpawnPoint.get_spawnpoints_in_hex(
-                self.scan_location, self.args.step_limit)
+        if not self.locations and not self.args.no_pokemon:
+            log.debug('Loading spawn points from database.')
+            spawns = SpawnPoint.select_in_hex_by_location(
```
sebastienvercammen: Whitespace above this line to separate it from the debug message for readability.

pogom/schedulers.py
```diff
@@ -352,30 +352,57 @@ def __init__(self, queues, status, args):
...
+                self.scan_location, self.step_limit)
+            log.debug('Loaded %s spawn points from database.' % len(spawns))
```
sebastienvercammen: Whitespace above this line to separate the debug message from the rest for readability.

pogom/schedulers.py
```diff
@@ -352,30 +352,57 @@ def __init__(self, queues, status, args):
...
+
+            for sp in spawns:
+                '''
+                # Skip bad spawnpoints.
+                if sp['missed_count'] > 10:
+                    continue
+
+                # Skip spawnpoints without a valid TTH.
+                if not SpawnPoint.tth_found(sp):
+                    continue
```
sebastienvercammen: Why are these code comments here?

pogom/schedulers.py
```diff
@@ -352,30 +352,57 @@ def __init__(self, queues, status, args):
...
+                del sp['earliest_unseen']
+                del sp['latest_seen']
+                del sp['kind']
+                del sp['links']
+                sp['spawnpoint_id'] = sp['id']
+                del sp['id']
+                sp['lat'] = sp['latitude']
+                del sp['latitude']
+                sp['lng'] = sp['longitude']
+                del sp['longitude']
```
sebastienvercammen: Instead of all these deletes, consider just creating a new object with only the values we need. The reference to the old `sp` will be released when the next iteration starts.

pogom/schedulers.py
```diff
@@ -387,77 +414,36 @@ def _generate_locations(self):
...
-
-        if self.args.verbose:
-            for i in self.locations:
-                sec = i['time'] % 60
-                minute = (i['time'] / 60) % 60
-                m = 'Scan [{:02}:{:02}] ({}) @ {},{}'.format(
-                    minute, sec, i['time'], i['lat'], i['lng'])
-                log.debug(m)
+        if not self.locations:
+            raise Exception('No availabe spawn points!')
```
sebastienvercammen: `availabe` -> `available`, and it can end in a dot rather than an exclamation mark.

pogom/transform.py
```diff
@@ -100,3 +100,40 @@ def jitter_location(location=None, maxMeters=10):
     distance = math.sqrt(random.random()) * (float(maxMeters))
     destination = fast_get_new_coords(origin, distance, bearing)
     return (destination[0], destination[1], location[2])
+
+
+def intermediate_point(pos1, pos2, f):
```
sebastienvercammen: `intermediate_point`, can we get a clearer function name?

friscoMad:
While I do think this is was a great amount of work and it is working (maybe we can improve the code but I think it is mostly ok) the big issue here is that there is no reason to resurrect spawn-scan, we even tried to kill it some time ago.
When it is working is faster and clearer than speed and also leads to workers moving less around the map. But it is complex to setup properly (needs -bh, -wph, -bh does not play well with geofence) and it needs an initial scan that currently can only be done with speed. Tell that to a newby, we can merge it but it will be a fringe scheduler as it was before, just that right now it works if you know how.
If SB come back and this helps the issue we need still a lot of work to make it usable for everyone until then I just think there is no reason for merging this.
That's why I am not approving it.

neskk:
@friscoMad This is more a "quality of life" update, functionality was a bit outdated and as you said in the right circumstances it works. IMHO the whole program is a bit like that. We just added some extra "humpf" to what was in place and made sure it was "safe" to use.

I've been moving between rigs and haven't had time to deal with the changes suggested by Seb. Hopefully tomorrow I can quickly perform the changes and then I hope this PR can get merged ;)

sebastienvercammen:
@friscoMad @neskk Fixing up the old `-ss` helps to have a cleaner version of spawn scanning code. In the future, this will let us develop new code more easily and will help to separate initial scan ("speed" scan) and spawnpoint scanning (2nd phase, once we have the spawnpoints).

Separating them is perfect: we can rewrite code to start in `-speed` and swap to `-ss` when the spawnpoints are complete. No more manually choosing which one you'd like to go for: only one optimal version for each.

Cleaner code will also help to integrate long-term solutions such as GMO timing, better request handling, etc.

We'll also be able to make more positive guarantees for the future, whereas today we can only guarantee that no one wants to touch `-speed`'s code.

For this to work, `-ss` needs to do its job and there is no reason `-bh` shouldn't work with geofencing. We should work with better abstraction layers if they're conflicting.

pogom/cluster.py
```diff
@@ -0,0 +1,115 @@
...
+            self.min_time = spawnpoint['time']
+
+        elif spawnpoint['time'] > self.max_time:
+            self.max_time = spawnpoint['time']
+            self.spawnpoint_id = spawnpoint['spawnpoint_id']
+            self.appears = spawnpoint['appears']
+            self.leaves = spawnpoint['leaves']
+
+    def new_centroid(self, spawnpoint):
+        f = len(self._spawnpoints) / (len(self._spawnpoints) + 1.0)
```
friscoMad: Avoid doing the len twice

pogom/cluster.py
```diff
@@ -0,0 +1,115 @@
...
+        new_centroid = intermediate_point(
+            (spawnpoint['lat'], spawnpoint['lng']), self.centroid, f)
+
+        return new_centroid
+
+
+def cost(spawnpoint, cluster, time_threshold):
+    min_time = min(cluster.min_time, spawnpoint['time'])
```
friscoMad: Make it a regular method and while I don't like the name I can think of a better one 😄

pogom/cluster.py
friscoMad: ¿can_contain?
Also make it a regular object method.

pogom/cluster.py
friscoMad: Probably just mix both or rename this to `format_clusters(clusters)` and call from the other function.

pogom/schedulers.py
```diff
@@ -352,30 +352,57 @@ def __init__(self, queues, status, args):
             self.step_distance = 0.070
 
         self.step_limit = args.step_limit
-        self.locations = False
+        self.locations = []
+
+        self.cluster_range = 60
+        if self.args.no_jitter:
```
friscoMad: Nice add 👍

pogom/schedulers.py
```diff
@@ -471,9 +457,42 @@ def schedule(self):
...
+        if last_action and self.args.kph:
+            now_date = datetime.utcnow()
+            worker_loc = [status['latitude'], status['longitude']]
+            meters = distance(step_location, worker_loc)
+            wait = int(max(meters / self.args.kph * 3.6
+                           - (now_date - last_action).total_seconds(), 0))
+            if wait > 0:
+                wait_msg = 'Moving {}m to step {}, arriving in {}s.'.format(
+                    int(meters), step, wait)
+
```
friscoMad: I think we should add a late check here iterating through locations if the account can not reach there before `leaves`
something like `if (now() + wait) > leaves`  put the item back and try with the next one

pogom/search.py
```diff
@@ -914,11 +915,12 @@ def search_worker_thread(args, account_queue, account_sets,
                 if leaves and now() > (leaves - args.min_seconds_left):
                     scheduler.task_done(status)
                     status['skip'] += 1
-                    status['message'] = messages['late']
-                    log.info(status['message'])
-                    # No sleep here; we've not done anything worth sleeping
-                    # for. Plus we clearly need to catch up!
-                    continue
+                    if time.time() > (leaves - args.min_seconds_left):
```
friscoMad: Why are we doing a the double if?

pogom/transform.py
```diff
@@ -100,3 +100,40 @@ def jitter_location(location=None, maxMeters=10):
...
+    x = (a * math.cos(lat1) * math.cos(lon1) +
+         b * math.cos(lat2) * math.cos(lon2))
+    y = (a * math.cos(lat1) * math.sin(lon1) +
+         b * math.cos(lat2) * math.sin(lon2))
+    z = a * math.sin(lat1) + b * math.sin(lat2)
+
+    lat3 = math.atan2(z, math.sqrt(x**2 + y**2))
+    lon3 = math.atan2(y, x)
+
+    def normalize(pos):
```
friscoMad: Why not just merging the subfunction in the return?

pogom/schedulers.py
neskk: To brain storm, I don't really know if they are a good idea or not :?

pogom/schedulers.py
neskk: Need to think about this, maybe later today I test your idea ;)

neskk: Problem here is the multi-threading. Since we're using a queue this code is ok because even if multiple workers query the queue it behaves nicely. Adding the item back kinda screws the sorting etc. Iterating the whole queue and figuring out what's the best item to schedule next is kinda already implemented in speed scan.

pogom/utils.py
```diff
@@ -458,7 +458,7 @@ def get_args():
     parser.add_argument('-el', '--encrypt-lib',
                         help=('Path to encrypt lib to be used instead of ' +
                               'the shipped ones.'))
-    parser.add_argument('-odt', '--on-demand_timeout',
+    parser.add_argument('-odt', '--on-demand-timeout',
```
friscoMad: Undocumented change?
I am pr standarize to regular dash but it didn't expected something like this in this PR 😄

friscoMad:
If I am not wrong `get_spawnpoints_in_hex` it is not longer in use with this changes in the app so maybe we can remove it in this PR.

Codewise I think everything looks ok, I am unable to test it properly right now, I hope someone can do test it for some time.
Also maybe we need to check docs to be sure everything is updated with this changes, at the very least the command-line.md should be updated.

pogom/utils.py
neskk: It triggered my OCD and I couldn't resist to fix it since in my quick lookup I didn't find more typos. Glad you're doing a standardization of these names, I also don't like the new `--enable-clean` being `-DC` :D
Maybe I should revert this, it's a bit out of place.

neskk:
@friscoMad I was just reviewing the code and there's a part of SpeedScan initial mapping that still uses `select_in_hex_by_location`. IMHO `get_spawnpoints` does "too" many things at once so it loses a bit o flexibility. I think it´s easier to get rid of `get_spawnpoints` instead of `select_in_hex_by_locations`.
@Alderon86 nice work with the documentation 👍

docs/extras/spawnpoint-scanning-scheduler.md
```diff
@@ -1,47 +1,39 @@
...
-### Dump scans from database then use the created file
+### Scans with Spawnpoint Clustering
 
 ```
-python runserver.py -ss YOURFILE.json -l YOURLOCATION -st STEPS --dump-spawnpoints
+python runserver.py -ss -ssct YOURVALUE -l YOURLOCATION -st STEPS
 ```
 
-Where YOURFILE.json is the file containing all the spawns, YOURLOCATION is the location the map should be centered at and also the center of the hex to get spawn locations from and -st sets the size of the clipping hexagon (hexagon is the same size as the scan of the same -st value)
+Where YOURLOCATION is the location the map should be centered at and also the center of the hex to get spawn locations from, -st sets the size of the clipping hexagon (hexagon is the same size as the scan of the same -st value), -ssct (Spawnpoint Cluster Time) sets a Time threshold for spawn point clustering. A Value around 200 is recommended.
```
fosJoddie: Description is lacking a time unit for spawnpoint cluster time. 
There should also absolutely be a part here that explain that the cluster time will try to schedule scans at the same position within `ssct` amount of seconds to catch multiple spawns at once.

docs/first-run/commandline.md
```diff
@@ -250,10 +250,13 @@
                             into local db). [env var: POGOMAP_NO_RAIDS]
       -nk, --no-pokestops   Disables PokeStops from the map (including parsing
                             them into local db). [env var: POGOMAP_NO_POKESTOPS]
-      -ss [SPAWNPOINT_SCANNING], --spawnpoint-scanning [SPAWNPOINT_SCANNING]
+      -ss, --spawnpoint-scanning
                             Use spawnpoint scanning (instead of hex grid). Scans
                             in a circle based on step_limit when on DB. [env var:
                             POGOMAP_SPAWNPOINT_SCANNING]
+      -ssct SS_CLUSTER_TIME, --ss-cluster-time SS_CLUSTER_TIME
```
fosJoddie: Description is missing units

neskk:
Rebased again, it's working with latest develop branch.

PR #2239, opened 2017-08-06
Fix tutorial, improve inventory parsing, fix releasing buddy

### LINKED ISSUE #2158: -tut not setting buddy
After the protobuf rework inventory parsing during tut will fail and not set your new friend as a buddy.
Also we need to test if inventory comes even when we are not asking for it and that it does contain a new poke as due to the last_timestamp maybe we need to do something to get it.

friscoMad:
## Description
This PR was meant to only fix tutorial but I did add some other fixes and refactors as adding them as another PR will only conflict with this.

So main thing, tutorial process follows current requests standards, based on node implementation.
This means adding additional subrequests to every request, adding some more get_player requests, etc. 
The only difference with node is that we are not getting the list of available options for avatar.

Also the main thing that was broken was setting buddy, in the request we had in tutorial encounter we tried to parse inventory while we were not requesting it so it was impossible to set a buddy.

So I just moved that part to login as there was no reason to keep it at tutorial and I needed to know if we already had a buddy and that was in player data.
So currently after login it will set a random buddy if you don't have one.

While I was at it I found other errors:

- Release code did not check if it was trying to release our buddy which will fail (it will also fail if we have favs but I don't expect that), this is still untested as I did not have an account with that many pokes.
- Inventory response was not being parsed in the get level rewards during login, as I have created a send_generic_request I just handled inventory there and used it thorought the whole login (I will be great to add it to other request but I have retrained to avoid scope creep.
- Inventory parsing was not logging the real total of items
- Step 6 get player is only done if no tutorial is needed.

Future work:

- use `send_generic_request` for all requests 

**Seb edit:** Fixes #2158.

## Motivation and Context
Currently tutorial was not setting buddy and found other issues during the process

## How Has This Been Tested?

Local instance with several fresh accounts 

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -154,9 +154,8 @@ def rpc_login_sequence(args, api, account):
     try:
         req = api.create_request()
         req.get_player(player_locale=args.player_locale)
-        response = req.call(False)
-
-        parse_get_player(account, response)
+        resp = req.call(False)
+        parse_get_player(account, resp)
 
         total_req += 1
         time.sleep(random.uniform(.53, 1.1))
@@ -180,16 +179,8 @@ def rpc_login_sequence(args, api, account):
         req = api.create_request()
         req.download_remote_config_version(platform=1,
                                            app_version=app_version)
-        req.check_challenge()
-        req.get_hatched_eggs()
-        req.get_inventory(last_timestamp_ms=0)
-        req.check_awarded_badges()
-        req.download_settings()
-        response = req.call(False)
-
-        parse_download_settings(account, response)
-        parse_new_timestamp_ms(account, response)
-        parse_inventory(account, response)
+        send_generic_request(req, account, settings=True, buddy=False,
+                             inbox=False)
 
         total_req += 1
         time.sleep(random.uniform(.53, 1.1))
@@ -220,16 +211,8 @@ def rpc_login_sequence(args, api, account):
                 paginate=True,
                 page_offset=page_offset,
                 page_timestamp=page_timestamp)
-            req.check_challenge()
-            req.get_hatched_eggs()
-            req.get_inventory(last_timestamp_ms=account[
-                'last_timestamp_ms'])
-            req.check_awarded_badges()
-            req.download_settings(hash=account[
-                'remote_config']['hash'])
-            response = req.call(False)
-
-            parse_new_timestamp_ms(account, response)
+            resp = send_generic_request(req, account, settings=True,
+                                        buddy=False, inbox=False)
 
             req_count += 1
             total_req += 1
@@ -243,13 +226,13 @@ def rpc_login_sequence(args, api, account):
 
             try:
                 # Re-use variable name. Also helps GC.
-                response = response['responses']['GET_ASSET_DIGEST']
+                resp = resp['responses']['GET_ASSET_DIGEST']
             except KeyError:
                 break
 
-            result = response.result
-            page_offset = response.page_offset
-            page_timestamp = response.timestamp_ms
+            result = resp.result
+            page_offset = resp.page_offset
+            page_timestamp = resp.timestamp_ms
             log.debug('Completed %d requests to get asset digest.',
                       req_count)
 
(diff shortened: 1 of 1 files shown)


Alderon86:
i like the generic_request part :D 
will be tested soonish!

pogom/account.py
```diff
@@ -410,22 +380,20 @@ def complete_tutorial(args, api, account):
...
-    time.sleep(random.uniform(0.5, 0.6))
-    req = api.create_request()
-    req.get_player_profile()
-    log.debug('Fetching player profile for %s...', account['username'])
-    req.call(False)
+        time.sleep(random.uniform(0.5, 0.6))
+        req = api.create_request()
+        req.get_player_profile()
+        log.debug('Fetching player profile for %s...', account['username'])
+        send_generic_request(req, account, inbox=False)
```
sebastienvercammen: @friscoMad Indendation change, can you confirm this is intentional?

pogom/account.py
```diff
@@ -700,15 +651,17 @@ def parse_inventory(api, account, api_response):
...
-        'Parsed %s player inventory: %d items, %d pokemons, %d available ' +
-        'eggs and %d available incubators.',
-        account['username'], parsed_items, parsed_pokemons, parsed_eggs,
-        parsed_incubators)
-    log.debug('Total amount in Inventory:' +
-              ' {} Items, {} pokemon, {} eggs, {} Incubator'.format(
-                int(parsed_items + item_count), len(account['pokemons']),
-                len(account['eggs']), len(account['incubators'])))
+    if log.isEnabledFor(logging.DEBUG):
+        log.debug(
```
sebastienvercammen: Why do we check `log.isEnabledFor` if we then call `log.debug` anyway (which will redirect the message to the proper log level, aka filtered if it's not DEBUG)?

sebastienvercammen:
Fixes #2158.

pogom/account.py
friscoMad: Intentional, at least in node implementation that request is only done if we need to complete this step.

pogom/account.py
friscoMad: It is just to avoid calculating the total items if we are not going to use it.
I don't think it is really an expensive calculation but we are doing it on every request during login (and soon in every request when we refactor all requests to use the send_generic_request) so I thought it make sense.

pogom/account.py
```diff
@@ -381,6 +336,16 @@ def rpc_login_sequence(args, api, account):
     log.info('RPC login sequence for account %s successful with %s requests.',
              account['username'],
              total_req)
+
+    time.sleep(random.uniform(3, 5))
+
+    if account['buddy'] == 0 and len(account['pokemons']) > 0:
```
michikrug: Is `account['pokemons']` already known here?

pogom/account.py
```diff
@@ -724,6 +677,7 @@ def clear_inventory(api, account):
     release_count = int(total_pokemon - 5)
     if total_pokemon > random.randint(5, 10):
         release_ids = random.sample(account['pokemons'].keys(), release_count)
+        release_ids.remove(account['buddy'])
```
michikrug: This will throw a `ValueError` if `account['buddy']` is not in `release_ids`.

pogom/account.py
friscoMad: Yes it comes with the first inventory request that is parsed, and we have a lot before this.

pogom/account.py
friscoMad: Make sense, we can fix it with `if account['buddy'] in release_ids:`, right?

michikrug: Correct

pogom/account.py
michikrug: Ok :)

pogom/account.py
```diff
@@ -967,6 +921,30 @@ def parse_level_up_rewards(api, account):
...
+    req.check_challenge()
+    req.get_hatched_eggs()
+    req.get_inventory(last_timestamp_ms=account['last_timestamp_ms'])
+    req.check_awarded_badges()
+
+    if settings:
+        if 'remote_config' in account:
+            req.download_settings(hash=account['remote_config']['hash'])
+        else:
+            req.download_settings()
```
Alderon86: im not sure if we should sent an empty request if remote_config is not in account

friscoMad: it is for the first download settings, currently we do not set the hash parameter and remote_config is only defined when we parse the a download setting response. That way we sent it empty the first request and then the last hash which is the intended behaviour.

I wanted to add a default remote_config with hash to None in reset account but I discarded it as I was unsure if it will would work due to pgoapi paramater handling.

Alderon86: okay good!
reseting remote config shouldnt be done every login, but seb is already working on a solution to store the values.

PR #2242, opened 2017-08-07
More permissive status-name and access logs moved into logs path.

neskk:
## Description

Improved the way we check if status-name is valid.
Keep things tidy. It makes more sense to output client access logs inside the same folder as the other logs so we can easily check in one place everything we need.

## Motivation and Context

**Status-Name**

It annoyed me that a simple ` - ` on the status name won't allow the map to start.
Using a simple regex to allow alpha-numeric characters and some special characters: `\s` (blank spaces), `-`, `_` and `.` on the status-name.

**Access Logs**
Seems logic that all logs are output to the same path. When using `-al` not it will output a timestamped filename ending in `_access.log`, e.g. `20170808_0720_access.log`.

## How Has This Been Tested?

Local instance.

## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -478,8 +478,14 @@ def set_log_and_verbosity(log):
 
     # Web access logs.
     if args.access_logs:
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
+        date = strftime('%Y%m%d_%H%M')
+        filename = os.path.join(
+            args.log_path, '{}_{}_access.log'.format(date, args.status_name))
+
         logger = logging.getLogger('werkzeug')
-        handler = logging.FileHandler('access.log')
+        handler = logging.FileHandler(filename)
         logger.setLevel(logging.INFO)
         logger.addHandler(handler)
 
--- a/runserver.py
+++ b/runserver.py
@@ -194,8 +194,9 @@ def main():
 
     args = get_args()
 
-    # Abort if status name is not alphanumeric.
-    if not str(args.status_name).isalnum():
+    # Abort if status name is not valid.
+    regexp = re.compile('^([a-zA-Z0-9\s\-\_\.]+)$')
+    if not regexp.match(args.status_name):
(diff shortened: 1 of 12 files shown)


runserver.py
```diff
@@ -194,8 +194,9 @@ def main():
 
     args = get_args()
 
-    # Abort if status name is not alphanumeric.
-    if not str(args.status_name).isalnum():
+    # Abort if status name is not valid.
+    regexp = re.compile('^([a-zA-Z0-9\s\-\_\.]+)$')
```
sebastienvercammen: Regex can be simplified:
```
^([\w\s\-.]+)$
```
where `\w` equals any alphanumeric character (`[a-zA-Z0-9_]`) and the dot doesn't need to be escaped because it's explicit in a character class.

To be tested, of course. It's 1.30am and I just got home from my birthday party. 😊

runserver.py
```diff
@@ -194,9 +194,10 @@ def main():
 
     args = get_args()
 
-    # Abort if status name is not alphanumeric.
-    if not str(args.status_name).isalnum():
-        log.critical('Status name must be alphanumeric.')
+    # Abort if status name is not valid.
+    regexp = re.compile('^([\w\s\-\.]+)$')
```
sebastienvercammen: Dot doesn't need to be escaped. I don't think we intentionally want the backslash character `\` to be allowed in status name.

runserver.py
```diff
@@ -194,9 +194,10 @@ def main():
...
+    if not regexp.match(args.status_name):
+        log.critical('Status name contains ilegal characters.')
```
sebastienvercammen: Typo:

> ilegal

runserver.py
neskk: Forgot about that. Was too late/early in the morning :)

tomballgithub:
Thank you, I like to use underscore (_) in my -sn for legibility..

PR #2243, opened 2017-08-07
Disable asset and item templates requests on login.

neskk:
## Description

It defeats a bit the purpose of having them in the first place but when you're running a tight on RPM it can potentially save you from a bottleneck.

## Motivation and Context

Run instance with `-nar`, `--no-asset-requests` to disable these requests.

## How Has This Been Tested?

Local instance.

## Types of changes

## Checklist:


Alderon86:
we should store them in the database instead of disabling it, so other workers can reuse the values.

neskk:
It would be a waste of db operations, I've thought about saving them in a `json` file, similarly to what node pogo api does, but afaik these values aren't used anywhere. Maybe on complete_tutorial we can assert the model's URL but that almost never changes.
If we start using these values for anything then we should store them, otherwise disabling them can help save a some RPMs.

sebastienvercammen:
Adding comments from Discord:

> [1:19 AM] Uncle Sebby: @Alderon86 I already have it on my list to add abstraction to the client requests to store account data (e.g. asset time, template time, the 32 requests) in a local cache folder (cache data per account, just md5 the account name for a filename). Having added pgoapiwrapper and pgoapirequestwrapper makes this super easy. Definitely filesystem over DB.
> [1:19 AM] Uncle Sebby: You can pick it up though, I'm still in Vienna till the evening of the 9th.

The requests should be complete, but only done once where possible.

michikrug:
I am storing the timestamps for each account in a JSON file on the disk so that the accounts only do the requests if actually something was changed. Thus, fresh accounts do it once and if nothing changes never again and additionally, it also survives restarts of the instance.

PR #2244, opened 2017-08-07
Move access logs (-al) into log's path.

neskk:
## Description

Keep things tidy. It makes more sense to output client access logs inside the same folder as the other logs so we can easily check in one place everything we need.

## Types of changes

## Checklist:


runserver.py
```diff
@@ -477,8 +477,14 @@ def set_log_and_verbosity(log):
 
     # Web access logs.
     if args.access_logs:
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
```
sebastienvercammen: This already gets done in `set_log_and_verbosity()` at the start of `main()`, although it would need to be moved out of the `if not args.no_file_logs` block.

neskk:
Merged into #2242

PR #2247, opened 2017-08-08
Fix sprite sizing in non-english locales.

### LINKED ISSUE #2228: Scaling Pokemon icons based on rarity does not work for non-english confugurations
Since the rarity string for each Pokemon is localized on server-side but the check for scaling the icons based on their rarity on the client-side is only done for english strings, it does not work, if the RocketMap language is set to something different than english.

## Expected Behavior
Pokemon icons should be scaled based on their rarity independent of the configured language.

## Current Behavior
Pokemon icons are not scaled based on their rarity when, i.e. setting the language to german (others not tested).

## Possible Solution

Get rid of localized rarity values.

1. configure RocketMap to use german language
2. open map in browser
3. look fore very/ultra rare pokemon
4. see that they are not bigger

pogo-excalibur:
## Description
Adds a new method `get_pokemon_rarity_index` to `utils.py` to provide a locale-inspecific rarity value.
This is passed to `raw_data` as `pokemon['pokemon_rarity_index']`.

## Motivation and Context
Sprite scaling based on rarity is currently hardcoded to english strings. Consequently icon scaling doesn't work in locales other than english. Fixes #2228.

(Arguably, the rarity values in `data/pokemon.json` shouldn't be hardcoded to english in the first place.)

## How Has This Been Tested?
Running a map with `locale: fr`.

frenchy

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -27,8 +27,9 @@ from cachetools import cached
 from timeit import default_timer
 
 from . import config
-from .utils import (get_pokemon_name, get_pokemon_rarity, get_pokemon_types,
-                    get_args, cellid, in_radius, date_secs, clock_between,
+from .utils import (get_pokemon_name, get_pokemon_rarity_index,
+                    get_pokemon_rarity, get_pokemon_types, get_args,
+                    cellid, in_radius, date_secs, clock_between,
                     get_move_name, get_move_damage, get_move_energy,
                     get_move_type, calc_pokemon_level)
 from .transform import transform_from_wgs_to_gcj, get_new_coords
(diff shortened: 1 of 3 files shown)


pogo-excalibur:
Apologies to @BrownSlaughter if you're already on this.

BrownSlaughter:
I have no had the time to start anything so its all good

friscoMad:
Nice find but I would change the solution 180 degrees, always return english data and localize in the client, this is not what we are doing with everything else but we should go that route in general.
Just return english rarity (or not rarity at all) and translate it for display using `i8ln` in client or just use `idToPokemon` that have all poke data from an ID (now it stores the rarity translated but you could store it in english too).

sebastienvercammen:
@pogo-excalibur @BrownSlaughter On-going discussion about @friscoMad's comments, what are your thoughts? Tagged "on hold" as long as we haven't come to a conclusion.

PR #2248, opened 2017-08-08
webhook bug fix

richardfoltin:
## Description
Fixes two bugs in the creation of webhooks:
- pokestops wasn't sent to webhooks because of a typo
- when there was only 1 item waiting in the webhook queue that item wasn't sent to webhooks until another item would be added to the queue

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2130,7 +2130,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                     wh_pokestop = pokestops[f.id].copy()
                     wh_pokestop.update({
                         'pokestop_id': b64encode(str(f.id)),
-                        'last_modified_time': f.last_modified_timestamp_ms,
+                        'last_modified': f.last_modified_timestamp_ms,
                         'lure_expiration': l_e,
                     })
                     wh_update_queue.put(('pokestop', wh_pokestop))
--- a/pogom/webhook.py
+++ b/pogom/webhook.py
@@ -68,6 +68,7 @@ def wh_updater(args, queue, key_caches):
     frame_interval_sec = (args.wh_frame_interval / 1000)
     frame_first_message_time_sec = default_timer()
     frame_messages = []
+    first_message = True
 
     # How low do we want the queue size to stay?
     wh_warning_threshold = 100
@@ -130,8 +131,9 @@ def wh_updater(args, queue, key_caches):
             now = default_timer()
             num_messages = len(frame_messages)
 
-            if num_messages == 1:
+            if num_messages == 1 and first_message:
                 frame_first_message_time_sec = now
+                first_message = False
 
             # If enough time has passed, send the message frame.
             time_passed_sec = now - frame_first_message_time_sec
@@ -144,6 +146,7 @@ def wh_updater(args, queue, key_caches):
                 send_to_webhooks(args, session, frame_messages)
 
                 frame_messages = []
+                first_message = True
 
             # Webhook queue moving too slow.
             if (not wh_over_threshold) and (


friscoMad:
This will fix the issue if there is a continous stream of wh that are discarded, we need to add also a timed get from the queue to wait at max X time to process the frame.
More precisely we need to use `get(True, timeout)` when there are items waiting, and without timeout when we have just sent a frame.

friscoMad:
@richardfoltin 
I tried to find the branch you used for the PR to add that code for you and it seems you did it online or at least I am unable to find it.

If you can't or don't want to update please tell us so we can help or in the worst case create a new PR based on this.

richardfoltin:
I hope I did it right. I'm not so good in python

friscoMad:
you have merged instead of rebasing and the PR is now a bit of a mess, also you will need some code to handle the when we get the exption, so is a bit more complex than I thought, I am trying to get it working and after that I can post to for you to copy.

richardfoltin:
I managed to revert the merge. Now how do I rebase?

friscoMad:
I don't think there will by any conflicts so, just do `git rebase origin/develop` (or upstream/develop or the RM remote name you used) then just do `git push -f`

friscoMad:
https://hasteb.in/fucideruco.py

This is the working solution with our requested changes, I needed to move some code to a function as the main loop was really too big and too indented to be legible.

richardfoltin:
@friscoMad commited the changes

friscoMad:
👍 I will test it in some minutes

friscoMad:
Working perfectly fine, thanks so much for the fix and the effort to improve it.

friscoMad:
@sebastienvercammen your suggestion had 2 issues:

1. On exception whtype, message = queue.get(true, frame_interval_sec) will leave whtype and message untouched so it will be skipped due to cache, it works but is work wasted and not clear just reading the code.

2. Without the branch in a setup with low volume of wh (the ones that had the problem in the first time) the wh thread will be woke up every half a second, to get the cache lock, and discard the same message again and again, so I think it is worth the branch.

Moving the code to a function was done as to fix 1 I needed to add another if and another indent and the function already had almost 200 lines and too many branches.

sebastienvercammen:
@friscoMad 

1. On exception, `whtype` and `message` don't exist (well, technically they do because Python defines it at function level, but they be referenced before assignment) and the loop starts over from the start. The problem is not that *"whtype and message are left untouched"* (by which I'm not even sure what you mean, the exception would simply jump to the catching block), it's that we still need to send the message if it contains a single message but timed out on the queue - which is why I said to use it as an example.
2. The solution to indentation (of which there isn't too much: without the `while True` and `try` blocks already there, there are hardly any additional indentation levels) is not to add more of `if: try: catch: else:`, followed by an `empty = True` inside the first block and `if empty:` afterwards.
3. If you want to reduce lines of code in a function, use proper abstraction. Don't just cut a large part out of a function, slap a function name on it and pass it all the parameters it requires from the previous scope. Use abstraction that makes sense in its implementation, for its user and for readability.

richardfoltin:
Travis build failed with the additional commits, so went back to my original pull request. I suppose my original commit fixes some of the bugs in the webhook creation. 

I don't really understand the additional problem, so I will leave this pull request in this state.

friscoMad:
1. As you said, defined at function level, we are inside a loop if there is an exception in get the values of the last loop are left in those vars, that's what I meant for left untouched.
2. Extra try catch is needed if we want a general failsafe as it was before. I added more ifs as the branches are needed but check if this version fits better: https://hasteb.in/gomafipiyi.py (used the else construct as it make sense to me having the except near the get, but I know I can add everything inside the try).
3. I was not trying to reduce the number of lines, just the complexity (https://en.wikipedia.org/wiki/Cyclomatic_complexity) as smaller/simpler functions are less errror prone, easier to understand and to mantain. It was not my PR and it was a simple fix with 4 lines, I thought that the split made sense but I guess I was wrong, sorry for that.

sebastienvercammen:
@friscoMad 

1. References to old objects are overwritten when the queue has an item and the other branch restarted `queue.get`.
2. Much better. Just move `num_messages = len(frame_messages)` up instead of getting the length twice.
3. You didn't reduce cyclomatic complexity, you increased it by adding more branches.

sebastienvercammen:
Due to work slowing down because I'm in the process of moving countries and this PR having been around for two weeks, I'm merging this PR as a hotfix for the bug. Any additional improvements should be in a separate PR.

PR #2258, opened 2017-08-11
Do a basic account rotation for encounters

friscoMad:
## Description

This does not handles more errors than before just splits the workload a bit better between all accounts in the file.
It just put the account to be used in a encounter at the bottom of the list, and as the accounts are checked from top to bottom that will make it a lot less probable to be used again soon.

## Motivation and Context

If RM is configured to do a low number of encounters then the same account is being used over an over again, if that account is banned or similar no encounters are finished properly.
Ciclyng over the whole list will allow to get some encounters done even with bad accounts.
Also accounts will be used less frequently which may or may not help with them getting shadowbanned.

## How Has This Been Tested?

Did just a couple of tests
I do not have a enc instance to properly test it.
I can see some possible weird situations (just putting in the bottom the wrong account, or skipping some accounts during checks) with thread context changes if the instance do call this function too much.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -1027,6 +1027,8 @@ class AccountSet(object):
                 account['last_coords'] = coords_to_scan
                 account['in_use'] = True
 
+                # Put the account the last one to force some rotation
+                self.sets[set_name].append(self.sets[set_name].pop(i))
                 return account
 
         # TODO: Instead of returning False, return the amount of min. seconds


LordJavi:
Nice one!

sebastienvercammen:
Can't be merged, it was a design decision not to rotate high-level accounts: rotating high-level accounts will indeed spread encounter requests over all accounts instead of just the ones needed to cover the area, currently resulting in a guaranteed account flag, followed by a shadowban (and in the past a permaban afterwards).

Instead of rotating, requests should be kept to the minimum amount of accounts, but handling of shadowbans/flags should be improved.

@friscoMad Is that rework something you'd like to do in this PR, or would you like me to close it?

friscoMad:
I don't see all those changes going in here, this was done as several people asked for it but I really don't mind to close it. It should be easy to just keep shadowbanned accounts "in_use" forever if you like that approach.

PR #2260, opened 2017-08-12
Removal of unnecessary "no encounter" information

sirlindqvist:
## Description
Got a ping from @BrownSlaughter earlier that people requested to strip down the Pokémon information window when encounter disabled.

## How Has This Been Tested?
On my own map.

image

## Types of changes

## Checklist:


static/js/map.js
```diff
@@ -543,32 +543,25 @@ function pokemonLabel(item) {
...
-            <span class='pokemon links exclude'><a href='javascript:excludePokemon(${id})'>Exclude</a></span>
-            <span class='pokemon links notify'><a href='javascript:notifyAboutPokemon(${id})'>Notify</a></span>
-            <span class='pokemon links remove'><a href='javascript:removePokemonMarker("${encounterId}")'>Remove</a></span>
           </div>
-      </div>
+        </div>
       <div class='pokemon container content-right'>
         <div>
-          <div class='pokemon disappear'>
+          <div class='pokemon disappear no encounter'>
```
friscoMad: I did suggest this also :D but finally I was happy as it was.
Anyway, use a class name not 2, no-encounter, no_encounter or noEncounter, not sure about what are the other class names, but using no.encounter in css is really weird.

dshoreman: One more vote for `no-encounter`. _pokemon.scss also uses `content-left` and `content-right`, and I'm sure snake-case classes are probably used over the rest of RM too. Two separate classes could open an ugly can of worms :D

Galacticaa: One more vote for no-encounter. _pokemon.scss also uses content-left and content-right, and I'm sure snake-case classes are probably used over the rest of RM too. Two separate classes could open an ugly can of worms :D

PR #2263, opened 2017-08-14
[REFACTOR] Use send_generic_request

friscoMad:
## Description

Main changes:

- Changed all requests (except one single call in captcha) to use send_generic_request
- Moved clear_dict_response to the send function, and moved all requests to its own module.
- Renamed all requests to the proto name.
- Added response logging at -vvv
- Merged parse_timestamp into parse_inventory
- Removed get_player_level as it is done in parse_inventory

Additional fixes/chages:

- Logger levels were not additive so -vvv was not -vv plus some more things, fixed that.
- Properly handled configuration result and now it raises an AccountBanned exception instead of nulltime

## Motivation and Context
As a continuation of #2239 it made sense to make all requests consistent, so this is it.

## How Has This Been Tested?

Limited tests with small instance (login, gmos and gym-info working).

## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -502,7 +502,7 @@ def spin_pokestop(api, account, args, fort, step_location):
             return False
 
         spin_result = response['responses']['FORT_SEARCH'].result
-        if spin_result is 1:
+        if spin_result == 1:
             log.info('Successful Pokestop spin with %s.',
                      account['username'])
             # Update account stats and clear inventory if necessary.
@@ -511,14 +511,14 @@ def spin_pokestop(api, account, args, fort, step_location):
             account['session_spins'] += 1
             incubate_eggs(api, account)
             return True
-        elif spin_result is 2:
+        elif spin_result == 2:
             log.debug('Pokestop was not in range to spin.')
-        elif spin_result is 3:
+        elif spin_result == 3:
             log.debug('Failed to spin Pokestop. Has recently been spun.')
-        elif spin_result is 4:
+        elif spin_result == 4:
             log.debug('Failed to spin Pokestop. Inventory is full.')
             clear_inventory(api, account)
-        elif spin_result is 5:
+        elif spin_result == 5:
             log.debug('Maximum number of Pokestops spun for this day.')
         else:
             log.debug(
@@ -565,11 +565,11 @@ def clear_inventory(api, account):
         release_response = release_p_response['responses']['RELEASE_POKEMON']
         release_result = release_response.result
 
-        if release_result is 1:
+        if release_result == 1:
             log.info('Sucessfully Released %s Pokemon', len(release_ids))
             for p_id in release_ids:
                 account['pokemons'].pop(p_id, None)
-        elif release_result != 1:
+        else:
             log.error('Failed to release Pokemon.')
 
     for item_id, item_name in items:
@@ -589,12 +589,12 @@ def clear_inventory(api, account):
 
             clear_response = resp['responses']['RECYCLE_INVENTORY_ITEM']
             clear_result = clear_response.result
-            if clear_result is 1:
+            if clear_result == 1:
                 log.info('Clearing %s %ss succeeded.', item_count,
                          item_name)
-            elif clear_result is 2:
+            elif clear_result == 2:
                 log.debug('Not enough items to clear, parsing failed.')
-            elif clear_result is 3:
+            elif clear_result == 3:
                 log.debug('Tried to recycle incubator, parsing failed.')
             else:
                 log.warning('Failed to clear inventory.')
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -626,10 +626,10 @@ def incubate_eggs(api, account):
 def parse_level_up_rewards(api, account):
     resp = level_up_rewards(api, account)
     result = resp['responses']['LEVEL_UP_REWARDS'].result
-    if result is 1:
+    if result == 1:
         log.info('Account %s collected its level up rewards.',
                  account['username'])
-    elif result is 2:
+    elif result == 2:
         log.debug('Account %s already collected its level up rewards.',
                   account['username'])
     else:
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -624,17 +624,17 @@ def incubate_eggs(api, account):
 
 
 def parse_level_up_rewards(api, account):
-    try:
-        resp = level_up_rewards(api, account)
-
-        if resp['responses']['LEVEL_UP_REWARDS'].result is 1:
-            log.info('Account %s collected its level up rewards.',
-                     account['username'])
-        else:
-            log.debug('Account %s already collected its level up rewards.',
-                      account['username'])
-    except Exception as e:
-        log.exception('Error during getting Level Up Rewards %s.', e)
+    resp = level_up_rewards(api, account)
+    result = resp['responses']['LEVEL_UP_REWARDS'].result
+    if result is 1:
+        log.info('Account %s collected its level up rewards.',
+                 account['username'])
+    elif result is 2:
+        log.debug('Account %s already collected its level up rewards.',
+                  account['username'])
+    else:
+        log.error('Error collecting rewards of account %s.',
+                  account['username'])
 
 
 # The AccountSet returns a scheduler that cycles through different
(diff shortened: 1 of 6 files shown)


pogom/apiRequests.py
```diff
@@ -0,0 +1,280 @@
...
+
+import logging
+
+from pgoapi.utilities import f2i, get_cell_ids
+from pgoapi.hash_server import BadHashRequestException, HashingOfflineException
+
+from .transform import jitter_location
+
+log = logging.getLogger(__name__)
+logResp = logging.getLogger(__name__+'.resp')
```
friscoMad: I don't like this but using the main logger with log.log(5, message) and enabling level 5 at runserver did not work (even setting a new level name).

pogom/apiRequests.py
```diff
@@ -0,0 +1,280 @@
...
+def catchException(task):
+
+    def _catch(function):
+
+        def wrapper(*args, **kwargs):
+            try:
+                return function(*args, **kwargs)
+            except Exception as e:
+                log.exception('Exception while %s with account %s: %s.', task,
+                              kwargs.get('account', args[1])['username'], e)
```
friscoMad: I also don't like this too much (using arg by postion), it seems to be ways to get argument by name using inspection but that seems really costly and all of our functions need the api and the account account for timestamp so keeping them in the same order shouldn't be a problem.

pogom/apiRequests.py
```diff
@@ -0,0 +1,280 @@
...
+    download_settings['template_time'] = template_time
+
+    account['remote_config'] = download_settings
+
+    log.debug('Download settings for account %s: %s.', account['username'],
+              download_settings)
+
+
+# Parse player stats and inventory into account.
+def parse_inventory(account, api_response):
```
friscoMad: I am not sure about moving the parse here but it is onl used here and leaving it at account lead me to a cyclic import issue.
If we change later to an account class we could move this back to the class.

pogom/account.py
```diff
@@ -755,197 +623,20 @@ def incubate_eggs(api, account):
...
-        response = response['responses']['LEVEL_UP_REWARDS']
-        result = response.result
-        if result is 1:
+        resp = level_up_rewards(api, account)
+
+        if resp['responses']['LEVEL_UP_REWARDS'].result is 1:
             log.info('Account %s collected its level up rewards.',
                      account['username'])
-        elif result != 1:
+        else:
```
neskk: I think it would be wise to at least have a debug message when status is 0. Since you're improving code a lot, just a little adjustment here would be good. Test if result is 1 (ok), test if result is 2 (already collected) and log an error otherwise.

pogom/apiRequests.py
neskk: This is the only thing that really makes me a bit itchy, maybe there's a more "clean" solution?

friscoMad: I really don't understand why it is needed, maybe you can give me a clue but using numbers even if it is supported by the docs does not seem to work, not sure why, I thought the problem was the level name in the logger format but giving it a name did not work either.

neskk: https://docs.python.org/2/library/logging.html#logging-levels
I did some tests, mostly trying the obvious stuff as you probably did and I think I got it working to some level.
Since `log.debug` numeric value is 10 and from what I understood you wanted a lower level for responses so I decided to use the "custom" level 5.
1. removed L12 `logResp = logging.getLogger(__name__+'.resp')`
2. changed L63 to `log.log(5, 'Response: %s', resp)`

On runserver.py @ set_log_and_verbosity():
```
-    logging.getLogger('pogom.apiRequests.resp').setLevel(logging.INFO)
+    logging.getLogger('pogom.apiRequests').setLevel(logging.INFO)
```
and
```
-        logging.getLogger('pogom.apiRequests.resp').setLevel(logging.DEBUG)
+        logging.getLogger('pogom.apiRequests').setLevel(5)
```
On the log file we can find the messages logged at level 5, they are just not print to the console. 
Changed L51 of runserver.py to: `stdout_hdlr.setLevel(5)` now it works properly and if `-vvv` is used responses are shown both in log files and console.

pogom/apiRequests.py
```diff
@@ -0,0 +1,280 @@
...
+            else:
+                if p_data.egg_incubator_id:
+                    # Egg is already incubating.
+                    continue
+                account['eggs'].append({
+                    'id': p_id,
+                    'km_target': p_data.egg_km_walked_target
+                })
+                parsed_eggs += 1
+    if log.isEnabledFor(logging.DEBUG):
```
neskk: I removed this line `if log.isEnabledFor(logging.DEBUG):` and it seems to be working well. I think I read somewhere a discussion about this but I don't remember what was the justification to have this.

pogom/account.py
friscoMad: I don't understand this comment it already logs if result is not 1 I just changed from 
```
if x = 1:
elif x != 1:
```
to:
```
if x=1:
else:
```
I think the beahiovour should be exactly the same as before. isn't it?

pogom/apiRequests.py
friscoMad: Yup this works without that line it is just to avoid executing the lambda when we are not going to use the result, that code is run in every single request and most people does not use debug so there is no need to waste those msecs in doing it 😀

pogom/apiRequests.py
friscoMad: Thanks so much 👍  I did not check the file so I thought that there was something inherently wrong on what I was trying to do, I have no idea we had an output filter in place.
I feel a lot better now

pogom/apiRequests.py
neskk: Aaaahhh, now it's obvious! But... having an `if` condition is also a bit of an overhead. I would probably change that debug line to just indicate how many items, pokes, eggs, incubators were parsed from current response. Having the totals is not that important and we can perhaps improve the status printer to display that kind of info.
Anyway I'm just in favor of simpler code and better separation of concerns ;)

pogom/account.py
neskk: The behavior remains the same, but I wanted an improvement... Basically I think it's useful to distinguish between `0` (error), `1` (ok) and `2` (already collected). As it is, when the response status code is `0` it is logged as "account already collected its level up rewards".
I guess it's OK to do it this way because if everything goes as it should, the response should never contain status code `0` but weird stuff can happen...

sirlindqvist:
Lot's of code, I'm not _that_ reviewer :) Running 72h+ without any noticeable errors.

pogom/apiRequests.py
sebastienvercammen: @friscoMad @neskk Individual item totals are irrelevant to RM, we can just remove the totals and the `if log.isEnabledFor(...)`.

pogom/apiRequests.py
```diff
@@ -0,0 +1,280 @@
...
+            parsed_items, parsed_pokemons, parsed_eggs, parsed_incubators)
+        total_items = reduce(lambda x, value: x + value,
+                             account['items'].itervalues(), 0)
+        log.debug('Total amount in Inventory: %d Items, %d pokemon, %d eggs,' +
+                  ' %d Incubator', total_items,
+                  len(account['pokemons']),
+                  len(account['eggs']), len(account['incubators']))
+
+
+def catchException(task):
```
sebastienvercammen: Consider renaming `catchException` to the less generic `catchRequestException`.

neskk:
Running smooth 👍

pogom/account.py
```diff
@@ -755,195 +623,18 @@ def incubate_eggs(api, account):
...
-    resp = req.call(False)
-    parse_inventory(account, resp)
-    parse_new_timestamp_ms(account, resp)
-    if settings:
-        parse_download_settings(account, resp)
-    clear_dict_response(resp)
-    return resp
+    resp = level_up_rewards(api, account)
+    result = resp['responses']['LEVEL_UP_REWARDS'].result
+    if result is 1:
```
sebastienvercammen: Are the additional checks with `is` rather than `==` necessary? If they are, can you add a comment above the code? And if they aren't, use `==` for consistency.

friscoMad:
@sebastienvercammen no more ❤️ for `is`

PR #2265, opened 2017-08-15
Show Expiration Times for Pokemon, Raids, Eggs, and Lures

tomballgithub:
## Description

Show the expiration time for pokemon, raids, and eggs.  
Change the countdown timer labels from hour/min/sec to h/m/s in 12-hr format.

## Motivation and Context

The map currently does not show the expiration time of pokemon, raids, or eggs.
This is useful information to directly present to the user versus requiring them to do math.

## How Has This Been Tested?

This has been tested on my public map for 2 weeks
My users have been testing the feature without issues.  It is very simply.
The operating system is Ubuntu 16.04 on a VPS

image

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -773,7 +773,7 @@ function pokestopLabel(expireTime, latitude, longitude) {
                 Lured Pokéstop
               </div>
               <div class='pokestop-expire'>
-                  <span class='label-countdown' disappears-at='${expireTime}'>00m00s</span> left
+                  <span class='label-countdown' disappears-at='${expireTime}'>00m00s</span> left (${moment(expireTime).format('HH:mm')})
               </div>
               <div>
                 <img class='pokestop sprite' src='static/images/pokestop//PokestopLured.png'>
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1851,7 +1851,7 @@ var updateLabelDiffTime = function () {
         if (disappearsAt.ttime < disappearsAt.now) {
             timestring = '(expired)'
         } else {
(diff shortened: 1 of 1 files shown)


sirlindqvist:
Shouldn't be 12hour format. Many people use 24h format.

Also, the format hour/min/sec was voted over h/m/s :)

gendelbendel:
Why not leave timestamp formatting up to config? Then people can choose if they want 24hr or 12hr, defaulting to one or the other.

tomballgithub:
I can change it to 24hr time but h/m/s/ saves space over hour/min/sec and that was one of the reasons given for the expiration time not being provided.

sebastienvercammen:
The agreed format was indeed 24-hour `hh:mm:ss left (hh:mm)`. Even in the US military time is a thing and people should get used to better/clearer formats... *like metric, come on US...*

tomballgithub:
Changed time to hh:mm:ss format

sirlindqvist:
I can't get the time in () to show at all on pokemons..??

tomballgithub:
Can you share an example? @sirlindqvist

tomballgithub:
image

tomballgithub:
Nice catch, fixed...

dripdropper:
I'm not a JavaScript expert, but it looks like you're using Moment.js, which appears to support locale-based time formatting.  If so, I think this will allow the time to be displayed based on the user's OS/browser settings (e.g., 12 or 24-hour):
moment().format('LT');
IMO, this is the "correct" solution from a UX standpoint.

Edit:  Upon deeper inspection, you might have to tell Moment.js to use the user's locale via something like:
var locale = window.navigator.userLanguage || window.navigator.language;
moment.locale(locale);

PR #2273, opened 2017-08-26
Close DB after queries to properly use the pool of connections

friscoMad:
## Description
The main change is trying to close the db after each query.
Also removed the db-max_connections param as it makes no sense, see https://github.com/RocketMap/RocketMap/pull/2273#issuecomment-325293182 for details.
Logging procedure is changed a bit to clean the status reset procedure and to avoid requesting a lot of db connnections at once.

## Motivation and Context
This is done as peewee by default will not close connections rendering the pool useless, using flaskUtils do close the db after a webserver request but for long lasting threads (like workers) there were no control to close them.
This change do reduce the number of open connections from 170 to 3-4 in my setup (no web) ~there is still a spike when all workers are created that I need to address otherwise~ it will be doable to set have at max 10 simultaneus connections in most of the setups (+ web server that heavily depends on user count).
Also it will reduce the number of open sockets so it will help the systems that have problems with too many open files.

## How Has This Been Tested?

Have been running for days in my main instance, Mysql is happier but couldn't see clear improvements in memory usage from RM or Mysql.

whm host show mysql processes - 66 0 23_2017-09-19_23-07-36

Connections using the previous version on main instance: 140 workers - 5 open connections and test instance with latest version 18 workers - 2 connections

whm host show mysql processes - 66 0 23_2017-09-19_23-13-03

Same config in latest develop 18 workers - 23 connections.

## Types of changes

## Checklist:

### DIFF
--- a/docs/common-issues/faq.md
+++ b/docs/common-issues/faq.md
@@ -155,22 +155,6 @@ InternalError(1054, u"unknown column 'cp' in 'field list'") or similar
 
 Only one instance can run when the database is being modified or upgraded. Run ***ONE*** instance of RM with `-cd` to wipe your database, then run ***ONE*** instance of RM (without `-cd`) to setup your database.
 
-#### Peewee max connections
-
-```
-ValueError: Exceeded maximum connections.
-```
-
-Try raising --db-max_connections, default is 5.
-
-#### MySQL max connections
-
-```
-OperationalError(1040, u'Too many connections')
-```
-
-You need to raise the maximum connections allowed on your MySQL server configuration, this is typically by setting `max_connections` in my.cnf or my.ini. Please use Google to find where this file is located on your specific operating system.
-
 #### SQLite query limit
 
 ```
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -89,6 +89,13 @@ def init_database(app):
 class BaseModel(flaskDb.Model):
 
     @classmethod
+    def database(cls):
+        return cls._meta.database
+
+
+class LatLongModel(BaseModel):
+
+    @classmethod
     def get_all(cls):
         results = [m for m in cls.select().dicts()]
         if args.china:
@@ -99,7 +106,7 @@ class BaseModel(flaskDb.Model):
         return results
 
 
-class Pokemon(BaseModel):
+class Pokemon(LatLongModel):
     # We are base64 encoding the ids delivered by the api
     # because they are too big for sqlite to handle.
     encounter_id = Utf8mb4CharField(primary_key=True, max_length=50)
@@ -230,9 +237,9 @@ class Pokemon(BaseModel):
 
         return pokemon
 
-    @classmethod
+    @staticmethod
     @cached(cache)
-    def get_seen(cls, timediff):
+    def get_seen(timediff):
         if timediff:
             timediff = datetime.utcnow() - timediff
         pokemon_count_query = (Pokemon
@@ -277,8 +284,8 @@ class Pokemon(BaseModel):
 
         return {'pokemon': pokemon, 'total': total}
 
-    @classmethod
-    def get_appearances(cls, pokemon_id, timediff):
+    @staticmethod
+    def get_appearances(pokemon_id, timediff):
         '''
         :param pokemon_id: id of Pokemon that we need appearances for
         :param timediff: limiting period of the selection
@@ -301,9 +308,9 @@ class Pokemon(BaseModel):
 
         return list(query)
 
-    @classmethod
-    def get_appearances_times_by_spawnpoint(cls, pokemon_id,
-                                            spawnpoint_id, timediff):
+    @staticmethod
+    def get_appearances_times_by_spawnpoint(pokemon_id, spawnpoint_id,
+                                            timediff):
         '''
         :param pokemon_id: id of Pokemon that we need appearances times for.
         :param spawnpoint_id: spawnpoint id we need appearances times for.
@@ -325,7 +332,7 @@ class Pokemon(BaseModel):
         return list(itertools.chain(*query))
 
 
-class Pokestop(BaseModel):
+class Pokestop(LatLongModel):
     pokestop_id = Utf8mb4CharField(primary_key=True, max_length=50)
     enabled = BooleanField()
     latitude = DoubleField()
@@ -423,7 +430,7 @@ class Pokestop(BaseModel):
         return pokestops
 
 
-class Gym(BaseModel):
+class Gym(LatLongModel):
     gym_id = Utf8mb4CharField(primary_key=True, max_length=50)
     team_id = SmallIntegerField()
     guard_pokemon_id = SmallIntegerField()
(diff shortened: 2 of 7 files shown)


pogom/search.py
```diff
@@ -786,18 +786,35 @@ def search_worker_thread(args, account_queue, account_sets,
...
+                'captcha':
+                    0,
+                'status':
+                    'active',
+                'message':
+                    'Switching to account {}.'.format(account['username'])
+            })
             log.info(status['message'])
-
             # New lease of life right here.
```
jagauthier: This comment doesn't make sense with the removal of the code below it.

pogom/search.py
```diff
@@ -1212,8 +1228,8 @@ def map_request(api, account, position, no_jitter=False):
     except BadHashRequestException:
         log.error('Invalid or expired hashing key: %s.',
                   api._hash_server_token)
-    except Exception as e:
-        log.exception('Exception while downloading map: %s', repr(e))
+    except Exception:
+        log.exception('Exception while downloading map')
```
jagauthier: How about a period at the end?

friscoMad:
I think you skipped the diff of the whole models.py as it appears hidden by default, there are a lot of close there.

About removal of db-max_connections, it made little sense, as it sets the limit in peewee Pooled class, to be (db-max_connections * accounts) not even workers default was 5 so it defaulted to 5 * accounts that was a really high limit except for small instances with 1-5 accounts, instances with an insane db-thread or just server instances where the number all the "system" threads/connections are a lot higher than the worker number.
Anyway setting max-connection in Pooled db just do a static check every time a connection is done and if we reached the limit it raises the exception there is no db reason for that, it just limits you to what you set, instead not setting the max-connection it just skips the check.
As we are telling our users to increase db-max_connections when the limit is hit it is a lot better to just remove the check and let it create as many connections as needed.

So for me even if this PR does not ends being merged that part should be added to RM anyway.

jagauthier:
You're right, It is hidden and I didn't see that.

pogom/models.py
```diff
@@ -62,17 +61,13 @@ def init_database(app):
...
-            connections *= len(args.accounts)
         db = MyRetryDB(
             args.db_name,
             user=args.db_user,
             password=args.db_pass,
             host=args.db_host,
             port=args.db_port,
-            max_connections=connections,
-            stale_timeout=300,
+            stale_timeout=30,
```
michikrug: `max_connections` should be set to `None` to not run into `ValueError('Exceeded maximum connections.')`

> max_connections (int) – Maximum number of connections. Provide None for unlimited.

pogom/search.py
```diff
@@ -786,18 +786,35 @@ def search_worker_thread(args, account_queue, account_sets,
...
-                account['username'])
+            prevStatus = WorkerStatus.get_worker(account['username'])
+            if prevStatus:
+                status.update(prevStatus)
+            else:
+                status.update({
+                    'username': account['username'],
+                    'last_modified': datetime.utcnow(),
+                    'last_scan_date': datetime.utcnow(),
+                    'latitude': None,
```
michikrug: Not setting an initial location seems to fail getting next_item for me...

michikrug: Getting a lot of 'Nothing to scan' wait messages

pogom/models.py
friscoMad: So true, thanks 👍

pogom/search.py
friscoMad: I am not seeing any problem with -speed and -gi, what is your scheduler/config?
I have commit 2058096 on my main instance but this change came later, my test instance running on latest code is not showing any error, I even did a -cd to be sure it was not hitting issues during initial scan but it is not complaining.

michikrug: I am currently running -speed -bh -wph 4 -st 8
HexSearch runs without problems.

I will also diff and check, if I have made some changes that may lead to that behavior.

sirlindqvist:
Running for 72h+ together with 2138, 2263, 2265, 2266, 2269, 2271, 2273. Didn't notice any errors.

sebastienvercammen:
Tagged pending changes because the PR needs to be rebased, and author tagged it as "WiP" in the description.

friscoMad:
Rebased and "done" but I still have some doubts on the proper way to do some things @sebastienvercammen :

 * Proper way to access db instance Model.meta._database is what I am using now but a global var would be cleaner, other option is to refactor all queries into a single class with db as a param or move all the close calls to be inside the logic (just before sleeps or network calls but it will be a mess)
 * Query methods should be classmethods, static or regular object methods all three of them work and I started moving from static to class for some methods as I thought it was needed but it is not the case.

docs/common-issues/faq.md
```diff
@@ -155,22 +155,6 @@ InternalError(1054, u"unknown column 'cp' in 'field list'") or similar
...
-
-Try raising --db-max_connections, default is 5.
-
-#### MySQL max connections
-
-```
-OperationalError(1040, u'Too many connections')
-```
-
-You need to raise the maximum connections allowed on your MySQL server configuration, this is typically by setting `max_connections` in my.cnf or my.ini. Please use Google to find where this file is located on your specific operating system.
```
sebastienvercammen: @friscoMad This second error could still occur, unless you found something in peewee that checks or edits MySQL's config?

friscoMad: It can but it shouldn't as connections are really shared with 140 workers I see a connection count of 4-5 from 150. Default mysql connection limit is 151 so you have a lot of room until you reach that limit, then only issue is the initial burst that I did limit moving the staggering lock.

friscoMad:
@sebastienvercammen 
Peewee pool does only frees the connection on close (https://github.com/coleifer/peewee/blob/master/playhouse/pool.py#L193)
You can call close on any db instance as they are all the same, the problem is that executing a query does open a connection if it is not opened but it is never closed unless you manually do so. With flask and other web frameworks what you do is create a hook that opens the db on request start and closes on end but for long standing threads there is no way to do the same.
The problem with `Model._meta.database.close()` is that we need to reach the db instance that we create in init_database. We can store the db in a global var,  pass it to every query method or do that, I don't like it but I thought it was the best option.

docs/common-issues/faq.md
friscoMad: Anyway we can leave it just in case.

sebastienvercammen:
@friscoMad We'll go for execution contexts:
```
with db.execution_context() as ctx:
    # A new connection will be opened or, if using a connection pool,
    # pulled from the pool of available connections. Additionally, a
    # transaction will be started.
    user = User.create(username='charlie')

# When the block ends, the transaction will be committed and the connection
# will be closed (or returned to the pool).

@db.execution_context(with_transaction=False)
def do_something(foo, bar):
    # When this function is called, a separate connection is made and will
    # be closed when the function returns.
```

hzpz:
I'm down from 33-36 to 7-8 database connections, this is awesome! Especially so since I use a managed database that limits the max connections. I have a lot of wiggle room now. Thanks!

PR #2276, opened 2017-08-27
Reset Gym Filter when disabled

### LINKED ISSUE #2268: Gym filter persists to raid filter even when disabled
## Expected Behavior

Within webview if you set a gym filter , eg team specific, empty slot etc then disable gyms then enable raid filters you should see the raids that match your filter.

## Current Behavior
The gym filter persists even when disabled/unchecked and you only see the raids that ALSO match the gym filter.

1. navigate to map, disable all options
2. enable gyms, set gym filter to only show mystic
3. disable gym filter
4. enable raid filter and note number/location of raids
5. disable raid filter
6 enable gym, set back to defaults (all teams, raids 1-5 and uncheck "empty slot")
7 disable gym
8 enable raids and compare number/locations to number from step 4

## Context
may not be obvious that you are not seeing all raids as they are still filtered

## Your Environment
windows, python 2.7

BrownSlaughter:
added a Gym Filter reset function and call it when the gym switch is toggled

## Description
added a Gym Filter reset function and call it when the gym switch is toggled

## Motivation and Context
currently the filter is not disabled properly when the switch is turned off which affected the showing of gyms like in [Issue 2268](https://github.com/RocketMap/RocketMap/issues/2268)

**Seb:** Fixes #2268.

## How Has This Been Tested?
Tested to see if the filter is reset when turned off
Tested to see if it fixed the raid showing issue once gyms were disabled.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2649,11 +2649,41 @@ $(function () {
             }
         }
     }
+
+    function resetGymFilter() {
+        Store.set('showTeamGymsOnly', 0)
+        Store.set('minGymLevel', 0)
+        Store.set('maxGymLevel', 6)
+        Store.set('showOpenGymsOnly', false)
+        $('#team-gyms-only-switch').val(Store.get('showTeamGymsOnly'))
+        $('#open-gyms-only-switch').prop('checked', Store.get('showOpenGymsOnly'))
+        $('#min-level-gyms-filter-switch').val(Store.get('minGymLevel'))
+        $('#max-level-gyms-filter-switch').val(Store.get('maxGymLevel'))
+        $selectTeamGymsOnly = $('#team-gyms-only-switch')
+        $selectTeamGymsOnly.select2({
+            placeholder: 'Only Show Gyms For Team',
+            minimumResultsForSearch: Infinity
+        })
+
+        $selectMinGymLevel = $('#min-level-gyms-filter-switch')
+        $selectMinGymLevel.select2({
+            placeholder: 'Minimum Gym Level',
+            minimumResultsForSearch: Infinity
+        })
+
+        $selectMaxGymLevel = $('#max-level-gyms-filter-switch')
+        $selectMaxGymLevel.select2({
+            placeholder: 'Maximum Gym Level',
+            minimumResultsForSearch: Infinity
+        })
+    }
+
     // Setup UI element interactions
     $('#gyms-switch').change(function () {
         var options = {
             'duration': 500
         }
+        resetGymFilter()
         var wrapper = $('#gym-sidebar-wrapper')
         if (this.checked) {
             lastgyms = false
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2655,11 +2655,14 @@ $(function () {
         Store.set('minGymLevel', 0)
         Store.set('maxGymLevel', 6)
         Store.set('showOpenGymsOnly', false)
+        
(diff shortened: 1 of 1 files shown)


static/js/map.js
```diff
@@ -2649,11 +2649,41 @@ $(function () {
             }
         }
     }
+
+    function resetGymFilter() {
+        Store.set('showTeamGymsOnly', 0)
+        Store.set('minGymLevel', 0)
+        Store.set('maxGymLevel', 6)
+        Store.set('showOpenGymsOnly', false)
+        $('#team-gyms-only-switch').val(Store.get('showTeamGymsOnly'))
```
sebastienvercammen: If you add a whiteline right above here...

static/js/map.js
```diff
@@ -2649,11 +2649,41 @@ $(function () {
...
+        $('#open-gyms-only-switch').prop('checked', Store.get('showOpenGymsOnly'))
+        $('#min-level-gyms-filter-switch').val(Store.get('minGymLevel'))
+        $('#max-level-gyms-filter-switch').val(Store.get('maxGymLevel'))
+        $selectTeamGymsOnly = $('#team-gyms-only-switch')
```
sebastienvercammen: ...aaaaand right above here, we'll be great friends. ❤️

static/js/map.js
```diff
@@ -2649,11 +2649,41 @@ $(function () {
...
+        $selectMinGymLevel.select2({
+            placeholder: 'Minimum Gym Level',
+            minimumResultsForSearch: Infinity
+        })
+
+        $selectMaxGymLevel = $('#max-level-gyms-filter-switch')
+        $selectMaxGymLevel.select2({
+            placeholder: 'Maximum Gym Level',
+            minimumResultsForSearch: Infinity
+        })
```
sebastienvercammen: Any reason you recreate the `.select2` instances instead of changing the attributes via [programmatic access](https://select2.github.io/examples.html#programmatic-control)?

BrownSlaughter: I copied the code from somewhere else and it worked :D

sebastienvercammen: @BrownSlaughter Don't, it's less efficient to recreate instances than to just change the attributes. 😛

tomballgithub:
It didn't work for me.  It simply hid the 'raids' button and when I undid -nr the 'raids' setting was still enabled.  I tried it on my personal instance and then installed the latest pull from Aug29 and tried it on a 2nd server.  Same results.

- Launch -os instance
- Load map
- Enable Raids, Disable Pokestops, Disable Pokemon (only Raids are showing)
- Close -os instance
- Relaunch -os instance with -NR switch
- Reload map
- Raids are still visible.  Raid switch is not visible.

BrownSlaughter:
@tomballgithub this is a fix for 2268 not the -nr 2275 issue

BrownSlaughter:
oops comment not close 👍

tomballgithub:
It didn't work for me.  I launched the instance and hit reset settings just to double confirm.  I enable raids/gyms but not pokemon/pokestops.  I changed the max gym level to 1.  Most of my raids disappeared (which seems wrong since gym filtering should be independent of raids).  I then turned off gyms and most of my raids remained hidden.  Other variations did the same

tomballgithub:
Works great!   I don't really like that the gym max level is applied to raids also but that's a different issue.

BrownSlaughter:
@tomballgithub it seemed to switch itself without needing the change just by getting the value so was unneeded

PR #2281, opened 2017-08-29
Make background DB cleaner more complete.

### LINKED ISSUE #2278: Only query Gyms that have been scanned recently (and are still active).
## Expected Behavior

In my area over the last month multiple gyms have been removed from the game.  I am receiving reports from users that the maps have disappeared but are still on the map.  The expected behavior would be gyms that have been removed from game are removed from the map.   You can detect these gyms by looking for gyms with an old 'last scanned' date

image

## Current Behavior

Gyms remain in the database and map unless database is cleared or gym manually removed

## Possible Solution

Allow the option to automatically remove gyms with last_scanned > X days
Or better yet detect if a gym is no longer present and remove it

1. Explained above

## Context

It causes a quality of life issue with map users.  They go to a gym that is not present

## Your Environment

* Version used:
* Environment name and version (e.g. Python 2.7):
* Operating System and version (desktop or mobile):
Ubunto 16.04 VPS

BrownSlaughter:
add removal of gym and gym details over 24 hours old to clean db loop

## Description
This will remove the gym and its deatils from the database if it has not been scanned in 24 hours

## Motivation and Context
If you no longer scan an area or a gym is removed by niantic the old gyms stay on the map and in the db but never updated.

**Seb:** Fixes #2278.

## How Has This Been Tested?
Had old gym info in db, the db cleaner loop removed them

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2654,6 +2654,20 @@ def clean_db_loop(args):
                              (datetime.utcnow() - timedelta(minutes=30)))))
             query.execute()
 
+            # Remove old gym Details.
+            query = (GymDetails
+                     .delete()
+                     .where(GymDetails.last_scanned <
+                            (datetime.now() - timedelta(days=1))))
+            query.execute()
+
+            # Remove old gym locations.
+            query = (Gym
+                     .delete()
+                     .where(Gym.last_scanned <
+                            (datetime.now() - timedelta(days=1))))
+            query.execute()
+
             # Remove active modifier from expired lured pokestops.
             query = (Pokestop
                      .update(lure_expiration=None, active_fort_modifier=None)
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2706,7 +2706,7 @@ def clean_db_loop(args):
             query = (HashKeys
                      .delete()
                      .where(HashKeys.expires <
-                            (datetime.now() - timedelta(days=1))))
+                            (datetime.utcnow() - timedelta(days=1))))
             query.execute()
 
             # If desired, clear old Pokemon spawns.
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2658,14 +2658,35 @@ def clean_db_loop(args):
             query = (GymDetails
                      .delete()
                      .where(GymDetails.last_scanned <
-                            (datetime.now() - timedelta(days=1))))
+                            (datetime.utcnow() - timedelta(days=7))))
             query.execute()
 
             # Remove old gym locations.
             query = (Gym
                      .delete()
                      .where(Gym.last_scanned <
-                            (datetime.now() - timedelta(days=1))))
+                            (datetime.utcnow() - timedelta(days=7))))
+            query.execute()
+
+            # Remove old raid Details.
+            query = (Raid
+                     .delete()
+                     .where(Raid.end <
+                            (datetime.utcnow() - timedelta(days=7))))
+            query.execute()
+
+            # Remove old gym members.
+            query = (GymMember
+                     .delete()
+                     .where(GymMember.last_scanned <
+                            (datetime.utcnow() - timedelta(days=7))))
+            query.execute()
+
+            # Remove old gym Pokemon.
+            query = (GymPokemon
+                     .delete()
+                     .where(GymPokemon.last_seen <
+                            (datetime.utcnow() - timedelta(days=7))))
             query.execute()
 
             # Remove active modifier from expired lured pokestops.


friscoMad:
Please also add pokestops as they also move around or get deleted some times.

elincognito:
It would also be nice a wh notification when a gym or pokestop is removed.

BrownSlaughter:
@friscoMad pokestop dont have a last_scanned so removal this way is not possible

BrownSlaughter:
@elincognito that is not the scope of this PR

tomballgithub:
But the feature works and I saw old gyms removed at first database scrub

friscoMad:
Ok some new thoughts pokestops have the last_updated field but the problem is that stops are not constantly updating like gyms so deleting based on this field would likely lead to false positives, they are only updated when a lure appears.

Anyway we could delete them anyway and let the map add them again, due to the long range for gyms/stops I don't see it a real problem but I would set the delta to at least 1 week to avoid doing this every day, anyway that would lead to having a lot of missing stops for some minutes just 1 week after the map was first started.

And on another issue, raids, gym members and gym pokes should also be cleaned with gym/gym details

tomballgithub:
I have old raid details in my db also.

sebastienvercammen:
Temporarily on hold while we discuss some SQL/SQLite things on Discord that are related to this PR.

tomballgithub:
Also, when removing a gym we should also remove the pokemon assigned to those gyms.  I am seeing orphaned rows in gymmember and gympokemon.

andreacassani:
I think there should be at least an option to enable/disable gym-details removal.
I for example run gym-info only once for every new area to retrieve gym names. This would require a new gym-info scan every 24 hours.

BrownSlaughter:
@acassani91 gym names are from GMO not gym_info you will get those if you scan within 1000m of a gym. This PR is to remove gym details that dont get scanned because you no long scan the same area or a gym is removed.

acassani91:
@BrownSlaughter Alright my bad, thanks for the reply!

friscoMad:
Sorry to say but gym name comes in gym_get_details that is what we send
with gym-info enabled.


2017-09-18 15:48 GMT+02:00 Andrea Cassani <notifications@github.com>:

friscoMad:
I think the only issue is with pokestops as we are not inserting them continously as we do with everything else, we would need to change the code to upsert everything instead of querying and only inserting the new ones. But I think it would be better to change the code to remove the missing ones instead of updating all and them removing the one that are not three anymore in a thread. That way we will avoid a lot of DB writes for no reason.

tomballgithub:
There is a tool out there called Worldpoole that queries the RM database to generate interesting reports using the data.  I just noticed that this shows historical data from the database and so if we prune it after 7 days it would impact that tool and the people that use it.  This PR has moved from just hiding gyms that are no longer in the game towards more database pruning.  Need to ensure we consider all aspects of those decisions.

Here is a snippet from the trainer report:
image

sebastienvercammen:
@tomballgithub Solved with #2305. If someone doesn't want data to be cleaned, they shouldn't enable it.

Those who want more fine-grained control should look into setting up automated backups or MySQL triggers or a few options to control data removal (e.g. age of data).

This lets RM take care of its own needs (removing old data to reduce/limit performance impact and deadlocks) while the host has other options if they want to keep all old data.

KartulUdus:
my problem with this:

Users have asked me to have a static map of the gyms available in the country.
I did a hex scan to find them and deleted the trainers|pokemon inside to have static points showing unassigned gyms.

would it be a horrible idea to have some kind of config switch to not delete old gyms if user opts for it?

pogo-excalibur:
Can a static map not be e.g. an image file?

fosJoddie:
I also have a much larger, static area with stops and gyms. It's about 2 counties, 50 000 square kilometers. A simple static picture wont really do it justice. Not to mention people who wanna geek out and see gym names.

I accept that when pokestops go become gyms, a cleaner needs to take care and remove such things (same when things get removed). But I'm scared to loose my "static" info that I scanned once and forgot.

tomballgithub:
Some of these last comments are interesting.  Maybe the solution would be to have configurable timeframes for the pruning of each item (although I realize there is opposition to having more configuration options)

pogom/models.py
```diff
@@ -2671,7 +2706,7 @@ def clean_db_loop(args):
             query = (HashKeys
                      .delete()
                      .where(HashKeys.expires <
```
neskk: `HashKeys.expires` should not be used here since it can be overwritten with `None`. I had several old HashKeys in my database and this PR didn't fix that issue.
Changing `HashKeys.expires` to `HashKeys.last_updated` works and is a better way to keep things clean.

friscoMad:
I don't see the issue here if you want static maps just create another DB with them and do whatever you want with them.
There is no reason reason to mix static and dynamic data in the same db, just enable clean in the dynamic instance and we are all fine.

If someone wants to do things like update gyms/stops once a months and then disable gym scanning/stop scanning during regular work while using the same db for everything, then just the solution is to not enable db-clean and do the cleanup by hand, trigger, cron job or whatever, he should be able to do so.

Adding a lot of new config options just for fringe use cases makes no sense.

fosJoddie:
@friscoMad  I don't agree with you that making a separate RM web instance and database to have static pokestop and gym info is trivial. ("just make a copy of the db").

That said, I can agree with you that there's maybe little to no practical use for having a bunch of more options just for some fringe use case like myself and @KartulUdus are discussing. 

I am giving this PR the thumbs up, after a rebase.

sebastienvercammen:
This PR is being closed in favor of #2383 but the discussion is still relevant. #2383 adds a few options for better control of data removal (including an option to disable certain cleaners).

PR #2284, opened 2017-08-30
Choose map service provider for navigation

sirlindqvist:
Option to be able to choose Apple Maps for navigation instead of default Google Maps.

## Motivation and Context
Created for iOS users having trouble with the Google Maps app.

When clicking a coordinate link a new window opens in Safari and asks for using the Google Maps app.

ios_google_maps

If taping 'use the app' the Apple AppStore opens even if the app already is on the phone. From there it's possible to click 'Open' and the Google Maps app loads, however it's not possible to start navigation. If taping 'no thanks' the question loops over and over again. 

Selecting Apple Maps instead as map service provider opens the native iOS maps app without any trouble at all.

options

## How Has This Been Tested?
On my own map. Sucessfully tested on both iOS 10.10.3 (iPhone 6S) and MacOS High Sierra public beta 7 (iMac mid 2011).

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -1017,6 +1017,10 @@ var StoreOptions = {
     'processPokemonIntervalMs': {
         default: 100,
         type: StoreTypes.Number
+    },
+    'mapServiceProvider': {
+        default: 'googlemaps',
+        type: StoreTypes.String
     }
 }
 
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -449,11 +449,11 @@ function openMapDirections(lat, lng) { // eslint-disable-line no-unused-vars
     var url = ''
     if (Store.get('mapServiceProvider') === 'googlemaps') {
         url = 'https://www.google.com/maps/?daddr=' + lat + ',' + lng
+        window.open(url, '_blank')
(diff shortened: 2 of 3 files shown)


sebastienvercammen:
To be discussed if "map service provider" should be an option, or detected via browser fingerprinting.

sirlindqvist:
I know you would ask that @sebastienvercammen :)

sirlindqvist:
Got fingerprinting working with client.js if we decide to go for this.

sebastienvercammen:
@sirlindqvist We're keeping it as is instead of browser fingerprinting so iOS users still have the option to install and use Google Maps on iOS.

sirlindqvist:
OK! I'll check the code and will do some small adjustments. Now, a new tab is opened in browser when Apple Maps is being used and the app opens on phone, shouldn't do that :)

friscoMad:
@sirlindqvist still need to do some fixes or is the PR good as it is now?

sirlindqvist:
Will fix the new tab issue

sirlindqvist:
@sebastienvercammen and @friscoMad this is ready for review.

PR #2288, opened 2017-09-03
Remove support for SQLite and Optimize database fields

friscoMad:
**BEWARE** 
This PR will modify your DB to a new version, migration code needs more tests and there is no rollback to previous DB versions, so only run this code after you have backed up your DB.

## Description

Remove support for SQLite, made mysql params required and added default host as 127.0.0.1
Updated all docs where it was refering to SQLite but we should review the initial install to be sure it is obvious that installing mysql is a must.
Also due to SQLite removal we can use int values for a lot of fields that were currently text, reducing greatly db size and probably mysql performance.

## Motivation and Context

SQLite is not enough for the data usage we do and just in small maps will work without issues, it also make it more complicated to evolve the code and improve so we are just removing it and force everyone to use mysql or mariadb, both are available in all enviroments and free to use, just a bit more complex to setup.

## How Has This Been Tested?

Tested in small instance regular operation, -cd, initial table creation.
Currently running in main instance. 
My main instance DB went from 1.6Gb to 970Mb after migration and 675Mb after optimization, migration took 10 minutes (3M pokes, 10.000 spawns).

Main instance DB pre update:
209 42 194 2 _ localhost _ frisco_pokemon _ phpmyadmin 4 7 3_2017-10-09_20-40-09

Same db after migration:
209 42 194 2 _ localhost _ frisco_pokemon _ phpmyadmin 4 7 3_2017-10-09_20-50-57

Same DB after migration and optimization:
209 42 194 2 _ localhost _ frisco_pokemon _ phpmyadmin 4 7 3_2017-10-09_20-55-45

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Live visualization of all the Pokémon (with option to show gyms, raids and Pok�
 * Filters
 * Independent worker threads (many can be used simultaneously to quickly generate a livemap of a huge geographical area)
 * Localization (en, fr, pt_br, de, ru, ko, ja, zh_tw, zh_cn, zh_hk)
-* DB storage (sqlite or mysql) of all found Pokémon
+* DB storage (mysql) of all found Pokémon
 * Incredibly fast, efficient searching algorithm (compared to everything else available)
 
 ## Information
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -39,8 +39,6 @@
 # Database settings
 ###################
 
-#db-type: sqlite                # sqlite (default) or mysql
-#db:                            # Database filename for SQLite. (default='pogom.db')
 #db-host:                       # Required for mysql ()
 #db-name:                       # Required for mysql
 #db-user:                       # Required for mysql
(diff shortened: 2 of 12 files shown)


friscoMad:
I still need to work on migration code, I have issues with peewee transactions, that's why I removed the RetryOperationalError (Travis error)
Anyway we need to evaluate if there is need for it, as I don't see any situation it will help.
Code works and migration gets completed but I haven't tested if RM runs after migration (it works with a new DB for sure).

friscoMad:
Ok forget about the last comment, we can not use transaction as all DDL (create, alter tables) clear them so I have removed all transactions in the code and reverted to use RetryOperationalError (still not sure if it will help in any case or it's just better to remove it @sebastienvercammen)
Migration of my main instance db went without issues but took 2min (1Gb and 2M pokes) still some more tests are needed to be sure.
Also will we support migrations forever? I think we could be ok with the last 5 or last 3 moths.

friscoMad:
Rebased and did some minor modifications:

- Removed encription lib arg
- Grouped db args

jaake77:
Testing. Seems good thus far. Just had to comment out -- db-type: mysql -- from my .ini files.

PR #2291, opened 2017-09-08
Make Push.js serviceWorker.min.js path relative

### LINKED ISSUE #2290: Push.js service worker only works for maps not running under subpath like 'xyz.org/map'
Push.js by default looks for the required serviceworker.min.js file at an absolute path on domain level like 'xyz.org/serviceworker.min.js'.
Since many of us run their maps under a subpath like 'xyz.org/map' the file can not be found.
Which then leads to no notifications on mobile browsers.

## Expected Behavior
Notifications should be shown on all browsers and also maps running under subpaths.

## Current Behavior
Notifications do not work on mobile browsers when the map runs under a subpaths.

## Possible Solution
Add option to Push.js configuration to point to the correct relative service worker path.
_serviceWorker: 'serviceWorker.min.js'_
Default is: _serviceWorker: '/serviceWorker.min.js'_

`$(function () {`
`    /* If push.js is unsupported or disabled, fall back to toastr`
`     * notifications. */`
`    Push.config({`
`        serviceWorker: 'serviceWorker.min.js',`
`        fallback: function (notification) {`
`            sendToastrPokemonNotification(`
`                notification.title,`
`                notification.body,`
`                notification.icon,`
`                notification.data.lat,`
`                notification.data.lon`
`            )`
`        }`
`    })`
`})`

michikrug:
## Description
Add option to Push.js configuration to point to the correct relative service worker path.
serviceWorker: 'serviceWorker.min.js'
Default is: serviceWorker: '/serviceWorker.min.js'

Fixes #2290

## Motivation and Context
Push.js by default looks for the required serviceworker.min.js file at an absolute path on domain level like 'xyz.org/serviceworker.min.js'.
Since many of us run their maps under a subpath like 'xyz.org/map' the file can not be found.
Which then leads to no notifications on mobile browsers.

## How Has This Been Tested?
Tested on own instance running under subpath '/map/'.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2282,6 +2282,7 @@ $(function () {
     /* If push.js is unsupported or disabled, fall back to toastr
      * notifications. */
     Push.config({
+        serviceWorker: 'serviceWorker.min.js',
         fallback: function (notification) {
             sendToastrPokemonNotification(
                 notification.title,


sebastienvercammen:
`./serviceWorker.min.js` should mean current directory, so while testing it's important to check how it actually behaves and not to forget that our build script puts the final built file in `static/dist/js/serviceWorker.min.js`, which gets served as `/js/serviceWorker.min.js` on the webserver.

michikrug:
@sebastienvercammen see https://github.com/Nickersoft/push.js/commit/b1b8cd3da957905d70b5d6a3377f5d4fef4f179b

It was changed to `/serviceWorker.min.js` and therefore is now not working anymore on all configurations.

And no: `/serviceWorker.min.js` is served at root path of the RocketMap instance.
`self.route("/serviceWorker.min.js", methods=['GET'])`

PR #2292, opened 2017-09-10
Remove (<amount> filtered) if no Pokemon are filtered

ibuckshot5:
## Description

Remove the <amount> filtered tag on GetMapObjects if the filter-out list is empty.

## Motivation and Context

Efficiency

## How Has This Been Tested?

Local installation

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2253,7 +2253,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
         # Helping out the GC.
         del forts
 
-    if filtered == 0:
+    if filtered:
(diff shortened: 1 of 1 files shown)


tomballgithub:
The only reason IMHO would be to add clarity.  By saying 'filtered' when filtering is not even enabled, it could cause confusion.  I do realize it does not make for elegant code.

jagauthier:
@tomballgithub The same logic could be used for pokemon, gyms, raids, etc.  All of those can be disabled.

sebastienvercammen:
PR closed. Logging info is most useful for debugging purposes and if an extra field that says "zero" when filtering is disabled confuses people, they should stop looking at technical logs.

PR #2294, opened 2017-09-11
Split proxy into two different lists, PTC and Niantic

AEtHeLsYn:
## Description
Recently PTC login server has started to ban IPs like Niantic did, but the banned IP may be not banned by Niantic and the other way around. With this PR we have two different lists that apply to login or to session when necessary.
Also, check login changed to avoid login in after 30 minutes, we use oauth token instead to refresh session.

## Motivation and Context
We can reuse some PTC banned proxies for  Niantic sessions and the other way around.

## How Has This Been Tested?
Local map

## Types of changes

## Checklist:


AEtHeLsYn:
Closing the PR as is, new approach needed.

PR #2302, opened 2017-09-18
Add proxy usage option, take 2

AEtHeLsYn:
## Description
With this PR we can choose if we want to use the proxies for all connecions, only for ptc login or only for niantic connections

## Motivation and Context
We can use proxies only for ptc login or for niantic sessions if we have an IP banned for one of them, not both.

## How Has This Been Tested?
Local map

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -333,6 +333,10 @@ def get_args():
     parser.add_argument('-px', '--proxy',
                         help='Proxy url (e.g. socks5://127.0.0.1:9050)',
                         action='append')
+    parser.add_argument('-pxu', '--proxy-usage',
+                        help=('Select which connections use proxy ' +
+                              '(all/ptc/niantic).'),
+                        type=str, default='all')
     parser.add_argument('-pxsc', '--proxy-skip-check',
                         help='Disable checking of proxies before start.',
                         action='store_true', default=False)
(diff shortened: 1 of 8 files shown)


pogom/utils.py
```diff
@@ -333,6 +333,10 @@ def get_args():
     parser.add_argument('-px', '--proxy',
                         help='Proxy url (e.g. socks5://127.0.0.1:9050)',
                         action='append')
+    parser.add_argument('-pxu', '--proxy-usage',
+                        help=('Select which connections use proxy ' +
+                              '(all/ptc/niantic).'),
+                        type=str, default='all')
```
friscoMad: Looking at the code I think that using a repeated value like wh-types it is better (also add valid values) that way the ifs would be `if 'ptc' in args.proxy_usage` instead of ìf args.proxy_usage != 'niantic'` also it is forward compatibe if in there future more proxy controls are needed (wno knows).
The problem is that I think you can not set the default value to both, but we can do that in code, set default to None and if there are no values defined then just add all of them.

AEtHeLsYn: In fact I chose that approach because I couldn't find a way to set default values to none in a repeated value. This way behaves the same as proxy rotation variable, even if you spell it wrong the default behaviour is the current one.

billyjbryant:
@AEtHeLsYn is this PR still active?

PR #2303, opened 2017-09-19
New config file for shared items across instances.

BrownSlaughter:
-scf flag added to allow for setting a second config file just for keys and any other shared settings

## Motivation and Context
For people who run multiple instances changing every config every month just to change the hash key is very time consuming. This allows you to just edit a single file and call it in the cli.

This works by allowing a second config to be called that just contains your keys and any other shared settings.

This is not a mandatory change, everything can still be run from config.ini if you only have 1 instance or dont need a shared config file.

## How Has This Been Tested?
started an instance with config.ini containing no keys and shared-config.ini containing my keys
`runserver.py -cf /config/config.ini -scf /config/shared-config.ini` 

## Types of changes

## Checklist:

- [ x] I have updated the documentation accordingly.


docs/first-run/configuration-files.md
```diff
@@ -48,6 +48,14 @@ The default configuration file is *config/config.ini* underneath the project hom
 
      python runserver.py -u randomjoe -p password1 -u bob -p password2 -l "seattle, wa" -st 5 -k MyGmapsKeyGoesHereSomeLongString -ps
 
+## Shared config
+
+You may now add settings to a shared config file rather than edit each config individually. This is useful in situations where you run multiple instances and have to change the hash key in each config every month. Add the shared settings to your shared-config.ini and call the shared config from the command line using the `-scf` flag when you start your instance.
```
friscoMad: The "now" part is irrelevant as it will be here forever, and I will try to make it clear that it's use is mainly for multiple instances. I would say something like: 
"You can add a second config file to use it as a shared config if you run several instances. In this file you can setup configuration shared by the different instances, like hash key, gmaps key, captcha or db settings. Add the shared settings..."

Did you test if it works configuring the -scf inside config.ini? (I don't expect it to work but who knows?

BrownSlaughter: no I have not tested it in the config.ini

sebastienvercammen:
@BrownSlaughter Some questions we need to answer before we can consider merging.

* RE: frisco: Does configuring `-scf` in the original config file work as expected? (I personally assume it will, no reason for it not to.)
* What happens if you configure both regular config and `-scf` and they both contain a config item with different values, which one takes precedence? If it's `-scf`: can you change it so the regular (non-shared) config takes precedence?
* Assuming configuring `-scf` in the regular config works: what happens if you configure the `-scf` filename in the regular config file, and in the `-scf` file you configure the regular config as `-scf` (i.e. intentionally trying to get the regular config to load `-scf` and loop `-scf` back to load the original config, ending in an infinite loop)?

BrownSlaughter:
> RE: frisco: Does configuring -scf in the original config file work? (I personally assume it will, no reason for it not to.)

You can not call a config from the config

> What happens if you configure both regular config and -scf and they both contain a config item with different values, which one takes precedence? If it's -scf: can you change it so the regular (non-shared) config takes precedence?

The config set with -cf always gets precedence

PR #2305, opened 2017-09-21
Replace --disable-clean with --enable-clean.

sebastienvercammen:
## Description
Reverse the behavior for the config item of the background database cleaner.

## Motivation and Context
The DB cleaner should only be enabled on one instance, which fits better with `-DC, --enable-clean`. Too many people skipped `--disable-clean`, enabling the DB cleaner on all their instances and inevitably running into MySQL deadlocks.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -389,7 +389,7 @@ def get_args():
                         help=('Get all details about gyms (causes an ' +
                               'additional API hit for every gym).'),
                         action='store_true', default=False)
-    parser.add_argument('--disable-clean', help='Disable clean db loop.',
+    parser.add_argument('-DC', '--enable-clean', help='Enable DB cleaner.',
                         action='store_true', default=False)
     parser.add_argument(
         '--wh-types',
--- a/docs/first-run/commandline.md
+++ b/docs/first-run/commandline.md
@@ -333,7 +333,7 @@
                             var: POGOMAP_WEBHOOK]
       -gi, --gym-info       Get all details about gyms (causes an additional API
                             hit for every gym). [env var: POGOMAP_GYM_INFO]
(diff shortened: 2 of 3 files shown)


pogom/utils.py
```diff
@@ -389,7 +389,7 @@ def get_args():
                         help=('Get all details about gyms (causes an ' +
                               'additional API hit for every gym).'),
                         action='store_true', default=False)
-    parser.add_argument('--disable-clean', help='Disable clean db loop.',
+    parser.add_argument('-DC', '--enable-clean', help='Enable DB cleaner.',
```
michikrug: -DC for --enable-clean seems confusing (is it meant to be short for "DB Cleaner"?)

friscoMad: +1 and caps are also weird

sebastienvercammen: @michikrug @friscoMad Correct, `-D` is used for the database (although currently sqlite). Caps aren't weird, they're common for differences in behavior.

We currently only have long forms for database info (e.g. `--db-host`) which could get a short form by extending the `-D` argument: `-DH` for host (and we have `-H` for web host), `-DP` for port, and so on.

For the future, this is what we're going for. It avoids absurd and arbitrary short forms - `pxo, ldur, ...`

FrostTheFox: I don't see a problem. -DatabaseCleaner -> -DC.

friscoMad: I found it confusing as we already have -dc also the previous behaviour was --disable-clean so -DC also seems to be a shorthand of it, and I do not see a problem to use D as a prefix for all database options or P for proxies but in that case -Dec and --db-enable-clean seems a better fit for me or if we prefrer a shorter version -Dc and --db-clean

PR #2312, opened 2017-09-25
Fix heatmap

### LINKED ISSUE #2307: Pokemon on heatmap not showing up
## Expected Behavior

Heatmap should show pokemon.

## Current Behavior

Pokemon on heatmap are not showing up. Using the developer console on Chrome I got two errors.
First when opening the stats page:
```
Uncaught TypeError: Cannot read property 'appendChild' of null
    at app.min.js:5
    at app.min.js:5
```
And the next error occured when opening the heatmap:
```
map.common.min.js:5 Uncaught ReferenceError: isNotifyPoke is not defined
    at setupPokemonMarkerDetails (map.common.min.js:5)
    at setupPokemonMarker (map.common.min.js:5)
    at Object.processAppearance (statistics.min.js:5)
    at Function.each (jquery.min.js:2)
    at Object.<anonymous> (statistics.min.js:5)
    at i (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at A (jquery.min.js:4)
    at XMLHttpRequest.<anonymous> (jquery.min.js:4)
```
## Possible Solution

-

1.  Open the heatmap...

## Your Environment

Client: Windows 10, Google Chrome Version 61.0.3163.100.
Server: Ubuntu 16.04, Python 2.7.

ChrisSsSsSsS:
this was listed as an Issue on 22.09.2017: #2307 

**Seb:** Fixes #2307.

## Description

There were two issues:
1. in app.js: `Uncaught TypeError: Cannot read property 'appendChild' of null`
2. `map.common.min.js:5 Uncaught ReferenceError: isNotifyPoke is not defined`

Here's are explanations why these two errors occurr:
1. The `TypeError` occurred because on the /stats page you don't have a gym-sidebar, but in the app.js it tries to append a child to the sidebar, without checking:
`$gymSidebar.appendChild($gymSidebarClose)` ([link](https://github.com/RocketMap/RocketMap/blob/develop/static/js/app.js#L100))
2. The `isNotifyPoke is not defined` error happens because the `setupPokemonMarkerDetails` function in `map.common.js` has a cross-reference to a function `isNotifyPoke` which is solely defined in `map.js`. This reference stems from the "scale by rarity" functionality and can be seen in this line
`var rarityValue = isNotifyPoke(item) || (upscaledPokemon.indexOf(item['pokemon_id']) !== -1) ? 29 : 2` ([link](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.common.js#L1102))

Here are my solutions:
1. I just added a null check
2. You could just add a dummy-function like `function isNotifyPoke(poke) { return false; }` to `statistics.js`, but this doesn't address the real issue: cross-referencing
So instead I pass an optional argument to the `map.common.js` function `setupPokemonMarker` and `updatePokemonMarker`, which toggles if the current item is "worthy of notification"(default: false). I updated the usages in `statistics.js` and `map.js` accordingly

## Motivation and Context

Without this change the heatmap won't work anymore.

## How Has This Been Tested?

I tested the map on my local server and the map seems unaffected(both non-notify and notify pkmn are shown) and on the /stats page the heatmap works like a charm

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1576,8 +1576,8 @@ function processPokemon(item) {
          !isExcludedPoke && isPokeAlive) {
         // Add marker to map and item to dict.
         if (!item.hidden) {
-            const isBounceDisabled = false
-            const scaleByRarity = true
+            const isBounceDisabled = Store.get('isBounceDisabled')
+            const scaleByRarity = Store.get('scaleByRarity')
             const isNotifyPkmn = isNotifyPoke(item)
 
             if (item.marker) {
@@ -1859,7 +1859,7 @@ function redrawPokemon(pokemonList) {
         var item = pokemonList[key]
 
         if (!item.hidden) {
-            const scaleByRarity = true
+            const scaleByRarity = Store.get('scaleByRarity')
             const isNotifyPkmn = isNotifyPoke(item)
 
             updatePokemonMarker(item, map, scaleByRarity, isNotifyPkmn)
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1576,10 +1576,14 @@ function processPokemon(item) {
          !isExcludedPoke && isPokeAlive) {
         // Add marker to map and item to dict.
         if (!item.hidden) {
+            const isBounceDisabled = false
+            const scaleByRarity = true
+            const isNotifyPkmn = isNotifyPoke(item)
+
             if (item.marker) {
-                updatePokemonMarker(item.marker, map)
+                updatePokemonMarker(item.marker, map, scaleByRarity, isNotifyPkmn)
             } else {
-                newMarker = setupPokemonMarker(item, map)
+                newMarker = setupPokemonMarker(item, map, isBounceDisabled, scaleByRarity, isNotifyPkmn)
                 customizePokemonMarker(newMarker, item)
                 item.marker = newMarker
             }
@@ -1855,7 +1859,10 @@ function redrawPokemon(pokemonList) {
         var item = pokemonList[key]
 
         if (!item.hidden) {
-            updatePokemonMarker(item, map)
+            const scaleByRarity = true
+            const isNotifyPkmn = isNotifyPoke(item)
+
+            updatePokemonMarker(item, map, scaleByRarity, isNotifyPkmn)
         }
     })
 }
--- a/static/js/statistics.js
+++ b/static/js/statistics.js
@@ -417,8 +417,8 @@ function closeOverlay() { // eslint-disable-line no-unused-vars
 function processAppearance(i, item) {
     var spawnpointId = item['spawnpoint_id']
     if (!((spawnpointId) in mapData.appearances)) {
-        const isBounceDisabled = true // we don't need this functionality in our heatmap..
-        const scaleByRarity = false   // ..neither this..
+        const isBounceDisabled = true // We don't need this functionality in our heatmap..
+        const scaleByRarity = false   // ..nor this..
         const isNotifyPkmn = false    // ..and especially not this
 
         if (item['marker']) {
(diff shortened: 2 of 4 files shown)


static/js/statistics.js
```diff
@@ -420,7 +420,7 @@ function processAppearance(i, item) {
         if (item['marker']) {
             item['marker'].setMap(null)
         }
-        item['marker'] = setupPokemonMarker(item, map, true, false)
+        item['marker'] = setupPokemonMarker(item, map, true, false, false)
```
sebastienvercammen: `true, false, false` are not very descriptive. Can you put them in separate variables just above it so we can use the variable names to know what they are w/o looking at the function description?

static/js/statistics.js
```diff
@@ -434,7 +434,7 @@ function redrawAppearances(appearances) {
         var item = appearances[key]
         if (!item['hidden']) {
             item['marker'].setMap(null)
-            const newMarker = setupPokemonMarker(item, map, true, false)
+            const newMarker = setupPokemonMarker(item, map, true, false, false)
```
sebastienvercammen: Same here as above.

static/js/map.js
```diff
@@ -1576,10 +1576,14 @@ function processPokemon(item) {
          !isExcludedPoke && isPokeAlive) {
         // Add marker to map and item to dict.
         if (!item.hidden) {
+            const isBounceDisabled = false
```
sebastienvercammen: We should get `isBounceDisabled` from `Store`: unlike `scaleByRarity` we don't include the `Store` value.

For `scaleByRarity`, we should remove this line from `map.common.js`:
```
scaleByRarity = scaleByRarity && Store.get('scaleByRarity')
```
move `Store.get('scaleByRarity')` out of the function (for dependency injection: no more function dependency on the `Store`), and use it where appropriate.

ChrisSsSsSsS: > We should get `isBounceDisabled` from `Store`: unlike `scaleByRarity` we don't include the `Store` value.
What exactly do you mean by that, should I add a store value for `isBounceDisabled` and move it out of the function like `scaleByRarity`?

sebastienvercammen: @ChrisSsSsSsS Yes, so its default value can be changed via `custom.js`.

ChrisSsSsSsS: @sebastienvercammen ok seems reasonable :thumbsup:

static/js/statistics.js
```diff
@@ -417,10 +417,14 @@ function closeOverlay() { // eslint-disable-line no-unused-vars
 function processAppearance(i, item) {
     var spawnpointId = item['spawnpoint_id']
     if (!((spawnpointId) in mapData.appearances)) {
+        const isBounceDisabled = true // we don't need this functionality in our heatmap..
+        const scaleByRarity = false   // ..neither this..
+        const isNotifyPkmn = false    // ..and especially not this
```
sebastienvercammen: Capitalize `We`, end with punctuation `this.`.

Also: 

> nor this

instead of

> neither this

static/js/statistics.js
```diff
@@ -433,8 +437,12 @@ function redrawAppearances(appearances) {
     $.each(appearances, function (key, value) {
         var item = appearances[key]
         if (!item['hidden']) {
+            const isBounceDisabled = true // we don't need this functionality in our heatmap..
+            const scaleByRarity = false   // ..neither this..
+            const isNotifyPkmn = false    // ..and especially not this
```
sebastienvercammen: Same comments as above.

static/js/statistics.js
JapuDCret: ohh man, how could I have possibly missed that...
it's updated now!

PR #2313, opened 2017-09-25
Periodically log resource usage.

sebastienvercammen:
## Description
Periodically (once per 60s) log resource usage.

## Motivation and Context
Helps with debugging and when trying to improve server configuration/setup.

## How Has This Been Tested?
Python REPL.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1129,4 +1129,5 @@ def log_resource_usage_loop(loop_delay_ms=60000):
     # Helper method to log to specific log level.
     def log_resource_usage_to_info():
         log_resource_usage(log.info)
+
     periodic_loop(log_resource_usage_to_info, loop_delay_ms)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1130,4 +1130,4 @@ def log_resource_usage_loop(loop_delay_ms=60000):
     def log_resource_usage_to_debug():
         log_resource_usage(log.debug)
 
-    periodic_loop(log_resource_usage_to_info, loop_delay_ms)
+    periodic_loop(log_resource_usage_to_debug, loop_delay_ms)
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1071,12 +1071,14 @@ def resource_usage():
                 'available': mem_usage.available,
                 'used': mem_usage.used,
                 'free': mem_usage.free,
-                'percent_used': mem_usage.percent
+                'percent_used': mem_usage.percent,
+                'process_percent_used': proc.memory_percent()
             },
             'CPU': {
                 'user': cpu_usage.user,
                 'system': cpu_usage.system,
-                'idle': cpu_usage.idle
+                'idle': cpu_usage.idle,
+                'process_percent_used': proc.cpu_percent(interval=1)
             },
             'NET': {
                 'bytes_sent': net_usage.bytes_sent,
@@ -1103,6 +1105,9 @@ def resource_usage():
                 'fans': psutil.sensors_fans()
             }
             usage['connections']['unix'] = len(proc.connections('unix'))
+            usage['num_handles'] = proc.num_fds()
+        elif platform == 'win32':
+            usage['num_handles'] = proc.num_handles()
 
     return usage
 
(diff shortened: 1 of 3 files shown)


runserver.py
```diff
@@ -317,12 +318,17 @@ def main():
...
         t.start()
 
-    # db cleaner; really only need one ever.
+    # Database cleaner; really only need one ever.
     if args.enable_clean:
         t = Thread(target=clean_db_loop, name='db-cleaner', args=(args,))
         t.daemon = True
         t.start()
 
+    # Let's log some periodic resource usage stats.
```
friscoMad: Just start the thread when it will be logged.

pogom/utils.py
```diff
@@ -1050,3 +1051,83 @@ def get_async_requests_session(num_retries, backoff_factor, pool_size,
...
+    while True:
+        # Do the thing.
+        f()
+        # zZz :bed:
+        time.sleep(loop_delay_ms / 1000)
+
+
+# Periodically log resource usage every 'loop_delay_ms' ms.
+def log_resource_usage_loop(loop_delay_ms=60000):
+    # Helper method to log to specific log level.
```
friscoMad: Why using so convoluted system instead of just doing log.debug in log_resource_usage?

runserver.py
sebastienvercammen: @friscoMad What?

pogom/utils.py
sebastienvercammen: @friscoMad They're layers of abstraction that make it simpler to work with. A generic method for looping, a method to log resource usage that's being plugged into the generic method, and a method to plug into the thread in the same structure we've used everywhere else.

Each one is a separate step of the process, there is no overlap or spaghetti code.

runserver.py
friscoMad: with default setup nothing is logged only when -v is used, so I was saying that we should only start the thread when the debug level is enabled.

pogom/utils.py
friscoMad: I see the reason for a generic loop maybe in the future we can remove threads and just have 1 task runner thread, but I really don't see why both log_resource_usage and log_resource_usage_to_debug are needed.
log_resource_usage seems the perfect point to log it and if we want to parametrize the level then that's fine but passing the string to a random function seems too much to me.

Both are fairly specific and I don't see log_resource_usage being reused in the near future, you know that I like to split and make smaller functions to make them more readable and easy to maintain but I think that's too much.

friscoMad:
In general I don't see the point to log soo much system data, the system IO, fans or memory usage provide little information for optimizing or benchmarking setups.
We could use proc functions 'pid', 'name', 'memory_full_info', 'memory_percent', 'cpu_percent', 'cpu_times', 'num_ctx_switches', 'num_fds'/'num_handles', 'connections' instead of the global ones.

pogom/utils.py
sebastienvercammen: @friscoMad `log_resource_usage_to_debug` is just a helper method local to `log_resource_usage_loop`, used to make it very easy to pass a scope where we're logging to the logging method of our choice.

Rather than adding more parameters to functions (which leads to a massive argument list in some cases, as is already happening in RM), we use a closure to solve that problem.

sebastienvercammen:
@friscoMad 

> I general I don't see the point to log soo much system data

Except for process ID, all data being logged (only once every minute, I should note) is dynamic and relevant to debugging scenarios.

You're correct about using `proc.pid` rather than `os.getpid()`, I forgot to swap them out (fixed on latest commit). All the other methods you mention don't report the same information as the methods I'm currently using.

`connections` and `num_threads` already use `proc`.

> the system IO, fans or memory usage provide little information for optimizing or benchmarking setups

Fans or temperatures are less often related to debugging, but it does happen. System I/O or memory usage..... those are fundamental.

friscoMad:
Maybe it's just me that I can't think of those scenarios and you can share with us but it seem more relevant to know the process memory usage or process cpu percent usage that the system information, we have no information of other processes and we can do little to improve them.

sebastienvercammen:
@friscoMad I added additional stats, hopefully they satisfy your scenario.

PR #2317, opened 2017-09-26
Don't use toastr when notifications are blocked.

sebastienvercammen:
## Description
When a client denies Notifications permission, we shouldn't show any notifications at all. We should assume the user intentionally blocked them so they wouldn't see any notifications at all.

toastr fallback is still enabled for browsers that don't support the Notifications API.

## Motivation and Context
By popular demand.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1909,9 +1909,9 @@ function sendNotification(title, text, icon, lat, lon) {
     /* Push.js requests the Notification permission automatically if
      * necessary. */
     Push.create(title, notificationDetails).catch(function () {
-        /* Push.js doesn't fall back automatically if the user denies the
-         * Notifications permission. */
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
@sebastienvercammen this was discussed in the original PR:
https://github.com/RocketMap/RocketMap/pull/2138#issuecomment-317709143

sebastienvercammen:
@pogo-excalibur Yes, it definitely was, but we've gotten reports about changing the behavior because it's more intuitive.

pogo-excalibur:
@sebastienvercammen fair enough :shrug:

pogo-excalibur:
Any reason for leaving the function?

sebastienvercammen:
@pogo-excalibur Clarification on the decision to change the behavior.

PR #2322, opened 2017-09-30
Don't allow -ns and -os to be used together

KartulUdus:
## Description
When a user forgets `only-server` in their config and launches with -ns (or vice versa)
there's an Uncaught exception:
```
2017-09-29 22:30:18,938 [        MainThread][     runserver][   ERROR] Uncaught exception
  File "runserver.py", line 505, in <module>
    main()
  File "runserver.py", line 415, in main
    while search_thread.is_alive():
UnboundLocalError: local variable 'search_thread' referenced before assignment
```

## Motivation and Context

Pebcak makes console unhappy and users confused.

## How Has This Been Tested?

Local test build in linux 14.04, Python 2.7.6.
launched with `-ns -os`
Tried `only-server` in config and `-ns` in console 
Tried `-os` & `-ns` alone to check that it still launches. 
## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -487,11 +487,6 @@ def get_args():
 
     args = parser.parse_args()
 
-    if args.only_server and args.no_server:
-        print("Error: You can't use no-server and only-server at the same "
-              "time, silly.")
-        sys.exit(1)
-
     if args.only_server:
         if args.location is None:
             parser.print_usage()
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -487,6 +487,11 @@ def get_args():
 
     args = parser.parse_args()
 
+    if (args.only_server and args.no_server):
+        print("Error: You can't use no-server and only-server at the same "
+              "time silly.")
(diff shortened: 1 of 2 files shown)


pogom/utils.py
```diff
@@ -487,6 +487,11 @@ def get_args():
 
     args = parser.parse_args()
 
+    if (args.only_server and args.no_server):
```
sebastienvercammen: Remove surrounding brackets, there's no need.

pogom/utils.py
```diff
@@ -487,6 +487,11 @@ def get_args():
...
+        print("Error: You can't use no-server and only-server at the same "
+              "time, silly.")
```
sebastienvercammen: For logging, use
```
log.critical("Error: You can't use no-server and only-server at the same time, silly.")
```

KartulUdus: Caused error in utils.py. Changed logging and moved to runsever.py as suggested

PR #2334, opened 2017-10-06
Add colons to boolean (true/false) options

blakewenloe:
## Description
Adds `:` or `=` to non-flag options so they are more easily discernible from flags.

## Motivation and Context
Easier for setting options quickly. We've been having to go through and change them for a while. It's easier just pull them with the colons as default.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -26,14 +26,14 @@
 #location:                      # Location, can be an address or coordinates.
 #step-limit:                    # Steps (default=10)
 #scan-delay:                    # Time delay between requests in scan threads. (default=12)
-#no-gyms                        # Disables gym scanning (default=False)
-#no-pokemon                     # Disables pokemon scanning. (default=False)
-#no-pokestops                   # Disables pokestop scanning. (default=False)
-#no-raids                       # Disables raid scanning (default=False)
-#print-status                   # Enable ENTER to switch between log view and a status view.
+#no-gyms:                        # Disables gym scanning (default=False)
+#no-pokemon:                     # Disables pokemon scanning. (default=False)
+#no-pokestops:                   # Disables pokestop scanning. (default=False)
+#no-raids:                       # Disables raid scanning (default=False)
+#print-status:                   # Enable ENTER to switch between log view and a status view.
                                 # Optionally add ": status" to start in status view and ": logs" to start in log view. (default=False)
 #status-page-password:          # Enables and protects the /status page to view status of all workers. (default=None)
-#captcha-solving                # Enables captcha solving. (default=False)
+#captcha-solving:                # Enables captcha solving. (default=False)
 
 
 # Database settings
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
Would you mind outlining what you mean when by 'flag options'?

Because from what I can see you've added colons to several things that *are* flags.

The commands I think could have colons: `account-max-spins`, `verbosity` and `log-path`.

ghost:
I suppose I should specify flags that take Boolean arguments.

ghost:
Or rather *any* arguments.

friscoMad:
Shouldn't it be the other way around, flags are false by default and enabling them (without value) sets them to true, you can use `flag: True` but it is the same as just `flag`

ghost:
Well, sort of. Here's my logic:

To set flags true, you just have to uncomment them in the config.

To set items that require boolean values, you have to check if there's a `default: false` or `default: true` at the end of the line and then add a colon + the value that you want them to be. If you uncomment them and don't set a value, you'll get an error. It wouldn't make sense to prefill the fields true/false value but it would make sense to at least make it clear by just having colons by default.

pogo-excalibur:
FYI: the config parser that we use (ConfigArgParse) considers *all* boolean fields to be flags.

pogo-excalibur:
FYI: the config parser that we use (ConfigArgParse) considers *all* boolean fields to be flags.

You don't need to (and perhaps shouldn't be: @sebastienvercammen would you like to weigh in on this?) set boolean parameters using `param: True` or `param: False`, you can just use `param` and `#param`.

ghost:
I think if it accepts a value, it should be a complete field that's ready to accept that value; without needing to enter the colons every time. If all fields were defaulted to true/false by uncommenting them, that would be more efficient. I agree with you there. There's just a few fields that take values that aren't boolean, so the rest of them being flags might streamline that.

tomballgithub:
A long time ago the sample config.ini did use booleans but it was purposefully changed to be the current method:  `param` and `#param`

ghost:
I just tried it. Defintely can't just uncomment the ones that are Boolean.

KartulUdus:
`gym-info` not `gym-info:` in config works

ghost:
Ah I see that now. Derp.

sebastienvercammen:
@pogo-excalibur @blakewenloe The fields you're describing are sometimes called binary fields: they only have two states (enabled or disabled). It's true that, in programming, this is usually implemented with the values `True` and `False`, but this isn't always the case.

ConfigArgParse (and argparse itself) are an abstraction layer on top of configuration files built in Python.

In essence, the configuration deals with items in terms of "enabled/disabled" rather than using programming- or language-specific terms, so the proper way to use it is to comment/uncomment those fields, rather than requiring "True" or "False".

The comments in the config example could be reworded to "enabled/disabled", and the `log-path` could have `logs/` as default value in the comments, to be clearer.

sebastienvercammen:
PR closed. A docs PR could still be made if someone picks it up.

PR #2336, opened 2017-10-10
Improve efficiency of L30 accounts

AEtHeLsYn:
## Description
With this PR we aim to improve efficiency of level 30 accounts by using the same account each parse loop instead of using several accounts if more than one encounter is needed. This is particularly helpful when scanning nests, as there may be some whitlisted pokemon together and we may need to use several level 30 accounts to encounter all pokemons before scanning the next step. 

## Motivation and Context
The change potentially reduces the usage of level 30 accounts. 

## How Has This Been Tested?
Local testing map

## Types of changes

## Checklist:


javisabalete:
What about this?

PR #2345, opened 2017-10-16
Change -DC to -ec (enable clean) to avoid confusion of -dc (display in console)

BrownSlaughter:
## Description
a simple flag change so -DC is not confused with -dc
seems strange to still have it as -DC when it not longer disables database cleaner anyway

## Motivation and Context

it doesn't solve anything really except my mild OCD

## How Has This Been Tested?
untested but such a simple change cant cause issues right? 

## Types of changes

- [ x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x ] I have updated the documentation accordingly.

### DIFF
--- a/docs/first-run/commandline.md
+++ b/docs/first-run/commandline.md
@@ -32,7 +32,7 @@
                     [--db-pass DB_PASS] [--db-host DB_HOST]
                     [--db-port DB_PORT]
                     [--db-threads DB_THREADS] [-wh WEBHOOKS] [-gi]
-                    [--enable-clean]
+                    [-ec] [--enable-clean]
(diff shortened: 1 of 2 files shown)


sebastienvercammen:
@BrownSlaughter The reason for the change in flag was extensively discussed in the original PR that introduced the rework, and on Discord.

Considering short-hand flags are built for the command-line, they are focused on the context of easily remembering them and making them as short and simple as possible.

For `-DC`, its short-hand form comes from prefixing all database-related CLI options with `-D` (since `-D` stands for database) and the `C` stands for its cleaner (capitalized since `-dc` was already in use).

Rather than remove `-DC`, all other options are planned to be reworked to the new and clearer forms, where long forms should be descriptive and short forms should be prefixed with their category. This will also change `-dc` into something else.

PR #2350, opened 2017-10-18
Set -ams default to 20 to prevent early Softbans

Alderon86:
## Description

80 spins by default is a bit too high.
## Types of changes

## Checklist:


friscoMad:
Yeah I think it is a lot more sensible value.

PR #2355, opened 2017-10-24
CLI tool to mass level accounts to level 2

friscoMad:
## Description

Small script that logs in, does a GMO and spin all stops in range.
It uses RM functions and params so it is really easy to run and manage.

Side change: Tools and Spawnpoint-Clustering folders renamed to be only lower case.

## Motivation and Context
-spin is nice but not enough for almost everyone, most people was using a external tool not longer working so I made this small one to handle our use case without the need of a bot tool.

## How Has This Been Tested?

Tested with my last batch of accounts, ~~not fully tested the latest commit as I did some changes to the logging and interrupt handling but should work~~ Fully tested now.

## Types of changes

## Checklist:


tools/levelup.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/python
```
Alderon86: is it really needed to create a new file? we could rewrite accounts.py to handle it also.

friscoMad: The script is small but I don't think it is a good idea to have a class/module inside RM that is executable on it's own.
Also we already have a code base that it is a mess and we should be rebasing to make each module clean and simple and not the other way around.

For example: While doing this I found out that inside setup_api we do check the console args to store in the worker status the proxy using index or full url, and we pass around the proxy url in the worker status info (should be display info only). It took me a while until I found what the hell was going on on a simple enough function like setup api that should only need the account and proxy params.

friscoMad:
Fixed a couple of issues especially the one that left some accounts in the last part of the file without leveling them properly.

sebastienvercammen:
@friscoMad Can you fix the issues with Travis?

friscoMad:
@sebastienvercammen Umm all the errors comes from the spanwpoint clustering files, not sure why it started complaining after the folder rename. I will try to fix everything this weekend.

Also I have been using this for some time and I would say that it works really nice but it some error happens accounts re not requeued and if you don't look to the traces you have no idea and could end adding accounts still at level 1 (and get catpchas), so maybe I will need to add some retry logic to the process.

ClarkyKent:
+1 for me. Works great.  Just need to make sure starting location have a pokestop.

jagauthier:
How do you envision this tool being integrated with RM?  This is currently a PR, which generally indicates it would be merged.  But if it's intended to be a standalone tool shouldn't it use standalone files?

friscoMad:
@jagauthier I don't see the process being included in the main RM unless we have an account management panel, it could be added during account start that if the account lvl is 1 it does the spin in a known location, but we will need to add a new param for the spin location.
And adding a new flag like -cd it doesn't make sense for me, flags can affect settings but shouldn't change the behaviour so heavily.
On the other hand being a stand alone script for me does not necesarily need to duplicate everything. we have modules not only because is great to organice code, but also because it makes it easy to reuse, that way the script should be up to date with little effort.
My opinion is that this is ok as a external script but if most people think it shoud be done differently I am open to change it.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
 ```
 
-To set the maximum number of Pokéstop spins per account per hour (default: 80), add -ams 30 to your cli parameters or edit your configuration file:
+This setting will spin a pokestop as the account see one that is spinnable (inside the radius) if the account is still at level 1, otherwise if the `account-max-spins` limit is not reached (default is 20 per account per hour prorrated to the minute) the account has a 50% chance to spin it.
+
+This setting could be enough for some maps with high density of stops, as the accounts will get near one soon enough to avoid the captcha, otherwise you will need to enable [catpcha handling](http://rocketmap.readthedocs.io/en/develop/first-run/captchas.html) to keep them working until there is a stop near.
+
+## Levelup tool
+
+Inside the RM folder there is a small python script that will go trought the account list, ask for the map at 1 location, and spin all stops in range (following `account-max-spins` limit), that way you can be sure that accounts are properly leveled before using them.
```
jaake77: Fix spelling on -- through

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
 ```
 
-To set the maximum number of Pokéstop spins per account per hour (default: 80), add -ams 30 to your cli parameters or edit your configuration file:
+This setting will spin a pokestop as the account see one that is spinnable (inside the radius) if the account is still at level 1, otherwise if the `account-max-spins` limit is not reached (default is 20 per account per hour prorrated to the minute) the account has a 50% chance to spin it.
+
+This setting could be enough for some maps with high density of stops, as the accounts will get near one soon enough to avoid the captcha, otherwise you will need to enable [catpcha handling](http://rocketmap.readthedocs.io/en/develop/first-run/captchas.html) to keep them working until there is a stop near.
+
+## Levelup tool
+
+Inside the RM folder there is a small python script that will go throught the account list, ask for the map at 1 location, and spin all stops in range (following `account-max-spins` limit), that way you can be sure that accounts are properly leveled before using them.
```
jaake77: through

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
-# Tutorial Completion
-## RocketMap's tutorial completion for new accounts
-RocketMap now completes the tutorial steps on all accounts on first log in, there is no more `-tut` argument to complete the tutorial.
+# Account leveling
```
sebastienvercammen: The old title includes "tutorial completion for new accounts". This is still relevant.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+RocketMap completes the tutorial steps on all accounts on first log in and sets a pokemon as buddy if it does not have one.
```
sebastienvercammen: > Pokémon

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
 
-It's recommended to enable pokestop spinning in the config to get your accounts to level 2.
+Accounts with level 1 will get captchas after some time and will stop working unless you setup [catpcha handling](http://rocketmap.readthedocs.io/en/develop/first-run/captchas.html):
+
+
+To avoid this it is recommended to level accounts at least to level 2, that is as simple as spin a pokestop and there are two ways to do so from RM:
```
sebastienvercammen: > this, it's

> to level 2. It's as simple as ...

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+ * Enabling pokestop spinning during regular operation
```
sebastienvercammen: > Pokéstop

> regular scanning

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+ * Using the levelup tool provided with RM
```
sebastienvercammen: > level-up tool

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+## Pokestop Spinning
```
sebastienvercammen: > Pokéstop

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
 
 To enable Pokéstop spinning, add pokestop-spinning to your configuration file, or -spin to your cli parameters.
 
 ```
 pokestop-spinning
 ```
 
-To set the maximum number of Pokéstop spins per account per hour (default: 80), add -ams 30 to your cli parameters or edit your configuration file:
+This setting will spin a pokestop as the account see one that is spinnable (inside the radius) if the account is still at level 1, otherwise if the `account-max-spins` limit is not reached (default is 20 per account per hour prorrated to the minute) the account has a 50% chance to spin it.
```
sebastienvercammen: > With this setting enabled, RM's scanner instances will try to spin a Pokéstop (50% chance to spin) if it's within range, the account is still level 1 and the `--account-max-spins` limit hasn't been reached (default is 20 per account per hour).

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+This setting could be enough for some maps with high density of stops, as the accounts will get near one soon enough to avoid the captcha, otherwise you will need to enable [catpcha handling](http://rocketmap.readthedocs.io/en/develop/first-run/captchas.html) to keep them working until there is a stop near.
```
sebastienvercammen: > a high density of Pokéstops

> until there is a Pokéstop within range.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+## Levelup tool
```
sebastienvercammen: > Level-up

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+Inside the RM folder there is a small python script that will go through the account list, ask for the map at 1 location, and spin all stops in range (following `account-max-spins` limit), that way you can be sure that accounts are properly leveled before using them.
```
sebastienvercammen: > In the tools folder is a small ... send a map request at a location, and spin all Pokéstops in range

> With this tool, you can make sure all accounts are level 2 before using them for scanning.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+The tool uses the same config file and options as RM (the ones that apply) so the setup and run is pretty simple, just change the location to some coodinates that are near 1 or more pokestops and change the worker setting to the number of simultaneus accounts you want to level up.
```
sebastienvercammen: > coordinates

> Pokéstops

> to the number of accounts you want to level up simultaneously.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+In the console you will see the initial level of each account, the spining and the final level, the script will end when all accounts have done the process or have failed 3 times.
```
sebastienvercammen: > the Pokéstop spinning and the final level. The script will end automatically when all accounts have finished the process or have failed 3 times.

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+To run it just make sure you are at RM root folder (it will not work otherwise) and just call:
```
sebastienvercammen: > To run the script, go to RM's root folder and execute:

docs/first-run/tutorial.md
```diff
@@ -1,17 +1,42 @@
...
+
+```
+python tools/levelup.py
+```
+
+All command line flags available in RM can be used here too (but not all will have any effect). So you could increase `account-max-spins` and change location and workers from the command line without needing to modify the config file with something like:
```
sebastienvercammen: > (but not all of them will have an effect)

tools/levelup.py
sebastienvercammen: 👍 Separate file is best.

docs/first-run/tutorial.md
friscoMad: Not sure  what you want me to edit, just add the subtitle?

docs/first-run/tutorial.md
friscoMad: Reworded this as the spin will have a 100% chance for lvl1 accounts and 50% for others.

PR #2359, opened 2017-10-25
Add team_id to raid webhook data

sLoPPydrive:
...so that notifications can also show the current gym owner team.

## Motivation and Context
Let users see in notifications whether they will get any bonus balls or not. And let them decide early if they want/need to take over the gym before the raid starts.

## How Has This Been Tested?
Works on my map.

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2246,6 +2246,7 @@ def parse_map(args, map_dict, step_location, db_update_queue, wh_update_queue,
                             wh_raid = raids[f.id].copy()
                             wh_raid.update({
                                 'gym_id': b64_gym_id,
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
Could you add the field to the docs as well?

And I think to the list of key fields (in pogom/webhooks.py).

friscoMad:
I don't think we want to trigger a new notification if the team changes sides during egg phase so I would avoid the change in webhooks.py

pogo-excalibur:
@friscoMad That seems reasonable. We can add it later only if there's demand for it.

But an update to the docs would still be good @sLoPPydrive .

PR #1604, opened 2016-12-05
Add Captcha to ps and status

onilton:
Add Captcha to ps and status.
This is just a clean-up of #1597, Also I have removed #1576 to keep the PR to single thing.

## Description
Added captcha count by worker to status and ps outputs

## Motivation and Context
Helps the user see captcha use

## How Has This Been Tested?
Personal

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2340, opened 2017-10-12
Added option to status page to only show Instances/Main Workers 

123FLO321:
## Description
Remove "Group by Workers" and ad "Show Instances" instead.

## Motivation and Context
This change will add the option to only show the instance-status without the worker-status.

## How Has This Been Tested?
Tested on local instance.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2372, opened 2017-11-07
Add total summary at top of status web page across all map instances

tomballgithub:
## Description

**_This PR is a rebase of #1723 which accidentally had its branch deleted and so I recreated it/_**

Adds total summaries at top of status web page across all map instances whether in group-by-mainworker mode or not. This was waiting on release of PR #1604 and basically adds up the stats for each main_worker and generates stats across all workers.

These total stats include: active workers, successes, failures, empties, skips, and captchas.

I had to muddle through learning quite a bit about javascript to make this work. I am open to suggestions on how to clean things up.

The per hr and per month stats use an average of the elapsed time of each instance, assuming that they are launched at approximately the same time. Any other method would be extravagant. If instances are launched random times from each other the top-level per hr and per month numbers are not trustworthy

## Motivation and Context

If you have multiple instances, there is not a way to see the totals across all instances. PR #1604 added per-main-worker support and this builds upon that

## How Has This Been Tested?

Tested on my Windows server with 10 instances of 5-25 workers each
Verified totals and calculations are correct
Verified behavior works in group-by-mainworker mode and other mode
**_I have been running this since January 2017 without issue_**

image

image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -40,7 +40,7 @@ args = get_args()
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 22
+db_schema_version = 23
 
 
 class MyRetryDB(RetryOperationalError, PooledMySQLDatabase):
@@ -1069,6 +1069,13 @@ class MainWorker(BaseModel):
     accounts_working = IntegerField()
     accounts_captcha = IntegerField()
     accounts_failed = IntegerField()
+    success = IntegerField(default=0)
+    fail = IntegerField(default=0)
+    empty = IntegerField(default=0)
+    skip = IntegerField(default=0)
+    captcha = IntegerField(default=0)
+    start = IntegerField(default=0)
+    elapsed = IntegerField(default=0)
 
     @staticmethod
     def get_account_stats():
@@ -3108,6 +3115,10 @@ def database_migrate(db, old_ver):
         db.execute_sql('ALTER TABLE `spawnpoint` '
                        'ADD CONSTRAINT CONSTRAINT_4 CHECK ' +
                        '(`latest_seen` <= 3600);')
+
+    if old_ver < 23:
+        db.drop_tables([WorkerStatus])
+        db.drop_tables([MainWorker])
 
     # Always log that we're done.
     log.info('Schema upgrade complete.')
--- a/static/js/status.js
+++ b/static/js/status.js
@@ -5,6 +5,24 @@ var showHashTable = true
 var showInstances = true
 var showWorkers = true
 var hashkeys = {}
+var statshash = 'summarystats'  /* unique statistics worker name */
+var active
+var success
+var failed
+var empty
+var skipped
+var captcha
+var mainWorkers
+var elapsedTotal
+var elapsedSecs
+var elapsedHours
+var sph
+var fph
+var eph
+var skph
+var cph
+var ccost
+var cmonth
 
 // Raw data updating
 var minUpdateDelay = 1000 // Minimum delay between updates (in ms).
@@ -167,6 +185,7 @@ function processHashKeys(i, hashkey) {
 }
 
 function parseResult(result) {
+    addTotalStats(result)
     if (showInstances) {
         $.each(result.main_workers, processMainWorker)
     }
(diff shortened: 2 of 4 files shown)


sebastienvercammen:
Hey @tomballgithub can you rebase? :)

tomballgithub:
Rebased @sebastienvercammen

static/js/status.js
```diff
@@ -322,6 +341,88 @@ function updateStatus() {
...
+    fph = 0
+    eph = 0
+    skph = 0
+    cph = 0
+    ccost = 0
+    cmonth = 0
+
+    $.each(result.main_workers, getStats)
+
+    if ((mainWorkers > 1) || !(showWorkers && showInstances) {
```
pogo-excalibur: Missing `)`

tomballgithub: Sorry, typo, fixed...

friscoMad:
The DB change is needed?
With the previous version the same information was shown so I am not sore why we need to store it.
Sorry if it is a dumb question, I haven't dived in too much to find how it works.

tomballgithub:
This information is not currently stored in the database.  I think you are referring to the text message strings that contain that information per instance.  I stored them directly in database to facilitate doing the math needed to aggregative the information.

This is a text string per instance:
image

tomballgithub:
Do you think it would be better to just parse those strings and pull the data out, avoiding the database accesses?

friscoMad:
We are accessing the DB anyway so I don't think it is a problem to add this in the overseer, just curious as it was not needed before and it was showing the same totals. but I get why you need it.

tomballgithub:
I have verified that everything works OK with the recent commit:  Added option to status page to only show Instances/Main Workers (#2340)

Current display for reference:
image

jagauthier:
Originally, I started with one instance.  I did not see the total status, so I checked and I see you only show that if there are multiple instances.  Great, so I fired up another.  The total status area showed up, but it's not at the top. It's between the two instances.

image

static/js/status.js
```diff
@@ -322,6 +341,88 @@ function updateStatus() {
...
+function addStatsWorker(hash) {
+    var worker = `
+    <div id="worker_${hash}" class="worker">
+    <span id="name_${hash}" class="name"></span>
+    <span id="method_${hash}" class="method"></span>
+    <span id="message_${hash}" class="message"></span>
+    </div>
+    `
+
+    $(worker).appendTo('#status_container')
```
Alderon86: with `prepend` instead of appendTo it should be always on top

tomballgithub:
Made recommended change based on feedback

tomballgithub:
Switched back to appendTo and rebased...

tomballgithub:
Agreed, and you can also just hit refresh

pogom/models.py
```diff
@@ -40,7 +40,7 @@
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 21
+db_schema_version = 22
```
sebastienvercammen: Should be 23 now due to #2416.

tomballgithub:
Sorry for missing this one.  Rebased...

tomballgithub:
Needed change made

static/js/status.js
```diff
@@ -322,6 +341,88 @@ function updateStatus() {
...
+    mainWorkers += 1
+
+    elapsedTotal += worker['elapsed']
+    elapsedSecs = elapsedTotal / (i + 1)
+    elapsedHours = elapsedSecs / 3600
+}
+
+function getActive(i, worker) {
+    active += 1
+}
```
sebastienvercammen: This is a weird one.
```
active = 0
$.each(result.workers, getActive)
```
Any reason you didn't do
```
active = result.workers.length
```
?

static/js/status.js
```diff
@@ -5,6 +5,24 @@ var showHashTable = true
...
+var elapsedTotal
+var elapsedSecs
+var elapsedHours
+var sph
+var fph
+var eph
+var skph
+var cph
+var ccost
+var cmonth
```
sebastienvercammen: Instead of separate vars for each stat, consider using a stats object to reference. It'll keep the scope clean.

static/js/status.js
```diff
@@ -322,6 +341,88 @@ function updateStatus() {
...
+
+    active = 0
+    success = 0
+    failed = 0
+    empty = 0
+    skipped = 0
+    captcha = 0
+    mainWorkers = 0
+    elapsedTotal = 0
+    elapsedSecs = 0
```
sebastienvercammen: Consider
```
elapsedSecs = 1
```
for
```
sph = (success * 3600 / elapsedSecs) || 0
fph = (failed * 3600 / elapsedSecs) || 0
eph = (empty * 3600 / elapsedSecs) || 0
skph = (skipped * 3600 / elapsedSecs) || 0
cph = (captcha * 3600 / elapsedSecs) || 0
```
or
```
elapsedSecs = Math.max(elapsedSecs, 1)
sph = (success * 3600 / elapsedSecs) || 0
...
```
if it doesn't interfere much with the stats.

static/js/status.js
sebastienvercammen: To fix the issue, rather than use append/prepend, consider adding an extra, static HTML container in a fixed position and target it by its unique ID.

static/js/status.js
tomballgithub: I don't know how to do this and would need to investigate if a required improvement

static/js/status.js
tomballgithub: I don't know how to do this and would need to investigate if a required improvement

static/js/status.js
tomballgithub: There is a lot I don't know with javascript and would need to investigate this is a required improvement.  It seems straightforward

static/js/status.js
tomballgithub: Let me look at this.  I believe you are trying to avoid a divide by 0 condition

static/js/status.js
sebastienvercammen: @tomballgithub No problem! The problem with creating more and more variables in the global scope, is that it "pollutes" it. This is a way of saying that we're adding too much into one scope, while it's generally best to keep variables within the scope they belong in, so different parts of the code don't get mixed.

Some references on scope:

* https://en.wikiversity.org/wiki/Introduction_to_Programming/Scope
* https://users.cs.cf.ac.uk/Dave.Marshall/PERL/node52.html

In our case, instead of having all these variables, my suggestion is to use an object to store all the statistics in:
```
var stats = {
    ...
    'skph': 0,
    'cph': 0,
    ...
}
```
and so on, which gives us an easy reference to work with. To access/update them, we'd use:

```
stats['cph'] += 1
```

sebastienvercammen:
Triple tested DB migration, build, lint, and final result. 👍

PR #2375, opened 2017-11-16
Update device_info devices and iOS versions

FrostTheFox:
Updated the possible iOS versions and iPhone model.
## Description

Removed iOS 8, as it will be deprecated for PoGo. Added iOS 11. Added iPhone 8/8+ and iPhone X. I believe the logic is correct as to which devices will have specific versions to use, I sourced device identifiers and internal names from theiphonewiki.com, iOS versions from Wikipedia. iOS 11.0 was left out, it *is* possible certain devices may have that version, but it would require a unique list for iPhone X and iPhone 8 (which I believe both shipped 100% with iOS 11.0.1 and above), so I just felt it was better to leave that one version out. Most people will be on higher versions.
## Motivation and Context

Niantic is removing support for iOS 8. https://support.pokemongo.nianticlabs.com/hc/en-us/articles/115014888368-Discontinued-support-for-iOS-8

## How Has This Been Tested?

Passed flake8. Not tested.
## Types of changes

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -945,7 +945,7 @@ def generate_device_info(identifier):
     # 10.0 was only for iPhone 7 and 7 Plus, and is rare.
     ios10 = ('10.0.1', '10.0.2', '10.0.3', '10.1', '10.1.1', '10.2', '10.2.1',
              '10.3', '10.3.1', '10.3.2', '10.3.3')
-    ios11 = ('11.0.1', '11.0.2', '11.0.3', '11.1', '11.1.1')
+    ios11 = ('11.0.1', '11.0.2', '11.0.3', '11.1', '11.1.1', '11.1.2')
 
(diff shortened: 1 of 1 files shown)


friscoMad:
Only iPhone 5s (wiki says it is 'N51AP','N53AP' but we have those as iPhone 6) and above can run iOS 11, so I will cut any iPhone below 6 from the list just in case.

FrostTheFox:
@friscoMad The iPhonex,x versions are NOT relative to the marketed name. The iPhone 5s is iPhone6,1 and 6,2 and they are N51AP and N53AP on the internal names respectively.

iPhone 5/5c are 5,1 - 5,2 - 5,3 - 5,4 and are excluded from iOS 11 as such.

https://www.theiphonewiki.com/wiki/Models This chart is very helpful.

friscoMad:
I thought it could be something like that.

PR #2377, opened 2017-11-19
Improved Notification Setting Controls

tomballgithub:
## Description

This adds 3 new interrelated settings requested by my users.  I realize some believe that there are 'already too many settings' but I believe these are actually useful and all users are not the same and as such these don't need to be hard-coded.  # 1 especially is a new feature versus the others being control of existing internal settings.

**1.  Hide pokemon which are not being notified (default: off)**
Once implemented, I unexpectedly saw the high value of this.  It hides all pokemon which are not being notified due to IV, Rarity, or Pokemon type.   It gives you a nice clean map with only the items you are interested in seeing (the notified pokemon)

**2. Disable browser popups.  (default: off)** 
Currently if you have notifications enabled you will get the popup notifications unless you go in and block them in the browser.  It can be very annoying if you set a notification incorrectly and get endless popups.  I have received many complaints from mobile users who want to see the pokemon on the map (bouncing) but not get popups.  This can be combined with # 1 above for example for a nice clean map with no popups

**3. Disable bouncing.  (default: off)** 
There is currently no way for the user to control the bouncing of notified pokemon.  A bouncing pokemon jumps out at you but it can also be difficult to click on it when it is moving.  This allows you to turn that off.  When combined with # 1 it allows you to show only notified pokemon and there is no real need for the bouncing

## Motivation and Context

Described above

## How Has This Been Tested?

I tested it on a development map to work out the kinks.
I then moved it to my production map without issue and reports have been good.

image

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -934,6 +934,14 @@ var StoreOptions = {
         default: false,
         type: StoreTypes.Boolean
     },
+    'hideNotNotified': {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    'showPopups': {
+        default: true,
+        type: StoreTypes.Boolean
+    },
     'playSound': {
         default: false,
         type: StoreTypes.Boolean
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -435,6 +435,9 @@ function initSidebar() {
     $('#scanned-switch').prop('checked', Store.get('showScanned'))
     $('#spawnpoints-switch').prop('checked', Store.get('showSpawnpoints'))
     $('#ranges-switch').prop('checked', Store.get('showRanges'))
+    $('#hideunnotified-switch').prop('checked', Store.get('hideNotNotified'))
+    $('#popups-switch').prop('checked', Store.get('showPopups'))
+    $('#bounce-switch').prop('checked', Store.get('isBounceDisabled'))
     $('#sound-switch').prop('checked', Store.get('playSound'))
     $('#pokemoncries').toggle(Store.get('playSound'))
     $('#cries-switch').prop('checked', Store.get('playCries'))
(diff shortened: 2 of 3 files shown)


static/js/map.common.js
```diff
@@ -934,6 +934,14 @@ var StoreOptions = {
...
         type: StoreTypes.Boolean
     },
+    'hideNotNotified': {
+        default: false,
+        type: StoreTypes.Boolean
+    },
+    'showPopups': {
+        default: true,
+        type: StoreTypes.Boolean
+    },
```
friscoMad: Missing isBounceDisabled default value.

static/js/map.js
```diff
@@ -2820,6 +2823,21 @@ $(function () {
         }
     })
 
+    $('#bounce-switch').change(function () {
+        Store.set('isBounceDisabled', this.checked)
+        location.reload()
```
friscoMad: Ok I thought I was doing something wrong but it makes some sense, you need to reload in this cases as it is easier than reprocessing everything but the UX is really weird, no one expects to get a reload just clickin in a switch.
The only solution I can think of is telling the user about the reload (alert) or finding a way to reload the map without reloading the page.

static/js/map.js
```diff
@@ -2820,6 +2823,21 @@ $(function () {
...
+    })
+
+    $('#hideunnotified-switch').change(function () {
+        Store.set('hideNotNotified', this.checked)
+        location.reload()
+    })
+
+    $('#popups-switch').change(function () {
+        Store.set('showPopups', this.checked)
+        location.reload()
```
friscoMad: This reload is not needed, previous notifications will dissapear with time and no new ones will appear.

static/js/map.js
friscoMad: Probably you can check how it is done with the pokemon hide and just apply it to everything that is not in the notify list.

static/js/map.js
```diff
@@ -2820,6 +2823,21 @@ $(function () {
...
     })
 
+    $('#bounce-switch').change(function () {
+        Store.set('isBounceDisabled', this.checked)
+        location.reload()
+    })
+
+    $('#hideunnotified-switch').change(function () {
+        Store.set('hideNotNotified', this.checked)
+        location.reload()
```
friscoMad: This is the more complex to solve but probably you can also iterate over all pokes and remove/add the bouncing.

tomballgithub:
Rebased...

tomballgithub:
Rebased...

PR #2380, opened 2017-11-28
Show Gym-Sidebar if Show-Raids active

123FLO321:
## Description
Also show Gym-Sidebar if Raids are shown.

## Motivation and Context
Some users where confused because the switch did not show but was still taken into affect.

## How Has This Been Tested?
Testet on locan instance and working as intended.

sc11

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2759,7 +2759,7 @@ $(function () {
             lastgyms = false
             wrapperGyms.hide(options)
             if (!switchRaids.prop('checked')) {
-               wrapperSidebar.hide(options)
+                wrapperSidebar.hide(options)
             }
         }
         buildSwitchChangeListener(mapData, ['gyms'], 'showGyms').bind(this)()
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
Would it not make more sense to have the toggle always there? I'm not sure why it was considered necessary to hide it.

captbunzo:
> Would it not make more sense to have the toggle always there? I'm not sure why it was considered necessary to hide it.

Agreed

123FLO321:
@pogo-excalibur @captbunzo No. It should be hidden when no Gyms are shown. Just like all other gym/stop/raud options are. But this option applies to Gyms and Raids

pogo-excalibur:
@123FLO321 True. But at present it's hidden unless `gym-info` is configured in the webserver's config, which seems strange to me.

But there's no need to cover that in this PR. I'll hopefully have time to test this later.

sebastienvercammen:
@pogo-excalibur The toggle to show via `-gi` can be discussed via Discord. Short-term, this PR does what we expect.

PR #2381, opened 2017-11-28
Docs/Config fixes

sirlindqvist:
Too many people use 'password' as db user password and then ask in #help for support :)

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/mysql.md
+++ b/docs/extras/mysql.md
@@ -35,7 +35,7 @@
    GRANT ALL PRIVILEGES ON rocketmapdb . * TO 'rocketmapuser'@'localhost';
    exit
    ```
-   You can change `rocketmapdb` to whatever you want the name of the database to be.
+   You can change `rocketmapdb` to whatever you want the name of the database to be. Also, be sure to change `'password'` to the password you want to use.
 5. If the database creation was successful it will tell you "Query OK, 1 row affected". If it doesn't echo that back at you then you either received an error message, or it just created a blank line. I've detailed how to fix common errors, and the blank line below.
    - **Blank line:**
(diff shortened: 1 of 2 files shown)


FrostTheFox:
I'm going to tag on a small little fix for the config file to this just so we don't have two super small PRs.

FrostTheFox:
Sorry to kinda hijack your PR, just adding a few things I need to update with it :P

docs/extras/mysql.md
```diff
@@ -44,6 +44,8 @@
    - **Error: "ERROR 1007 (HY000): Can't create database 'rocketmapdb'; database exists"**
      The rocketmapdb database already exists.
      If you're trying to start a fresh database you'll need to execute `DROP DATABASE rocketmapdb`, and then run `CREATE DATABASE rocketgomapdb`. If you want to keep the rocketmapdb but start a new one, change the name.
+   - **(1045, u"Access denied for user 'rocketmapuser'@'localhost' (using password: YES)")**
+     You are using **password** as your password for the database user **rocketmapuser**. Simply run `CREATE USER 'rocketmapuser'@'localhost' IDENTIFIED BY 'password';` and replace `'password'` to the password you want to use.
```
sebastienvercammen: Not

> are

but

> might be

And what's the effect of running `CREATE USER` again on a user that already exists? `SET PASSWORD` is built to change passwords for existing users.

FrostTheFox:
@sebastienvercammen According to https://dev.mysql.com/doc/refman/5.7/en/set-password.html, `ALTER USER` is the preferred method for setting account passwords, but regardless, you're right it is redundant to run a CREATE USER when we've already created it. Changing now

PR #2382, opened 2017-12-01
0.83.2

FrostTheFox:
## Description

Update API to 0.83.2. Fix fatal logic for out of date pgoapi version, check if pgoapi version < RM expected version instead of !=m, and check if it's 8300, if it isn't, then it's lower and not current and we should stop.
## Motivation and Context

0.83.2 is forced. Version integer is NOT 8302, it's 8300, meaning that the logic to decode the 0.83.2 format to a single integer would not work, since it will do `if 8300 != 8302` or `if 8300 > 8302` which of course both fail. Yes, this is more manual, but for now, it's a solution and probably the smallest one, albeit the most manual. I'm willing to take the time to change it myself if it means it works how it's intended, but hopefully the next update realigns the integer with the app version.
## How Has This Been Tested?

Not tested. ***PLEASE TEST***
## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -123,7 +123,7 @@ def check_login(args, account, api, proxy_url):
 # Simulate real app via login sequence.
 def rpc_login_sequence(args, api, account):
     total_req = 0
-    app_version = int(args.api_version.replace('.', '0'))
+    app_version = PGoApi.get_api_version()
 
     # 1 - Make an empty request to mimick real app behavior.
     log.debug('Starting RPC login sequence...')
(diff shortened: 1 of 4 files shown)


sebastienvercammen:
- Added a version int map, so we can map our `--api-version` to a different pgoapi version int.
- We now use pgoapi's internal version int for requests instead of `--api-version`.

The reasoning behind both is that we want pgoapi to take care of everything related to the API and its requests, and RM should be the layer on top. `--api-version` will still affect the watchdog, but requests will use pgoapi's version int.

This breaks the ability to force a specific version int to be used, in case the API isn't updated, but we've been talking with Chrales to maintain pgoapi again and if it's ever necessary we can push a PR ourselves to update the API where necessary.

A PR is a better update process than keeping a config argument for the sake of being able to force something.

PR #2383, opened 2017-12-03
Database Cleanup (#2281 rework)

neskk:
Keeping the database size under control is an issue because some tables never get cleanup.
## Description

Pulled the changes from #2281, rebased it and added a SpawnPoint cleanup routine.
Since SpawnPoint are a bit tricky and are connected to other three tables (ScanSpawnPoint, SpawnPointDetectionData and ScannedLocation) I tried to cascade the cleanup so that all relevant information is kept and only incorrect data is deleted.

**Update**
Removed spawnpoint integrity checks, because they were slow and computationally expensive.
Simplified the spawnpoint cleanup routine so that it removes old SpawnPoint that haven't got a valid scan in a while. SpawnpointDetectionData is also purged of old entries. All related tables are updated so that the data maintains integrity.

**-DC, --db-cleanup**: previously `-DC, --enable-clean`, changed to have consistency between the parameters involved in the database cleanup routine.
- **-DCw, --db-cleanup-worker**: controls how long will inactive worker status remain in the database.
  - Default value is 30 minutes, can be disabled with 0.
- **-DCp, --db-cleanup-pokemon**: previously `-pd, --purge-data`, has the same meaning as before - controls the age of Pokemon spawns that remain in the database.
  - Default value is 0 *disabled*. Recommended 720 hours (30 days).
- **-DCg, --db-cleanup-gym**: lets you control the maximum age of Gym Details, Gym Members and Gym Pokemon that stay on the database.
  - Default is 8 hours. It's up to you to define how long data becomes outdated. Can be disabled.
- **-DCs, --db-cleanup-spawnpoint**: controls the maximum age of Pokemon Spawnpoints **without valid scans** and also controls how long SpawnpointDetectionData can remain in the database.
  - Default value is 720 hours (30 days).
- **-DCf, --db-cleanup-fort**: controls how long can Gyms and Pokestops remain in database without being scanned.
  - Default value is 0 *disabled*.

## Motivation and Context

Big databases clog up servers and can have a detrimental impact on the performance of the scanners.

## How Has This Been Tested?

MySQL tests went fine... Didn't even need to disable foreign key checks.

```
[db-cleaner][models][   DEBUG] Regular database cleanup started.
[db-cleaner][models][   DEBUG] Completed regular cleanup in 0.004396 seconds.
[db-cleaner][models][   DEBUG] Beginning cleanup of old worker status.
[db-cleaner][models][   DEBUG] Completed cleanup of old worker status in 0.001754 seconds.
[db-cleaner][models][   DEBUG] Beginning cleanup of old pokemon spawns.
[db-cleaner][models][   DEBUG] Deleted 0 old Pokemon entries.
[db-cleaner][models][   DEBUG] Completed cleanup of old pokemon spawns in 0.001303 seconds.
[db-cleaner][models][   DEBUG] Beginning cleanup of old gym data.
[db-cleaner][models][   DEBUG] Deleted 0 old GymDetails entries.
[db-cleaner][models][   DEBUG] Deleted 0 old Raid entries.
[db-cleaner][models][   DEBUG] Deleted 0 old GymMember entries.
[db-cleaner][models][   DEBUG] Deleted 0 old GymPokemon entries.
[db-cleaner][models][   DEBUG] Completed cleanup of old gym data in 0.003231 seconds.
[db-cleaner][models][   DEBUG] Beginning cleanup of old spawnpoint data.
[db-cleaner][models][   DEBUG] Found 0 old SpawnPoint entries.
[db-cleaner][models][   DEBUG] Deleted 0 old SpawnpointDetectionData entries.
[db-cleaner][models][   DEBUG] Found 0 ScannedLocation entries from old spawnpoints.
[db-cleaner][models][   DEBUG] Deleted 0 ScanSpawnPoint entries from old spawnpoints.
[db-cleaner][models][   DEBUG] Deleted 0 old SpawnPoint entries.
[db-cleaner][models][   DEBUG] Deleted 0 ScanSpawnPoint entries from old scan locations.
[db-cleaner][models][   DEBUG] Deleted 0 ScannedLocation entries from old spawnpoints.
[db-cleaner][models][   DEBUG] Completed cleanup of old spawnpoint data in 0.008164 seconds.
[db-cleaner][models][   DEBUG] Beginning cleanup of old forts.
[db-cleaner][models][   DEBUG] Deleted 0 old Gym entries.
[db-cleaner][models][   DEBUG] Deleted 0 old Pokestop entries.
[db-cleaner][models][   DEBUG] Completed cleanup of old forts in 0.002356 seconds.
[db-cleaner][models][    INFO] Full database cleanup completed.
````

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2652,36 +2652,31 @@ def db_updater(q, db):
 
 
 def clean_db_loop(args):
-    if args.db_type == 'mysql':
-        step = 250
-    else:
-        step = 50
+    step = 250
     cycle = 0
     while True:
         try:
+            cycle += 1
             db_cleanup_regular()
+            log.info('Regular database cleanup complete.')
 
-            if cycle % 10 != 0:
-                cycle += 1
-                log.info('Regular database cleaning complete.')
-                time.sleep(60)
-                continue
-            else:
-                cycle = 1
+            # Run once every 10 minutes.
+            if cycle % 10 == 0:
+                # If desired, clear old Pokemon spawns.
+                if args.db_cleanup_pokemon > 0:
+                    db_clean_pokemons(args.db_cleanup_pokemon)
 
-            # If desired, clear old Pokemon spawns.
-            if args.db_cleanup_pokemon > 0:
-                db_clean_pokemons(args.db_cleanup_pokemon)
+                # Remove old gym data.
+                if args.db_cleanup_gym > 0:
+                    db_clean_gyms(args.db_cleanup_gym)
 
-            # Remove old Gym details and locations.
-            if args.db_cleanup_gym > 0:
-                db_clean_gyms(args.db_cleanup_gym)
+                # Remove old and extinct spawnpoint data.
+                if args.db_cleanup_spawnpoint > 0:
+                    db_clean_spawnpoints(step, args.db_cleanup_spawnpoint)
 
-            # Remove old and extinct Spawnpoints.
-            if args.db_cleanup_spawnpoint > 0:
-                db_clean_spawnpoints(step, args.db_cleanup_spawnpoint)
+                log.info('Full database cleaning complete.')
+                cycle = 0
 
-            log.info('Full database cleaning complete.')
             time.sleep(60)
         except Exception as e:
             log.exception('Exception in clean_db_loop: %s', repr(e))
@@ -2707,7 +2702,7 @@ def db_cleanup_regular():
                 Pokestop.lure_expiration < datetime.utcnow()))
         query.execute()
 
-        # Remove old (unusable) captcha tokens
+        # Remove old (unusable) captcha tokens.
         query = (Token
                  .delete()
                  .where((Token.last_updated <
@@ -2727,70 +2722,70 @@ def db_cleanup_regular():
 def db_clean_pokemons(age_hours):
     with Pokemon.database().execution_context():
         log.info('Beginning cleanup of old Pokemon spawns.')
-        start = datetime.utcnow()
+        start_timer = default_timer()
         query = (Pokemon
                  .delete()
                  .where((Pokemon.disappear_time <
                          (datetime.utcnow() - timedelta(hours=age_hours)))))
         rows = query.execute()
-        diff = datetime.utcnow() - start
+        time_diff = default_timer() - start_timer
         log.info('Completed cleanup of old Pokemon spawns. ' +
-                 '%i deleted in %f seconds.', rows, diff.total_seconds())
+                 '%d deleted in %.6f seconds.', rows, time_diff)
 
 
 def db_clean_gyms(age_hours, gyms_age_days=30):
     with Gym.database().execution_context():
-        log.info('Beginning cleanup of old Gym data.')
-        start = datetime.utcnow()
-        # Remove old gym Details.
+        log.info('Beginning cleanup of old gym data.')
+        start_timer = default_timer()
+        # Remove old GymDetails.
         query = (GymDetails
                  .delete()
                  .where(GymDetails.last_scanned <
                         (datetime.utcnow() - timedelta(hours=age_hours))))
         rows = query.execute()
-        log.info('Deleted %d old Gym Details.', rows)
+        log.info('Deleted %d old GymDetails.', rows)
 
-        # Remove old raid Details.
+        # Remove old Raid.
         query = (Raid
                  .delete()
                  .where(Raid.end <
                         (datetime.utcnow() - timedelta(hours=age_hours))))
         rows = query.execute()
-        log.info('Deleted %d old Raid Details.', rows)
+        log.info('Deleted %d old Raid.', rows)
 
-        # Remove old gym members.
+        # Remove old GymMembers.
         query = (GymMember
                  .delete()
                  .where(GymMember.last_scanned <
                         (datetime.utcnow() - timedelta(hours=age_hours))))
         rows = query.execute()
-        log.info('Deleted %d old Gym Members.', rows)
+        log.info('Deleted %d old GymMember.', rows)
 
-        # Remove old gym Pokemon.
+        # Remove old GymPokemon.
         query = (GymPokemon
                  .delete()
                  .where(GymPokemon.last_seen <
                         (datetime.utcnow() - timedelta(hours=age_hours))))
         rows = query.execute()
-        log.info('Deleted %d old Gym Pokemons.', rows)
+        log.info('Deleted %d old GymPokemon.', rows)
 
-        # Remove old gym locations.
+        # Remove old Gym.
         query = (Gym
                  .delete()
                  .where(Gym.last_scanned <
                         (datetime.utcnow() - timedelta(days=gyms_age_days))))
         rows = query.execute()
-        log.info('Deleted %d old Gym locations.', rows)
+        log.info('Deleted %d old Gym.', rows)
 
-        diff = datetime.utcnow() - start
-        log.info('Completed cleanup of old Gym data in %f seconds.',
-                 diff.total_seconds())
+        time_diff = default_timer() - start_timer
+        log.info('Completed cleanup of old gym data in %.6f seconds.',
+                 time_diff)
 
 
-def db_clean_spawnpoints(step, age_hours, missed=3):
+def db_clean_spawnpoints(step, age_hours, missed=5):
     with SpawnPoint.database().execution_context():
-        log.info('Beginning cleanup of old Spawnpoint data.')
-        start = datetime.utcnow()
+        log.info('Beginning cleanup of old spawnpoint data.')
+        start_timer = default_timer()
 
         # Select old SpawnPoint.
         query = (SpawnPoint
@@ -2885,9 +2880,9 @@ def db_clean_spawnpoints(step, age_hours, missed=3):
         log.info('Deleted %d ScannedLocation from old SpawnPoint.',
                  num_rows)
 
-        diff = datetime.utcnow() - start
-        log.info('Completed cleanup of old Spawnpoint data in %f seconds.',
-                 diff.total_seconds())
+        time_diff = default_timer() - start_timer
+        log.info('Completed cleanup of old spawnpoint data in %.6f seconds.',
+                 time_diff)
 
 
 def bulk_upsert(cls, data, db):
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2719,7 +2719,7 @@ def clean_db_loop(args):
                 query = (HashKeys
                          .delete()
                          .where(HashKeys.expires <
-                                (datetime.now() - timedelta(days=1))))
+                                (datetime.utcnow() - timedelta(days=1))))
                 query.execute()
 
                 # If desired, clear old Pokemon spawns.
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1092,19 +1092,38 @@ class MainWorker(BaseModel):
     elapsed = IntegerField(default=0)
 
     @staticmethod
-    def get_account_stats():
+    def get_account_stats(age_minutes=30):
+        stats = {'working': 0, 'captcha': 0, 'failed': 0}
+        timeout = datetime.utcnow() - timedelta(minutes=age_minutes)
         with MainWorker.database().execution_context():
             account_stats = (MainWorker
                              .select(fn.SUM(MainWorker.accounts_working),
                                      fn.SUM(MainWorker.accounts_captcha),
                                      fn.SUM(MainWorker.accounts_failed))
+                             .where(MainWorker.last_modified >= timeout)
                              .scalar(as_tuple=True))
-        dict = {'working': 0, 'captcha': 0, 'failed': 0}
-        if account_stats[0] is not None:
-            dict = {'working': int(account_stats[0]),
+            if account_stats[0] is not None:
+                stats.update({
+                    'working': int(account_stats[0]),
                     'captcha': int(account_stats[1]),
-                    'failed': int(account_stats[2])}
-        return dict
+                    'failed': int(account_stats[2])
+                })
+        return stats
+
+    @staticmethod
+    def get_recent(age_minutes=30):
+        status = []
+        timeout = datetime.utcnow() - timedelta(minutes=age_minutes)
+        with MainWorker.database().execution_context():
+            query = (MainWorker
+                     .select()
+                     .where(MainWorker.last_modified >= timeout)
+                     .order_by(MainWorker.worker_name.asc())
+                     .dicts())
+
+            status = [dbmw for dbmw in query]
+
+        return status
 
 
 class WorkerStatus(LatLongModel):
@@ -1139,18 +1158,18 @@ class WorkerStatus(LatLongModel):
                 'longitude': status.get('longitude', None)}
 
     @staticmethod
-    def get_recent():
+    def get_recent(age_minutes=30):
         status = []
+        timeout = datetime.utcnow() - timedelta(minutes=age_minutes)
         with WorkerStatus.database().execution_context():
             query = (WorkerStatus
                      .select()
-                     .where((WorkerStatus.last_modified >=
-                             (datetime.utcnow() - timedelta(minutes=5))))
-                     .order_by(WorkerStatus.username)
+                     .where(WorkerStatus.last_modified >= timeout)
+                     .order_by(WorkerStatus.username.asc())
                      .dicts())
 
-            for s in query:
-                status.append(s)
+            status = [dbws for dbws in query]
+
         return status
 
     @staticmethod
@@ -2681,7 +2700,7 @@ def clean_db_loop(args):
             cycle += 1
             time.sleep(60)
         except Exception as e:
-            log.exception('Database cleanup failed: %s', e)
+            log.exception('Database cleanup failed: %s.', e)
 
 
 def db_cleanup_regular():
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2677,14 +2677,35 @@ def clean_db_loop(args):
                 query = (GymDetails
                          .delete()
                          .where(GymDetails.last_scanned <
-                                (datetime.now() - timedelta(days=1))))
+                                (datetime.utcnow() - timedelta(days=7))))
                 query.execute()
 
                 # Remove old gym locations.
                 query = (Gym
                          .delete()
                          .where(Gym.last_scanned <
-                                (datetime.now() - timedelta(days=1))))
+                                (datetime.utcnow() - timedelta(days=7))))
+                query.execute()
+
+                # Remove old raid Details.
+                query = (Raid
+                         .delete()
+                         .where(Raid.end <
+                                (datetime.utcnow() - timedelta(days=7))))
+                query.execute()
+
+                # Remove old gym members.
+                query = (GymMember
+                         .delete()
+                         .where(GymMember.last_scanned <
+                                (datetime.utcnow() - timedelta(days=7))))
+                query.execute()
+
+                # Remove old gym Pokemon.
+                query = (GymPokemon
+                         .delete()
+                         .where(GymPokemon.last_seen <
+                                (datetime.utcnow() - timedelta(days=7))))
                 query.execute()
 
                 # Remove old (unusable) captcha tokens
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2673,6 +2673,20 @@ def clean_db_loop(args):
                         Pokestop.lure_expiration < datetime.utcnow()))
                 query.execute()
 
+                # Remove old gym Details.
+                query = (GymDetails
+                         .delete()
+                         .where(GymDetails.last_scanned <
+                                (datetime.now() - timedelta(days=1))))
+                query.execute()
+
+                # Remove old gym locations.
+                query = (Gym
+                         .delete()
+                         .where(Gym.last_scanned <
+                                (datetime.now() - timedelta(days=1))))
+                query.execute()
+
                 # Remove old (unusable) captcha tokens
                 query = (Token
                          .delete()
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1784,12 +1784,12 @@ class Token(BaseModel):
                     token_ids.append(t.id)
                     tokens.append(t.token)
                 if tokens:
-                    log.debug('Retrived Token IDs: {}'.format(token_ids))
-                    result = DeleteQuery(Token).where(
-                        Token.id << token_ids).execute()
-                    log.debug('Deleted {} tokens.'.format(result))
+                    log.debug('Retrieved Token IDs: %s', token_ids)
+                    query = DeleteQuery(Token).where(Token.id << token_ids)
+                    rows = query.execute()
+                    log.debug('Claimed and removed %d captcha tokens.', rows)
         except OperationalError as e:
-            log.error('Failed captcha token transactional query: {}'.format(e))
+            log.exception('Failed captcha token transactional query: %s.', e)
 
         return tokens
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pogom/models.py [hunk 1]
```diff
@@ -2639,63 +2639,244 @@ def db_updater(q, db):
...
+                         (datetime.utcnow() - timedelta(days=1))) |
+                        (HashKeys.last_updated <
+                         (datetime.utcnow() - timedelta(days=7)))))
+        query.execute()
+
+
+def db_clean_pokemons(age_hours):
+    with Pokemon.database().execution_context():
+        log.info('Beginning cleanup of old Pokemon spawns.')
+        start = datetime.utcnow()
```
sebastienvercammen: A better way to determine time passed in Python is [timeit.default_timer](https://docs.python.org/2/library/timeit.html#timeit.default_timer).

pogom/models.py
```diff
@@ -2639,63 +2639,244 @@ def db_updater(q, db):
...
+    with Gym.database().execution_context():
+        log.info('Beginning cleanup of old Gym data.')
+        start = datetime.utcnow()
+        # Remove old gym Details.
+        query = (GymDetails
+                 .delete()
+                 .where(GymDetails.last_scanned <
+                        (datetime.utcnow() - timedelta(hours=age_hours))))
+        rows = query.execute()
+        log.info('Deleted %d old Gym Details.', rows)
```
sebastienvercammen: We want to use either the correct English spelling

> Gym details

or the model's name

> GymDetails

pogom/models.py
```diff
@@ -2639,63 +2639,244 @@ def db_updater(q, db):
...
+        query = (SpawnPoint
+                 .select(SpawnPoint.id)
+                 .where((SpawnPoint.last_scanned <
+                         (datetime.utcnow() - timedelta(hours=age_hours))) &
+                        (SpawnPoint.missed_count > missed))
+                 .dicts())
+        old_sp = [(sp['id']) for sp in query]
+
+        num_records = len(old_sp)
+        log.debug('Found %d old SpawnPoint.', num_records)
```
sebastienvercammen: To stay consistent with your other logging up above that's more specific about the format (´%i deleted´):

> Found %i old SpawnPoint entries.
  
(or "old spawnpoints" or similar)

pogom/utils.py
```diff
@@ -391,8 +387,21 @@ def get_args():
                         help=('Get all details about gyms (causes an ' +
                               'additional API hit for every gym).'),
                         action='store_true', default=False)
-    parser.add_argument('-DC', '--enable-clean', help='Enable DB cleaner.',
+    parser.add_argument('-dbc', '--db-cleanup',
```
sebastienvercammen: The new format is to use a single letter as prefix. An uppercase letter for emphasis/importance, a lowercase letter for the actual setting. e.g. `-Dc` (yes, rather than `-DC`) for database cleaner.

In this case a `-D` prefix would be more appropriate, since "database" isn't emphasized as "DataBase" and it's not a combination of two words.

We've already decided against making all options just a combination of random letters picked from the full name because it gets inconsistent and confusing very fast. Using a consistent prefix like `-D` for database items is easy to remember and easy to use.

All old settings still retain their old config names/references until we've finished other, more important work.

pogom/utils.py
```diff
@@ -388,6 +384,22 @@ def get_args():
               'queue falls behind.'),
         type=int,
         default=1)
+    group = parser.add_argument_group('Database Cleanup')
```
Alderon86: Why do we need an extra arg for every cleaning option? i dont like to spam my config with commands, that *could* be merged into one.

neskk: All settings have reasonable default values so they don't need to be defined.
I don't understand how you're supposed to fit all arguments into one, that's not possible. 😄 
The only argument that is disabled by default is pokemon cleanup and I did this because `purge_data` was also set by default to 0. Since `--db-cleanup-pokemon` is a direct replacement for it, I left this behavior unchanged.
So the number of "required" arguments didn't change: 1. for general cleanup and 2. params to enable pokemon cleanup. Same as before...

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
-
-                # Remove expired HashKeys
-                query = (HashKeys
-                         .delete()
-                         .where(HashKeys.expires <
-                                (datetime.now() - timedelta(days=1))))
-                query.execute()
+            cycle += 1
+            db_cleanup_regular()
+            log.info('Regular database cleanup complete.')
```
sebastienvercammen: We have a log entry for regular cleanup but not for the specific ones on lines 2615 - 2624. I would add a log entry for each one, but move all these logs to `DEBUG` instead of `INFO`, and keep the

> Full database cleaning complete.

on `INFO` so it's always shown that it has completed, but the host has the option to enable `-v` for details.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
+                    db_clean_gyms(args.db_cleanup_gym)
+
+                # Remove old and extinct spawnpoint data.
+                if args.db_cleanup_spawnpoint > 0:
+                    db_clean_spawnpoints(step, args.db_cleanup_spawnpoint)
+
+                log.info('Full database cleaning complete.')
+                cycle = 0
+
             time.sleep(60)
```
sebastienvercammen: Instead of a fixed 60s sleep and a 10-minute timer with the cycles (counting by minutes passed), it would be clearer to have configurable options for the timers (in minutes) and use those to sleep the appropriate amount of time.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
         except Exception as e:
             log.exception('Exception in clean_db_loop: %s', repr(e))
```
sebastienvercammen: This is left over from an older time when people didn't know how to properly use the `logging` library so we forced them to always use `repr()` because they were confusing string concatenation with logging (= exceptions galore), but could you remove the `repr() in this PR? :)

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
 
 
+def db_cleanup_regular():
+    with MainWorker.database().execution_context():
+        query = (MainWorker
+                 .delete()
+                 .where((MainWorker.last_modified <
+                         (datetime.utcnow() - timedelta(minutes=30)))))
```
sebastienvercammen: Same as above, move the hard-coded `30` to the top and give it a descriptive variable name. To fix the issue of line length in PEP8, you could use a separate variable for

```
(datetime.utcnow() - timedelta(minutes=30))
```

which would also solve the issue of duplicate code, since you use the same 30-minute time difference just below for WorkerStatus.

These things are easier to catch/avoid if we don't use hardcoded values and use variables at the top. We'd instantly see that both are the same.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
+                         (datetime.utcnow() - timedelta(minutes=2)))))
+        query.execute()
+
+        # Remove expired HashKeys.
+        query = (HashKeys
+                 .delete()
+                 .where((HashKeys.expires <
+                         (datetime.utcnow() - timedelta(days=1))) |
+                        (HashKeys.last_updated <
+                         (datetime.utcnow() - timedelta(days=7)))))
```
sebastienvercammen: Store `datetime.utcnow()` in a variable at the top before you start, there are too many function calls here that can be avoided.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
+                log.info('Full database cleaning complete.')
+                cycle = 0
+
             time.sleep(60)
         except Exception as e:
             log.exception('Exception in clean_db_loop: %s', repr(e))
 
 
+def db_cleanup_regular():
+    with MainWorker.database().execution_context():
```
sebastienvercammen: Add a comment above this line to clarify that the execution context will automatically commit on success or rollback on error. Not everyone knows about this.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
+        query = (SpawnPoint
+                 .select(SpawnPoint.id)
+                 .where((SpawnPoint.last_scanned <
+                         (datetime.utcnow() - timedelta(hours=age_hours))) &
+                        (SpawnPoint.missed_count > missed))
+                 .dicts())
+        old_sp = [(sp['id']) for sp in query]
+
+        num_records = len(old_sp)
+        log.debug('Found %d old SpawnPoint.', num_records)
```
sebastienvercammen: 👍 Use of `SpawnPoint` makes it clear what data was altered.

If you want to make it "better English", you could reword it to

> Found %d old SpawnPoint entries.

and do the same for all other logs. It'll keep the clarity and it'll be better English.

pogom/models.py
```diff
@@ -2609,63 +2601,239 @@ def db_updater(q, db):
...
+        num_rows = 0
+        for i in range(0, num_records, step):
+            query = (SpawnPoint
+                     .delete()
+                     .where((SpawnPoint.id <<
+                             old_sp[i:min(i + step, num_records)])))
+            num_rows += query.execute()
+        log.info('Deleted %d old SpawnPoint.', num_rows)
+
+        sl_delete = list(sl_delete)
```
sebastienvercammen: Why do you make `sl_delete` a `set` to then convert it to a `list`? You could use `list.append` instead of `set.add`.

pogom/utils.py
sebastienvercammen: 👍 I'm OK with the extra arguments. If we want configurability, we need configuration items. Doesn't mean all of them should be in everyone's configs if we have sensible defaults.

pogom/utils.py
```diff
@@ -389,6 +385,22 @@ def get_args():
...
+                       action='store_true', default=False)
+    group.add_argument('-dbcp', '--db-cleanup-pokemon',
+                       help=('Clear Pokemon from database X hours ' +
+                             'after they disappear (0 to disable).'),
+                       type=int, default=0)
+    group.add_argument('-dbcg', '--db-cleanup-gym',
+                       help=('Clear Gym details from database X ' +
+                             'hours after being scanned (0 to disable).'),
+                       type=int, default=3)
+    group.add_argument('-dbcs', '--db-cleanup-spawnpoint',
```
sebastienvercammen: Your arguments use the old format, where we arbitrarily selected a few letters and made all config items lowercase. Somewhere between a few weeks and a few months ago we've decided to go with a more context-appropriate choice, using the long, descriptive form for configuration files and short forms with contexts/grouping for the CLI:

* Descriptive long forms. This is 👍 in your case.
* For the short forms, use an uppercase letter for scope/namespace/group. In the case of database items, this is `-D`. This promotes easier usage on the command-line: all config items relating to the database will start with `-D`.
* For the short form, use lowercase letters to denote the configuration item we're referring to.

Examples:

```
-Du: Database user
-Dp: Database password
-DC: Enable DB cleanup
-DCg: DB cleanup gym settings
```

pogom/models.py
neskk: I've thought about this but here's why I didn't do dynamic/assignable timers to control the loop:
Some cleanup operations should be ran as soon as possible, for example:
 - Captcha tokens cleanup: it clears tokens older than 120 seconds. Running this query once a minute is the bare minimum to enforce max age of tokens. In worse case scenario it will leave tokens up to 179 secs in the database (120 s + 60 s loop sleep). For this reason I feel these parameters shouldn't be under control of the user if they mess with some hard coded values, i. e. captcha token validity window doesn't really change so there's no point allowing user to customize their max age.
- I've created the "full database cleanup" thing because there are queries that are not so time sensitive so they don't need to run every minute. Stuff like old Pokemon/Gym/Spawnpoint cleanup can be done less frequently. The queries in this section of the cleaner deal with entries with ages between a couple of hours to 30+ days. I've also decided to leave this value hardcoded but it's assigned to a self explanatory variable `full_cleanup_period_minutes = 10` so that if in the future someone is looking at this code it can easily make sense of whats going on. Again the logic is that these settings are important values and there isn't a good reason to let users change them.

neskk:
@sebastienvercammen sorry for forced update over your commit, I forgot to pull 😢
Added that last dot in log.exception.
I have a bunch of scripts I use to keep git branches rebased and pushed and today it messed up this because I hadn't pull from origin and remote tracking branch isn't set to origin.

pogom/models.py
neskk: Didn't read this suggestion for some reason, sorry :|
Added a comment with a link for peewee site and "resumed" description of what execution context does.

pogom/models.py
```diff
@@ -2573,61 +2587,285 @@ def db_updater(q, db):
 
 
 def clean_db_loop(args):
+    step = 250
+    cycle = 0
+
+    # Every 10 minutes (10 cycles) run full database cleanup.
+    full_cleanup_period_minutes = 10
```
sebastienvercammen: Dubious: it's only 10 minutes because your cycle time is 60 seconds. If anyone would change the 60-second sleep without noticing this variable/comment, it would change the frequency of the cleanup.

To solve, don't rely on your sleep/cycle time. Use `default_timer()` to time it separately and track it as an actual time difference.

neskk:
@sebastienvercammen I wish you would devote the same time/attention to other PRs you merge, because it looks like you're extra picky with me. To prove my point: https://github.com/RocketMap/RocketMap/commit/6d879afd9c399dff6b32766434e627fc40238118#diff-6c84a64c790b52a5ff2656d6bb388fefR862
Extra thread, could at least include or refactor proxy refresher code but no; Can't even figure out why there's that `max()` in there; The method is called `dynamic_loading_refresher` but then the thread is called `dynamic-enclist`...

sebastienvercammen:
@neskk Because #2445 will already do all that you're arguing about...

pogom/models.py
```diff
@@ -2573,61 +2587,286 @@ def db_updater(q, db):
 
 
 def clean_db_loop(args):
+    step = 250
+
+    # Run regular database cleanup once every minute.
+    regular_cleanup_interval = 60
+    # Run full database cleanup once every 10 minutes.
+    full_cleanup_timer = default_timer()
+    full_cleanup_interval = 600
```
sebastienvercammen: Rename both intervals, add `_sec` at the end to clarify that they're both in seconds.

Similar to the code you had just before (minute), and consistency with the rest of the code.

pogom/models.py
```diff
@@ -2576,61 +2590,286 @@ def db_updater(q, db):
 
 
 def clean_db_loop(args):
+    step = 250
```
friscoMad: As I see the steps are for IN clauses but MySQL has no limit on this, so maybe we can remove it.

pogom/utils.py
```diff
@@ -461,6 +479,9 @@ def get_args():
                               'STATUS_NAME as main worker name.'))
     parser.add_argument('-spp', '--status-page-password', default=None,
                         help='Set the status page password.')
+    parser.add_argument('-spf', '--status-page-filter',
+                        help='Filter worker status inactive for X minutes.',
+                        type=int, default=30)
```
sebastienvercammen: * Add an opt group for the status page.
* `Sf, status-page-filter` (alternative: `SPf`).
* `Sp, status-page-password` (alternative: `SPp`).
* Update description, add: 
   > 0 to show all.
   > or: 0 to disable.

pogom/models.py
```diff
@@ -1116,18 +1135,18 @@ def db_format(status, name='status_worker_db'):
                 'longitude': status.get('longitude', None)}
 
     @staticmethod
-    def get_recent():
+    def get_recent(age_minutes=30):
         status = []
```
sebastienvercammen: You can remove this, since:

```
status = [dbws for dbws in query]
```

and

```
>>> a = []
>>> [x for x in a]
[]
```

pogom/models.py
```diff
@@ -1069,19 +1069,38 @@ class MainWorker(BaseModel):
...
                     'captcha': int(account_stats[1]),
-                    'failed': int(account_stats[2])}
-        return dict
+                    'failed': int(account_stats[2])
+                })
+        return stats
+
+    @staticmethod
+    def get_recent(age_minutes=30):
+        status = []
```
sebastienvercammen: You can remove this, since:

```
status = [dbmw for dbmw in query]
```

and

```
>>> a = []
>>> [x for x in a]
[]
```

pogom/models.py
```diff
@@ -1739,12 +1758,12 @@ def get_valid(limit=15):
                     token_ids.append(t.id)
                     tokens.append(t.token)
                 if tokens:
-                    log.debug('Retrived Token IDs: {}'.format(token_ids))
-                    result = DeleteQuery(Token).where(
-                        Token.id << token_ids).execute()
-                    log.debug('Deleted {} tokens.'.format(result))
+                    log.debug('Retrieved Token IDs: %s', token_ids)
```
sebastienvercammen: Punctuation in log:

```
log.debug('Retrieved Token IDs: %s.', token_ids)
```

pogom/models.py
```diff
@@ -1739,12 +1758,12 @@ def get_valid(limit=15):
...
+                    query = DeleteQuery(Token).where(Token.id << token_ids)
+                    rows = query.execute()
+                    log.debug('Claimed and removed %d captcha tokens.', rows)
```
sebastienvercammen: Not good: a `Token.get_valid()` method should only be responsible for retrieving entries from the database. What its caller wants to do with it afterwards is still unknown, so the responsibility for claiming/deleting is the caller's, not for the method that just retrieves data.

pogom/models.py
sebastienvercammen: ☝️ Good point. And if there is any advantage to limiting the step, we should still consider something higher than just 250.

pogom/models.py
```diff
@@ -2610,61 +2624,286 @@ def db_updater(q, db):
...
+                # Remove old and extinct spawnpoint data.
+                if args.db_cleanup_spawnpoint > 0:
+                    db_clean_spawnpoints(step, args.db_cleanup_spawnpoint)
+
+                # Remove old pokestop and gym locations.
+                if args.db_cleanup_forts > 0:
+                    db_clean_forts(args.db_cleanup_forts)
+
+                log.info('Full database cleanup completed.')
+                full_cleanup_timer = default_timer()
```
sebastienvercammen: Duplicate code, you call `default_timer()` twice:

```
if (default_timer() - full_cleanup_timer > full_cleanup_secs):
   ...
   full_cleanup_timer = default_timer()
```

replace with:

```
now = default_timer()
if (now - full_cleanup_timer > full_cleanup_secs):
   ...
   full_cleanup_timer = now
```

but only if the cleanup's duration is insignificant (i.e. up to a handful of seconds for a 60-second or 10-minute refresh).

pogom/utils.py
```diff
@@ -390,6 +386,30 @@ def get_args():
               'queue falls behind.'),
         type=int,
         default=1)
+    group = parser.add_argument_group('Database Cleanup')
+    group.add_argument('-DC', '--db-cleanup',
+                       help='Enable regular database cleanup thread.',
+                       action='store_true', default=False)
+    group.add_argument('-DCw', '--db-cleanup-worker',
+                       help=('Clear worker status from database after X ' +
+                             'minutes of inactivity (0 to disable).'),
```
sebastienvercammen: Change the notation of

> (0 to disable)

to the new format we're gradually implementing for consistency

> 0 to disable. Default: x.

pogom/models.py
neskk: There are queries that use `where(ids << list_with_many_ids)`.
MySQL still has limits to how many variables/query length you can use. I can set it higher, but it think it's a good practice to have a maximum.

pogom/models.py
neskk: I don't understand your point, query isn't executed until you actually try to iterate it. That's why this was done...

pogom/models.py
neskk: This was code that I did previously. Time between query and delete is very important and it must be done within a transaction. We want to get the token and immediately delete it so that other instances don't get hold of the same token. Time between to function calls could actually lead to double spending problem. Usually things have a reason behind being done in a particular way, saying "not good" is easy and I understand your point, but IMHO it doesn't make sense in this case. The only use case for this method requires a delete query afterwards, no point in separating concerns.

pogom/models.py
neskk: It's not actually duplicate code because it's the *most* correct way to do this, but I guess a small optimization can be made.

neskk:
@sebastienvercammen I appreciate your feedback and I'm sorry for sounding impolite. Sometimes I see code merged that seems lacking some attention to details. I guess this is not often at all thanks to your standards and reviewing work which is good. The PR that I complained about seemed a bit "rushed", but since you said there's going to be a rework soon I guess it's alright and a nice addition overall.

pogom/models.py
neskk:
```
...
  File "D:\Workspace\Python\PokemonGo-Map\pogom\app.py", line 570, in post_status
    return jsonify(d)
  File "C:\Program Files (x86)\Python27\lib\site-packages\flask\json.py", line 266, in jsonify
    (dumps(data, indent=indent, separators=separators), '\n'),
  File "C:\Program Files (x86)\Python27\lib\site-packages\flask\json.py", line 126, in dumps
    rv = _json.dumps(obj, **kwargs)
  File "C:\Program Files (x86)\Python27\lib\json\__init__.py", line 251, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "C:\Program Files (x86)\Python27\lib\json\encoder.py", line 209, in encode
    chunks = list(chunks)
  File "C:\Program Files (x86)\Python27\lib\json\encoder.py", line 434, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "C:\Program Files (x86)\Python27\lib\json\encoder.py", line 408, in _iterencode_dict
    for chunk in chunks:
  File "C:\Program Files (x86)\Python27\lib\json\encoder.py", line 442, in _iterencode
    o = _default(o)
  File "D:\Workspace\Python\PokemonGo-Map\pogom\app.py", line 589, in default
    return list(iterable)
TypeError: 'ResultIterator' object is not iterable
```

sebastienvercammen: @neskk Sorry, I meant that you could remove the `status = []`, initializing `status` to an empty list, since `status` is being overwritten afterwards anyway.

pogom/models.py
sebastienvercammen: @neskk 👍 from me if you set it to [500](https://github.com/RocketMap/RocketMap/blob/develop/pogom/models.py#L2689).

pogom/models.py
neskk: I decided to leave this in place and added a `try .. except`.
The whole reason I added `status = []` was to make sure that it's not undefined if an error occurs. This makes more sense with the `try .. except` block. I actually don't know exactly how the front-end/flask handles exceptions so I guess is better to be safe. Most of the times only OperationalError should occur, perhaps I should change `except Exception` to `except OperationalError`...

sebastienvercammen: @neskk You won't need the `try: except:` and you don't need the variable definition/initialization of `status = []` at the top.

The `ResultIterator` you get from the query will either contain results or it'll return no results and be an empty iterable.

I used this code to demonstrate what happens when you use list comprehension in Python on an empty iterable:

```
>>> a = []
>>> [x for x in a]
[]
```

Rather than an exception, it simply returns an empty list.

neskk: Are you absolutely sure we don't need the `try .. except`? Pretty much every query to the database can result in an exception.

neskk: @sebastienvercammen I have a commit with this code removed but I think it makes sense to leave it in because database operation can always raise exceptions if connection is unavailable/offline.
I think this PR is ready, I have been running it without problems for the past month.

pogom/utils.py
sebastienvercammen: Waiting for this last tiny change, and for the last commits to be picked up by reviewers. I've pinged them for this PR.

neskk:
Update:
@jaake77 thanks for testing ;)

I pushed a commit that wasn't suppose to be included, it would remove the `try .. except` around `get_recent()` database queries. I read a bit about flask error handling and from what I figured out there aren't any error handlers attached to flask, so when a database exception occurs it may crash the instance. Because of this and since these methods are used directly by flask front end routes I think it's safer to have these `try .. except` blocks.
I rebased and removed the commit, and all I done recently was text changes so this should be good to go.

fosJoddie:
Forgot to mention I also added in some old spawnpoint data that was also cleaned up good as far as I could tell

PR #2385, opened 2017-12-05
Update Proxy Check links/headers

FrostTheFox:
Updated the links and headers used for proxy checking based on self MITMs.
## Description

Niantic check has been changed to plfe/version, because it works the same for checking version, is cleaner because you can fully emulate the app's request, no POSTs. PTC has been changed because the previous link used is no longer used in PoGo's login flow (something I plan to take up the API Maintainers on).

The headers used are from my iPhone X running iOS ~~11.1.2~~ EDIT: 11.2, I just updated to make sure we have the latest details. I know these UAs are different from what is in the APIs used, but I have no idea what iOS version they are from and I know these are modern and from an actual device. I also plan to take them up on changing this in the APIs.
## Motivation and Context

X-Unity-Version needed updating, and I noticed multiple discrepancies with the real requests.
## How Has This Been Tested?

Not tested in code. Links have been checked to make sure they produce appropriate ban statuses, and headers and requests are once again pulled from an actual device running PoGo. PLEASE TEST WITH KNOWN BANNED PROXIES ON BOTH NIANTIC AND PTC. Passes flake8 locally.
n/a
## Types of changes

## Checklist:


pogom/proxy.py
```diff
@@ -103,22 +103,28 @@ def start_request_futures(ptc_session, niantic_session, proxy, timeout):
         proxy_test_ptc_url,
         proxies={'http': proxy, 'https': proxy},
         timeout=timeout,
-        headers={'User-Agent': ('pokemongo/1 '
-                                'CFNetwork/811.4.18 '
-                                'Darwin/16.5.0'),
-                 'Host': 'sso.pokemon.com',
-                 'X-Unity-Version': '5.5.1f1',
-                 'Connection': 'close'},
+        headers={'Host': 'sso.pokemon.com',
```
RamiroWincor: Host is not needed, it comes from the request url and I think keep-alive is sent regardless.

RamiroWincor: In fact we use Connection: close to try to force urllib3 to close the request instantly to avoid the max fd open issue with proxies

pogom/proxy.py
```diff
@@ -103,22 +103,28 @@ def start_request_futures(ptc_session, niantic_session, proxy, timeout):
...
 
     # Send request to nianticlabs.com.
-    future_niantic = niantic_session.post(
+    future_niantic = niantic_session.get(
         proxy_test_url,
-        '',
         proxies={'http': proxy, 'https': proxy},
         timeout=timeout,
-        headers={'Connection': 'close'},
+        headers={'Host': 'pgorelease.nianticlabs.com',
```
RamiroWincor: Just the same thing

pogom/proxy.py
FrostTheFox: It is correct that requests will probably add the host tag itself, but it's for peace of mind and to make sure we know it's sending the appropriate header.

The app sends keep-alive and the goal is to emulate the app, but I suppose for the purpose of keeping proxies from killing themselves it can be dropped.

PR #2386, opened 2017-12-05
Check if check_challenge exists before setting captcha_url/send download_settings with every request

FrostTheFox:
~~According to niico, check_challenge is GONE~~ (EDIT: It's back in the latest version), and download_settings is sent with every request "except the first getplayer".
## Description

I have no idea what further needs to be changed, if anything, but this should send download_settings with the majority of the requests, since a whole lot of everything seems to call apiRequests.py, which eventually loops everything through send_generic_request (as far as I can tell, once again, I could be missing things, please change or let me know if I am.)
## Motivation and Context

This was changed, and it needs to be reflected in our requests.
## How Has This Been Tested?

Not tested, as always, please fully test. Passes flake8 locally.
n/a
## Types of changes

## Checklist:

### DIFF
--- a/pogom/account.py
+++ b/pogom/account.py
@@ -496,10 +496,10 @@ def spin_pokestop(api, account, args, fort, step_location):
         time.sleep(random.uniform(2, 4))  # Don't let Niantic throttle.
 
         # Check for reCaptcha.
-        captcha_url = response['responses']['CHECK_CHALLENGE'].challenge_url
-        if len(captcha_url) > 1:
-            log.debug('Account encountered a reCaptcha.')
-            return False
+        # captcha_url = response['responses']['CHECK_CHALLENGE'].challenge_url
+        # if len(captcha_url) > 1:
+        #     log.debug('Account encountered a reCaptcha.')
+        #     return False
 
         spin_result = response['responses']['FORT_SEARCH'].result
         if spin_result == 1:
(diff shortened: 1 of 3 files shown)


FrostTheFox:
Once again, I'm not very experienced in the api request portion of the code, but I *think* those need commented out, because wouldn't they complain as the check_challenge part of the response is now gone? The only thing I can think of is if some weird protobuf nonsense makes it still have an empty key there.

Please let me know if I'm wrong about any of this or if more needs done, I just wanted to get the ball rolling because no one else has opened a PR on this subject and it needs done.

(so it is known, my intention was to simply cut off anything that called or relied upon check_challenge to start, so that we block entry points for the captcha code, and then we can think long-term about what to do with the leftovers.)

friscoMad:
There is a missed check in models.py in the encounter handling.
Anyway as said in Discord I would prefere to switch to a construct like this one: 
https://github.com/RocketMap/RocketMap/blob/9a5a1fdaf776de7c1be199a429284094320d6629/pogom/captcha.py#L183

This way we will leave the feature there and we can clean up everything if we remove all captcha handling later on.

FrostTheFox:
Holding pending more info on reports check_challenge is back in 0.85.1

Alderon86:
for now, it improved the lifetime of my accounts, but lets see how this check_challange thing goes

FrostTheFox:
Since the next API version is likely to be forced soon and check_challenge is back in that one, I'm hesitant to remove req.check_challenge right now.

I'll leave the download_settings fix and the ifs, because that makes things cleaner if we do need to remove check_challenge again (which could definitely happen because niantic is weird).

PR #2282, opened 2017-08-29
Dynamic Rarity

BrownSlaughter:
Tired of seeing Mr mime as ultra rare when you see 100's an hour?
This will set rarity for all pokemon from what you actually scan!

## Description
sets rarity group dependant what you actually see.
Updated pokemon.json

## Motivation and Context
Rarity is area dependent, what one person scans in America is not valid for what someone scans in England so having a 1 size fits all pokemon.json does not work for all. 

Future pokemon will automatically get set a rarity without needing to update pokemon.json

Scaling by rarity will also be more accurate because of this.

## How Has This Been Tested?
looked on map and checked it works like it did before changes but dynamically

Tested on a clean DB everything starts as ultra rare because its never seen them before but quickly changes during initial scan

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2387, opened 2017-12-07
Added "Exclude Pokemon by Rarity" feature to hide unwanted rarities.

trunksbomb:
Hidden rarities are chosen by a dropdown on the options tab, where each successive option contains all the rarities of the last option.
For example, the "Common" option hides all commons, the "Uncommon" option hides commons and uncommons, etc.

**Edit by Seb:** This PR doesn't work if the locale is not set to English, however this is intentional. We're moving Pokémon data localization from the back-end to the front-end, which will fix our localization problems/conflicts and send less data via raw_data. Once this localization change is finished, this PR will automatically work for all locales.

## Description
This was a requested feature by my local PoGo group. In the options menu, underneath the "Excluded Pokemon" list, you are now able to specify rarities to be hidden by default. To keep it simple (both for me to develop and users to use), the choices are limited to successive rarities. That is, the first and default option is to exclude no rarities; option 2 is to exclude all commons; option 3 excludes commons and uncommons; option 4 excludes common, uncommon, and rare- so on and so forth.
Behind the scenes, it behaves very similarly to the existing `excludedPokemon` list. While processing Pokemon, their rarity is checked against the current `excludedRaritiesList` and hidden if matched. When a change is made to `select` element, the entire `excludedPokemonByRarity` array is dumped and rebuilt on the next go round. This behavior is similar to how `excludedPokemon` works but is a bit more heavy handed. I'm unsure if this would cause any performance impacts.

## Motivation and Context
Most people (in my group at least) don't care to see a thousand Pidgey and Ratata cluttering their maps, and the current solution is to either blacklist those Pokemon so nobody can see them at all, or clutter up the "Hide Pokemon" list with all the undesired Pokemon.

## How Has This Been Tested?
I tested this in my local development environment and my production environment.
The Store setting (`excludedRarity`) persists throughout map reloads and behaves as intended when going up through the hidden rarities and back down.

Is a GFYcat appropriate? https://gfycat.com/gifs/detail/SaneLinearDogfish
At the time this was made, reversing through the list of hidden rarities did not function and so it is not present in the GFY.

## Types of changes

## Checklist:


Alderon86:
i think u should put this on hold till `dynamic raritiy` is merged, because rarity is different like everywhere.

trunksbomb:
From what I'm seeing with #2282 , this PR would neither affect nor be affected by dynamic rarity, at least not directly. If dynamic rarity updates the `pokemon_rarity` property then it will just use whatever rarity was dynamically set for your area. I don't see any reason to put it off because of that particular PR, unless I'm missing something.

trunksbomb:
I accidentally closed this PR earlier while doing an upstream fork merge.

static/js/map.js
```diff
@@ -1327,10 +1339,14 @@ function clearStaleMarkers() {
     $.each(mapData.pokemons, function (key, value) {
         const isPokeExpired = mapData.pokemons[key]['disappear_time'] < Date.now()
         const isPokeExcluded = excludedPokemon.indexOf(mapData.pokemons[key]['pokemon_id']) !== -1
-
-        if (isPokeExpired || isPokeExcluded) {
+        const isRarityExcluded = mapData.pokemons[key].hasOwnProperty('pokemon_rarity') ? (excludedRaritiesList[Store.get('excludedRarity')].indexOf(mapData.pokemons[key]['pokemon_rarity'].toLowerCase()) !== -1) : false
```
sebastienvercammen: A bit too long to be readable. :)

```
// Limit choice to our options [0, 5].
const excludedRarityOption = Math.min(Math.max(Store.get('excludedRarity'), 0), 5)
const excludedRarity = excludedRaritiesList[excludedRarityOption]
const hasRarity = mapData.pokemons[key].hasOwnProperty('pokemon_rarity')
// Not beautiful code with null as fallback, but it's more readable than a one-liner.
const rarity = hasRarity ? mapData.pokemons[key]['pokemon_rarity'].toLowerCase() : null
const isRarityExcluded = (hasRarity && excludedRarity.indexOf(rarity) !== -1)
```

static/js/map.js
```diff
@@ -1327,10 +1339,14 @@ function clearStaleMarkers() {
...
+        if (isPokeExpired || isPokeExcluded || isRarityExcluded) {
             const oldMarker = mapData.pokemons[key].marker
 
+            if (isRarityExcluded && excludedPokemonByRarity.indexOf(mapData.pokemons[key]['pokemon_id']) === -1) {
```
sebastienvercammen: Same here: for readability and to avoid duplicate code (a few lines lower), you'll want to keep a reference to the Pokémon ID.

```
const pokemon_id = mapData.pokemons[key]['pokemon_id']
```

and higher up, you'll want to keep a direct reference to the Pokémon object because we use it multiple times (make sure to edit my other example code to use this reference):

```
const pokemon = mapData.pokemons[key]
```

static/js/map.js
```diff
@@ -1567,13 +1583,14 @@ function processPokemonChunked(pokemon, chunkSize) {
 
 function processPokemon(item) {
     const isExcludedPoke = excludedPokemon.indexOf(item['pokemon_id']) !== -1
+    const isExcludedRarity = item.hasOwnProperty('pokemon_rarity') ? (excludedRaritiesList[Store.get('excludedRarity')].indexOf(item['pokemon_rarity'].toLowerCase()) !== -1) : false
```
sebastienvercammen: Same as above, readability is a problem.

static/js/map.js
```diff
@@ -1592,6 +1609,8 @@ function processPokemon(item) {
         } else {
             oldMarker = item.marker
         }
+    } else if (!isExcludedPoke && isExcludedRarity && isPokeAlive && excludedPokemonByRarity.indexOf(item['pokemon_id']) === -1) {
+        excludedPokemonByRarity.push(item['pokemon_id'])
```
sebastienvercammen: Same as above: readability. Simplify.

sebastienvercammen:
Sorry for barging in and pushing the changes myself, but we've decided to put this PR on somewhat of a fast track to be merged with today's changes.

PR #2395, opened 2017-12-12
Update de.json

Netfreak25:
adding translation for german moves

## Description
Adding the translation for each move for the german language

## Motivation and Context
Names of the Pokémon and Types of the moves have already been translated, but german players are also used to the german name of the move, which is completely different in most of the cases

## How Has This Been Tested?
Tested on my Map 

screen shot 2017-12-12 at 20 51 33
screen shot 2017-12-12 at 20 53 48

## Types of changes

## Checklist:


Netfreak25:
I have to admit that this is a bad example that the names of the moves are completely different in german

SwenWacker:
"Psychic": "Psycho",
"Psychic": "Psychokinese",
see #2019

PR #2399, opened 2017-12-16
Add user interface option to Notify by Pokemon Level

tomballgithub:
## Description

This adds a user interface option to notify by pokemon level.  This is virtually the same coding as the notify by pokemon perfection.

If the user enters a pokemon level between 1 and 40, the map will notify them if a pokemon meeting that level is found.  If both a level and perfection notification are configured, a pokemon will only be notified if it meets both criteria.

## Motivation and Context

This was request in the PR channel and I have had multiple users request this feature.  It is especially interesting now with weather-boosting causing pokemon with level 35 to be found in the wild

## How Has This Been Tested?

I tested the code on my Rocketmap server user interface under Ubuntu 16.04.
I tweaked it until it worked as desired.

image

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -869,6 +869,9 @@ var StoreOptions = {
     'excludedRarity': {
         default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
         type: StoreTypes.Number
+    'remember_text_level_notify': {
+        default: '',
+        type: StoreTypes.Number
     },
     'showRaids': {
         default: false,
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -862,13 +862,14 @@ var StoreOptions = {
         default: [], // Common, Uncommon, Rare, Very Rare, Ultra Rare
         type: StoreTypes.JSON
     },
+    'excludedRarity': {
+        default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
+        type: StoreTypes.Number
+    },
     'remember_text_perfection_notify': {
         default: '',
         type: StoreTypes.Number
     },
-    'excludedRarity': {
-        default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
-        type: StoreTypes.Number
     'remember_text_level_notify': {
         default: '',
         type: StoreTypes.Number
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -862,16 +862,16 @@ var StoreOptions = {
         default: [], // Common, Uncommon, Rare, Very Rare, Ultra Rare
         type: StoreTypes.JSON
     },
-    'excludedRarity': {
-        default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
-        type: StoreTypes.Number
-    },
     'remember_text_perfection_notify': {
         default: '',
         type: StoreTypes.Number
     },
     'remember_text_level_notify': {
         default: 0,
+        type: StoreTypes.Number
+    },
+    'excludedRarity': {
+        default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
         type: StoreTypes.Number
     },
     'showRaids': {
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -871,7 +871,7 @@ var StoreOptions = {
         type: StoreTypes.Number
     },
     'remember_text_level_notify': {
-        default: '',
+        default: 0,
         type: StoreTypes.Number
     },
     'showRaids': {
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -1055,8 +1055,8 @@ function isNotifyPerfectionPoke(poke) {
         const perfection = getIv(poke['individual_attack'], poke['individual_defense'], poke['individual_stamina'])
         const level = getPokemonLevel(poke['cp_multiplier'])
 
-        const hasHighIV = notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection
-        const hasHighLevel = notifiedMinLevel > 0 && level >= notifiedMinLevel
+        hasHighIV = notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection
+        hasHighLevel = notifiedMinLevel > 0 && level >= notifiedMinLevel
 
         const shouldNotifyForIV = (hasHighIV && notifiedMinLevel <= 0)
         const shouldNotifyForLevel = (hasHighLevel && (hasHighIV || notifiedMinPerfection <= 0))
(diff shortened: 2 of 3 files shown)


static/js/map.js
```diff
@@ -1014,13 +1016,20 @@ function playPokemonSound(pokemonID, cryFileTypes) {
...
 
-    if (poke['individual_attack'] != null) {
+    if (poke['individual_attack'] != null && poke['cp_multiplier'] !== null) {
         const perfection = getIv(poke['individual_attack'], poke['individual_defense'], poke['individual_stamina'])
+        const level = getPokemonLevel(poke['cp_multiplier'])
+
         hasHighIV = notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection
+        hasHighLevel = notifiedMinLevel > 0 && level >= notifiedMinLevel
+
+        hasHighAttributes = (hasHighIV && notifiedMinLevel === 0) || (hasHighLevel && notifiedMinPerfection === 0) || hasHighLevel && hasHighIV
```
sLoPPydrive: Change this to
```
hasHighAttributes = (hasHighIV && !(notifiedMinLevel > 0)) || (hasHighLevel && !(notifiedMinPerfection > 0)) || hasHighLevel && hasHighIV
```
...to make it always work.

tomballgithub: Fixed...

tomballgithub:
FYI sloppydrive's users found the needed code improvement earlier but otherwise this is working well.

darkelement1987:
Is it also possible for you to build in 'Exclude/hide pokemon by iv/level'? I would love that.

tomballgithub:
@darkelement1987 Yes there is another PR that allows you to hide pokemon that are not being notified.  So you could set the minimum level and IV to what you want and then 'hide unnotified pokemon' and it will only show the high level ones (excluding the others)  https://github.com/RocketMap/RocketMap/pull/2377

darkelement1987:
Awesome!

2018-01-15 19:02 GMT+01:00 tomballgithub <notifications@github.com>:

tomballgithub:
Rebased...

static/js/map.js
```diff
@@ -2739,12 +2749,24 @@ $(function () {
             $textPerfectionNotify.val(notifiedMinPerfection)
             Store.set('remember_text_perfection_notify', notifiedMinPerfection)
         })
+        $textLevelNotify.on('change', function (e) {
+            notifiedMinLevel = parseInt($textLevelNotify.val(), 10)
+            if (isNaN(notifiedMinLevel) || notifiedMinLevel <= 0) {
+                notifiedMinLevel = ''
+            }
+            if (notifiedMinLevel > 40) {
```
Alderon86: Wild pokemon can be max lvl 30 and 35 with weather boost, just a tiny notice.

static/js/map.common.js
```diff
@@ -869,6 +869,9 @@ var StoreOptions = {
     'excludedRarity': {
         default: 0, // 0: none, 1: <=Common, 2: <=Uncommon, 3: <=Rare, 4: <=Very Rare, 5: <=Ultra Rare
         type: StoreTypes.Number
+    'remember_text_level_notify': {
```
fosJoddie: js object literal syntax error, `npm run build` will fail

tomballgithub: Thank you, fixed.  Was introduced during rebase.

static/js/map.js
tomballgithub: Using level 40 allows this code to be future proof.

static/js/map.js
```diff
@@ -1050,13 +1052,20 @@ function playPokemonSound(pokemonID, cryFileTypes) {
...
 
-    if (poke['individual_attack'] != null) {
+    if (poke['individual_attack'] != null && poke['cp_multiplier'] !== null) {
         const perfection = getIv(poke['individual_attack'], poke['individual_defense'], poke['individual_stamina'])
+        const level = getPokemonLevel(poke['cp_multiplier'])
+
         hasHighIV = notifiedMinPerfection > 0 && perfection >= notifiedMinPerfection
+        hasHighLevel = notifiedMinLevel > 0 && level >= notifiedMinLevel
+
+        hasHighAttributes = (hasHighIV && !(notifiedMinLevel > 0)) || (hasHighLevel && !(notifiedMinPerfection > 0)) || hasHighLevel && hasHighIV
```
sebastienvercammen: You can simplify

```
hasHighAttributes = (hasHighIV && !(notifiedMinLevel > 0)) || (hasHighLevel && !(notifiedMinPerfection > 0)) || hasHighLevel && hasHighIV
```

to

```
const shouldNotifyForIV = (hasHighIV && notifiedMinLevel <= 0)
const shouldNotifyForLevel = (hasHighLevel && (hasHighIV || notifiedMinPerfection <= 0))

hasHighAttributes = shouldNotifyForIV || shouldNotifyForLevel
```

Note that I used `<= 0` because it's the reverse of `>`. If `<` is unnecessary, change it to `==`.

static/js/map.js
```diff
@@ -2739,12 +2749,24 @@ $(function () {
             $textPerfectionNotify.val(notifiedMinPerfection)
             Store.set('remember_text_perfection_notify', notifiedMinPerfection)
         })
+        $textLevelNotify.on('change', function (e) {
+            notifiedMinLevel = parseInt($textLevelNotify.val(), 10)
+            if (isNaN(notifiedMinLevel) || notifiedMinLevel <= 0) {
+                notifiedMinLevel = ''
```
sebastienvercammen: The default should be `0` rather than an empty string and the `<= 0` should be changed with `< 0`. `0` can be a valid option, to disable it. Make sure to keep the behavior consistent with "Notify by IV".

static/js/map.js
sebastienvercammen: I would remove this check and set the upper limit to `35` in the input in the HTML template. If anyone tampers with it to allow values above 35, the code is right to assume that they only want to be notified of levels at or above whichever level they've set.

static/js/map.js
tomballgithub: I agree but note that I copied the exact '' assignment also used for `remember_text_perfection_notify`

static/js/map.js
tomballgithub: This was due to a bug caused by the '' assignment as default

static/js/map.js
tomballgithub: There doesn't seem to be a simple way to do this in the template.  Min/Max values are only applicable if input type is number not text.  Changing to number causes a behavior change in that there are now scroll bars up/down and the input box gets resized.  I'll leave it as-is but change to 35 maximum even though I'd prefer 40 to be future proof.

tomballgithub:
Should be all rebased and fixed

static/js/map.js
```diff
@@ -1046,13 +1048,23 @@ function playPokemonSound(pokemonID, cryFileTypes) {
...
+        hasHighLevel = notifiedMinLevel > 0 && level >= notifiedMinLevel
+
+        const shouldNotifyForIV = (hasHighIV && notifiedMinLevel <= 0)
+        const shouldNotifyForLevel = (hasHighLevel && (hasHighIV || notifiedMinPerfection <= 0))
+
+        hasHighAttributes = shouldNotifyForIV || shouldNotifyForLevel
     }
 
-    return hasHighIV
+    return hasHighAttributes
```
sebastienvercammen: The current code will cause a problem when "Notify by IV" is enabled and "Notify by level" is disabled, and the Pokémon has IV data but no `cp_multiplier` (technically not possible since they're both inserted during scan, but logically they're unrelated so they should be separated).

The problem lies in your condition:
```
var hasHighAttributes = false

if (poke['individual_attack'] != null && poke['cp_multiplier'] !== null) {
    ...
}

return hasHighAttributes
```

tomballgithub:
Approved via code review only, not testing of latest changes

PR #2404, opened 2017-12-19
Dynamic loading of Encounter's list and webhook

ClarkyKent:
## Description
This will update the list of pokemon IDs in file provided to option enc_whitelist_file. This will update every 30 secs.

## Motivation and Context
With weather change, rare pokemon can become pretty common. This permits to change IV scanning list without need of stoping the map.

## How Has This Been Tested?
personal map

## Types of changes

## Checklist:


package.json
```diff
@@ -51,4 +51,4 @@
     "url": "https://github.com/RocketMap/RocketMap/issues"
   },
   "homepage": "https://github.com/RocketMap/RocketMap#readme"
-}
+}
```
pogo-excalibur: Would you mind removing this from the PR?

Just for the sake of tidiness.

pki0:
I really like this idea. Would it be possible to do this with L2 accounts, L30 accounts and webhook_whitelist too? Would be awesome.

friscoMad:
I have some suggestions:

- Only parse the file if it has been updated, reading the same file every 30 secs forever seems a bit overkill
- Try to make the function somewhat generic to support other files as already suggested (accounts are more difficult to handle) but there is something similar for proxies.
- Even more we could create a cron module as we have several cron tasks to try to reduce the thread number (just for the cron functions if they are several).
- We should try to avoid using args everywhere, if you only need some args values, pass those values

ClarkyKent:
@pkiel for L2 and L30 there is different approach. It will be in there in future with PR's like "Accs in DB".
For webhook, yes it can be added.

ClarkyKent:
@friscoMad 

> Only parse the file if it has been updated, reading the same file every 30 secs forever seems a bit overkill

God idea , i will do that.

> Try to make the function somewhat generic to support other files as already suggested (accounts are more difficult to handle) but there is something similar for proxies.

That is the idea. to have a generic function , since i was only doing whitelist , so started from this. In future i will make changes to this function to adapt new changes.

> We should try to avoid using args everywhere, if you only need some args values, pass those values

Ok, will do that as well

pogom/utils.py
```diff
@@ -839,6 +839,25 @@ def i8ln(word):
...
+    while True:
+        # Wait 60 secs before refresh,
+        time.sleep(60)
+
+        try:
+            # IV/CP scanning.
+            if enc_whitelist_file:
+                if (os.stat(enc_whitelist_file).st_mtime + 60) > time.time():
+                    with open(enc_whitelist_file) as f:
+                        enc_whitelist = frozenset([int(l.strip()) for l in f])
```
sebastienvercammen: `enc_whitelist` is a local variable, you need to change `args.enc_whitelist`.

pogom/utils.py
```diff
@@ -839,6 +839,25 @@ def i8ln(word):
...
+# Thread function for periodical enc list updating.
+def enc_list_refresher(enc_whitelist_file):
+    while True:
+        # Wait 60 secs before refresh,
+        time.sleep(60)
+
+        try:
+            # IV/CP scanning.
+            if enc_whitelist_file:
+                if (os.stat(enc_whitelist_file).st_mtime + 60) > time.time():
```
sebastienvercammen: Consider using
```
os.path.getmtime(...)
```
instead ([docs](https://docs.python.org/2/library/os.path.html#os.path.getmtime)).

jorgitoh:
Hi. Im using something similar. I copied the code from sloppy fork:

https://github.com/sLoPPydrive/RocketMap/commit/d4e447f3e41ae50a175f492fb8a724b40daf560b
https://github.com/sLoPPydrive/RocketMap/commit/818794549873d8305860e0810829179b37dbcde3

Works perfect. Maybe we should adjust the check file timer.

sebastienvercammen:
@jorgitoh This PR is the first in a series to abstract and standardize the reloading at runtime of configuration files.

This PR focuses on reloading the whitelist, the next will focus on moving similar background tasks to a single thread/method.

Sloppy has clarified on Discord that he has his own approach to things. But if he feels like building the PR, he's welcome to.

pogom/utils.py
```diff
@@ -841,17 +841,25 @@ def i8ln(word):
...
+                file_modified_time_sec = os.path.getmtime(enc_whitelist_file)
+                time_diff_sec = current_time_sec - file_modified_time_sec
+                if (time_diff_sec < refresh_time_sec):
+                    args = get_args()
                     with open(enc_whitelist_file) as f:
-                        enc_whitelist = frozenset([int(l.strip()) for l in f])
-                    log.info('New enc list is ' + enc_whitelist)
+                        args.enc_whitelist = frozenset(
+                            [int(l.strip()) for l in f])
+                    log.info('New enc list is ' + args.enc_whitelist)
```
sebastienvercammen:
```
log.info('New enc list is %s.', args.enc_whitelist)
```

pogom/utils.py
```diff
@@ -839,6 +839,33 @@ def i8ln(word):
...
+            if enc_whitelist_file:
+                current_time_sec = time.time()
+                file_modified_time_sec = os.path.getmtime(enc_whitelist_file)
+                time_diff_sec = current_time_sec - file_modified_time_sec
+                if (time_diff_sec < refresh_time_sec):
+                    args = get_args()
+                    with open(enc_whitelist_file) as f:
+                        args.enc_whitelist = frozenset(
+                            [int(l.strip()) for l in f])
+                    log.info('New enc list is %s' + args.enc_whitelist)
```
fosJoddie: I'm getting an error about frozen sets here, suggest changing to 
`log.info('New enc list is %s.' + args.enc_whitelist)`

sebastienvercammen: > log.info('New encounter whitelist is: %s.', args.enc_whitelist)

pogom/utils.py
```diff
@@ -839,6 +839,33 @@ def i8ln(word):
...
+
+            log.info('Regular encounter list refresh complete.')
```
sebastienvercammen: > Updated encounter whitelist.

pogom/utils.py
```diff
@@ -839,6 +839,33 @@ def i8ln(word):
...
+        except Exception as e:
+            log.exception('Exception while refreshing enc_list: %s.', e)
```
sebastienvercammen: > Exception occurred while updating encounter whitelist: %s.

runserver.py
```diff
@@ -391,6 +392,14 @@ def main():
...
             log.info('Periodical proxies refresh disabled.')
 
+        if args.encounter:
+            t = Thread(target=enc_list_refresher,
+                       name='dynamic-enclist', args=(args.enc_whitelist_file,))
+            t.daemon = True
+            t.start()
+            log.info('Encounter is enabled.')
+        else:
+            log.info('Encounter is disabled.')
```
sebastienvercammen: > Encounters are enabled.

> Encounters are disabled.

sebastienvercammen:
Still pending changes per @ClarkyKent's request.

PR #2408, opened 2017-12-22
Update custom.js.example to disable marker clustering even when js load order make the library load before custom.js

fosJoddie:
Updates the custom.js example so that if the client happens to be loading for the first time and also at the same time happens to load the marker cluster library before custom.js it will still disable the clusttering without needing a page reload.

## Description
Changed markerCluster definition to be more explicit since we reference it in a different file/context

Checks if markerCluster is defined when custom.js loads and disables clustering directly on the markerCluster object if it is available + needed

If there is other obvious places/things that should utilize the same "hack", it probably should be in this PR as well

## Motivation and Context
Some of my users are paranoid and runs RM in incognito mode, but would always complain that marker clustering was enabled until they reloaded the page.

## How Has This Been Tested?
Opened my local map in incognito mode and saw that there was in fact no pokemon clustering

(your mind sees a beautiful image of a lot of pokemon in a map where there would be grouping indicators before - you sigh in awe at the awesomeness of technology)
 
## Types of changes

## Checklist:

### DIFF
--- a/static/js/custom.js.example
+++ b/static/js/custom.js.example
@@ -1,8 +1,8 @@
 $(function () {
     'use strict'
 
-    //marker cluster might have loaded before custom.js
-    const isMarkerClusterLoaded = !!markerCluster;
+    // Marker cluster might have loaded before custom.js
+    const isMarkerClusterLoaded = typeof window.markerCluster !== 'undefined' && !!window.markerCluster
 
     /* Settings. */
 
@@ -84,9 +84,9 @@ $(function () {
     if (disableClusters) {
         Store.set('maxClusterZoomLevel', -1)
 
-        if(isMarkerClusterLoaded) {
-			markerCluster.setMaxZoom(-1)
-		}
+        if (isMarkerClusterLoaded) {
+            window.markerCluster.setMaxZoom(-1)
+        }
     }
 
     // Google Analytics.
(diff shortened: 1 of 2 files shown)


static/js/custom.js.example
```diff
@@ -1,6 +1,8 @@
 $(function () {
     'use strict'
 
+    //marker cluster might have loaded before custom.js
+    const isMarkerClusterLoaded = !!markerCluster;
```
sebastienvercammen:
```
// Marker cluster might have loaded before custom.js.
const isMarkerClusterLoaded = typeof markerCluster !== 'undefined' && !!markerCluster;
```

static/js/custom.js.example
```diff
@@ -81,6 +83,10 @@ $(function () {
 
     if (disableClusters) {
         Store.set('maxClusterZoomLevel', -1)
+
+        if(isMarkerClusterLoaded) {
```
sebastienvercammen: Space
```
if (...)
```

static/js/custom.js.example
```diff
@@ -1,8 +1,8 @@
 $(function () {
     'use strict'
 
-    //marker cluster might have loaded before custom.js
-    const isMarkerClusterLoaded = !!markerCluster;
+    // Marker cluster might have loaded before custom.js
```
sebastienvercammen: Missing a dot at the end of your comment. 😳

PR #2410, opened 2017-12-22
Skip spawnpoints with missed_count > 5.

sebastienvercammen:
## Description
Hides the "missed x times in a row" error message for spawnpoints that have been missed more than 5 times.

## Motivation and Context
We missed one instance where we iterate spawnpoints, causing the error spam.

## How Has This Been Tested?
Untested. The fix seems very simple, but we must double-check the effects of the code that is being skipped. Since the spawnpoint is invalid (missed > 5) no additional parsing should be necessary, but we shouldn't assume that's the case.

Approve only if you've run, tested and reviewed the code.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2278,6 +2278,9 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
     # Look for spawnpoints within scan_loc that are not here to see if we
     # can narrow down tth window.
     for sp in ScannedLocation.linked_spawn_points(scan_location['cellid']):
+        if sp['missed_count'] > 5:
+                continue
+
         if sp['id'] in sp_id_list:
             # Don't overwrite changes from this parse with DB version.
             sp = spawn_points[sp['id']]


pogom/models.py
```diff
@@ -2278,6 +2278,9 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
     # Look for spawnpoints within scan_loc that are not here to see if we
     # can narrow down tth window.
     for sp in ScannedLocation.linked_spawn_points(scan_location['cellid']):
+        if sp['missed_count'] > 5:
```
friscoMad: Maybe we should just skip the logging if count > 5 but also set missed_count to 0 if it does appear, it could happend due to SBs or bad requests.

sebastienvercammen: @friscoMad As it is, we skip the spawnpoint on several occasions when missed_count > 5. In this PR we'll standardize the behavior so it happens for the error message as well, and in the next PR we'll reset it to 0 if we find it + change the hardcoded 5 to be a configurable number.

Issue #2396, opened 2017-12-13
Bad handling with L30 accounts


urinella:
Right now we have two stages while accounts are getting banned.
First the account is totally blind and causes bad scans. After a lot of scans the account gets temporary banned. Normal scanning can escape from this first stage with --max-empty.

But L30 accounts only get's "status 0" and try to encounter again and again.
`There was an error encountering Pokemon ID XY with account XXXXX: 0.`

## Expected Behavior
If an account from the L30 pool getting status = 0, ignore this account in the current pool and try to reencounter the Pokémon with the next L30 account.

## Current Behavior
Stage 1 banned accounts getting status = 0 and trying to encounter all Pokémon until the account is fully temp banned.

## Possible Solution
Modify the exception handling to reencounter a Pokémon and kick status = 0 accounts out of the pool.

## Steps to Reproduce (for bugs)
1. Encounter enough Pokémon for 24h
2. Wait

PR #2415, opened 2017-12-28
Retry encounter at peaks and bans

pkiel:
## Description

With an infinity speed-limit it is possible to encounter Pokémon every second with one account. This means up to 86.400 Encounters during the 24 hour account lifetime.

**First problem:**
If you encounter a lot of Pokémon it is possible that multiple Pokémon from your encounter-list spawns at the same time. Then you get the following error:
`No L30 accounts are available, please consider adding more. Skipping encounter.`
I call it: _Peak-Error_.
This PR avoids this problem and reencounter the Pokémon up to 10 times.

**Second problem:**
If the high-level account gets blind or banned it skips the encounter with status = 0.
See: #2396 .
This PR also avoids this problem and reencounter the Pokémon up to 10 times.

## Motivation and Context

In the past I was encountering 180.000 Pokémon a day with 7 accounts.
Now I can use 2-3 accounts to encounter all of my Pokémon.

This saves a lot of accounts, breaking the _Peak-Error_ and the _status = 0 error_.

I set the retries to a limit of 10. This was enough for me. I never hit a greater count then 5 on my map. 
I also set a random sleep-timer for the encounter to have a small break.

All in all it would be better to create an encounter-queue but maybe this is the first fix to handle the problems I wrote above.

This is an example for the retry messages:

```
2017-12-28 14:13:07,001 [  search-worker-29][        models][    INFO] Retry: 1 for encountering a Pokemon.
2017-12-28 14:13:11,407 [  search-worker-29][        models][    INFO] Retry: 2 for encountering a Pokemon.
2017-12-28 14:13:24,312 [  search-worker-29][        models][    INFO] Retry: 3 for encountering a Pokemon.
2017-12-28 14:13:29,421 [  search-worker-29][        models][    INFO] Retry: 4 for encountering a Pokemon.
2017-12-28 14:13:56,861 [ search-worker-205][        models][    INFO] Retry: 1 for encountering a Pokemon.
2017-12-28 14:14:52,921 [  search-worker-52][        models][    INFO] Retry: 1 for encountering a Pokemon.
```
## How Has This Been Tested?

The PR runs locally on a map with 200 workers.

## Types of changes

## Checklist:


pogom/models.py
```diff
@@ -2026,6 +2028,20 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
             if args.encounter and (pokemon_id in args.enc_whitelist):
                 pokemon_info = encounter_pokemon(
                     args, p, account, api, account_sets, status, key_scheduler)
+                # Retry the encounter if no accounts available
+                while not pokemon_info:
```
friscoMad: Consider using do-while instead of a regular while to simplify the logic (only 1 encounter_pokemon call)

friscoMad:
How does this fix the blind error? The same account will be used again and again until 10 retries are done unless we add some other code to remove blinded accounts from the list.

Alderon86:
it doesnt handle blinding and i think this souldnt be adressed by this pr.

pki0:
@friscoMad @Alderon86 You're correct. That's why I wrote it can **'avoid'** the problem. In my case I was running 7 accounts and when the ban came on the first accounts I had not a single missing IV scan. It's more a wheelchair for this problem then a fix.
This problem needs another fix of cause.

pogom/models.py
pkiel: I think I can not follow you, sorry.
Do you mean removing the first call and set pokemon_info = [] before starting the while loop?
Or do you have an example for me please?

friscoMad: Ohh python does not have the do...while construct, maybe something like (this should replace the code frome line 2030):

```python
enc_retry_count = 0
while true:
  poke_info = encounter_pokemon()
  enc_retry_count += 1
  if enc_retry_count >= 10:
    log.error('No L30....')
  if poke_info or enc_retry_count >= 10:
    break
  time.sleep(X)

**Comment by friscoMad** [comment]
After giving more thought to this I am not really sure we should retry encounters in all cases.
Despawn poke (pokemon fled) error can occur when new spawnpoints appear and a niantic down or hash error probably shouldn't retry.
If we only want to retry on the not enough accounts error I would add it in the account set next function as it seems the perfect place to handle the issue (we can do that a blocking call until we have a valid account) but if we also want to retry other errors like status 0 or 3 (previously SB was error 3, I think you get 0 when the account is banned) I think we shoud add that logic inside the encounter pokemon function.

**Comment by pki0** [comment]
@friscoMad Yes. Status 3 & 4 should not be retried at all.
We could return the status to pokemon_info as well but also another approach to retry the encounter would be nice. This was just the first idea I had. 

Even better would be a queue like the normal scans do. Then the queue is filling up and scans that last > 60  seconds in the queue are getting killed or something like that.
But I think I'm not able to write this queue. Maybe someone who follows this discussion wants to do a new PR with a queue or rewrite of the encounter function.

File: pogom/models.py [hunk 2]
```diff
@@ -2026,6 +2028,20 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
...
+                    if encounter_retry_count >= 10:
+                        log.error('No L30 accounts are available, please' +
+                                  ' consider adding more. Skipping encounter' +
+                                  ' after 10 retries.')
+                        break
+                    log.debug('Retry: %s for encountering a Pokemon.',
+                              encounter_retry_count)
+                    pokemon_info = encounter_pokemon(
```
fosJoddie: One way to solve the problem of retryability is to let encounter_pokemon return an additional flag `can_retry` which could be used to show if rerunning `encounter_pokemon` might yield a better result

like so: 

`pokemon_info, can_retry = encounter_pokemon(...`

Then in encounter_pokemon you would (for now) return `False, True` anywhere where we today just return `False`

So if we are unable to get a high level account: 
```
        if not hlvl_account:
            return False, True
```

pki0:
So I added the suggestions.

- encounter_pokemon(...) is now in the code only once.
- Added a flag for can_retry = False on captcha, and status != 1 and != 0
- Also changed the waiting time a bit which worked well for me.

pogom/models.py
```diff
@@ -2292,6 +2309,7 @@ def encounter_pokemon(args, pokemon, account, api, account_sets, status,
     hlvl_account = None
     pokemon_id = None
     result = False
+    can_retry = True
```
sebastienvercammen: Dubious: the `encounter_pokemon` method shouldn't be the one that tells its caller whether it can retry or not.

pki0:
Okay, so I found a better method for the encounter by removing the whole "in_use" operation from the account_set. With this I am able to encounter 17k spawnpoints with 4-5 accounts. Working on a method to do all encounters with one account.
I don't know anything about getting proxies banned with this method so I think RM will never accept this.
So this method is not a point for me anymore and I'll close this PR.

PR #2417, opened 2017-12-29
show raid details even if image is missing

BrownSlaughter:
## Description
Right now if a new raid boss pops up we see no details about the raid (look at walimer for an example) until the id is added to `pokemonWithImages` and only see the gym colour and raid level
before
This PR changes it so even if a new raid boss spawns and the map has not been updated it will still see details of the raid
after

## Motivation and Context

It is annoying seeing that there is a raid but not knowing what it is until the map and image has been updated

## How Has This Been Tested?
Removed raid id 71 from `pokemonWithImages` to simulate a new raid boss then tested details show before and after PR.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -75,7 +75,6 @@ const unownForm = ['unset', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K
 const pokemonWithImages = [
     2, 3, 5, 6, 8, 9, 11, 28, 31, 34, 38, 59, 62, 65, 68, 71, 73, 76, 82, 89, 91, 94, 103, 105, 110, 112, 123, 125, 126, 129, 131, 134, 135, 136, 137, 139, 143, 144, 145, 146, 150, 153, 156, 159, 243, 244, 245, 248, 249, 250, 302, 303, 359, 382, 383, 384
 ]
-var raidImage
 
 /*
  text place holders:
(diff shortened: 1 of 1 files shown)


static/js/map.js
```diff
@@ -678,7 +678,21 @@ function gymLabel(gym, includeMembers = true) {
         if (isRaidStarted) {
             // Set default image.
             image = `
-                <img class='gym sprite' src='static/images/raid/${gymTypes[gym.team_id]}_${raid.level}_unknown.png'>
+                <div class='raid container'>
```
friscoMad: The only difference between this code and the one inside the next if is just the image and level and timer that can be shown anyway so just store the image in a variable (using the one in the array or the unkown one) and use a single template for both cases.

static/js/map.js
```diff
@@ -75,7 +75,7 @@ const unownForm = ['unset', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K
 const pokemonWithImages = [
     2, 3, 5, 6, 8, 9, 11, 28, 31, 34, 38, 59, 62, 65, 68, 71, 73, 76, 82, 89, 91, 94, 103, 105, 110, 112, 123, 125, 126, 129, 131, 134, 135, 136, 137, 139, 143, 144, 145, 146, 150, 153, 156, 159, 243, 244, 245, 248, 249, 250, 302, 303, 359, 382, 383, 384
 ]
-
+var raidImage
```
sebastienvercammen: This global definitely shouldn't be here. 😳 

Put it at the top in the function where you use it.

PR #2422, opened 2018-01-04
Correctly rotate hash keys

friscoMad:
## Description
Previously we had keyscheduler logic all over the place but we can move all to a single point in apiRequest and it makes a lot more sense.

## Motivation and Context
Did this changes for the future api but most of them can be ported back to current api so it will rotate the keys more evenly (instead of changing the key per gmo it does change per request).

The next step should be doing the same for proxies buy I wanted this merged before adding more things.

## How Has This Been Tested?

Just some minutes in my test instance, but my main instance is running the branch from where I ported the code.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/apiRequests.py
+++ b/pogom/apiRequests.py
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 
 import logging
-from datetime import datetime
 
 from pgoapi.hash_server import HashServer
 from pgoapi.utilities import get_cell_ids
--- a/pogom/apiRequests.py
+++ b/pogom/apiRequests.py
@@ -2,15 +2,44 @@
 # -*- coding: utf-8 -*-
 
 import logging
+from datetime import datetime
 
+from pgoapi.hash_server import HashServer
 from pgoapi.utilities import get_cell_ids
 from pgoapi.hash_server import BadHashRequestException, HashingOfflineException
 
 log = logging.getLogger(__name__)
+_key_scheduler = None
 
 
 class AccountBannedException(Exception):
     pass
+
+
+def configure(key_scheduler):
+    global _key_scheduler
+    _key_scheduler = key_scheduler
+
+
+def execute_request(req):
+    if _key_scheduler:
+        key = _key_scheduler.next()
+        log.debug('Using key {} for this scan.'.format(key))
+        req.__parent__.activate_hash_server(key)
+    resp = req.call(False)
+    if _key_scheduler:
+        update_hash_stats()
+    return resp
+
+
+def update_hash_stats():
+    # Update hashing key stats in the database based on the values
+    # reported back by the hashing server.
+    key = HashServer.status.get('token', None)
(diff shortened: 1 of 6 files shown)


pogom/apiRequests.py
```diff
@@ -3,16 +3,44 @@
 
 import logging
 
+from pgoapi.hash_server import HashServer
 from pgoapi.utilities import get_cell_ids
 from pgoapi.hash_server import BadHashRequestException, HashingOfflineException
 
 log = logging.getLogger(__name__)
+_key_scheduler = None
```
sebastienvercammen: We shouldn't introduce globals.

pogom/apiRequests.py
```diff
@@ -3,16 +3,44 @@
...
 
 
 class AccountBannedException(Exception):
     pass
 
 
+def configure(key_scheduler):
+    global _key_scheduler
+    _key_scheduler = key_scheduler
```
sebastienvercammen: This method is only here because of the global, we should use a proper scope.

pogom/schedulers.py
```diff
@@ -1213,3 +1213,25 @@ def current(self):
     def next(self):
         self.curr_key = self.key_cycle.next()
         return self.curr_key
+
+    def update_hash_stats(self, key, remaining, maximum, expiration):
```
Alderon86: i dont see the request for the token anymore? did you triple check with several keys and instances that none of the values jumps around?

friscoMad: I did not triple check anything and does not have several keys to test 😄 
I am not really sure about what are your fears, the current desing of api does not return the data with the response so it is not thread safe, we just get the data after the request has been done but if there was a context switch the data we retrieve it is always from the last request, at least we are not mixing the key from the request with the data we are reading as they could be different, we just store the data (token and headers) the api has when this is called.
Proper solution would need to change the API to be thread safe.

Alderon86: yup and because we store just the last request the token is needed to identify which key made the request, otherwise its failing to allocate the correct key for the value, i have some spare time left tomomorrow and will check if this is going ok or not.

pogom/apiRequests.py
```diff
@@ -3,16 +3,44 @@
...
+
+
+def update_hash_stats():
+    # Update hashing key stats in the database based on the values
+    # reported back by the hashing server.
+    key = HashServer.status.get('token', None)
+    remaining = HashServer.status.get('remaining', 0)
+    maximum = HashServer.status.get('maximum', 0)
+    expiration = HashServer.status.get('expiration', None)
+    _key_scheduler.update_hash_stats(key, remaining, maximum, expiration)
```
friscoMad: @Alderon86 I am getting the key directly from the status so even if it is not the one that made the request we store the data correctly.

friscoMad:
Closing this as I will no longer be supporting my PRs, if any one wants to take over please do so.

PR #2416, opened 2017-12-29
Update to fix CONSTRAINT_4 error for spawnpoint table

Kneckter:
## Description

RM tries to write DB records with a 'latest_seen=3600' instead of 3599 like the constraint allows for. Changing the Check statement to '=<3600' allows for these records to be written instead of throwing "InternalError(4025, u'CONSTRAINT CONSTRAINT_4 failed for rocketmapdbadmin.spawnpoint')".

Note that this code change will not update existing DBs and will only apply the new constraint to maps that run the -cd flag. 

An FAQ can be added if anyone wants to run the mysql commands to update their DB. The following commands should allow anyone to do it:

```command
mysql -u rocketmapuser -p
```
```mysql
USE 'rocketmapdb';
ALTER TABLE spawnpoint DROP CONSTRAINT CONSTRAINT_4;
ALTER TABLE spawnpoint ADD CONSTRAINT CONSTRAINT_4 CHECK (`latest_seen` <= 3600);
```

## Motivation and Context

This error kept occurring on my map and many others. Most have seen this as a resource issue and the only solution was to decrease step size and run a -cd. Search the discord channels, there are 6 pages of results but no real solutions. If this is implemented, people will be able to run their large steps without having to restart a DB if they run into a constraint issue.

## How Has This Been Tested?

Using Ubuntu 17.04 with the GUI desktop installed. Running a speedscan with an -st 200 geofenced to 36723 steps with 600 workers and 10 DB threads. 

**I have not tested the python code change, only the DB constraint commands.**

I have been running this -st 200 with 50K spawnpoints started to get error 4025 when I hit about 75%. I pulled up HeidiSQL to see what was being written. Since I have spawnpoints with 'latest_seen=3599' and saw that RM wanted to write one with 'latest_seen=3600', I changed the constraint on the spawnpoints table with the above commands. Since then, I have not received the error and completed the initial scan. There are about 150 spawnpoints that have 'latest_seen=3600'. 

As my TTH search got smaller, so did the total 'latest_seen=3600' DB records. I was down to 26 after a night so this change should not cause any issues with searching.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2423, opened 2018-01-04
Fix 3600 being inserted

friscoMad:
## Description
Should fix #2416 with a proper initialization value.
Thanks @fosJoddie for the tip on the code and @Kneckter for raising this issue and getting some traces.

## Motivation and Context
Current initialization values can lead to an invalid value being inserted in DB, proper initialization range is 0,0 still I don't see why it should end with this range, so probably there are more issues down the line that lead to this error, it would be great if someone can point to a spawnpoint having this issue so we can check it properly.

## How Has This Been Tested?

Not tested

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1681,7 +1681,7 @@ class SpawnpointDetectionData(BaseModel):
         # If more than one disparate union, take the largest as our starting
         # point.
         union = reduce(lambda x, y: x if (x[1] - x[0]) % 3600 >
-                       (y[1] - y[0]) % 3600 else y, union, [0, 3600])
+                       (y[1] - y[0]) % 3600 else y, union, [0, 0])
         sp['latest_seen'] = union[1]
         sp['earliest_unseen'] = union[0]
         log.info('1x60: appear %d, despawn %d, duration: %d min.',


pogom/models.py
```diff
@@ -1681,7 +1681,7 @@ def classify(sp, scan_loc, now_secs, sighting=None):
         # If more than one disparate union, take the largest as our starting
         # point.
         union = reduce(lambda x, y: x if (x[1] - x[0]) % 3600 >
-                       (y[1] - y[0]) % 3600 else y, union, [0, 3600])
+                       (y[1] - y[0]) % 3600 else y, union, [0, 0])
```
sebastienvercammen: RE; Discord

Since the old code is meant to be the difference in time (between seconds after the hour), perhaps it's best to keep returning `[0, 3600]` and apply a `min(union[1], 3599)` so it never returns more than 3599 (yet still uses 3600 seconds in an hour to do its calculations).

Adding it here on Github so it can be discussed publicly.

sebastienvercammen:
PR closed in favor of #2416.

PR #2376, opened 2017-11-16
Accounts in Database

Alderon86:
## Description
This is a combination of an old Pr "F-HealthyAccountDB" (can be found on my repo) and Neskk's approach.
Added "on-the-fly" support for csv files. No instance restarting is needed, just change your .csv.

Parses given .csv file into database to use the defined number of workers via -w to start scanning.
Accounts are locked to instances.
Csv Parser got rewritten to also add the account level (usefull for l30's and above)
Removed couple of commands like -u and -p because its defined via csv.
Accounts removed from the csv also will get removed from db.
TODO's are marked properly.
## Motivation and Context

coming soon
## How Has This Been Tested?

locale instance
## Screenshots (if appropriate):
coming soon
## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

thanks @neskk for inspiration

PR #1819, opened 2017-02-13
Configurable Beehive

neskk:
This PR enables more control over how the beehive is configured, giving the ability to disable or force a worker count in specified hives.

## Description

Beehive mode is great because it saves a lot of memory when compared to using a separate instances to achieve this kind of map coverage. Currently we're unable customize the beehive worker distribution and this limits what we can do in our setups leading to some frustration and resources wasted when trying to setup a "beehive of beehives".
- `--beehive` is now an integer representing the number of leaps desired, e.g. `-bh 1` produces a beehive of 7 hexes (leap 1).
- `--workers_per_hive` is now used to force the population of each hive, meaning that if it's set to a value greater than zero it will throw an error if not enough workers are available to ensure that each active hive gets the desired amount of workers.
- `--beehive-workers` this is the new addition, it's a list of pairs `<hive index>-<worker count>`, e.g. `-bhw 1-0, 2-3, 3-0` will disable index 1 and 3 and force 3 workers in hive index 2.
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2424, opened 2018-01-06
Accounts in Database + Account Manager

neskk:
[**Alpha Stage**] - Probably still has some bugs to iron out.
Reworked pretty much every place where accounts are used and created a single entity responsible for handling all operations: an account manager.

## Description

Accounts are now saved in the database, their status is kept updated.
AccountManager handles all the queues and is responsible for maintaining account information updated.
When an error/captcha is encountered accounts will remain allocated to current instance - preserving old behavior.
Captcha code refactored into the account manager, hopefully we can have a generic way of handling captchas no matter which request is being done.

## Motivation and Context

Managing CSV files with accounts is kinda boring.
Account handling code was scattered across the project, multiple queues and variables were required, now it should be much simpler to control accounts, allowing more complex logic.

## TODOs:
- Shadow ban detection and handling.
- Provide a configuration parameter to control how account rotation works - currently hard-coded to maximize account reusage.
- Web endpoint to display accounts in the database and provide administration interface:
  - Insert accounts (import accounts from a CSV through web-browser)
  - Export accounts to CSV file.
  - Clear all accounts from the database.
- Configurable parameter to allow captchas and failed accounts to be immediately released from instance - they won't remain allocated so they won't show up on status printer, but can allow to have a dedicated instance just to solve captchas.
- Stuff like `device_info`, `proxy_url` and perhaps `asset_download_status` (future work - cache downloaded assets in json) could be kept in the database to maximize the consistency of usage.
- Add captcha handling in a generic way to all requests. (optional)

## How Has This Been Tested?

Local environment using SQLite.

## Types of changes

## Checklist:


sebastienvercammen:
Pending merge of #2376.

pogom/models.py
```diff
@@ -2345,28 +2374,41 @@ def encounter_pokemon(args, pokemon, account, api, account_sets, status,
...
             log.warning('Expected account of level 30 or higher, ' +
                         'but account %s is only level %d',
-                        hlvl_account['username'], encounter_level)
+                        hlvl_account['username'], hlvl_account['level'])
+            account_manager.failed_account(hlvl_account, 'not hlvl')
             return False
 
+        # Make a GMO request before the encounter.
+        if using_db_account:
+            response = gmo(hlvl_api, hlvl_account, scan_location)
```
friscoMad: Don't do this, at least not without a flag, currently ppl is tping accounts over the place and it works for the encounter, adding a gmo will take more time to do the encounter and will not give any interesting info.

runserver.py
```diff
@@ -403,14 +430,19 @@ def main():
...
                   heartbeat, db_updates_queue, wh_updates_queue)
 
         log.debug('Starting a %s search thread', args.scheduler)
         search_thread = Thread(target=search_overseer_thread,
                                name='search-overseer', args=argset)
         search_thread.daemon = True
         search_thread.start()
+    else:
+        # Check if we are able to solve captchas.
+        if args.captcha_solving and not args.hash_key:
```
friscoMad: OS can do captcha solving now?

pogom/accountManager.py
```diff
@@ -0,0 +1,722 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class AccountManager(object):
+    def __init__(self, args, db_queue, wh_queue, high_level=30):
+
+        self.args = args
+        self.dbq = db_queue
+        self.whq = wh_queue
```
friscoMad: It seems it is not used

pogom/accountManager.py
```diff
@@ -0,0 +1,722 @@
...
+from .pgoapiwrapper import PGoApiWrapper
+
+
+log = logging.getLogger(__name__)
+
+
+class AccountManager(object):
+    def __init__(self, args, db_queue, wh_queue, high_level=30):
+
+        self.args = args
```
friscoMad: Avoid using args, expand the needed parameters

pogom/accountManager.py
```diff
@@ -0,0 +1,722 @@
...
+                              name='account-manager')
+        self.manager.daemon = True
+        self.manager.start()
+
+    def run_manager(self):
+        self.thread_id = 0
+        cycle = 0
+        time.sleep(60)
+        while True:
+            # Run once every 15 seconds.
```
friscoMad: This will not run every 15 secs, it will run every 11m 15 sec if the cycle thing is ok

friscoMad:
Some thoughts as I haven't done a proper review:
 * Workerstatus should be merged with accounts or at least deduplicate data between the 2
 * Captcha tokens should be stored in account table
 * I prefer to keep Captcha handling in a single module, make it a service as the account manager and make the manager talk with the captcha service but I will keep the account manager to just manage accounts.

pogom/models.py
neskk: Last night @Alderon86 explained this to me... I thought it would be a red flag for accounts to encounter mons without even doing a GMO before. I would have thought that it was a blatant misbehavior that would be picked quickly by the game.
I will revert this change or at least add an option to enable it by choice.

runserver.py
neskk: Hummm, this is still a WIP. I'm thinking that it should be possible to have an `--only-server` dedicated to captcha solving that would pick up captcha'd accounts from the database and would just listen for tokens (manual captcha solving). Right now this check makes little sense because I  disabled AccountManager when launching in `--only-server`.
I think that having a flag that tells AccountManager to immediately release captcha'd accounts (if automatic captcha is not enabled) and then allow `-os` instances that have `--captcha-solving` enabled to pick up these accounts into their captcha queue. We would have just a single instance querying the database for tokens and that its `--hostname` could be set as the `--manual-captcha-domain`. This only makes sense in multi instance setups but it's a bit hard to provide so many configuration/behavioral options in a single software.

pogom/accountManager.py
neskk: It is used often to push updates to the Account model and to be available to captcha solvers.
@friscoMad I've seen your PR #2422 and I was thinking that maybe Account manager could hold the `key_scheduler` instead of being defined as a global variable in apiRequests. I was a bit tired of having to pass db/wh update queue to every function. `args` is also for convenience, there's some spots where I managed to avoid passing the variable to the function by having this local reference in the AccountManager object. This way, if you pass the `account_manager` around you will have everything you need. Originally I had put AccountManager in models.py and took advantage of having `args` declared as a global variable. When I moved it outside of models.py - because it's not really a database model, AM isn't saved in the database - I thought if I could avoid having a local reference to `args` and it's tremendously useful to have it, making the function calls a bit more concise and readable.

pogom/accountManager.py
neskk: As I've explained in the above comment, I don't feel it's that bad to save a reference... This probably has low impact on memory/performance and it's very useful to have it available. There are too many variables that are accessed from args, maybe if, down the road, the scope of `args` usage inside AccountManager is reduced to a half-a-dozen variables I would expand them all. For the time being, and if I'm not wrong, accessing a pointer to a pointer isn't an avoidable code pattern and does not compromise performance.

pogom/accountManager.py
neskk: Hum... you probably didn't read with enough attention, this is the same logic I've used in #2383, you've probably misunderstood how it works.
Sleep timer is **always** 15 seconds, some times it only runs the first bit (hence the "Run once every 15sec"), every 4 cycles it calls some more functions and then sleep 15 secs (hence the "Run once every minute" - 4 x 15 secs) and, finally, every 40 cycles ( 40 x 15 secs = 600 /60 = 10min) it will call account monitoring routine.

neskk:
@friscoMad `WorkerStatus` and `MainWorker` could be reworked yes, but then the scope of this PR would increase greatly and we would have to rework the interface. I'm not against this, I'm just saying that it will affect even more things all at once and would require even more work. Since WorkerStatus has some duplicate information about Account we could fetch that information from the Account table. But there's a tradeoff... there's one more "foreign key" and the query for WorkerStatus becomes more complex. WorkerStatus table records never grow too much, they're cleaned once they reach 30min without being updated and Accounts aren't updated nearly as often they do. I decided to keep them as they are for the time being, until we really decide how we're handling this.
I can see Captcha as being a separate module but since AccountManager had all the required "things" around I went for the easiest route. Maybe not it's easier to refactor it back to a separate module that has a reference to AccountManager to interact with its queues.
Finally, Captcha Tokens being saved in Account table makes little sense for me, it would be a VARCHAR row that would be null all over. I know that nulls don't occupy space but I don't see why this a good reason to do this.
Thank you @friscoMad for taking time to look at this PR, I always appreciate constructive critics and feedback 👍

pogom/accountManager.py
friscoMad: Yup, my bad it works but it is a bit convoluted, use something like this, it is a lot clearer:
```python
counter = 0
While true:
  counter += 1
  captcha_manager()

  if counter % 4 == 0:
    account_keeper()
    account_recycler()

  if counter % 40 == 0:
    monitor()
    counter = 0

  time.sleep(15)
```

FrostTheFox:
I know some of this has been stated over Discord, but I want to make it clear going forward.

You have already been removed from the #pr channel in Discord twice for issues with our processes, decisions and methods of collaboration. I previously told you (#1819) to remain professional on GitHub or you will be blocked from the repository. You had not said anything else unprofessional on GitHub, so I did not have any problem allowing you to open PRs and work on the GitHub repository.

Fast forward to now, the plan was for #2376 to be merged first, as it had a tested and proven database model, and one that we had been discussing inside #pr for a little while. You would rebase any work on behavior off of said PR, that way, we have the database model in place, solidified and tested, and we can build the behavior around it, and test it separately. You were *told* this by Seb.

Instead, what was done, was this PR was opened in direct competition of #2376, completely disregarding the original plan. This is completely against what we aim to do here at RM, not only in terms of behavior on the repository, but our testing preferences as well. You recommended skipping #2376 because "their starting points are pretty much the same and I got account rotation to work correctly using the database. I'm not saying PR #2424 is absolutely ready and flawless but I think it's already more stable and reliable than this one, for instance". You previously mentioned that you had talked with Alderon about reworking some old code from Arengo. All this begs the question, why the hell, would you decide to open a PR in competition with #2376 instead of collaborating with Alderon on it? If they really are of the same starting points, it should have been no problem?

In no particular order, a few replies:

"Nice way of saying thanks to me for my work because it seems really not important to you guys."
This is the type of attitude that we don't tolerate in RM. Contributions are highly appreciated, but when they're done in a way that causes drama and trouble for the rest of the team, and then complaining we don't appreciate your work when we condone the drama is highly counter productive and doesn't make us very happy. It often takes time where we could be working on actual code.

"Plus, collaboration doesn't not have it's credits properly accounted for as I told you last week when you merge spawn-spawn rework which had 90% of the code done and debugged by me."
We don't work on this project to have our names on the front of the commits page. Issues with credit for code should be taken up with GitHub. It's not our fault that the opener of the PR is the name displayed. Also, we have our team of reviewers and devs, all of whom are tasked with reviewing PRs and approving them. Their approvals are what determines the mergability of PRs, does their names get put on every PR that they approve? Nope.

"Sometimes I wonder why I keep spending my time on this project since all you do @sebastienvercammen is hold a grudge towards me... but it's fine, I forgive you :)"
I originally was the one who in my discussions with Seb was partially responsible for giving you the 2nd chance in PR, I originally talked to you about the issue in Discord and tried to level with you. I was the one who warned you to stop taunting @sebastienvercammen on the GitHub. This behavior of calling someone specific on "having a grudge" is an attempt at deflecting blame and accountability for behaviors of an individual. I gladly accept criticism where it is due, and I often receive it outside and inside of the context of RM, and I don't think I have ever not noted the criticism. Even if I felt the person was just trying to troll me or something. Constructive criticism is valid regardless of the context and subject, whether it be on code or behavior. My point here is, that, because someone is telling you that your attitude is not going to be tolerated, and then you don't change that attitude, doesn't make them have a "grudge" against you, especially when their decisions are made with in coordination and with approval of other members of the team.

You claim that you talked to Alderon for a bit, and it "immediately became clear that it would be easier just to go to 2424 and I think Alderon realized that as well". If this were the case, why would Alderon have not come to us and expressed his desire to continue with 2424 and abandon 2376? (I'm not blaming Alderon here, I'm just saying if this were truly the case and you were in collaboration with Alerdon, then that is what should have happened).

Finally, you seem to be confused as to why we are preferring to separate things in the first place. It's a very simple guideline and it's one we've tried to follow for a while now. When you stare at a 500 line change PR, which say adds two new features. It's a large blob of changes to look at and test, especially when you mix multiple features in with it. But, if you split it into a couple PRs, each for the specific feature, it means that each PR's code can be reviewed separately, and then tested separately. It keeps the procedure more smooth when we have our review team, as we can simply hand them a list of PRs (which, in a perfect world, would each be a single feature), and if there's an issue with feature A, they can just say "PR #x has x issue", and it can be commented on that PR, meanwhile, feature B (PR 2) might be perfectly fine and can be already merged into the main repository while A (PR 1) is being patched. Where as, one PR for both A and B, means reviewers need to say "Feature A in PR #x is broke", and that delays the entirety of the PR until the issue is fixed. This is not a super unique standpoint held in RM, it's standard project management.

And finally, the driving point behind all of this wall of text, is that you have received multiple 2nd chances, and warnings to change your behavior. Apparently, they are not being taken as they should be. Therefore, I am here to extend to you your final warning. This warning is NOT from Seb, this warning is not from anyone else on the RM team. This warning is from myself. Stop causing drama, stop causing issues, stop exhibiting anti-collaborative behavior, or I will not hesitate to block you from collaborating on the repository. We normally allow anyone to collaborate, but continued incidents is going to result in blocks.

This is your choice. If you want to continue to help and collaborate with us, change your attitude, and work with the other members, not against them. I'm not going to allow more of the time of our team to be wasted on any issues like this, so should they arise in the future, immediate action will be taken.

neskk:
@FrostTheFox 
- I didn't put my PR here in competition with #2376. Everyone is spinning this as me trying to overshadow that work when it's not true.
- I certainly didn't want any of this drama and was unaware of it since I'm out of PR.
- I understand that it's best to split functionality between PRs but some stuff makes sense together.
- I'm not doing this for bragging rights.
- I also don't want to waste more time with this discussion, I didn't start it, I have explained my reasons for posting a new PR and if I went against any rule I would accept any decision you guys made and moved forward immediately.
- Everyone jumps the gun to attack me, sometimes without reason but ok... I'll get over it.

FrostTheFox:
* I use the word competition because the PRs overlap in scope.
* Which is great but it doesn't discount that it happened.
* Accounts in database and an account manager are two different things which should be separate, imo. If you don't feel that way, well, that's just how you feel I guess? I'll admit I haven't looked at the code intricately, but I'm sure that accounts in the database as basic functionality is certainly within reach without a lot of account manager work. Anyway, that's not the real point, the point isn't necessarily about the splitting, it's about us making our plan known, and having decided that's what we want to do and then it being subsequently ignored. 
* So why bring it up in the first place?
* I don't either, so this is going to be my last reply.

My reply was to explain why, as of now, I feel it is necessary to give you a final warning. I could have just left you the the last sentence, but I prefer to give context. Therefore, I am done here and any further comments will be related to code.

neskk:
Perhaps if everyone looked more at the code this whole discussion would be avoided.
#2376 was far from "well tested" and fully functional, those claims are untrue. People have the right to disagree with me, I said that it made much more sense to put the account manager instead of some hacked functions glued together just to preserve old structures.
I did bring up the PR credits because, this time (unlike spawn_scan), I would like my efforts to be properly accounted for. I know it's an issue with GitHub, I would have never brought it up in public, I was talking privately with Seb and he published those messages out of context: last week I told him that Contributors tab didn't track the work I did in spawn_scan. Still... I would never refuse to collaborate because of this and, at this point, I really don't care if you believe me or not.
My branch initially moved passed the point where #2376 was very fast, and as far as I know it's not a crime not wanting to fix somebody else's PR after I had already done that work. I haven't got much time now, I'm sickly tired of this subject, I didn't ask for this drama, I simply submitted a PR, explained the differences, similarities and reason for doing so. If it was so wrong, you guys had the power to close it, but you can't force me to do something and then tell everyone that I'm not being "collaborative" because it's easy to say that while in your shoes.

pogom/utils.py
```diff
@@ -973,6 +890,29 @@ def generate_device_info(identifier):
     return device_info
 
 
+def generate_instance_id(args):
```
KartulUdus: I believe using process id breaks builds in systemd.
Restarting rocketmap.service doesn't change it and I got stuck in "duplicate instance detected" until I manually cleared

neskk: That message tells you that an instance with that same `instance_id` was running 30 seconds ago.
1. If you're **restarting** the instance, allow 30 seconds to cool-off between shutdown and startup and you won't get that message.
2. Maybe they are indeed conflicting instances in which case you shouldn't be manually clearing them up. `generate_instance_id` should return a different `instance_id` for each one of your running instances. It hashes location, worker count, status name and several other configuration parameters to make sure we get an unique value.

Either you didn't wait the 30 secs between shutdown > launch or you're trying to launch an instance in the same location, with same step size, worker count and scanning parameters. 
**Status name is ignored in `generate_instance_id` if it's not set**. Only after generating `instance_id` will it be updated to match the process ID. This is done so that you don't get a different `instance_id` each time you launch your instance without a defined `status-name`.

neskk:
I'm re-writing account allocation, current method is imperfect and I think this will solve some weird issues.
Should be pushed in a couple of hours.

pogom/utils.py
KartulUdus: tyvm 

to anyone running this on systemd with 
restart=always, should now add a:
`RestartSec=30`

pogom/account.py
```diff
@@ -29,6 +29,14 @@ class LoginSequenceFail(Exception):
     pass
 
 
+# Keep track of different types of bans using a single field.
+class AccountBanned:
```
friscoMad: Use Enum for this

pogom/account.py
```diff
@@ -108,12 +118,19 @@ def check_login(args, account, api, proxy_url):
...
     if num_tries > args.login_retries:
-        log.error(
-            ('Failed to login to Pokemon Go with account %s in ' +
-             '%d tries. Giving up.'),
-            account['username'], num_tries)
-        raise TooManyLoginAttempts('Exceeded login attempts.')
+        log.error('Failed %d times to login to Pokemon Go with account %s.',
+                  num_tries, account['username'])
+        account['login_attempts'] = login_attempts + 1
+        if account['login_attempts'] < 5:
```
friscoMad: >

pogom/account.py
```diff
@@ -484,31 +504,34 @@ def pokestop_spinnable(fort, step_location):
     return in_range and not pause_needed
 
 
-def spin_pokestop(api, account, args, fort, step_location):
+def spin_pokestop(args, account_manager, status, api, account, fort):
     if not can_spin(account, args.account_max_spins):
```
friscoMad: Args is not used

friscoMad:
I still have not done a full review but I have some ideas:
For simplicity I would try to aggregate all in the account instance (status, parent_manager, api, proxy...) we are just passing everything around together so it makes sense to keep it together, all is account information.
Account Manager could be just a module or a singleton that you could import and use anywhere but I am not sure what's the best way to do that in the most pythonic way.

I see you did a lot of work in captcha handling, still captcha should be handled in apiRequest.py as it can appear at any time (captchaed accounts get the message during the login process for example) and after a captcha all requests should be retried (if the captcha was solved), that may need more changes over your changes, I am not saying that you should add that right now in the PR but take into account if possible to make the evolution as easy as possible.

neskk:
@friscoMad I totally agree with you in what you said, I tried to leave room for changes in the future and I think account manager can accommodate what you're thinking.
1. AccountManger - Singleton that handles all account data, it can be done I avoided putting more things inside for now so that this PR doesn't grow even more. I like this idea and may execute it in the future.
2. Captcha rework - I tried to create a generic method to handle captchas... I agree with you, captcha detection should go inside apiRequests but then apiRequests would have to "know" about AccountManager. This way I keep changes to apiRequests to a minimum. I'm happy with current solution, it works fine but it can still be improved.

pogom/account.py
neskk: I feel like Enum is a bit of an overkill for this, I've thought about this and feel it's unnecessary for the time being, maybe later if we have a more generic way of managing bans we change this.

pogom/account.py
neskk: Fixed ;)

tobias86aa:
hlvl-workers will the system detect if they are shadow banned or banned? Will the account get mark as banned or shadowbanned?

neskk:
@tokar86a currently, **only shadow-ban is not verified for hlvl-workers**. This means, accounts will be marked as temporary/permanent ban but never shadow-ban.
Since *normal* Shadow-Ban detection requires a get_map_objects request to check for rare mons, this check can't be done with hlvl-workers because the don't make this request (GMO).

neskk:
I understand it's hard to test and read the changes when there's so much stuff.
I've been testing this for a while, the recent changes (encounters retry, etc.) introduced a couple errors but I think it's all working properly now.
I don't expect this to be completely bug free yet but I've tested most part of things. Until next week I won't probably work on this unless it's a bugfix. I appreciate any comments/feedback that you guys might have, I feel this is almost finished 💯

pogom/utils.py
neskk: Adjusted this cool-down timer to 10 secs.

pogom/account.py
neskk: Enum doesn't "let" me do things like `if account['banned'] <= AccountBanned.Shadowban` which are useful. I tried to change the code and all the database queries would have to use `AccountBanned.Clear.value` which is also annoying. For the time being and, as I discussed with Seb some time ago, I think I will let this stay a simple class which is more flexible than Enum.

neskk:
Updated Account model, for those of you who are already on this database schema version, you can update schema manually with:
```
ALTER TABLE `pogomap`.`account` 
CHANGE COLUMN `fail` `failed` SMALLINT(6) NOT NULL DEFAULT 0 ;
```
This will preserve current table information and put your database ready for this latest commit.

pogom/models.py
```diff
@@ -40,7 +42,7 @@
 flaskDb = FlaskDB()
 cache = TTLCache(maxsize=100, ttl=60 * 5)
 
-db_schema_version = 22
+db_schema_version = 23
```
sebastienvercammen: Latest RM is `23`, this should be `24`.

Please note: #2461 will also increase to 24 and it's a smaller PR so might be merged sooner. Keep this in mind.

pogom/accountManager.py
```diff
@@ -0,0 +1,941 @@
...
+        elif not self.args.hlvl_scan:
+            # Allow high-level accounts to be allocated for scanning.
+            conditions &= (Account.level < self.high_level)
+
+        try:
+            with Account.database().execution_context():
+                query = (Account
+                         .select()
+                         .where(conditions)
+                         .order_by(Account.last_modified.desc())
```
fosJoddie: I'm not sure if it is the order by last_modified or not, but in my test instance the logic will consistently take level 0 accounts from the db before reusing those accounts that are level 1 or more.  

The end result is that every restart of instances are taking in more new accounts before it has to

image

Screnshot shows how the manager picked 4 old accounts and one new, when there was clearly some still good older accounts available. The level 1 account was fetched and tutorialed right after the instance started.

I suggest adding an order by level where it selects anything about 0 before 0 level.

neskk: There are a couple of details we need to consider. I'm confident that current logic/query works properly, what I think happened in your setup is that another instance, or perhaps that same instance using -asi rotated accounts and initialized that level 1 account. If an account is picked and used it will have its `last_modified` updated and that what counts. From my understanding even if you used an account just for a couple of hours it will be flagged too and eventually banned so sorting by level won't help much.

billyjbryant:
This is going to need to be rebased soon as #2533 #2528 and #2537 will all change the `db_schema_version`

PR #2427, opened 2018-01-07
Add location marker + mobile Pokémon chunk processing controls in custom.js.

### LINKED ISSUE #2425: Location marker can be moved, possibly messing up map experience.
Problem comes from our help channel on Discord.

## Expected Behavior
Gmaps location marker should be static (if `--no-fixed-location` is disabled).

## Current Behavior
It can be moved. Even though it doesn't affect the scanner, it can cause issues (especially on mobile) where the marker gets moved off screen and the map pans towards it.

## Possible Solution
Allow marker to be static, or even disabled.

sebastienvercammen:
## Description
* Added options to show/hide the scan location and visitor location markers, and to make it (un-)draggable in Store.
* Added options to custom.js to control the Store settings for the markers.
* Updated comments to be consistent, and not to refer to "same as desktop".
* Added options to custom.js to allow Pokémon processing chunk size and interval for mobile devices.
* Made sure that code referencing locationMarker and searchMarker only runs when there is one.

## Motivation and Context
Fixes #2425.

## How Has This Been Tested?
Untested.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -169,9 +169,7 @@ class Pogom(Flask):
         if args.on_demand_timeout > 0:
             self.control_flags['on_demand'].clear()
 
-        search_display = True if (args.search_control and
-                                  args.on_demand_timeout <= 0) else False
-
+        search_display = (args.search_control and args.on_demand_timeout <= 0)
         scan_display = False if (args.only_server or args.fixed_location or
                                  args.spawnpoint_scanning) else True
 
--- a/static/js/custom.js.example
+++ b/static/js/custom.js.example
@@ -1,14 +1,13 @@
 $(function () {
     'use strict'
 
-    // Marker cluster might have loaded before custom.js.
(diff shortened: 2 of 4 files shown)


friscoMad:
Can't we just tie no-fixed-location to location being draggable?
I know it is disabled by default as no-fixed-location but needing to create the custom.js just to enable it seems a bit weird.
Also I am not sure if this should be stored client side, maybe the show location marker option but being draggable or not seems a map owner decission.

sebastienvercammen:
@friscoMad 

* If `--no-fixed-location` is enabled, some search controls will be shown in the sidebar. It also enables the back-end to change scanning location if it receives the appropriate POST request. Whether a marker should be visible and movable or not can be picked separately from this behavior with the current commit.
* Location marker is by design front-end (client-side), it's added by the client's JS. Even if the user decides to change the Store's settings via his console, it won't have any real effect.
* If "Follow my location" is enabled, the marker will never be draggable. If "follow me" gets disabled again, it'll fall back to the custom setting.

PR #2430, opened 2018-01-08
Add warning on -novc

FrostTheFox:
Add a disclaimer that bad things can happen in -novc and that you won't receive support.
## Description

tl;dr log.warning('bad stuff can happen yo, no support 4 u')
## Motivation and Context

Make the dangers of -novc known
## How Has This Been Tested?

not tested, but i mean, c'mon, it's a log.warning.
n/a
## Types of changes

## Checklist:

### DIFF
--- a/runserver.py
+++ b/runserver.py
@@ -280,8 +280,8 @@ def main():
         if not args.only_server:
             log.warning('You are running RocketMap in No Version Check mode. '
                         'If you don\'t know what you\'re doing, this mode '
-                        'can have consequences, and you will not receive support '
-                        'running in NoVC mode. You have been warned.')
+                        'can have consequences, and you will not receive '
+                        'support running in NoVC mode. You have been warned.')
 
     position = extract_coordinates(args.location)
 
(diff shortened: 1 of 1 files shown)


runserver.py
```diff
@@ -252,6 +252,12 @@ def main():
 
     set_log_and_verbosity(log)
 
+    if args.no_version_check:
```
friscoMad: Just warn on scanner instances not need to warn on -os instances, move it to can_start_scanning

FrostTheFox: I purposefully didn't put it in can_start_scanning because I was trying to keep that for checking stuff that would cause a sys.exit.

FrostTheFox:
(2nd travis fail is because travis is having network issues)

runserver.py
```diff
@@ -276,6 +276,13 @@ def main():
     if not args.no_server and not validate_assets(args):
         sys.exit(1)
 
+    if args.no_version_check:
```
pogo-excalibur: Could these be combined?

PR #2431, opened 2018-01-09
Update speed-scheduler.md

PhoenixHugs:
Add note about not using Speed Scheduling for Gym only scans

## Description
Add a small note warning users to not use the speed scheduling option unless they are scanning for pokemon as the speed scheduler only starts scans where pokemon are spawning so if disabled stops scanning after the 1st scan.

## Motivation and Context
To help other users avoid the lengthy process of figuring out why there gym only scanners aren't working as expected.

## How Has This Been Tested?
N/A

N/A

## Types of changes
Change to documentation only for Speed Scheduling

## Checklist:

- [ X ] My change requires a change to the documentation.
- [ X ] I have updated the documentation accordingly.

### DIFF
--- a/docs/scanning-method/speed-scheduler.md
+++ b/docs/scanning-method/speed-scheduler.md
@@ -14,6 +14,8 @@ Speed Scheduler is an alternative scheduler to Hex Scan or Spawnpoint Scan with 
 
 To use Speed Scheduler, always put -speed in the command line or set `speed-scan` in your config file.
 
+Please note: Speed Scheduler does not work for gym only scanning.
+
 ## Commands and configs
 
 What command line args should I use for Speed Scheduler?
--- a/docs/scanning-method/speed-scheduler.md
+++ b/docs/scanning-method/speed-scheduler.md
@@ -14,7 +14,7 @@ Speed Scheduler is an alternative scheduler to Hex Scan or Spawnpoint Scan with 
 
 To use Speed Scheduler, always put -speed in the command line or set `speed-scan` in your config file.
 
-Speed Scheduler is optomized for scanning for Pokemon so it does not work well for gym/pokestop only scanning. If you are interested in scanning for just gyms/pokestops consider using Hex Scheduler.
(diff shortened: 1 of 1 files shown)


PhoenixHugs:
I'm asking for this change to be added so that other users don't make the same mistake that I did and add the -speed option to a gym only instance.  I'm setting up my first map and decided to scale back and do gyms only until I got a better grasp at creating my own PTC accounts.  When I scaled back to gyms only, my scanner basically quit working.  When using speed scheduling with a gym only map, the workers will do one initial scan then more or less stop looking for scans to perform.

pogo-excalibur:
Clarifying this is a good idea. 👍 

It would be great if you could fill in the PR form - these are useful for when we need to reference PRs.

PhoenixHugs:
PR Form updated.

docs/scanning-method/speed-scheduler.md
```diff
@@ -14,6 +14,8 @@ Speed Scheduler is an alternative scheduler to Hex Scan or Spawnpoint Scan with
 
 To use Speed Scheduler, always put -speed in the command line or set `speed-scan` in your config file.
 
+Please note: Speed Scheduler does not work for gym only scanning.
```
pogo-excalibur: It might be worthwhile elaborating slightly here:
'Speed Scheduler is optomized for scanning for Pokemon so it does not work well for gym/pokestop only scanning. If you are interested in scanning for just gyms/pokestops consider using [Hex Scheduler](LINK).'

docs/scanning-method/speed-scheduler.md
```diff
@@ -14,6 +14,8 @@ Speed Scheduler is an alternative scheduler to Hex Scan or Spawnpoint Scan with
 
 To use Speed Scheduler, always put -speed in the command line or set `speed-scan` in your config file.
 
+Speed Scheduler is optomized for scanning for Pokemon so it does not work well for gym/pokestop only scanning. If you are interested in scanning for just gyms/pokestops consider using Hex Scheduler.
```
sebastienvercammen: > optomized

😭 

also

> Pokéstop

PR #2436, opened 2018-01-12
Add captcha reporting to level-up script

friscoMad:
## Description
Add captcha reporting to level-up script

## Motivation and Context
It seems that it is possible to get captchas even with fresh accounts, at least we should report that failures

## How Has This Been Tested?

Tested with a batch of accounts that did not have any problem, so more testing is needed.

## Types of changes

## Checklist:

### DIFF
--- a/tools/levelup.py
+++ b/tools/levelup.py
@@ -28,11 +28,15 @@ class FakeQueue:
 
 def get_location_forts(api, account, location):
     response = gmo(api, account, location)
+    if len(response['responses']['CHECK_CHALLENGE'].challenge_url) > 1:
+        log.error('account: %s got captcha: %s', account['username'],
+                  response['responses']['CHECK_CHALLENGE'].challenge_url)
+        return ([], True)
     cells = response['responses']['GET_MAP_OBJECTS'].map_cells
     forts = []
     for i, cell in enumerate(cells):
         forts += cell.forts
-    return forts
+    return (forts, False)
 
 
 def level_up_account(args, location, accounts, errors):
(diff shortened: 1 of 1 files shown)


friscoMad:
I have done more tests this weekend, it still does not handle captchas received during login, so previously captchaed accounts are not handled but it shouldn't happen with fresh accounts.

tools/levelup.py
```diff
@@ -131,7 +139,8 @@ def level_up_account(args, location, accounts, errors):
 
 account_queue.join()
 log.info('Process finished')
-if len(errors) > 0:
-    log.info('Some accounts did not finish properly:')
-    for account in errors:
-        log.info('\t%s', account['username'])
+for error_type in errors:
+    if len(errors[error_type]) > 0:
+        log.info('Some accounts did not finish properly (%s):', error_type)
```
sebastienvercammen: Log as warning instead of info? Thoughts?

tools/levelup.py
```diff
@@ -28,11 +28,15 @@ def put(self, something):
...
+    if len(response['responses']['CHECK_CHALLENGE'].challenge_url) > 1:
+        log.error('account: %s got captcha: %s', account['username'],
+                  response['responses']['CHECK_CHALLENGE'].challenge_url)
+        return ([], True)
     cells = response['responses']['GET_MAP_OBJECTS'].map_cells
     forts = []
     for i, cell in enumerate(cells):
         forts += cell.forts
-    return forts
+    return (forts, False)
```
sebastienvercammen: Instead of `(forts, captcha)` as return value, maybe we can use a standard like Golang?

```
return (error, return)
```

where we can re-use `error` to be a captcha error, an unexpected response (timeout, banned, blocked by anticheat), and so on.

tools/levelup.py
```diff
@@ -77,7 +85,7 @@ def level_up_account(args, location, accounts, errors):
                 log.exception('Exception in worker: %s retrying.', e)
```
sebastienvercammen: > Exception in worker: %s. Retrying.

tools/levelup.py
```diff
@@ -64,7 +68,11 @@ def level_up_account(args, location, accounts, errors):
             check_login(args, account, api, status['proxy_url'])
             log.info('Account %s level %d', account['username'],
```
sebastienvercammen: > Account %s, level %d.

PR #2439, opened 2018-01-15
Sync the "Exclude Rarity" dropdown with the stored value on load

animanmaster:
## Description
This PR syncs the "Exclude Rarity" dropdown with the value stored locally upon load.    

## Motivation and Context

I noticed that coming back to the map in a separate session/window will show "None Selected" in the "Exclude Rarity" dropdown, even though a filter _is_ actually being applied. 

## How Has This Been Tested?

Tested with my local map with the following steps:

1. Make sure we are in the initial state (Make sure `Store.get('excludedRarity')` returns 0 and "None Selected" is chosen in the dropdown).
2. Change "Exclude Rarity" to something else ("Uncommon and below", for example).
3. Refresh the page.
4. Check the value shown in "Exclude Rarity"

Expected: Selection should be retained.  

With no rarity exclusion applied:

screenshot 2018-01-15 at 12 51 28 pm

After choosing "Uncommon and below":
screenshot 2018-01-15 at 12 51 56 pm

Upon refresh (before the change in this PR):
screenshot 2018-01-15 at 12 54 15 pm

Upon refresh (after the change in this PR):
screenshot 2018-01-15 at 2 01 26 pm

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2742,6 +2742,7 @@ $(function () {
 
         // recall saved lists
         $selectExclude.val(Store.get('remember_select_exclude')).trigger('change')
+        $selectExcludeRarity.val(Store.get('excludedRarity')).trigger('change')
         $selectPokemonNotify.val(Store.get('remember_select_notify')).trigger('change')
         $selectRarityNotify.val(Store.get('remember_select_rarity_notify')).trigger('change')
         $textPerfectionNotify.val(Store.get('remember_text_perfection_notify')).trigger('change')


static/js/map.js
```diff
@@ -2742,6 +2742,7 @@ $(function () {
 
         // recall saved lists
         $selectExclude.val(Store.get('remember_select_exclude')).trigger('change')
+        $selectExcludeRarity.val(Store.get('excludedRarity')).trigger('change');
```
Alderon86: Remove the `;` to make travis happy.
Otherwise, good catch!

animanmaster: Whoops, will do. Thanks! :D

PR #2440, opened 2018-01-16
Potential EX gym detection

KartulUdus:
The generally accepted theory on where EX raids appear is OSM data light green parks. 
This PR gives the ability to store which gyms can possibly spawn an EX raid within an area
## Description

* New column in gym table (boolean for park or not)
* New requirement for overpy (pip install -r requirements.txt --upgrade)
* New arg for populating Gym.park

Uses existing geofencing functions
Using a single area geofence is required 

Launch with 
`python runserver.py -exg -gf geofences/yourgeofence.txt`
Parks and other applicable green areas will be pulled within the bounding box of your geofence.
Existing gyms within the geofence will be compared against areas
Will update Gym.park=True if the gym is inside an area

## Motivation and Context

My discord was making CSV files of gyms manually and using tools like overpass-turbo to generate maps with the data.
The query was easy enough to implement with overpy to map directly
## How Has This Been Tested?

Hardly at all, I mostly photoshopped the examples below.
screen shot 2018-01-16 at 03 09 25

The switch:
screen shot 2018-01-19 at 00 08 00

Switch when turned off or on:
screen shot 2018-01-19 at 00 27 58
screen shot 2018-01-18 at 23 59 35

## Types of changes

## Checklist:

- [x] I have made the thing work with the front-end


ghost:
```
2018-01-15 20:34:32,979 [        MainThread][           osm][    INFO] Finding border points from geofence
2018-01-15 20:34:32,979 [        MainThread][           osm][    INFO] Finding parks within zone
2018-01-15 20:34:51,293 [        MainThread][     runserver][   ERROR] Uncaught exception
  File "runserver.py", line 531, in <module>
    main()
  File "runserver.py", line 268, in main
    exgyms(args.geofence_file)
  File "/home/RocketMap/pogom/osm.py", line 62, in exgyms
    gyms = Gym.get_gyms(south, west, north, east)
  File "/home/RocketMap/pogom/models.py", line 499, in get_gyms
    for g in results:
  File "/usr/local/lib/python2.7/dist-packages/peewee.py", line 2967, in __iter__
    return iter(self.execute())
  File "/usr/local/lib/python2.7/dist-packages/peewee.py", line 2960, in execute
    self._qr = ResultWrapper(model_class, self._execute(), query_meta)
  File "/usr/local/lib/python2.7/dist-packages/peewee.py", line 2656, in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "/usr/local/lib/python2.7/dist-packages/playhouse/shortcuts.py", line 208, in execute_sql
    cursor.execute(sql, params or ())
  File "/usr/lib/python2.7/dist-packages/MySQLdb/cursors.py", line 226, in execute
    self.errorhandler(self, exc, value)
  File "/usr/lib/python2.7/dist-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorvalue
OperationalError: (1054, "Unknown column 't1.park' in 'field list'")
```

KartulUdus:
there's a database migration, run normally first

ghost:
Whoops okay I upgraded. Now i'm getting this error:
```
2018-01-15 20:38:33,709 [        MainThread][           osm][    INFO] Finding border points from geofence
2018-01-15 20:38:33,709 [        MainThread][           osm][    INFO] Finding parks within zone
2018-01-15 20:38:45,531 [        MainThread][     runserver][   ERROR] Uncaught exception
  File "runserver.py", line 531, in <module>
    main()
  File "runserver.py", line 268, in main
    exgyms(args.geofence_file)
  File "/home/RocketMap/pogom/osm.py", line 62, in exgyms
    gyms = Gym.get_gyms(south, west, north, east)
  File "/home/RocketMap/pogom/models.py", line 528, in get_gyms
    p['pokemon_name'] = get_pokemon_name(p['pokemon_id'])
  File "/home/RocketMap/pogom/utils.py", line 872, in get_pokemon_name
    return i8ln(get_pokemon_data(pokemon_id)['name'])
  File "/home/RocketMap/pogom/utils.py", line 862, in get_pokemon_data
    args.root_path,
AttributeError: 'Namespace' object has no attribute 'root_path'
```

ghost:
I'm noticing that gyms are being set back to `park=0` after they've been set as `park=1` in the DB. I ran `select name, park from gym join gymdetails on gym.gym_id=gymdetails.gym_id where park=1;` and got 40 in the first set, and then 15 the next time I ran it.

**Edit:** When running the command in the terminal, it spits out about 60 eligible gyms, but saves 4 of them as `park=1` to the DB.

pogom/osm.py
```diff
@@ -0,0 +1,79 @@
...
+                         'lon': float(node.lon)})
+
+        for gym in gyms.items():
+
+            gympoint = {'lat': float(gym[1]['latitude']),
+                        'lon': float(gym[1]['longitude'])}
+            # Check if gyms falls within a designated park and update if park
+            if Geofences.is_point_in_polygon_custom(gympoint, data):
+                gymname = Gym.get_gym(gym[0])['name'].encode('utf8')
+                log.info('gym:{} could spawn legendary raid'.format(gymname))
```
blakewenloe: Just a nitpick, but they can all technically spawn legendary raids. I'd change this to `could spawn ex raid` or `eligible for ex raid`.

KartulUdus:
Interesting, then it seems that it's getting overridden with default once scanned. 
I'll take a look later today

ghost:
I may be completely off-center, but do you think that, after running your command, the next time those gyms are scanned (which is pretty quick), they update the park back to default/0?

blakewenloe:
I may be completely off-center, but do you think that, after running your command, the next time those gyms are scanned (which is pretty quick), they update the park back to default/0?

ghost:
That's something I think I can test. I'll turn off all of my other instances and try it.

fosJoddie:
Interesting PR, I'll give this one a test when I can.

If you are scanning normally while this script is running, I'm pretty sure that the changes WILL be overwritten by the normal scanner upserting gyms. Not really a bug if that is the case. This tool should be ran while normal scanner is not running.

jorgitoh:
osm.py", line 77, in exgyms
    gymname = Gym.get_gym(gym[0])['name'].encode('utf8')
AttributeError: 'NoneType' object has no attribute 'encode'

I think not all gyms have "name". Maybe we need to run first with -gi to get all the data from gyms?.

ghost:
Confirming it works perfectly when I have the other instances (aside from `-os`) turned off.

SkOODaT:
its on a gym scan its setting it back to zero its defaulting the value back, im already starting to think how to stop this ..... the check is too much to do entirely every time..... prolly best thing to do is have it to the check every time it scans a gym for detailed infos singularly, you could drop the required geofence file that way too and check the gym based off its lat and long

Really like the idea tho yould be nice if we could add something similar to some other maps ive seen to bound off said parks on the front end, this is also nice for webhooks to send the data with gyms, eggs and raids

jorgitoh:
@SkOODaT Same problem. Any instance reset the value of park to 0. Any idea?.

ghost:
If workers don't actually pull the park info for the gyms outside of running this particular script, wouldn't it make sense to stop them from updating the parks column when running a normal instance?

SkOODaT:
yes but with the way data is gathered i dont think its possible to "stop" the update when the worker searches ... my gym is_in_battle stuff has this same issue, there for populating the gym table with a singular check uesing overpy to osm would make sense it would just update the table with the data everytime, i guess it might be a slight waste of resource tho, im no pro coder

this is out of someone elses repo but gyms could just be checks like this: 
        result = api.query("""
            [out:json];
            is_in({},{});
            area._[leisure="park"];
            out;
            """.format(gym['lat'],gym['lng']))

jorgitoh:
Im comparing the result with http://overpass-turbo.eu/ and current invitation for ex raids. For example. This code return false on http://www.openstreetmap.org/way/9939231

KartulUdus:
@SkOODaT rerunning it every time would be a quick way to get banned from OSM I believe.
I've been told that more than `leisure="park"` is considered a _park_  
@jorgitoh My sources claim that the EX raids are currently based on green parks from 2016 july, could you add that to your query to see if the park is there?

jorgitoh:
@KartulUdus its weird. Im using this
http://overpass-turbo.eu/

The query its similar to ur query:

[out:json][timeout:25];
// gather results
(
  // query part for: “leisure=park”
  node["leisure"="park"]({{bbox}});
  way["leisure"="park"]({{bbox}});
  relation["leisure"="park"]({{bbox}});
  // query part for: “leisure=playground”
  node["leisure"="playground"]({{bbox}});
  way["leisure"="playground"]({{bbox}});
  relation["leisure"="playground"]({{bbox}});
  // query part for: “leisure=recreation_ground”
  node["leisure"="recreation_ground"]({{bbox}});
  way["leisure"="recreation_ground"]({{bbox}});
  relation["leisure"="recreation_ground"]({{bbox}});
  // query part for: “landuse=recreation_ground”
  node["landuse"="recreation_ground"]({{bbox}});
  way["landuse"="recreation_ground"]({{bbox}});
  relation["landuse"="recreation_ground"]({{bbox}});
  // query part for: “landuse=grass”
  node["landuse"="grass"]({{bbox}});
  way["landuse"="grass"]({{bbox}});
  relation["landuse"="grass"]({{bbox}});
  // query part for: “landuse=meadow”
  node["landuse"="meadow"]({{bbox}});
  way["landuse"="meadow"]({{bbox}});
  relation["landuse"="meadow"]({{bbox}});
  // query part for: “landuse=greenfield”
  node["landuse"="greenfield"]({{bbox}});
  way["landuse"="greenfield"]({{bbox}});
  relation["landuse"="greenfield"]({{bbox}});
);
// print results
out body;
>;
out skel qt;

I have a few gyms in park with values in false.

Maybe the year. Im 100% about like 10 gyms with active ex this week.

SkOODaT:
I was just uesing the one as example the rest need to be added of course lol ... And are you shure it cause a ban ...if so then we need say a json to store the data in and call on it I'm pretty sure your not going to be able to stop the gym table from zeroing the data without data to replace it :)

KartulUdus:
@jorgitoh could you add `[date:"2016-07-10T00:00:00Z"]` to the top of your query and see if it was a park then?

I got an idea how to solve the returning to 0 but only an idea so far ;)

jorgitoh:
Yes, with ur the date the website http://overpass-turbo.eu/ shows the park correctly:

worksfine

Maybe an error in the code?

pogom/osm.py
```diff
@@ -0,0 +1,79 @@
...
+import json
+import logging
+from geofence import Geofences
+from models import Gym, init_database
+from flask import Flask
+from utils import get_args
+
+args = get_args
+
+app = Flask(__name__)
```
friscoMad: app can be None in this case not need to instance Flask

friscoMad:
Due to the current replace-into query in bulk updater there is no way to keep the park field in the db, hopefully the db insert change comming after devkat exclusive time can keep that info, otherwise we would need to change how we insert the gyms in db or do a query before any gym insert.

On the other hand as I said in a previous comment, I prefer to not add more params for tools that are only done once a external script seems a better fit. On the other hand we can make this to run every day or so to keep data updated and avoid the param.

sebastienvercammen:
@friscoMad 👍 That code is already in two-week testing, it'll be released on the 26th.

pogo-excalibur:
Would it not possibly have been easier to have done this as a front end change, using an overlay layer to indicate nearby parks (with the determination of EX-elegible gyms left to users)?

What advantage are you aiming for by having this done at the backend/db level?

pogom/osm.py
KartulUdus: You are indeed correct sir, changed now

KartulUdus:
* changed log message to `<Gym> is eligible for legendary raid`
 * removed Flask from osm.py
 * changed log message so that by default it would use gym_name, but if none is found (args.gym_info == false) it defaults back to gym_id
* up to date with current develop  
* applied forehead to hard surface for how to pass old Gym.parks value along the insert, alas to no avail

SkOODaT:
Here you all go i slaved away on some Sql :S lol
The Fix add this right above your gym webhooks in models.py

```
                with Gym.database().execution_context():
                    Query = Gym.select().where(Gym.gym_id == f.id).dicts()
                    park_id = None
                    for gym in list(Query):
                        park_id = gym['park']
                    log.debug(park_id)
```
and add 
```
                        'park':
                            park_id
```
onto the webhooks and database stuff  (your adding it 2 times)
it will preserve whatever is on the gym table 
all it does is grab the var we have collected with KartulUdus OSM stuff off the gym table before its reset and readds it lol 
if you put it before the webhooks and add it to webhooks too it will send the park tag out on gyms aswell will be useful for gym egg and raid alerts
This is also why its useful to do it right here in the database multiple things can use it 
just needed some time to think it over :D

SkOODaT:
oh also remove 
`park = BooleanField()`
you dont need it its there 2 times just use the one with the default value

KartulUdus:
thats what I was _trying_ to do @SkOODaT 😄 , query it before making new insert and use old value. 

It is an extra query which then hopefully can be removed after the 26/01/2018 magic.
Still need a switch for the FE to only show these gyms, stay tuned

SkOODaT:
i have a feature suggestion too, now what we need is a way for it to just check on server startup every 30 days or something that shouldn't be too hard... some front end work would be nice as well some way for gyms to be distinguished maby we need a potiential ex raid gym icon or something 
throwing some ideas out there :)

~edit that would work a front end switch too, with the data in the database it could be added onto the gym tip too in some way

~edit 2 the query is pretty simple it looks for that specific gym then looks to see what data accompanys it i am uesing it with other things as well lol, i use lure deployer info on pokestops and i use a similar query

SkOODaT:
`'park': park_id `needs to be added onto eggs/raid webhooks too i missed that ..... and if you want it in the extended gym info webhooks entire thing needs to be but again down in the gym phaseing

SkOODaT:
You should also take A.R.T and I's Weather S2Cells and mod it to work with raid S2cells show them on the front end when your new switch is on, just a suggestion :)

KartulUdus:
will add to eggs/raid possibly today 

I think we shouldn't add too many switches tho. perhaps cell size could be a dropdown in upcoming weather pr #2420 ?

ghost:
I agree. Adding cell size here is probably cramming too many features into one PR.

SkOODaT:
well the only reason i say put it into OSM is because A.R.T has the weather stuff all separated into weather.js and weather.py (doesnt really go with it).... also the reason adding the ex raid s2cells to show and be beneficial would to be able to use that strategy where 2 are in each cell etc 

> the number of gyms that hold an EX Raid at one square is either 1 or 2 only.

showing the s2 cells for raids in combination with only showing ex raid potential gyms would allow end users to easily ID all the gyms that should get raided at to maximize potential ex raid passes for that square :)
~build it right into the switch toggle ... gym parks only = s2 cells for raids come on etc, 
this PR isnt really all that big yet a exraid.js and some s2 front end cell code wouldn't be alot 

just a suggestion tho :)

KartulUdus:
Someone from my discord discovered edge cases. 

The gym allegedly doesn't have to be Exactly within a park, but an s20 cell that touches a park area. 

For example a couple of edge cases I'm missing : 
image

perhaps we should include some rounding to allow such gyms if only X meters away from a park polygon

KartulUdus:
[x] raid and gym only switches 
[x] scanning doesn't overwrite park
[x]  "Close enough" (by corners of s2 lvl 20 cell) to geofence 
[x] weird behaviour where raid egg image disappears if `Park Raids Only` activated

jorgitoh:
I have a problem here:

map1osm
map2osm

Both gyms are not considered (or are not inside) with ur code.

darkelement1987:
You ran with -exg parameter?

jorgitoh:
@darkelement1987 yes

KartulUdus:
@jorgitoh 
I did a check on this area and got that 

Monumento a los Congresales de Tucumán - is not eligible for ex raid 
However, Fuente Deprecada & Glorieta are 

the monument, although in seemingly a park area, does not have any of it's s2 lvl 20 cell corners inside the fence 

sources: 
http://overpass-turbo.eu/
http://s2.sidewalklabs.com/regioncoverer/?center=-34.559197%2C-58.451587&zoom=20
screen shot 2018-01-23 at 09 01 29
screen shot 2018-01-23 at 09 02 00
screen shot 2018-01-23 at 09 02 03
screen shot 2018-01-23 at 09 02 40

jorgitoh:
Thank you. So all the s2 lvl 20 cell corners must be inside of the fence?.

Ur code doesnt return "Fuente Deprecada" as an elegible gym.

KartulUdus:
could be your geofence or not latest pull or gym not in database when running -exg
at least 1 cell corner needs to be inside a park for it to be counted as a park 

I ran it locally a few minutes before those screenshots and Fuente Deprecada & Glorieta were eligible in that park

sebastienvercammen:
Posted on Discord:

> [6:16 PM] p_brent: it was discovered that all EX raid are in level 20 cells which have the cell centre inside the park
> [6:16 PM] p_brent: not just border is contained in cell
> [6:23 PM] p_brent: so all calculations need to reference cell center instead of the gyms actual location

Example

Source: https://www.reddit.com/r/TheSilphRoad/comments/7t5vn1/new_discovery_for_nonsponsored_ex_raid_eligible/

KartulUdus:
did that.
Checking Tallinn has gone from 42s->60 minutes -> 5 minutes -> 75s  
now check like illustrated:
screen shot 2018-01-27 at 11 31 06
1 = center of s2 cell
2-5 = corners of s2 cell
6 = gym location

If point 1 is in park, it's saved as a park

waiting for _top secret_ database changes to touch up the rest

sebastienvercammen:
#2464 is now merged, meaning auto-increment IDs in the tables will now stay the same, since we no longer rely on peewee's `REPLACE INTO`.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "Gruntfile.js",
   "dependencies": {
     "babel-preset-env": "^1.6.0",
-    "eslint": "^3.1.1",
+    "eslint": "^3.19.0",
```
pogo-excalibur: Would you mind outlining the resaon for this?

pogom/models.py
```diff
@@ -622,6 +623,10 @@ def get_gym(id):
 
         return result
 
+    @staticmethod
+    def is_gym_park(id, park):
```
pogo-excalibur: I don't know if this is just the convention I'm used to, but I expect `is_*` methods to fetch a Boolean property. This seems to be setting a property, so I'd probably use something like `link_gym_to_park` or `set_gym_park`.

pogom/osm.py
```diff
@@ -0,0 +1,107 @@
...
+from geofence import Geofences
+from models import Gym, init_database
+from utils import get_args
+from s2sphere import LatLng, Cap, RegionCoverer
+
+args = get_args
+app = None
+log = logging.getLogger(__name__)
+db = init_database(app)
+earthCircumferenceMeters = 1000 * 40075.017
```
pogo-excalibur: Why perform a multiplication here?
And this should be a constant: `EARTH_CIRCUMFERENCE_METERS`.

pogom/osm.py
```diff
@@ -0,0 +1,107 @@
...
+    east = max(fence, key=lambda ev: ev['lon'])['lon']
+    log.info('Finding parks within zone')
+    ex_gyms = ex_query(south, west, north, east)
+
+    gyms = Gym.get_gyms(south, west, north, east)
+    log.info('Checking {} gyms against {} parks'.format(len(gyms),
+                                                        len(ex_gyms.ways)))
+
+    for gym in gyms.items():
+        gympoint = {'lat': float(gym[1]['latitude']),
```
pogo-excalibur: Not a big issue, but why use a dict here rather than two variables?
Consistency?

pogom/osm.py
```diff
@@ -0,0 +1,107 @@
...
+                    gymname = Gym.get_gym(gym[0])['name'].encode(
+                        'utf8')
+                except AttributeError:
+                    gymname = gym[0]
+                log.info('{} is eligible for EX raid'.format(gymname))
+                Gym.is_gym_park(gym[0], True)
+                break
+
+
+def get_s2_cell_center(lat, lng, radius, parent_level):
```
pogo-excalibur: This method seems remarkably complex for the task, particularly as `radius` seems to always be `0`.
Could you outline why getting the `CellId` of the `LatLng` object and then getting the cell's center isn't sufficient?

pogom/models.py
pogo-excalibur: It may also be worth discussing whether we should store *which* park (feature id) the gym is in.

pogom/osm.py
KartulUdus: I believe the geofencing function `def is_point_in_polygon_custom`
expects the point to have a 'lat' & 'lon' key

pogom/osm.py
KartulUdus: Capitalized now

pogom/models.py
KartulUdus: method renamed as recommended. 

unfortunately, not all parks return a 'name' for the way. would now know how to exactly tackle that

package.json
KartulUdus: whoopsie, must of done a `npm too many things` reverted now

pogom/osm.py
pogo-excalibur: Right. But `gympoint` doesn't get passed to `is_point_in_polygon_custom` - it only gets passed to `get_s2_cell_center`.

static/js/map.js
```diff
@@ -1797,12 +1802,26 @@ function processGym(i, item) {
         }
     }
 
+    if (Store.get('showParkGymsOnly')) {
+        if (item.park === false) {
```
pogo-excalibur: Consider `if (!item.park) {`.

static/js/map.js
```diff
@@ -1204,7 +1208,8 @@ function updateGymMarker(item, marker) {
             scaledSize: new google.maps.Size(48, 48)
         })
         marker.setZIndex(google.maps.Marker.MAX_ZINDEX + 1)
-    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') && raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
+    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') &&
```
pogo-excalibur: You'll need to use something like: `&& (!Store.get('showParkRaidsOnly') || item.park)`.
(untested code - so treat with caution)

And it'll also need to be added to the first `if` level at L1201.

pogo-excalibur: And I'd suggest adding it to the *end* of the current chain.

templates/map.html
```diff
@@ -152,6 +162,16 @@ <h3>Open Spot</h3>
                   </label>
                 </div>
               </div>
+              <div class="form-control switch-container" id="park-gyms-only-wrapper">
+                <h3>Gyms In Park</h3>
```
pogo-excalibur: Perhaps this could be something more informative like 'Only Show Gyms In Parks'?

darkelement1987: "Park Gyms Only" ?

ghost:
Just attempted to merge this and hit an error (as below), running latest version of RM. Looks as though it needs a default value setting for bool field (false).

2018-01-29 23:05:21,290 [        MainThread][     runserver][   ERROR] Uncaught exception
  File "runserver.py", line 579, in <module>
    main()
  File "runserver.py", line 330, in main
    db = startup_db(app, args.clear_db)
  File "runserver.py", line 216, in startup_db
    verify_database_schema(db)
  File "C:\xampp\htdocs\RocketMap\pogom\models.py", line 2920, in verify_database_schema
    if not database_migrate(db, db_ver):
  File "C:\xampp\htdocs\RocketMap\pogom\models.py", line 3262, in database_migrate
    migrator.add_column('gym', 'park', BooleanField()))
  File "C:\Python27\lib\site-packages\playhouse\migrate.py", line 684, in migrate
    operation.run()
  File "C:\Python27\lib\site-packages\playhouse\migrate.py", line 144, in run
    getattr(self.migrator, self.method)(*self.args, **kwargs))
  File "C:\Python27\lib\site-packages\playhouse\migrate.py", line 152, in inner
    return fn(self, *args, **kwargs)
  File "C:\Python27\lib\site-packages\playhouse\migrate.py", line 222, in add_column
    raise ValueError('%s is not null but has no default' % column_name)
ValueError: park is not null but has no default

static/js/map.js
```diff
@@ -1208,8 +1208,8 @@ function updateGymMarker(item, marker) {
             scaledSize: new google.maps.Size(48, 48)
         })
         marker.setZIndex(google.maps.Marker.MAX_ZINDEX + 1)
-    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') &&
-        raidLevel >= Store.get('showRaidMinLevel') && !Store.get('showParkRaidsOnly') && raidLevel <= Store.get('showRaidMaxLevel')) {
+    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') || !Store.get('showParkRaidsOnly' &&
+        raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel'))) {
```
pogo-excalibur: What's happened here :) ?

static/js/map.js
```diff
@@ -1204,7 +1208,8 @@ function updateGymMarker(item, marker) {
             scaledSize: new google.maps.Size(48, 48)
         })
         marker.setZIndex(google.maps.Marker.MAX_ZINDEX + 1)
-    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') && raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
+    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') || !Store.get('showParkRaidsOnly' &&
```
pogo-excalibur: What's happened here 😄 ?

KartulUdus: it breaks 🙄

static/js/map.js
pogo-excalibur: Map has been 'fixed', but the issue with raid eggs not appearing will now be back.
I'd suggest trying the code snippet I posted above.

pogo-excalibur: The map has been 'fixed', but the issue with raid eggs not appearing will now be back.

You'll need to use something like: `(!Store.get('showParkRaidsOnly') || item.park)`.
(untested code - so treat with caution)

And it'll also need to be added to the first `if` level at L1201.

KartulUdus: when added to 1201, it never removed the !park gyms 

I believe this is resolved now with 
` && (!Store.get('showActiveRaidsOnly') || !Store.get('showParkRaidsOnly')) &&`

sebastienvercammen:
@KartulUdus Can you rebase?

pogom/utils.py
```diff
@@ -496,6 +496,10 @@ def get_args():
                               'environment and auto-upload to ' +
                               'hastebin.com.'),
                         action='store_true', default=False)
+    parser.add_argument('-exg', '--ex-gyms',
+                        help=('Fetch OSM parks within geofence and ' +
+                              'mark gyms that are in sed areas'),
```
fosJoddie: possible typo? I suggest changing to `Fetch OSM parks within geofence and flag gyms that are candidates for ex raids.`

KartulUdus: done + also added to help= that `Only required once per area.`

KartulUdus: Also added periods to comments and log messages. 

Holding fire on documentation until specified where to what.

docs/extras/exclusive-raids.md
```diff
@@ -0,0 +1,58 @@
+# Exclusive Raids
+
+EX-raids in Pokemon Go appear at gyms in parks, playgrounds and leisure areas.
```
sebastienvercammen: > Pokémon

docs/extras/exclusive-raids.md
```diff
@@ -0,0 +1,58 @@
...
+
+RocketMap includes a built-in tool to check if existing gyms fall inside these areas.
+
+## How to check your area
+
+  - Scan your area normally until you have all gyms you wish to check.
+  - Run a single instance with a [geofence](geofences.md) and the `-exg` flag.
```
sebastienvercammen: Include the long form the first time, for clarity.

> `-exg` (`--ex-gyms`)

docs/extras/exclusive-raids.md
```diff
@@ -0,0 +1,58 @@
...
+
+Once your check is completed, you can check for gyms or raids in confirmed park areas by these toggles:  
+  ![Sample Image](../_static/img/exgr.png)  
+
+  ![Sample Image](../_static/img/exgg.png)
+
+## Extra
+
+More specifically, a gym that could spawn an exclusive raid will have the center of its s2 level 20 cell inside a confirmed park area.
+You can manually check for s2 cells [here](http://s2.sidewalklabs.com/regioncoverer/).
```
sebastienvercammen: http://s2.sidewalklabs.com/regioncoverer/

image

pogom/models.py
```diff
@@ -2112,7 +2118,15 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
                 b64_gym_id = str(f.id)
                 gym_display = f.gym_display
                 raid_info = f.raid_info
+                # try to get recorded value for Gyms.park, default to false.
+                try:
+                    park = Gym.select(Gym.park).where(
+                        Gym.gym_id == f.id).dicts()[0]['park']
+                except IndexError:
+                    park = False
```
sebastienvercammen: Don't need to rely on the exception, you can just use the dict's `.get('key', default)`:

```
park = Gym.select(Gym.park).where(Gym.gym_id == f.id).dicts()[0].get('park', false)
```

pogom/models.py
```diff
@@ -2112,7 +2118,15 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
                 b64_gym_id = str(f.id)
                 gym_display = f.gym_display
                 raid_info = f.raid_info
+                # try to get recorded value for Gyms.park, default to false.
```
sebastienvercammen: Capitalize

> Try

pogom/osm.py
```diff
@@ -0,0 +1,97 @@
+import json
+import logging
+import math
+from overpy import Overpass
+from geofence import Geofences
+from models import Gym, init_database
+from utils import get_args
+from s2sphere import LatLng, CellId
+
+args = get_args
```
sebastienvercammen: * `args` is unused.
* Should have been `args = get_args()` if you'd have used it.

pogom/osm.py
```diff
@@ -0,0 +1,97 @@
...
+app = None
+log = logging.getLogger(__name__)
+db = init_database(app)
+EARTH_CIRCUMFERENCE_METERS = 400750170
```
sebastienvercammen: What are `app` and `db` for? Unused.

pogom/osm.py
```diff
@@ -0,0 +1,97 @@
...
+    out skel qt;
+    """.format(s, w, n, e))
+
+    return result
+
+
+def exgyms(geofence):
+    # Parse geofence file.
+    log.info('Finding border points from geofence.')
+    f = json.loads(json.dumps(Geofences.parse_geofences_file(geofence, '')))
```
sebastienvercammen: * Use a more descriptive name than `f` (e.g. `geofences`).
* Doing both `json.loads(json.dump())` will have the original as result, so why do any at all?

pogom/osm.py
```diff
@@ -0,0 +1,97 @@
...
+
+        for way in ex_gyms.ways:
+            data = []
+            for node in way.nodes:
+                data.append({'lat': float(node.lat),
+                             'lon': float(node.lon)})
+            if Geofences.is_point_in_polygon_custom(s2_center, data):
+                # Try to get Gym name, but default to id if missing.
+                try:
+                    gymname = Gym.get_gym(gym[0])['name'].encode('utf8')
```
sebastienvercammen: No need to re-query the database.

`Gym.get_gyms(...)` already returns all gym details, including gym name.

pogom/osm.py
```diff
@@ -0,0 +1,97 @@
...
+    );
+    out body;
+    >;
+    out skel qt;
+    """.format(s, w, n, e))
+
+    return result
+
+
+def exgyms(geofence):
```
sebastienvercammen: * Function name `exgyms` isn't descriptive enough.
* I would separate the function into several separate steps so we don't have one big blob that's supposed to do everything, e.g.:
```
# Assumes .get_first_geofence is added to Geofences.
def update_ex_gyms(geofence):
    fence = Geofence.get_first_geofence()
    ...
    ex_gyms = ex_query(south, west, north, east)
    gyms = Gym.get_gyms(south, west, north, east)
    ...
    confirmed_ex_gyms = filter(__gym_is_ex_gym, gyms)
    # Assumes .set_gyms_in_park is added to Gym.
    Gym.set_gyms_in_park(confirmed_ex_gyms, True)


def __gym_is_ex_gym(gym, ex_gyms):
    ex_gyms = ex_gyms.ways

    for way in ex_gyms.ways:
        ... # Copy code here that's currently in exgyms.
        return True

    return False
```

static/js/map.js
sebastienvercammen: ☝️

pogom/osm.py
KartulUdus: ✅

pogom/osm.py
KartulUdus: ✅

pogom/osm.py
KartulUdus: ✅

pogom/osm.py
KartulUdus: 🙅‍♂️ 
it's needed for gyms = Gym.get_gyms
without init_database, it can't query where existing gyms are.
see Friscos review above

pogom/models.py
KartulUdus: ✅

pogom/models.py
KartulUdus: ✅

docs/extras/exclusive-raids.md
KartulUdus: 👌 
This was intended as an extra, should someone not trust the outcome of the check.

docs/extras/exclusive-raids.md
KartulUdus: ✅

docs/extras/exclusive-raids.md
KartulUdus: ✅

pogom/osm.py
KartulUdus: 😊 
...
✅

pogom/models.py
```diff
@@ -626,6 +628,12 @@ def get_gym(id):
 
         return result
 
+    @staticmethod
+    def set_gyms_in_park(gyms, park):
+        for gym in gyms:
+            Gym.update(park=park).where(
+                Gym.gym_id == str(gym['gym_id'])).execute()
```
sebastienvercammen: Replace with:

```
# Instead of doing this, you can also only pass the gym_ids as an argument rather than gyms. Up to you.
gym_ids = [gym['gym_id'] for gym in gyms]
# Execute a single query on a list rather than running a query for each item in the list.
Gym.update(park=park).where(Gym.gym_id << gym_ids).execute()
```

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Query Overpass for known gym areas.
+    api = Overpass()
+    result = api.query("""
+    [out:json]
+    [date:"2016-07-10T00:00:00Z"]
+    [timeout:620]
+    [bbox:{},{},{},{}];
+    (
+    //Tags that are confirmed to classify gyms as 'parks' for EX Raids
```
sebastienvercammen: Write comments as:

> // Capitalized sentence, with punctuation.

*(space and dot)*

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+    west = min(fence, key=lambda ev: ev['lon'])['lon']
+    north = max(fence, key=lambda ev: ev['lat'])['lat']
+    east = max(fence, key=lambda ev: ev['lon'])['lon']
+    log.info('Finding parks within zone.')
+    ex_gyms = ex_query(south, west, north, east)
+    gyms = Gym.get_gyms(south, west, north, east)
+    log.info('Checking {} gyms against {} parks.'.format(len(gyms),
+                                                         len(ex_gyms.ways)))
+    confirmed_ex_gyms = filter(lambda gym: __gym_is_ex_gym(gym, ex_gyms),
+                               gyms.values())
```
sebastienvercammen: Add an empty line and a comment above the `filter` call for clarity:

```
# Retrieve list of confirmed EX gyms to update the DB.
confirmed_ex_gyms = ...
```

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+
+    return result
+
+
+def update_ex_gyms(geofence):
+    # Parse geofence file.
+    log.info('Finding border points from geofence.')
+    geofence = Geofences.parse_geofences_file(geofence, '')
+    fence = geofence[0]['polygon']
+    # Figure out borders for bounding box.
```
sebastienvercammen: Add an empty line before this comment.

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+    south = min(fence, key=lambda ev: ev['lat'])['lat']
+    west = min(fence, key=lambda ev: ev['lon'])['lon']
+    north = max(fence, key=lambda ev: ev['lat'])['lat']
+    east = max(fence, key=lambda ev: ev['lon'])['lon']
+    log.info('Finding parks within zone.')
```
sebastienvercammen: Add an empty line before this log.

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+    Gym.set_gyms_in_park(confirmed_ex_gyms, True)
+
+
+def __gym_is_ex_gym(gym, ex_gyms):
+    gympoint = [float(gym['latitude']),
+                float(gym['longitude'])]
+    s2_center = get_s2_cell_center(gympoint[0], gympoint[1], 20)
+    for way in ex_gyms.ways:
+        data = []
+        for node in way.nodes:
```
sebastienvercammen: Empty line before this for loop (separate blocks).

pogom/osm.py
```diff
@@ -0,0 +1,93 @@
...
+            data.append({'lat': float(node.lat),
+                         'lon': float(node.lon)})
+        if Geofences.is_point_in_polygon_custom(s2_center, data):
```
sebastienvercammen: Empty line before this `if`.

static/js/map.js
sebastienvercammen: Still missing this ☝️ (read previous comment)

KartulUdus:
I think that's the all of it now @sebastienvercammen 

- spaces punctuation line breaks and stuff
- update by array after -exg check 
- found and fixed a nasty where console would error when finding a new gym 
- updated query and doc to include farmyard 
- added park as `true|false` in `gym` webhook ( I want to `!raid snorlax park d2000` )

pogom/osm.py
```diff
@@ -0,0 +1,102 @@
+import logging
+import math
+from overpy import Overpass
+from geofence import Geofences
+from models import Gym, init_database
+from s2sphere import LatLng, CellId
+
+app = None
+log = logging.getLogger(__name__)
+db = init_database(app)
```
sebastienvercammen: Add a comment above this to clarify that you're doing it for side effects (= the effect of the function that isn't a return value).

pogom/osm.py
```diff
@@ -0,0 +1,102 @@
...
+
+def __gym_is_ex_gym(gym, ex_gyms):
+    gympoint = [float(gym['latitude']),
+                float(gym['longitude'])]
+    s2_center = get_s2_cell_center(gympoint[0], gympoint[1], 20)
+
+    for way in ex_gyms.ways:
+        data = []
+        for node in way.nodes:
+            data.append({'lat': float(node.lat),
```
sebastienvercammen: The block `for way in ex_gyms.ways:` will be done every time `__gym_is_ex_gym` is called, and it'll always be on the same data, so it'll unnecessarily be repeated.

Build the `way.nodes` data list outside of the function instead, and pass it as `__gym_is_ex_gym(gym, way_nodes)` instead so it's only done once.

pogom/osm.py
KartulUdus: done

pogom/osm.py
KartulUdus: no.

I moved that loop you don't like outside __gym_is_ex_gym alas, then I iterated through the list twice. 

that would of not worked regardless as an overpy impediment there gave me an error later:
``
    ex_gyms.ways[index]['polygon'] = data
TypeError: 'Way' object does not support item assignment
```

this was by moving the thing *Back* to update_ex_gyms where it resided before.

```
# Generate polygons that is_point_in_polygon_custom accepts for each park.
for index, way in enumerate(ex_gyms.ways):
    data = [{'lat': float(e.lat), 'lon': float(e.lon)} for e in way.nodes]
    ex_gyms.ways[index]['polygon'] = data
```
Please see an [example of a result](https://hastebin.com/ihimihafut.json) from overpy 
the object returned from overpy is a scattered mess of nodes and to know which nodes make up a park, you need to iterate over the "way" first 

I did make the for loop prettier inside __gym_is_ex_gym and ran a test against an area

File: docs/extras/exclusive-raids.md [hunk 33]
```diff
@@ -0,0 +1,62 @@
+# Exclusive Raids
```
pogo-excalibur: Niantic seem to have switched from using the term 'exclusive' to *only* using the term 'EX'.
I think it would be reasonable for us to follow this.

pogom/osm.py
```diff
@@ -0,0 +1,100 @@
...
+    gympoint = [float(gym['latitude']),
+                float(gym['longitude'])]
+    s2_center = get_s2_cell_center(gympoint[0], gympoint[1], 20)
+
+    for way in ex_gyms.ways:
+        data = [{'lat': float(e.lat), 'lon': float(e.lon)} for e in way.nodes]
+
+        if Geofences.is_point_in_polygon_custom(s2_center, data):
+            gymname = gym['name'] or gym['gym_id']
+            log.info('{} is eligible for EX raid.'.format(
```
pogo-excalibur: 'is eligible for EX raid**s**' (add an 's' on the end)

pogo-excalibur: Or perhaps better:
```
Found {count} EX-elegible gyms:
> Gym 1
> Gym 2
> ...
```

pogom/utils.py
```diff
@@ -496,6 +496,11 @@ def get_args():
                               'environment and auto-upload to ' +
                               'hastebin.com.'),
                         action='store_true', default=False)
+    parser.add_argument('-exg', '--ex-gyms',
+                        help=('Fetch OSM parks within geofence and flag ' +
+                              'gyms that are candidates for ex raids. ' +
```
pogo-excalibur: 'gyms that are candidates for ex' -> 'gyms that are candidates for EX'

pogom/utils.py
```diff
@@ -1309,7 +1314,6 @@ def get_debug_dump_link():
     result = '''#######################
 ### RocketMap debug ###
 #######################
-
```
pogo-excalibur: Is this intentional?

pogo-excalibur:
Question for discussion:
Should we remove gyms that are listed as being in parks, but which have the attribute `enabled: false` ?

In my experience these gyms do not spawn raids, and so cannot host an EX raid.

docs/extras/exclusive-raids.md
```diff
@@ -0,0 +1,62 @@
...
+
+EX-raids in Pokémon Go appear at gyms in parks, playgrounds and leisure areas.
+
+RocketMap includes a built-in tool to check if existing gyms fall inside these areas.
+
+## How to check your area
```
pogo-excalibur: Add a note in here that only the first region in the geofence file will get checked.

pogo-excalibur:
Issue while testing with no ex-eligible gyms nearby: `ProgrammingError: (1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '))' at line 1")`

This is caused by:
```
@staticmethod
def set_gyms_in_park(gyms, park):
    gym_ids = [gym['gym_id'] for gym in gyms]
    Gym.update(park=park).where(Gym.gym_id << gym_ids).execute()
```

With no ex-eligible gyms you get: `gyms = []`.

So the SQL query syntax:
```
UPDATE 'gym' SET 'park' = %s WHERE ('gym'.'gym_id' IN ())
```

becomes invalid because of the `IN ()` at the end.

pogo-excalibur:
The bug I found above is still present.

You've handled the case where there are *no gyms at all*, but that's not what was causing the issue.

The bug occurs when there *are* gyms in the geofence, but *none* of them is EX-eligible.

pogom/osm.py
```diff
@@ -0,0 +1,109 @@
+import logging
+import math
+from overpy import Overpass
+from geofence import Geofences
+from models import Gym, init_database
+from s2sphere import LatLng, CellId
+
+# Specify app for a database connection and log to show module name in logs
```
sebastienvercammen: Dot.

pogom/osm.py
```diff
@@ -0,0 +1,109 @@
...
+EARTH_CIRCUMFERENCE_METERS = 400750170
+
+
+def earthMetersToRadians(meters):
+    return (2 * math.pi) * (float(meters) / EARTH_CIRCUMFERENCE_METERS)
+
+
+def ex_query(s, w, n, e):
+
+    # Query Overpass for known gym areas.
```
sebastienvercammen: No empty line above this comment.

pogom/models.py
```diff
@@ -2112,7 +2119,15 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
                 b64_gym_id = str(f.id)
                 gym_display = f.gym_display
                 raid_info = f.raid_info
+                # Try to get recorded value for Gyms.park, default to false.
+                park = False
+                gym_by_id = Gym.select(Gym.park).where(
+                    Gym.gym_id == f.id).dicts()
+                if gym_by_id:
+                    park = gym_by_id[0]['park']
```
sebastienvercammen: These 5 lines could be simplified as a new method on `Gym`, e.g. `Gym.is_in_park()` which returns `True` or `False`. Can return `False` if the `Gym` isn't in the DB.

For proper abstractions, you would implement a method that accepts a list of gyms and does a single `IN ()` query, then use that method to query for single gyms.

```
get_all(x):
    return SELECT * ... IN (x)

get_one(x):
    return get_all(x)

or:

get_one(x):
    return SELECT * ... WHERE x = x
```
(Depends on whether `IN` versus `=` has any performance effect on single items.)

Up to you.

pogom/osm.py
```diff
@@ -0,0 +1,109 @@
...
+    # Figure out borders for bounding box.
+    south = min(fence, key=lambda ev: ev['lat'])['lat']
+    west = min(fence, key=lambda ev: ev['lon'])['lon']
+    north = max(fence, key=lambda ev: ev['lat'])['lat']
+    east = max(fence, key=lambda ev: ev['lon'])['lon']
+
+    log.info('Finding parks within zone.')
+    ex_gyms = ex_query(south, west, north, east)
+    gyms = Gym.get_gyms(south, west, north, east)
+    if (gyms):
```
sebastienvercammen: Don't surround `gyms` with parentheses.

pogom/osm.py
```diff
@@ -0,0 +1,109 @@
...
+                                                        len(ex_gyms.ways)))
+        # Retrieve list of confirmed EX gyms to update the DB.
+        confirmed_ex_gyms = filter(lambda gym: __gym_is_ex_gym(gym, ex_gyms),
+                                   gyms.values())
+
+        if confirmed_ex_gyms:
+            Gym.set_gyms_in_park(confirmed_ex_gyms, True)
+        else:
+            log.info('No EX-eligible gyms found.')
+    else:
```
sebastienvercammen: To solve nesting

```
if x:
    if y:
        a()
    else:
        b()
else:
    c()
```

use this code style:

```
# No gyms at all.
if not gyms:
    a()

confirmed_ex_gyms = ...

# No EX-eligible gyms.
if not confirmed_ex_gyms:
    b()

# At this point, we have a set of EX-eligible gyms and we can continue.
```

runserver.py
```diff
@@ -330,6 +331,15 @@ def main():
 
     args.root_path = os.path.dirname(os.path.abspath(__file__))
 
+    if args.ex_gyms:
+        if args.geofence_file == '':
+            log.critical('A geofence is required to find EX-gyms.')
+            sys.exit(1)
+        else:
```
sebastienvercammen: No need for this `else`: the first branch of the conditional exits the process, so anything after that line of code is assured to have a geofence file and `-exg` enabled.

runserver.py
```diff
@@ -330,6 +331,15 @@ def main():
 
     args.root_path = os.path.dirname(os.path.abspath(__file__))
 
+    if args.ex_gyms:
+        if args.geofence_file == '':
```
sebastienvercammen: Replace this line with:

```
# Geofence is required.
if not args.geofence_file:
```

static/js/map.js
```diff
@@ -1204,7 +1208,8 @@ function updateGymMarker(item, marker) {
             scaledSize: new google.maps.Size(48, 48)
         })
         marker.setZIndex(google.maps.Marker.MAX_ZINDEX + 1)
-    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && !Store.get('showActiveRaidsOnly') && raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
+    } else if (item.raid && item.raid.end > Date.now() && Store.get('showRaids') && (!Store.get('showActiveRaidsOnly') || !Store.get('showParkRaidsOnly')) &&
+        raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')) {
```
sebastienvercammen: This `if` is way too long. Clarify:
```
const hasActiveRaid = item.raid && item.raid.end > Date.now()
const raidLevelVisible = raidLevel >= Store.get('showRaidMinLevel') && raidLevel <= Store.get('showRaidMaxLevel')
const showRaidSetting = Store.get('showRaids') && (!Store.get('showActiveRaidsOnly') || !Store.get('showParkRaidsOnly'))

if (hasActiveRaid && raidLevelVisible && showRaidSetting) {
    ...
}
```

pogom/osm.py
KartulUdus: .

pogom/osm.py
KartulUdus: ✅

pogom/models.py
KartulUdus: it's ok. We're over-engineering a simple task that the end-human would only run once (per -cd) for a minute or so

pogom/osm.py
KartulUdus: new habits develop swift

runserver.py
KartulUdus: Thanks, makes sense 
✅

runserver.py
KartulUdus: I thought I tried that first to no avail 🤔 
anyway: 
✅

static/js/map.js
KartulUdus: did the things. ✅ 

Thought maybe to change `let raidLevel = getRaidLevel(item.raid)` to also be a const, alas i'm not sure if it get's reassigned somewhere of if const would break it so I left it as `let`.
let, var, const is used on map.js depending on moon phases and wind direction so better not touch.

pogom/models.py
sebastienvercammen: @KartulUdus Code style and proper abstractions aren't overengineering.

In this case it's OK to leave as is since no one else needs it (yet), but the moment we need the same code anywhere else, we'll have to implement it anyway.

PR #2445, opened 2018-01-17
[WIP] Add a job module to handled repeated tasks

friscoMad:
## Description
Add a job module based on schedule and in the future move al threaded repeated tasks to use this system.

## Motivation and Context
Add a easy way to configure some functions to run at specified schedules so we can avoid creating a thread for each one and doing custom code for each one.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/pogom/captcha.py
+++ b/pogom/captcha.py
@@ -34,75 +34,52 @@ from .utils import now
 log = logging.getLogger(__name__)
 
 
-def captcha_overseer_thread(args, account_queue, account_captchas,
-                            key_scheduler, wh_queue):
-    solverId = 0
-    while True:
-        # Run once every 15 seconds.
-        sleep_timer = 15
+def start_solvers_task(args, account_queue, account_captchas, key_scheduler,
+                       wh_queue):
+    tokens_needed = len(account_captchas)
+    if tokens_needed > 0:
+        tokens = Token.get_valid(tokens_needed)
+        tokens_available = len(tokens)
+        solvers = min(tokens_needed, tokens_available)
+        log.debug('Captcha overseer running. Captchas: %d - Tokens: %d',
+                  tokens_needed, tokens_available)
+        for i in range(0, solvers):
+            hash_key = None
+            if args.hash_key:
+                hash_key = key_scheduler.next()
 
-        tokens_needed = len(account_captchas)
-        if tokens_needed > 0:
-            tokens = Token.get_valid(tokens_needed)
-            tokens_available = len(tokens)
-            solvers = min(tokens_needed, tokens_available)
-            log.debug('Captcha overseer running. Captchas: %d - Tokens: %d',
-                      tokens_needed, tokens_available)
-            for i in range(0, solvers):
-                hash_key = None
-                if args.hash_key:
-                    hash_key = key_scheduler.next()
+            t = Thread(target=captcha_solver_thread,
+                       args=(args, account_queue, account_captchas,
+                             hash_key, wh_queue, tokens[i]))
+            t.daemon = True
+            t.start()
 
-                t = Thread(target=captcha_solver_thread,
-                           name='captcha-solver-{}'.format(solverId),
-                           args=(args, account_queue, account_captchas,
-                                 hash_key, wh_queue, tokens[i]))
-                t.daemon = True
-                t.start()
+        # Hybrid mode
+        if args.captcha_key and args.manual_captcha_timeout > 0:
+            tokens_remaining = tokens_needed - tokens_available
+            # Safety guard
+            tokens_remaining = min(tokens_remaining, 5)
+            for i in range(0, tokens_remaining):
+                account = account_captchas[0][1]
+                last_active = account['last_active']
+                hold_time = (datetime.utcnow() -
+                             last_active).total_seconds()
+                if hold_time > args.manual_captcha_timeout:
+                    log.debug('Account %s waited %ds for captcha token ' +
+                              'and reached the %ds timeout.',
+                              account['username'], hold_time,
+                              args.manual_captcha_timeout)
+                    if args.hash_key:
+                        hash_key = key_scheduler.next()
 
-                solverId += 1
-                if solverId > 999:
-                    solverId = 0
-                # Wait a bit before launching next thread
-                time.sleep(1)
+                    t = Thread(target=captcha_solver_thread,
+                               args=(args, account_queue, account_captchas,
+                                     hash_key, wh_queue))
+                    t.daemon = True
+                    t.start()
 
-            # Adjust captcha-overseer sleep timer
-            sleep_timer -= 1 * solvers
-
-            # Hybrid mode
-            if args.captcha_key and args.manual_captcha_timeout > 0:
-                tokens_remaining = tokens_needed - tokens_available
-                # Safety guard
-                tokens_remaining = min(tokens_remaining, 5)
-                for i in range(0, tokens_remaining):
-                    account = account_captchas[0][1]
-                    last_active = account['last_active']
-                    hold_time = (datetime.utcnow() -
-                                 last_active).total_seconds()
-                    if hold_time > args.manual_captcha_timeout:
-                        log.debug('Account %s waited %ds for captcha token ' +
-                                  'and reached the %ds timeout.',
-                                  account['username'], hold_time,
-                                  args.manual_captcha_timeout)
-                        if args.hash_key:
-                            hash_key = key_scheduler.next()
-
-                        t = Thread(target=captcha_solver_thread,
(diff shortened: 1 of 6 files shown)


friscoMad:
I want to reorganize requirements in a different PR so users don't need to install all dev dependencies.
I was unable to create tests for the threaded function 😢 

Not sure about module naming and class naming, we already use scheduler for a lot of things but the dissonance between file and class bothers me, maybe we could go with jobScheduler for both.

I did had a decorator also but it is too narrow in scope (no args, so no class methods) that it was even a pain to tests if needed I will readd it.
 
@sebastienvercammen please do some initial review on the concept and module before I change all repeated tasks on RM code.

sebastienvercammen:
@friscoMad RE: our discussion on Discord a while ago:

* We don't want to build our own resource/task scheduler that would have to deal with task priorities, auto multithreading, auto multiprocessing, and so on. However:
* We set up a single `Thread` that handles our list of tasks.
* And we can divide our current "tasks" into several categories:
  1. Tasks that are CPU-bound. -> CPython will take care of scheduling.
  1. Tasks that are net I/O bound (background requests). -> Concurrency achieved with requests-futures.
  1. Tasks to watch files for changes (fs monitoring). -> CPU time minimized by using `watchdog` and only reacting to file changes.
  1. And finally, tasks that need their own `Thread` so they don't interfere with other tasks. -> CPython takes care of scheduling/switching.
     * Proxy refresher.
     * Database cleanup.
     * Dynamic rarity.
* Each task has specific requirements for when it needs to be run (e.g. via `Class.can_run()`), which will include a time (run at) or a frequency (every x seconds/minutes). Tracking time should be done by the scheduler rather than the individual tasks so we only have time tracking code once.

`check_forced_version` belongs in the 2nd type, since we can start a request and give it a callback for when it finishes.

The ideal way to implement these is via a handful of classes: a base Task/Job class and at least one class per type of Task. Each type will have its own parameters to properly handle its resources (whether that's a request's future and callback, a file path for fs monitoring, or starting a Thread for its task).

Having these classes will implement a standard and will avoid problems when people try to add new tasks and have to code the details themselves each time. They also help avoid duplicate code.

friscoMad:
@sebastienvercammen Please check the 3rd orientation based on your input.
I found that there are no 4 plain types of tasks, just 2 types of triggers and 2 types of payloads(CPU or IO).
Creating the 4 combinations seemed a bit overkill so I created the 2 type of task by type of process and let it configure the trigger on init.
I have not changed any current code from the 2nd orientation to be sure you think this approach is correct before continuing, so you should only check tasks.py.

If this is correct I will add some tests and change the code to use this classes.

pogom/tasks.py
```diff
@@ -0,0 +1,183 @@
...
+            sched.do(self._run, self)
+        else:
+            self._filename = filename
+            self._last_modification = os.path.getmtime(filename)
+            # Currently there is no reason for allow configuration of this time
+            # we should change this type of trigger to use OS filesystem
+            # watchers.
+            schedule.every().minute().do(self._check_file(), self)
+
+    def _check_file(self):
```
pogo-excalibur: I'm not sure that this method belongs here. If we're making a class-level distinction between CPU-bound and IO-bound tasks then this would fit better under IOTask.

But I'm thinking that IOTask perhaps needs to be subclassed by FileTask (or similar) with this method.

friscoMad: That's what I tried to explain in the description a file check it does use IO but it should complete the really fast, the thing is what type of task are we running when the file changes, for example a change in the proxy file triggers a IOTask (as it does check the new proxies) but a change in a white_list file it is a CPUTask (just change some vars).

That's the reasoning for having this in the class, you can choose between 2 type of triggers (time based or file based) for any type of workload (CPU  or IO)

pogom/tasks.py
```diff
@@ -0,0 +1,183 @@
...
+    not skipped.
+    """
+    pass
+
+
+class IOTask(BaseTask):
+    """
+    IOTask should be used for tasks that DOSE do any kind of IO, like disk
+    access, network requests or DB queries they are run in their own thread to
+    avoid skipping other tasks.
```
sebastienvercammen: > they are run in their own thread to avoid skipping other tasks

Unfortunately this is a misunderstanding. Not all I/O tasks need a separate thread. I/O tasks can rely on the OS to switch processing to something else while it's waiting for an I/O response.

In addition, some I/O tasks might rely on a library which already implements multiprocessing or multithreading to achieve concurrency in the best way required for that library (e.g. [requests-futures](https://github.com/ross/requests-futures)).

I/O should not automatically be tied to requiring a new Thread. Tasks requiring a new Thread are a different kind, and requiring a Thread could be necessary for any Task (made easier by more/better subclassing).

pogom/captcha.py
```diff
@@ -34,75 +34,52 @@
...
-                        solverId += 1
-                        if solverId > 999:
-                            solverId = 0
-                        # Wait a bit before launching next thread
-                        time.sleep(1)
-                    else:
-                        break
-
-        time.sleep(sleep_timer)
+def start_solvers_task(args, account_queue, account_captchas, key_scheduler,
```
sebastienvercammen: Better as a subclass, not a function. The scope and behavior will be much clearer when reading this code without having to look at the entire codebase (e.g. nothing about this function implies it runs as an infinite loop).

friscoMad:
As it is clear that after 3 days of work and 3 different approaches there is no understanding.
It seems that is difficult to understand that any task that do not returns the control almost inmediatly to the task scheduler thread (it really doesn't matter if does not block the CPU using multithreading, requests or whatever) will block other tasks for running and thus need to be run in a separate thread.

So I will stop wasting my time and close this PR, if someone wants to take over feel free to do so.

sebastienvercammen:
@friscoMad What you say "doesn't matter", does matter. requests_futures would use its own ThreadPoolExecutor, handle itself in the background entirely, and return control to the task scheduler thread near immediately, without needing yet another Thread spawned by ourselves.

Not to mention that CPython is single-threaded, threads spawned don't allow for parallel execution, so any Thread spawned always takes control away from (read: pauses) the main thread while CPython switches between threads to give it some processing time.

But your attitude makes it clear you don't wish to work on this, and that's OK. This PR can be picked up by someone else.

RamiroWincor:
I think I have demostrate previously that I understand the GIL and the "fake" parallelization on CPython, that was the reason to start working on this.

I understand that IO can be misleading maybe it could be blocking/non blocking.

Also take into account that a FWTask could also be blocking: load and check proxies or handle account DB changes.

By my understanding you are currently asking for this class hierarchy:

1. BaseTask: empty run method?
2. FWTask(BaseTask): Filewatching 
3. FWBlockingTask(¿FSTask or BlockingTask?): FileWatching + Thread
4. ScheduledTask(BaseTask): Scheduled
5. BlockingTask(ScheduledTask): Scheduled + Thread
6. NetworkTask(ScheduledTask): Seriously not sure whats the difference between this and 4

That has 2 issues: 
 
- I really don't get what 6 is supposed to do differently that 4
- It would need some code duplication or global functions as there is no way to create 3 just with inheritance, that's how I ended with 2 classes (Thread or not) with 2 triggers (Sched or FS), you can switch it to threading or not being the parameter but that was what I did in my second attempt and you said it was a no go.

I said I will not to work in this anymore there is no need to evaluate my attitude, as told in #pr I am leaving pogo world.
Even if I am no longer motivated, I wanted to complete all the PRs I had pending.
But after 3 attempts to do something you see good enough and 1 month to get this reviewed all I got is that this is not ok with no clear indication on how to fix it, it seems that only you know how to do it properly so I gave up.

sebastienvercammen:
@RamiroWincor @friscoMad *(Same person, different account)*

Your attitude doesn't refer to not wanting to work on it, but to:

> It seems that is difficult to understand that ...  
>   
> ... that only you know how to do it properly so I gave up

There's no need to insult the people you work with.

Your reactions to the review/feedback are inappropriate. The reviews were code reviews, not a personality review.

Your examples of class structure, which you get unnecessarily angry about, are wrong because you're still misunderstanding the difference between classes for functionality/hierarchy (base classes) and the additional layer on top of those for enclosure/readability/usability.

Time to review can also not have been a problem. This PR has never been on our #pr priority/review list so we could focus on more urgent things (something that is visible by all #reviewers and #pr), and the moment you pinged me personally for a review I posted it. And this is your reaction.

Either way, you've made your point clear, and unnecessarily personal. We won't expect you to make any future attempts.

friscoMad:
yup it was me, I forgot to log off from my work account.
I won't continue the discussion as it leads to nowhere just wanting to point out that this PR has been in the list in the section of "needs changes" for several weeks.

PR #2452, opened 2018-01-20
Add option to disable kph and hlvl-kph.

sebastienvercammen:
## Description
Setting `--kph` or `--hlvl-kph` to `0` will disable the appropriate speed checks.

## Motivation and Context
* Gyms can still be scanned even if an account is blinded (speed ban).
* High-level account encounters don't do scan/GMO calls, only encounters.
* Avoids discussions like: 
```
[12:45 PM] r4ms1ck: the default ? or hihger hlvl-kph: 25
[12:46 PM] Konturka: @r4ms1ck I've mine at 99999999999999999999
```

## How Has This Been Tested?
Untested. 🙇 

## Types of changes
- [x] New feature (non-breaking change which adds functionality).

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -77,7 +77,7 @@
 #workers-per-hive:              # Only referenced when using --beehive. Sets number of workers per hive. (default=1)
 #workers:                       # Number of search worker threads to start. (default=#accounts)
 #spawn-delay:                   # Number of seconds after spawn time to wait before scanning to be sure the Pokemon is there. (default=10)
-#kph:                           # Set a maximum speed in km/hour for scanner movement. (default=35)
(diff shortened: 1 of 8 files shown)


sebastienvercammen:
Last commit changes some things:

* Made minor edits to improve clarity/readability/consistency.
* Edited the logging of toggles so they're always logged instead of only when they're changed.
* Replaced `len(args.wh_types) == 0` with `not args.wh_types`, as recommended per `len-as-condition (C1801)`. Explanation on why this is more efficient [here](https://stackoverflow.com/questions/43121340/why-is-the-use-of-lensequence-in-condition-values-considered-incorrect-by-pyli#comment73322508_43121340).
* Added logging for kph/hlvl-kph on startup.

tomballgithub:
Approved

PR #2454, opened 2018-01-20
Add toggle to show/hide Pokémon stats.

daangroot:
## Description

This PR brings a toggle which will show or hide pokemon stats (IV, CP, weight,) on the map.

## Motivation and Context

Some of my users did not want to see pokemon stats, so I thought lets make a toggle for that. :)

## How Has This Been Tested?

I have tested it on my main instance.
Testing environment: Windows, Google Chrome.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2893,7 +2893,8 @@ $(function () {
         markerCluster.repaint()
     })
     $('#pokemon-stats-switch').change(function () {
-        Store.set('showPokemonStats', this.checked)
+        Store.set('showPokemonStats, this.checked)
+		updateMap()
     })
     $('#scanned-switch').change(function () {
         buildSwitchChangeListener(mapData, ['scanned'], 'showScanned').bind(this)()
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -2894,7 +2894,11 @@ $(function () {
     })
     $('#pokemon-stats-switch').change(function () {
         Store.set('showPokemonStats', this.checked)
-		updateMap()
+        redrawPokemon(mapData.pokemons)
+        redrawPokemon(mapData.lurePokemons)
+
+        // We're done processing the list. Repaint.
+        markerCluster.repaint()
     })
     $('#scanned-switch').change(function () {
         buildSwitchChangeListener(mapData, ['scanned'], 'showScanned').bind(this)()
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -643,6 +643,17 @@ function pokemonLabel(item) {
     return contentstring
 }
 
+function updatePokemonLabels(pokemonList) {
+    $.each(pokemonList, function (key, value) {
+        var item = pokemonList[key]
+
+        item.marker.infoWindow = new google.maps.InfoWindow({
+            content: pokemonLabel(item),
+            disableAutoPan: true
+        })
+    })
+}
+
 function isOngoingRaid(raid) {
     return raid && Date.now() < raid.end && Date.now() > raid.start
 }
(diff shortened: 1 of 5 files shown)


static/js/map.js
```diff
@@ -2891,6 +2905,24 @@ $(function () {
         buildSwitchChangeListener(mapData, ['pokemons'], 'showPokemon').bind(this)()
         markerCluster.repaint()
     })
+    $('#pokemon-stats-switch').change(function () {
+        Store.set('showPokemonStats', this.checked)
+        var options = {
+            'duration': 500
+        }
+        var wrapper = $('#notify-perfection-wrapper')
```
sebastienvercammen: Use `const` instead of `var` to denote items whose reference won't change.

When using a variable to reference a jQuery object, you can add a `$` to the front of the variable name to make it instantly clear to anyone reading your code.

```
const $wrapper = $('#notify-perfection-wrapper')
```

and if you ever need to reference the DOM element again from the jQuery object:

```
const wrapper = $wrapper[0]
```

static/js/map.js
```diff
@@ -642,6 +644,19 @@ function pokemonLabel(item) {
     return contentstring
 }
 
+function updatePokemonLabel(item) {
```
sebastienvercammen: Move this function and `updatePokemonLabels` to `map.common.js` near `updatePokemonMarker`.

static/js/map.js
```diff
@@ -996,7 +1011,7 @@ function getNotifyText(item) {
     var find = ['<prc>', '<pkm>', '<atk>', '<def>', '<sta>']
     var replace = [((iv) ? iv.toFixed(1) : ''), item['pokemon_name'], item['individual_attack'],
         item['individual_defense'], item['individual_stamina']]
-    var ntitle = repArray(((iv) ? notifyIvTitle : notifyNoIvTitle), find, replace)
+    var ntitle = repArray(((Store.get('showPokemonStats') && iv) ? notifyIvTitle : notifyNoIvTitle), find, replace)
```
sebastienvercammen: Lots of these items could/should be `const`.

To shorten this line, you could move `Store.get('showPokemonStats')` to its own variable just above it.

```
const showStats = Store.get('showPokemonStats')
const ntitle = repArray(((showStats && iv) ...
```

static/js/map.js
```diff
@@ -2891,6 +2905,24 @@ $(function () {
...
+            wrapper.show(options)
+        } else {
+            wrapper.hide(options)
+        }
+        updatePokemonLabels(mapData.pokemons)
+        redrawPokemon(mapData.pokemons)
+        redrawPokemon(mapData.lurePokemons)
+
+        // We're done processing the list. Repaint.
+        markerCluster.repaint()
```
sebastienvercammen: These redrawing functions are *really heavy* and should be avoided. Are you sure it's necessary, considering you're not changing the markers but only the content of the labels (which aren't added/drawn until you've clicked one)?

templates/map.html
```diff
@@ -257,6 +257,18 @@ <h3>Exclude Rarity</h3>
...
+              <div class="onoffswitch">
+                <input id="pokemon-stats-switch" type="checkbox" name="pokemon-stats-switch" class="onoffswitch-checkbox" checked>
+                <label class="onoffswitch-label" for="pokemon-stats-switch">
+                <span class="switch-label" data-on="On" data-off="Off"></span>
+                <span class="switch-handle"></span>
+                </label>
+              </div>
+            </div>
+              {% endif %}
               {% endif %}
```
sebastienvercammen: Indentation seems a bit off here.

daangroot:
I've removed the code for changing pokemon animations, because I couldn't get it to work. For now this PR only updates the info window labels, prevents notifications based on IV and redraws the pokemon who were upscaled because of IVs.

static/js/map.js
daangroot: I implemented this suggestion, but it broke the code. All other wrappers used are also vars.

sebastienvercammen: @daangroot If you only implemented it in this specific position, it shouldn't break, `wrapper` isn't being changed.

All the other wrapper could also use `const` if they never change its reference.

If it does break for you, can you post a screenshot of the error in your console?

daangroot: You are right. The first time I tested it I must have forgotten to put `$`  infront of `wrapper` when hiding and showing it.

PR #2459, opened 2018-01-24
Dynamic rarity

BrownSlaughter:
Tired of seeing Mr mime as ultra rare when you see 100's an hour?
This will set rarity for all pokemon from what you actually scan!

You can use `-rh` or  `--rarity-hours` to set the time period it uses to calculate rarity - default is 48 hours

I deleted my repo so had to close and reopen the PR
I also reduced the time frame it checks from all in db down to 2 days to try help with huge databases

**Seb edit:**

* Works on all instances except scanners only `-ns`.
* Dynamic rarity is disabled by default, similar to the database cleaner. It should be enabled on a single instance. (If enabled on multiple instances, you will create problems.)
* Reworked DB query and indexes to be more efficient.
   * 10 million Pokémon entries without limit = 5 seconds.
   * 10 million Pokémon entries filtered by the last 48h = **0.5 seconds**.
* Config items are now
   * `-Rh, --rarity-hours`, number of hours of Pokémon data to include. Decimals allowed. Default 48. 0 for all data.
   * `-Rf, --rarity-update-frequency`, how often to update, in minutes. Decimals allowed. Default 0. 0 to disable).
* Moved Pokémon rarity to front-end. Removes the need to cache data on back-end and makes it possible to serve the JSON file via nginx/apache2 (e.g. via symlink). Less overhead per request.

When pulling, remember to:
```
npm run build
```

## Description
sets rarity group dependant what you actually see.
Updated pokemon.json

## Motivation and Context
Rarity is area dependent, what one person scans in America is not valid for what someone scans in England so having a 1 size fits all pokemon.json does not work for all. 

All gen 3 and all future pokemon will automatically get set a rarity without needing to update pokemon.json

Scaling by rarity will also be more accurate because of this.

## How Has This Been Tested?
looked on map and checked it works like it did before changes but dynamically

Tested on a clean DB everything starts as ultra rare because its never seen them before but quickly changes during initial scan

## Types of changes

## Checklist:

### DIFF
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -5,15 +5,13 @@ import calendar
 import logging
 import gc
 
+from datetime import datetime
+from s2sphere import LatLng
+from bisect import bisect_left
 from flask import Flask, abort, jsonify, render_template, request,\
     make_response, send_from_directory
 from flask.json import JSONEncoder
 from flask_compress import Compress
-from datetime import datetime
-from s2sphere import LatLng
-from bisect import bisect_left
-from threading import Lock
-from timeit import default_timer
 
 from .models import (Pokemon, Gym, Pokestop, ScannedLocation,
                      MainWorker, WorkerStatus, Token, HashKeys,
@@ -26,58 +24,6 @@ from .blacklist import fingerprints, get_ip_blacklist
 log = logging.getLogger(__name__)
 compress = Compress()
 
-rarity_lock = Lock()
-
-
-def get_rarity(pokemon_id):
-    args = get_args()
-    raritytime = args.raritytime
-    # Data shared by several threads. Only one should be here to check/update.
-    with rarity_lock:
-        # Can't call it "now" because utils has a now() method...
-        now_time = default_timer()
-
-        rarity_is_loaded = hasattr(get_rarity, 'last_seen')
-        should_refresh = False
-
-        if rarity_is_loaded:
-            # Refresh once in a while.
-            rarity_refresh_seconds = 60 * 60  # Once an hour.
-            time_diff = now_time - get_rarity.last_seen_time
-            should_refresh = time_diff > rarity_refresh_seconds
-
-        # Load or refresh.
-        if not rarity_is_loaded or should_refresh:
-            log.info('Updating dynamic rarity...')
-            get_rarity.last_seen = Pokemon.get_seen(raritytime)
-            get_rarity.last_seen_time = now_time
-            log.info('Updated dynamic rarity.')
-
-    # State checking is done. Code here is thread safe.
-    seen = get_rarity.last_seen
-    total = seen['total']
-    found = 0
-    spawn_group = ''
-    for pokemon in seen['pokemon']:
-        if pokemon['pokemon_id'] == pokemon_id:
-            found = 1
-            pokemon_count = pokemon['count']
-    if found == 0:
-        pokemon_count = 0
-    spawn_rate = round(100 * pokemon_count / float(total), 4)
-    if spawn_rate < 0.01:
-        spawn_group = 'Ultra Rare'
-    elif spawn_rate < 0.03:
-        spawn_group = 'Very Rare'
-    elif spawn_rate < 0.5:
-        spawn_group = 'Rare'
-    elif spawn_rate < 1:
-        spawn_group = 'Uncommon'
-    else:   # anything >= 1
-        spawn_group = 'Common'
-
-    return spawn_group
-
 
 def convert_pokemon_list(pokemon):
     args = get_args()
@@ -88,7 +34,7 @@ def convert_pokemon_list(pokemon):
     pokemon_result = []
     for p in pokemon:
         p['pokemon_name'] = get_pokemon_name(p['pokemon_id'])
-        p['pokemon_rarity'] = get_rarity(p['pokemon_id'])
+        p['pokemon_rarity'] = Pokemon.get_rarity(p['pokemon_id'])
         p['pokemon_types'] = get_pokemon_types(p['pokemon_id'])
         p['encounter_id'] = str(p['encounter_id'])
         if args.china:
--- a/pogom/app.py
+++ b/pogom/app.py
@@ -11,13 +11,14 @@ from flask.json import JSONEncoder
 from flask_compress import Compress
 from datetime import datetime
 from s2sphere import LatLng
-from pogom.utils import get_args
 from bisect import bisect_left
+from threading import Lock
+from timeit import default_timer
 
 from .models import (Pokemon, Gym, Pokestop, ScannedLocation,
                      MainWorker, WorkerStatus, Token, HashKeys,
                      SpawnPoint)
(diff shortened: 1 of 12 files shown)


jaake77:
Have you considered a tool to run on the side to update rarity? (not dynamic)

BrownSlaughter:
@jaake77 I dont like having lots of tools and scripts separate to RM, it is much easier to do it as part of the map

sebastienvercammen:
@BrownSlaughter Looks like this PR is currently missing the majority of its code?

sebastienvercammen:
I've pushed an update to reduce the effect of the `get_rarity` function on performance.

* It caches the result locally.
* It refreshes the data once every hour. No separate update thread needed. Separating the updating into its own Thread is ideally done in #2445, so `get_rarity` calls don't interrupt the response if an update is in progress.
* Code is thread safe. If there's an on-going update, a thread will have to wait for the lock.
* Added logging.
* The lock had to be added to the global namespace for `get_rarity` to be thread safe, otherwise we could have a race condition inside the method where multiple threads are trying to set up a Lock the first time.

pogom/app.py
```diff
@@ -11,20 +11,69 @@
...
 
 log = logging.getLogger(__name__)
 compress = Compress()
 
+rarity_lock = Lock()
+
+
+def get_rarity(pokemon_id):
+    # Data shared by several threads. Only one should be here to check/update.
+    with rarity_lock:
```
friscoMad: get_seen() is already cached in models.py does it makes any sense to cache again the results here?

sebastienvercammen: @friscoMad The cache local to get_rarity is a reference to the same data, so we're just using one extra reference for the added security of making sure that it's cached. `get_rarity` can be unaware of what `Pokemon.get_seen()` does, both DB and `get_rarity` code can be changed in the future, and we'll be able to make safe assumptions (that it's being cached).

The alternative is to add a comment to explain what's happening in the background. Both are 👍, I just picked this one for clarity and for being able to see the explicit thread safety.

sebastienvercammen:
Last update:

* Simplified query. No subqueries, nothing unnecessary. Only two fields are retrieved: Pokémon ID and spawn count.
* Uses DB's `GROUP BY` and `COUNT()` instead of doing it ourselves.
* No manipulating any records. Straight from DB to our `get_rarity`, so no overhead per row.
* Cached list is sorted. Now uses binary search for `O(log n)` search instead of the list iteration (`O(n)`) we used to have.
* Using `0` for `--rarity-hours` will use all data in the DB.
* Updated docs.

sebastienvercammen:
Last update:

* Rebased to latest RM commit.
* Moved the dynamic rarity updater to its own thread, in the exact same style as the other "background tasks", easy to pick up for #2445.
* Added `-Rf, --rarity-update-frequency` to set how often it should update (in minutes).
* Renamed `-rh` to `-Rh` to be consistent with our new config style (using capitalized letters for grouping/namespace).
* Export rarities to a static JSON file. Can now be served by a production webserver (e.g. nginx) for better performance. Also removes the need to do any memory caching.
* Moved `pokemon_rarity` to the front-end. This fits perfectly with our upcoming internationalization and data rework which will move more data from the back-end to the front-end (Pokémon moves, types, localization), reducing overhead per request.
   * Completely removed caching/searching. O(log n) for binary search was good, but the new lookup format for front-end is O(1) *(in most cases, because O(1) assumes a perfect hash table)* and done by the client.
   * Also helped to reduce/simplify some code thanks to `getPokemonRarity`.
* Fixed a cyclic import. These problems can also be avoided by importing the full module rather than only importing a specific name (e.g. `import pogom.utils` vs `from pogom.utils import get_pokemon_name`).

Currently untested, we have proper testing coming in a few hours.

pogom/models.py
```diff
@@ -198,6 +198,29 @@ def get_active_by_id(ids, swLat, swLng, neLat, neLng):
...
+    # Returns a dict:
+    #   { 'pokemon': [ {'pokemon_id': '', 'count': 1} ], 'total': 1 }.
+    @staticmethod
+    def get_spawn_counts(hours):
+        # Allow 0 to query everything.
+        if hours:
+            hours = datetime.utcnow() - timedelta(hours=hours)
+
+        query = (Pokemon
+                 .select(Pokemon.pokemon_id,
```
friscoMad: We should use the same trick to avoid using the pokemon_id index, in my tests current version in a 800K pokes db takes about 18-2secs, except if I use -Rh to 12 or lower that it goes to 0.12s.
Using the trick it takes between 0,40s and 0.60s. Using get_seen instead of get_spawn_counts does yield a different number or rows but it is mostly the same time to get the data.

pogom/utils.py
```diff
@@ -136,6 +137,16 @@ def get_args():
     parser.add_argument('-sd', '--scan-delay',
                         help='Time delay between requests in scan threads.',
                         type=float, default=10)
+    parser.add_argument('-Rh', '--rarity-hours',
```
friscoMad: Maybe we can use a optgroup for rarity options.

runserver.py
```diff
@@ -457,6 +463,18 @@ def main():
                                name='search-overseer', args=argset)
         search_thread.daemon = True
         search_thread.start()
+    else:
+        # -os only.
```
friscoMad: This shouldn't be -os only, it should be not be run on -ns.
There is no reason to let user running a single instance for map and web without this, it is a single thread that goes online 1 every hour and spend most of the time waiting for a DB query to complete (IO bound) the impact should be minimal, a lot less that flaks threads.

pogom/models.py
sebastienvercammen: Currently investigating other alternatives to using the "trick" to avoid using the pokemon_id index, e.g. reworking our current indexes to get all data directly from the indexes rather than having to load extra data or do a table search.

If none of the other options have better performance than the trick, we'll use that trick.

runserver.py
```diff
@@ -203,6 +203,12 @@ def can_start_scanning(args):
         log.critical(api_version_error)
         return False
 
+    # Only allow dynamic rarity on -os.
```
friscoMad: This need to be gone :(

sebastienvercammen: Comments will kill this PR. :(

sebastienvercammen:
Update:

* Allow dynamic rarity to run on all instances except scanner-only (`-ns`).
* Moved config items to an option group.
* Updated DB querying and DB indexes to speed up the code.
   * Original update used to take ~112 seconds for 798 461 entries.
   * This update took ~5 seconds for 10 million unfiltered entries and 0.5 seconds for 10 million filtered entries (filtered by last 48 hours only).
   * All other code is unaffected by the dropped old `disappear_time` index: since `disappear_time` is the leftmost prefix of the index it'll also be used for queries using only `disappear_time` (source: [MySQL docs](https://dev.mysql.com/doc/refman/5.7/en/mysql-indexes.html)).

darkelement1987:
THANKS SO MUCH! You know how much i've waited for this! :D

PR #2463, opened 2018-01-26
Add Log Filename

roberthbelter:
## Description

This change add a option to include a log file name in configuration files and command line options.
## Motivation and Context

The reason for this change is to make it easier to use monitoring and automation scripts.
## How Has This Been Tested?

I tested on an implementation of this map

config file:
log-path: /var/log/rocketmap/gyms
log-filename: gyms.log

Log:
$ tail /var/log/rocketmap/gyms/gyms.log
2018-01-26 12:08:34,776 [  search-worker-60][        search][    INFO] Searching at -30.012241,-51.152141,11.154882.
2018-01-26 12:08:34,982 [  search-worker-77][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 4, raids: 0.
2018-01-26 12:08:35,201 [  search-worker-50][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 2, raids: 1.
2018-01-26 12:08:35,233 [  search-worker-71][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 1, raids: 1.
2018-01-26 12:08:35,250 [  search-worker-53][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 6, raids: 1.
2018-01-26 12:08:35,262 [  search-worker-96][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 6, raids: 1.
2018-01-26 12:08:35,275 [  search-worker-43][        search][    INFO] Searching at -29.987884,-51.168296,11.550637.
2018-01-26 12:08:35,287 [  search-worker-72][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 9, raids: 2.
2018-01-26 12:08:35,434 [   search-worker-2][        search][    INFO] Searching at -30.000062,-51.160219,10.421407.
2018-01-26 12:08:35,547 [  search-worker-57][        models][    INFO] Parsing found Pokemon: 0 (0 filtered), nearby: 0, pokestops: 0, gyms: 1, raids: 0

## Types of changes

## Checklist:

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -185,6 +185,7 @@
 #verbosity                      # Show debug messages from RocketMap and PGoApi. Values are 1,2,3 (default=0)
 #no-file-logs                   # Disables logging to files except for access.log. (default=False)
 #log-path:                      # Defines the path logs are saved at. (default=logs/)
+#log-filename:                  # Defines the log filename to be saved. (default='{}_{}.log'.format(strftime('%Y%m%d_%H%M'),str(os.getpid()))
 #no-version-check               # Disable API version check. (default=False)
 #version-check-interval:        # Interval to check API version in seconds (Default: in range [60, 300]).
 #mock:                          # Mock mode - point to a fpgo endpoint instead of using the real PogoApi,
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -185,7 +185,7 @@
 #verbosity                      # Show debug messages from RocketMap and PGoApi. Values are 1,2,3 (default=0)
 #no-file-logs                   # Disables logging to files except for access.log. (default=False)
 #log-path:                      # Defines the path logs are saved at. (default=logs/)
-#log-filename:                  # Defines the log filename to be saved. (default='{}_{}.log'.format(strftime('%Y%m%d_%H%M'),str(os.getpid()))
+#log-filename:                  # Defines the log filename to be saved. The default generates yyyymmdd_HHMM_statusname.log. (default=%Y%m%d_%H%M_<SN>)
 #no-version-check               # Disable API version check. (default=False)
 #version-check-interval:        # Interval to check API version in seconds (Default: in range [60, 300]).
 #mock:                          # Mock mode - point to a fpgo endpoint instead of using the real PogoApi,
--- a/docs/first-run/commandline.md
+++ b/docs/first-run/commandline.md
@@ -393,7 +393,7 @@ variables which override config file values which override defaults.
                             Defines the log filename to be saved. Allows date
                             formatting, and replaces <SN> with the instance's
                             status name. Read the python time module docs for
-                            details. Default: %Y%m%d_%H%M_<SN>. [env var:
+                            details. Default: %Y%m%d_%H%M_<SN>.log. [env var:
                             POGOMAP_LOG_FILENAME]
       --dump                Dump censored debug info about the environment and
                             auto-upload to hastebin.com. [env var: POGOMAP_DUMP]
--- a/docs/first-run/commandline.md
+++ b/docs/first-run/commandline.md
@@ -388,6 +388,9 @@ environment variables which override config file values which override defaults.
                             logs. [env var: POGOMAP_NO_FILE_LOGS]
       --log-path LOG_PATH   Defines directory to save log files to. [env var:
                             POGOMAP_LOG_PATH]
+      --log-filename FILENAME
+                            Defines the log filename to be saved. [env var:
+                            POGOMAP_LOG_FILENAME]
       --dump                Dump censored debug info about the environment and
                             auto-upload to hastebin.com. [env var: POGOMAP_DUMP]
       -exg, --ex-gyms       Fetch OSM parks within geofence and flag gyms that are
--- a/docs/first-run/commandline.md
+++ b/docs/first-run/commandline.md
@@ -18,8 +18,8 @@
                     [-ld LOGIN_DELAY] [-lr LOGIN_RETRIES] [-mf MAX_FAILURES]
                     [-me MAX_EMPTY] [-bsr BAD_SCAN_RETRY]
                     [-msl MIN_SECONDS_LEFT] [-dc] [-H HOST] [-P PORT]
-                    [-L LOCALE] [-c] [-m MOCK] [-ns] [-os] [-sc] [-nfl]
-                    -k GMAPS_KEY [--skip-empty] [-C] [-cd] [-np] [-ng] [-nr]
+                    [-L LOCALE] [-c] [-m MOCK] [-ns] [-os] [-sc] [-nfl] -k
+                    GMAPS_KEY [--skip-empty] [-C] [-cd] [-np] [-ng] [-nr]
                     [-nk] [-ss] [-ssct SS_CLUSTER_TIME] [-speed] [-spin]
                     [-ams ACCOUNT_MAX_SPINS] [-kph KPH] [-hkph HLVL_KPH]
                     [-ldur LURE_DURATION] [-px PROXY] [-pxsc]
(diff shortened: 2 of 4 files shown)


config/config.ini.example
```diff
@@ -185,6 +185,7 @@
 #verbosity                      # Show debug messages from RocketMap and PGoApi. Values are 1,2,3 (default=0)
 #no-file-logs                   # Disables logging to files except for access.log. (default=False)
 #log-path:                      # Defines the path logs are saved at. (default=logs/)
+#log-filename:                  # Defines the log filename to be saved. (default=DATE[Ymd_HM]_STATUSNAME.log)
```
sebastienvercammen: Dubious. The example you use as default is just a static string, people might thing you've coded string replacement to replace those names by their proper values (date, status name).

```
DATE[Ymd_HM]_STATUSNAME.log
```

docs/first-run/commandline.md
```diff
@@ -389,6 +389,9 @@ which override config file values which override defaults.
                             logs. [env var: POGOMAP_NO_FILE_LOGS]
       --log-path LOG_PATH   Defines directory to save log files to. [env var:
                             POGOMAP_LOG_PATH]
+      --log-filaname FILENAME
```
sebastienvercammen: Typo:

> filaname

`commandline.md` is generated with `python runserver.py -h`. Don't forget to remove sensitive data.

runserver.py
```diff
@@ -498,8 +498,12 @@ def set_log_and_verbosity(log):
         os.mkdir(args.log_path)
     if not args.no_file_logs:
         date = strftime('%Y%m%d_%H%M')
-        filename = os.path.join(
-            args.log_path, '{}_{}.log'.format(date, args.status_name))
+        if not args.log_filename:
```
sebastienvercammen: Invalid condition: the default value of `args.log_filename` is `DATE[Ymd_HM]_STATUSNAME.log`, which will always evaluate to `True`.

roberth-belter-azion:
Right! Sorry for not understanding this part of code yet. I fixed a few things, but I still need to set the configuration example (not really good) and I did not figure out how to use the arg (not yet parsed) status_name in --log-filename default. If you can help :)

pogo-excalibur:
I think that this would be more useful if it could do the text substitution that Seb suggested.

docs/first-run/commandline.md
roberthbelter: My bad, fixed!

runserver.py
roberthbelter: Fixed too... in parser, the default value is None. But, after the argument parser, the default value of filename is defined by the code below:

    if args.log_filename is None:
        if args.status_name is None:
            args.log_filename = '{}_{}.log'.format(strftime('%Y%m%d_%H%M'),
                                                   str(os.getpid()))
        else:
            args.log_filename = '{}_{}.log'.format(strftime('%Y%m%d_%H%M'),
                                                   args.status_name)

I think this is the best way to use the status_name argument inside the log-filename if it is not defined.

pogom/utils.py
```diff
@@ -508,6 +512,14 @@ def get_args():
 
     args = parser.parse_args()
 
+    if args.log_filename is None:
```
sebastienvercammen: You could simply do
```
args.log_filename = strftime(args.log_filename)
```
to add support for date parsing/substitution.

Optionally, you could add custom placeholders (e.g. `<pid>, <name>` for PID and status_name) and replace them yourself before applying `strftime`. Note: if you go this route, don't do:
```
strftime(custom_replace(x))
```
since `custom_replace`'s result could contain characters that would be replaced by strftime (e.g. due to status_name). To avoid this, split the string into pieces in your `custom_replace` so that each custom placeholder is its own piece. Once everything custom is replaced, apply `strftime` to the other pieces and `[].join('')` them all together.

pogom/utils.py
```diff
@@ -508,6 +512,14 @@ def get_args():
...
+        if args.status_name is None:
+            args.log_filename = '{}_{}.log'.format(strftime('%Y%m%d_%H%M'),
+                                                   str(os.getpid()))
+        else:
+            args.log_filename = '{}_{}.log'.format(strftime('%Y%m%d_%H%M'),
+                                                   args.status_name)
```
sebastienvercammen: Could be simplified as:

```
log_tail = args.status_name if args.status_name else str(os.getpid())
args.log_filename = '{}_{}.log'.format(strftime(...), log_tail)
```

tomballgithub:
Great idea.

pogom/utils.py
roberthbelter: Fixed... I think the new code solves the other problem too

pogom/utils.py
```diff
@@ -520,6 +524,11 @@ def get_args():
 
     args = parser.parse_args()
 
+    if args.log_filename is None:
+        date = strftime('%Y%m%d_%H%M')
+        log_tail = args.status_name if args.status_name else str(os.getpid())
+        args.log_filename = '{}_{}.log'.format(date, log_tail)
```
sebastienvercammen: Now the `log_filename` is hardcoded and can never be changed?

roberthbelter: In this case, the following situation is occurring:

The user can set the parameter via command line or configuration file.
If it is not defined, parser will set it to None and in this part of the code, if it is None (because the user has not defined it), a default value will be set. This is defined here because the default value can be changed by the status_name parameter, if it exists.

In resume, this part of the code only sets the default value if the user do not define it.

sebastienvercammen: @roberthbelter No worries, I completely overlooked the condition. A bit silly.

roberth-belter-azion:
Well, in fact, I removed the `log_tail`. The `status_name` argument has already been parseed and can be used here. About `strftime()`, the motivation of this PR is precisely to remove it and and allow the user to define a static name. Why this? My motivation, in particular, is the use of monitoring scripts. If the process crashes and restarts, any monitoring script must also be restarted and re-adapted to the new file name.

pogo-excalibur:
But if you use `strftime(args.log_filename)` then those who want to use the substitution sequences can (and we can use them in the default), and those who don't want to use substitution sequences don't have to.

config/config.ini.example
```diff
@@ -185,6 +185,7 @@
 #verbosity                      # Show debug messages from RocketMap and PGoApi. Values are 1,2,3 (default=0)
 #no-file-logs                   # Disables logging to files except for access.log. (default=False)
 #log-path:                      # Defines the path logs are saved at. (default=logs/)
+#log-filename:                  # Defines the log filename to be saved. (default='{}_{}.log'.format(strftime('%Y%m%d_%H%M'),str(os.getpid()))
```
fosJoddie: trying to decipher this default makes my head impode.

But should this not be "simply" `%Y%m%d_%H%M_<SN>`  ? 

The code says it runs `strftime()` on the file name, then replaces the `<SN>` part with status name after that.

This default suggests you should be writing python.

Maybe a better way would be to refer to the docs for details on how to customize, and say the default produces files like: `yyyy-mm-dd_HHMM-statusname.log`.

docs/first-run/commandline.md
```diff
@@ -388,6 +388,9 @@ environment variables which override config file values which override defaults.
                             logs. [env var: POGOMAP_NO_FILE_LOGS]
       --log-path LOG_PATH   Defines directory to save log files to. [env var:
                             POGOMAP_LOG_PATH]
+      --log-filename FILENAME
```
fosJoddie: Should at the very least define the default here, and perhaps link to `https://docs.python.org/2/library/time.html` for the strftime arguments possible

PR #2464, opened 2018-01-26
devkat: DB upsert performance improvements.

sebastienvercammen:
## Description
* Changes our bulk upsert code to use `INSERT ... ON DUPLICATE KEY UPDATE` instead of relying on peewee, which only supports `REPLACE INTO`.
* Also leverages the cursor's `.executemany`, allowing the DB to properly prepare for a batched query.

Long-term, we can look at increasing the batch step (currently 250) and analyze the performance with `executemany`.

## Motivation and Context
`REPLACE INTO` first deletes the old entry before inserting the new one. This causes several issues:

* Unnecessary overhead.
* The auto-increment ID for the row doesn't stay the same, it gets a new one when inserted. #2440 is affected by this.
* Table indexes have to be updated each time.
* Strongly affects DB lock wait times and eventually causes deadlocks once it reaches its limit.

## How Has This Been Tested?
Tested by devkat's grapefruit patrons as well as a 2-week testing period by our NAG patrons, per our usual patron-only release/testing schedule.

Feedback:

> stock RM:
> Upserted to ScannedLocation, 7957 records (upsert queue remaining: 0) in 1.981721 seconds.
> 
> new:
> Upserted to ScannedLocation, 7957 records (upsert queue remaining: 0) in 0.909359 seconds.

> [9:04 PM] -censored-: Jesus H Crust that upsert commit really helped. Changed from 16 db threads and still getting db-queues and eventually locks to 1 db-thread and smooth sailing.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2689,28 +2689,37 @@ def bulk_upsert(cls, data, db):
     # properly escaped. We use placeholders for 
     # VALUES(%s, %s, ...) so we can use executemany().
     row_fields = rows[0].keys()
-    fields = ['`'+conn.escape_string(f)+'`' for f in row_fields]    
     table = '`'+conn.escape_string(cls._meta.db_table)+'`'
-    placeholders = ['%s' for field in fields]
-    assignments = ['`{x}` = VALUES(`{x}`)'.format(
-        x=conn.escape_string(x)
-    ) for x in row_fields]
-    
+
     # Store defaults so we can fall back to them if a value
     # isn't set.
     defaults = {}
 
-    for f in row_fields:
-        if f in cls._meta.defaults:
-            field_default = cls._meta.defaults[f]
-        else:
-            field_default = None
+    for f in cls._meta.fields.values():
+        field_name = f.db_column
+        field_default = cls._meta.defaults.get(f, None)
 
         # peewee's defaults can be callable, e.g. current time.
         if callable(field_default):
             field_default = field_default()
 
-        defaults[f] = field_default
+        if field_default is not None:
+            # Make sure ones w/ a default are in our query.
+            if field_name not in row_fields:
+                row_fields.append(field_name)
+
+        defaults[field_name] = field_default
+
+    # Sort our keys. We'll do the same for row.items() later.
+    row_fields = sorted(row_fields)
+
+    # Assign fields, placeholders and assignments after defaults
+    # so our lists/keys stay in order.
+    fields = ['`'+conn.escape_string(f)+'`' for f in row_fields]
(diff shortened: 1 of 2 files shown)


pogom/models.py
```diff
@@ -2641,23 +2641,122 @@ def clean_db_loop(args):
...
+        x=escaped_field
+    ) for escaped_field in escaped_fields]
+
+    # We build our own MySQL query because peewee only supports
+    # REPLACE INTO for upserting, which deletes the old row before
+    # adding the new one, giving a serious performance hit.
+    query_string = ('INSERT INTO {table} ({fields}) VALUES'
+                    + ' ({placeholders}) ON DUPLICATE KEY UPDATE'
+                    + ' {assignments}')
+
```
friscoMad: Maybe we can benchmark this part as maybe it is not needed but all that meta and peewee classes sounds more or less expensive calls to get always the same result per class so maybe we can cache this string and as we do a lot of inserts it can help.
Also this part maybe fits better in is own function for clarity

pogom/models.py
```diff
@@ -2641,23 +2641,122 @@ def clean_db_loop(args):
...
+    num_rows = len(rows)
     i = 0
+
+    # This shouldn't happen, ever, but anyways...
+    if num_rows < 1:
+        return
+
+    # We used to support SQLite and it has a default max 999 parameters,
+    # so we need to limit how many rows we insert for it.
     step = 250
```
friscoMad: Not sure but maybe we can remove the steps also, I did a small research and it seems that mysql has a 10K parameter limit, we are using 2x each value, but even then it is 5K params and I am not sure what queries can insert so many elements (steps maybe)

pogom/models.py
sebastienvercammen: @friscoMad I agree that we could cache some things per model. I'm not sure how expensive it is exactly, I primarily benchmarked the DB changes. Definitely something we can look into while reviewing.

But I'm not sure that separating it into its own function is the best idea: it's purely to build our query, and it would be irrelevant in any scope outside of the upsert.

pogom/models.py
sebastienvercammen: @friscoMad Per the provider docs:

Parameters limits:
```
Oracle: 64 000
MySQL: 65535
PostgreSQL: 34464
Sqlite: 999
```

We can definitely increase it, that's why I added it as a TODO in the description of the PR.

friscoMad:
Just as information query creation step does take between 0.0002 and 0.0004 seconds, not big deal.

Issue #2458, opened 2018-01-24
Fix for CONSTRAINT_4 doesn't get applied by MySQL


pogo-excalibur:
## Expected Behavior
DB migrations should run without errors on all supported DBs.

## Current Behavior
Running migration 22 (#2416) on a MySQL database produces an error (i.e. the migration doesn't get applied, but the version gets updated so RM still runs) with the error:
```
ProgrammingError: (1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONSTRAINT CONSTRAINT_2' at line 1")
```

Note: the concern here is the error message - the migration gets applied 'correctly'.

## Possible Solution
MySQL doesn't support CHECK constraints, so the migration is not actually required for MySQL DBs.

The error is raised because MySQL also doesn't used the `DROP CONSTRAINT` syntax used in the PR.

It may therefore be necessary to detect the DB type and only apply the migration to MariaDB databases.

Alternatively, as this doesn't cause any particular issues we could just ignore it - most users have probably updated by now, and new users don't have to run DB migrations.

## Steps to Reproduce (for bugs)

1. Have a version of the database prior to v22 (I used v21 for testing)
2. Update to a version which contains #2416
3. Run the DB migration

## Context

It's a relatively small issue - particularly as out docs recommend usinf MariaDB - but I've seen it come up a couple of times in `#help` so there are at least a couple of users using MySQL (and who have therefore not got the fix for the `CONSTRAINT_4` issue).

## Your Environment

MySQL DB, rather than MariaDB.

sebastienvercammen:
Confirmed, but the problem isn't that MySQL doesn't support CHECK constraints. MySQL parses but ignores them.

The problem is that MySQL doesn't know `DROP CONSTRAINT`, causing a syntax error.

Since the constraints aren't applied in MySQL this won't cause any issues with databases, but we should handle it better.

dongemus:
FYI, I got this error migrating to 22 using MariaDB. I have 4 machines running mariadb so I can't remember which it was, but they all have 10.1xx or 10.2xx

sebastienvercammen:
@dongemus I've just tested on MariaDB and it worked properly. Could you reproduce the problem on MariaDB and paste the logs?

dongemus:
InfiniteChaos had it happen on MariaDB as well, he posted about it in Devkat #help 

[...]

PR #2466, opened 2018-01-27
Make db migration transactional

codename-art:
## Description

Make database migration by transactions for each version.  This will avoid corruption of the database on failed migration. It will also allow continue the migration from the correct version in case of migration of several versions.

## Motivation and Context

Exception during migration can lead to schema corruption or data loss. Old migration method set new version value despite exceptions.

## How Has This Been Tested?

Tested on #2458 

### Case 1: failed migration on first version change
New DB version: 23
DB version before migration: 21
Add test value in 22 migration: 
```python
db.execute_sql('INSERT INTO `versions` VALUES (\'test\', \'1\');')
```
Run server:
image
Error rised in migration_22
Versions table:
image
Thats correct, inserted data was rolledback on error.

### Case 2: failed migration on second version change
New DB version: 23
DB version before migration: 21
Add test value 1 in 22 migration:
```python
db.execute_sql('INSERT INTO `versions` VALUES (\'test\', \'1\');')
```
Comment out `DROP CONSTRAINT` in 22 migration

Add test value 2 in 23 migration:
```python
db.execute_sql('INSERT INTO `versions` VALUES (\'test\', \'2\');')
```
And `DROP CONSTRAINT` into 23 migration for failing.
image
Error rised in migration_23.
Versions table:
image
Thats correct, migration from 21 to 23 was sucessefull, but data inserted in migration_23 was rolledback on error.

### Case 3: no errors
New DB version: 23
DB version before migration: 21
Add test value 1 in 22 migration:
```python
db.execute_sql('INSERT INTO `versions` VALUES (\'test\', \'1\');')
```
Comment out `DROP CONSTRAINT` in 22 migration
Add test value 2 in 23 migration:
```python
db.execute_sql('INSERT INTO `versions` VALUES (\'test\', \'2\');')
```
image
image

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2776,9 +2776,6 @@ def verify_database_schema(db):
 
 
 def database_migrate(db, old_ver):
-    # Update database schema version.
-    Versions.update(val=db_schema_version).where(
-        Versions.key == 'schema_version').execute()
 
     log.info('Detected database version %i, updating to %i...',
              old_ver, db_schema_version)
@@ -2786,12 +2783,12 @@ def database_migrate(db, old_ver):
     # Perform migrations here.
     migrator = MySQLMigrator(db)
 
-    if old_ver < 2:
+    def migrate_2(db, migrator):
         migrate(migrator.add_column('pokestop', 'encounter_id',
                                     Utf8mb4CharField(max_length=50,
                                                      null=True)))
 
-    if old_ver < 3:
+    def migrate_3(db, migrator):
         migrate(
             migrator.add_column('pokestop', 'active_fort_modifier',
                                 Utf8mb4CharField(max_length=50, null=True)),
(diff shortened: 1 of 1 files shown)


pogom/models.py
```diff
@@ -3085,10 +3079,58 @@ def database_migrate(db, old_ver):
...
-    if old_ver < 23:
+    def migrate_23(db, migrator):
         db.drop_tables([WorkerStatus])
         db.drop_tables([MainWorker])
 
-    # Always log that we're done.
+    migrations = {
+        2: migrate_2,
+        3: migrate_3,
+        4: migrate_4,
```
sebastienvercammen: This is a no-no: there's no need to define an infinite amount of functions and store them in a `1 = 1` lookup table.

Instead, use an execution context from peewee to wrap a transaction around the migrations.

friscoMad:
I did try to do something similar with the SQLite PR but found that in MySQL DDL statements (table alters, creation, etc) can not be rolled back with a transaction so anyway we can end in an inconsistent state.
It will revert to the previous state if the first statement crash, but if the first one goes ok and the second crashes the first one will not be rolled back and the version will be changed to the old one.

codename-art:
@friscoMad You are right, I forgot that DDL perform implicit commit.
In the current form, rollback is useless.
Only need a correct update of the value in Versions table, for avoid skipping of next migrations, but it is a task for another PR.

PR #2471, opened 2018-01-31
Update custom.js: add default Pokémon to "Hide Pokémon".

OfficialWiddin:
A lot of people have been asking on how to exclude some Pokémon by default, so having it in the wiki is a good idea.

## Description

I've added a description and code for how to exclude Pokémon by default.  
The code was provided by Sebby.

## Motivation and Context

A lot of people are asking for it in #help.

## How Has This Been Tested?

The code was provided by Sebby so I haven't tested it.

## Types of changes

## Checklist:

### DIFF
--- a/docs/extras/custom-js.md
+++ b/docs/extras/custom-js.md
@@ -31,23 +31,21 @@ Setting options in this way forces that setting on page load, so even if a user 
 If you don't want to revert back to what you have set in custom.js every time, you can do it like in the example below.  
 
 * **Add default Pokémon to "Hide Pokémon".**  
-First we need to retrieve what has already been set so we don't overwrite anything excluded in the interface.
+First we need to retrieve what has already been set so we don't overwrite anything excluded in the interface. Then add the Pokémon we want to hide by default **by adding the Pokémon IDs to the excludedPokemon-array, e.g. [13, 16] as shown in the example below**. Finally we have to tell the script to store the value to set it.
+
 ```
-const totalExcluded = Store.get('remember_select_exclude')
-```
-And then add the Pokémon we want to hide by default.
-```
-const excludedPokemon = [13,16] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3])
-```
-Finally we have to tell the script to store the value to set it.
-```
+const totalExcluded = Store.get('remember_select_exclude') // Retrieve what has already been set.
+
+const excludedPokemon = [13, 16] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3])
+
+// If the new default Pokémon doesn't exist then add it to the list.
 excludedPokemon.forEach((id) => {
     if (totalExcluded.indexOf(id) === -1) {
         totalExcluded.push(id)
     }
 })
 
-Store.set('remember_select_exclude', totalExcluded)
+Store.set('remember_select_exclude', totalExcluded) //Tell the script to store the value to set it.
 ``` 
 When you load the map Weedle and Pidgey will be hidden as default.
 
--- a/docs/extras/custom-js.md
+++ b/docs/extras/custom-js.md
@@ -13,7 +13,7 @@ Place your custom code into 'custom.js' in the folder 'static/js'. *Examples are
 Examples for a MOTD, google analytics and disabling scaling icons by rarity are included in custom.js.example!
 The example below is how to set default options usually done by editing core files.
 
-**Set a new map style by default.**
+* **Set a new map style by default.**  
 First we set a variable for it and pick which style we want as default.
 ```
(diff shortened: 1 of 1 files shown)


docs/extras/custom-js.md
```diff
@@ -22,9 +22,33 @@ Next we have to tell the script to store the value to set it.
...
+If you don't want to revert back to what you have set in custom.js every time, you can do it like in the example below.  
+
+* **Add default Pokémon to "Hide Pokémon".**  
+First we need to retrieve what has already been set so we don't overwrite anything excluded in the interface.
+```
+const totalExcluded = Store.get('remember_select_exclude')
+```
+And then add the Pokémon we want to hide by default.
+```
+const excludedPokemon = [13,16] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3])
```
pogo-excalibur: [13,16] -> [13, 16]

docs/extras/custom-js.md
```diff
@@ -22,9 +22,33 @@ Next we have to tell the script to store the value to set it.
...
+Setting options in this way forces that setting on page load, so even if a user changes the setting it will revert back to what you have set in custom.js every time, keep this in mind when forcing settings.  
+
+If you don't want to revert back to what you have set in custom.js every time, you can do it like in the example below.  
+
+* **Add default Pokémon to "Hide Pokémon".**  
+First we need to retrieve what has already been set so we don't overwrite anything excluded in the interface.
+```
+const totalExcluded = Store.get('remember_select_exclude')
+```
+And then add the Pokémon we want to hide by default.
```
sebastienvercammen: Don't split up the code, you're expecting everyone to understand that they need to combine them in the proper order. Unfortunately, that's something we know we shouldn't expect. Always assume the person that's reading the docs has zero code knowledge.

It will be easier to understand if you give them a full block of code they can copy, and explain what field they have to change. An example block of code with code comments, and some surrounding text to explain what the code does.

docs/extras/custom-js.md
```diff
@@ -22,9 +22,31 @@ Next we have to tell the script to store the value to set it.
...
+const excludedPokemon = [13, 16] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3])
+
+// If the new default Pokémon doesn't exist then add it to the list.
+excludedPokemon.forEach((id) => {
+    if (totalExcluded.indexOf(id) === -1) {
+        totalExcluded.push(id)
+    }
+})
+
+Store.set('remember_select_exclude', totalExcluded) //Tell the script to store the value to set it.
```
sebastienvercammen: Space

> //Tell

docs/extras/custom-js.md
```diff
@@ -22,9 +22,31 @@ Next we have to tell the script to store the value to set it.
...
+* **Add default Pokémon to "Hide Pokémon".**  
+First we need to retrieve what has already been set so we don't overwrite anything excluded in the interface. Then add the Pokémon we want to hide by default **by adding the Pokémon IDs to the excludedPokemon-array, e.g. [13, 16] as shown in the example below**. Finally we have to tell the script to store the value to set it.
+
+```
+const totalExcluded = Store.get('remember_select_exclude') // Retrieve what has already been set.
+
+const excludedPokemon = [13, 16] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3])
+
+// If the new default Pokémon doesn't exist then add it to the list.
+excludedPokemon.forEach((id) => {
```
pogo-excalibur: I'm not sure we use this new-style function declaration anywhere else.

sebastienvercammen: @pogo-excalibur Ah yes my personal habit is to use ES6, that's how this creeped into this example when I pasted it in #help on our Discord.

I'm open for any thoughts on this, I'm biased on this matter. I'll invite the rest on Discord to contribute their thoughts as well.

pogo-excalibur: I personally prefer the new style, but I don't use it when working on RM code for consistency.

friscoMad: I am pro ES6, we have babel to handle the transpiling and code is a lot cleaner.
I am not sure if we have arrow functions on other parts of map.js but I think that I added some ES6 code when helping sirlindqvist with the new UI, anyway map.js needs a big rewrite.

OfficialWiddin: So should i change it so it looks like the other functions or leave it as it is now?

fosJoddie: I say the example should be in this new style as long as the syntax is valid, and then aim to move the rest of the code towards that same more modern style of coding in future rewrites etc.

PR #2475, opened 2018-02-01
Generation Addition and added to pokemon label window

BrownSlaughter:
This adds generation info to RM and adds a small label to the pokemon label window

## Motivation and Context
This PR is a precursor to things like "exclude by gen" and "generation % on stats page"

## Screenshots (if appropriate):
gitgen
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2476, opened 2018-02-01
Improve look of Pokémon marker label

daangroot:
## Description
Small improvement of style and formatting of the Pokémon marker label.
See screenshots below for the changes made.

## Motivation and Context

I  did not really like the look of the label, so I thought lets make it more beautiful  :)

## How Has This Been Tested?

Tested on my main setup.

Old:
old

New:
new

## Types of changes

## Checklist:


darkelement1987:
I like it! The way it shows IV's is much better like this.

ghost:
I like this, minimal and easier to read, how would it account for the recent PR #2475 (adding gen info to marker labels), possibly next to level?

Alderon86:
@daangroot add `margin-top: 5px;` to the # nav part to make the changes @sebastienvercammen suggested for fat fingers.
Without Iv:
noniv
With Iv:
iv

daangroot:
@sebastiaanvercammen I don't see why a top margin would help. There is nothing above it that could be pressed, so nothing can go wrong.

Alderon86:
@daangroot u could press something like exclude or notify for e.g.

daangroot:
Oh yes sorry, did not think about the no iv label.

daangroot:
I added a 5px top margin to the navigate class as suggested.

daangroot:
@Markair I think it's best tahat

static/sass/layout/_pokemon.scss
```diff
@@ -79,11 +79,13 @@
   }
 
   &.navigate {
+    position: absolute;
```
Alderon86: why do we need that? im not a fan of `position: absolute`.

static/sass/layout/_pokemon.scss
```diff
@@ -94,6 +96,6 @@
     display: block;
     height: 48px;
     width: 48px;
-    margin-bottom: 11px;
+    margin-bottom: 15px;
```
Alderon86: Is there a reason why we would have to increase it?
Also, can u provide a screenshot with the difference compared to my simple change?

dripdropper:
You should consider leaving the ADS labels on the IVs to eliminate confusion.  Many (not all) tools use ADS, but the in-game appraisal uses SAD.
Otherwise it looks like good, especially the size reduction, which helps to keep things on-screen.  It's too bad the Google marker API is too stupid to place the InfoWindow in a direction that keeps it on-screen.

daangroot:
@dripdropper I removed them because it looked better, but I can understand that it might cause confusion, so I will add them back.

static/sass/layout/_pokemon.scss
daangroot: That's how it was done before. It can be done differently but this is the easiest way.

static/sass/layout/_pokemon.scss
daangroot: To align the links on the left with the location link. I think it looks better this way.
image

fosJoddie:
nice changes, looks neat and triggers no warning or epileptic episodes

jaake77:
Werks Gud. Thanks for your efforts.

sirdemoncze:
Thanks. It works perfect 👍.

billyjbryant:
Looks good and works.

static/js/map.js
```diff
@@ -506,7 +506,7 @@ function initSidebar() {
 }
 
 function getTypeSpan(type) {
-    return `<span style='padding: 2px 5px; text-transform: uppercase; color: white; margin-right: 2px; border-radius: 4px; font-size: 0.6em; vertical-align: middle; background-color: ${type['color']}'>${type['type']}</span>`
+    return `<span style='padding: 2px 5px; text-transform: uppercase; color: white; margin-right: 2px; border-radius: 4px; font-size: 0.75em; vertical-align: middle; background-color: ${type['color']}'>${type['type']}</span>`
```
sebastienvercammen: Move all the inline CSS to the stylesheet, and remove the `getTypeSpan` function. Since the HTML will be much shorter, it'll be clearer if it's no longer a separate function.

daangroot:
Most up-to-date look of the label:
image

billyjbryant:
Looking at the Type box code and testing locally

billyjbryant:
Adding `padding-top: 0.45em` (or `padding-top:2.25%`) on the .type element fixes it for mobile browsers but breaks on desktop.  this could probably be fixed by adding the necessary overrides to `mobile.scss`

daangroot:
@sebastienvercammen I've tried a lot of things to make the type text centered, but I couldn't fix it.

daangroot:
Maybe we can merge it and fix it afterwards. I can make an issue for this problem, so other people are aware of it and can maybe fix it.

sebastienvercammen:
@daangroot We won't merge without the fix. The PR is instead tagged as "help wanted".

daangroot:
Hopefully the final look:

image

PR #2479, opened 2018-02-06
Status Page: Add account busy/idle to instances.

Levonestral:
## Description

Adds a busy/idle accounts count display on status screen.

## Motivation and Context

Although many people already know their usage for accounts per instance, those that are just starting may or may not know how many they need. 

In the status screen you already see the display of "Nothing to scan" which shows that account is idle allowing you to determine how many aren't being used to their full potential. Assuming this statement is correct, I've made these changes to allow a quick glance at your account usage vs idle states so you can more easily make adjustments.

## How Has This Been Tested?

Tested on smaller map. 10 instances, 400 accounts.

image

## Types of changes

## Checklist:


dripdropper:
Can/should this be changed to show level 0-29 and level 30+ accounts separately, considering the added costs and value of level 30+ accounts?

Levonestral:
That's a good question and I fully agree that additional information would be of great use.

High level accounts can be included in the regular accounts list, but can also be included through the separate list of files (typical). That second list of accounts is not managed the same way as the others and isn't currently (best I can tell) not displayed in the status window at all. (not even their counts)

I'm not sure (at this point) it would be worth the effort to attempt to make the changes needed to include those high level accounts in the status page right now, with the accounts-in-database PR #2424 potentially offering much better access to that information in the near future?

I think there are several options:

1. Move forward with this PR "as is" displaying and counting only the regular accounts, but revisit the display once the other PR is completed.
2. Postpone, or potentially delete this PR and revisit the whole concept once we get better understanding of what is available later.
3. Go ahead now with an attempt at changes that would allow for the display of those high level accounts and hope that there aren't too many conflicts/redundency with 2424 later.

I'm good with any of the options but figured it was worth additional thought.

dripdropper:
Thanks for the thoughtful response.  I am also fine with any option.
If the code is working well, perhaps the simplest option is to move forward with this PR "as is".  It's hard to say when a PR as complex as #2424 will be complete.  This code seems quite simple, so it can probably be changed or removed easily whenever #2424 is accepted.

sebastienvercammen:
👍 For now, we're keeping the PR as is to avoid conflicts with other PRs that change the architecture. Merging as is will already make monitoring scanning slightly easier.

neskk:
I've looked at this and I think its a good idea to have this stats although I don't like the way stats are obtained (looking at status message). I may add something like this in the future to #2424 meanwhile I think it's OK to have this.
In the future I think it would be great to have dynamic worker allocation, meaning that the instance would adjust its worker count accordingly to how many spawnpoints, etc.

static/js/status.js
```diff
@@ -369,10 +371,16 @@ function getStats(i, worker) {
     elapsedHours = elapsedSecs / 3600
 }
 
+function getAccountStats(i, account) {
+    idle += (account['message'] === 'Nothing to scan.') ? 1 : 0
+    busy += (account['message'] !== 'Nothing to scan.') ? 1 : 0
+}
```
sebastienvercammen: I would move away from this type of "side effect counting" (using a callback method to update a global). Just do it the traditional way:
```
// result.workers contains the accounts
const idle = result.workers.reduce((accumulator, account) {
    if (account['message'] === 'Nothing to scan.') {
        accumulator += 1
    }
}, 0)
const busy = result.workers.length - idle
```
Just an example. Of course you can also separate the reducer from `reduce` for readability, use a regular counting loop, ...

Levonestral: Good idea. I've made the updates.

tomballgithub:
Since you are making this line longer I would go ahead and reduce the number of decimal places on the two captcha areas to compensate.  Do it like the 'all instances' just above it in your photo  $0.0 instead of $0.000000.  Goal is to keep this from getting too long.

Levonestral:
Good suggestion, although I'm hesitant in making a change to something that was there for a specific reason. That said, the total up top has that applied, so if others agree, I can certainly make the change:

```
        'Captchas: {} ({:.1f}/hr)|${:.5f}/hr|${:.3f}/mo | ' +
```

to

```
        'Captchas: {} ({:.1f}/hr)|${:.1f}/hr|${:.1f}/mo | ' +
```

tomballgithub:
I think it is a holdover from long ago when captcha costs were a given and
highly scrutinized.

Levonestral:
Ah works for me then. I've made the change. Thanks!

pogom/search.py
```diff
@@ -655,27 +656,39 @@ def get_stats_message(threadStatus, search_items_queue_array, db_updates_queue,
...
-        'Fails: {} ({:.1f}/hr) | Empties: {} ({:.1f}/hr) | ' +
-        'Skips {} ({:.1f}/hr) | Captchas: {} ({:.1f}/hr) (${:.1f}/hr, ' +
-        '${:.1f}/mo) | Elapsed: {:.1f}h'
-    ).format(overseer['active_accounts'], overseer['success_total'], sph,
+        'Total active: {}, busy: {}, idle: {} | ' +
+        'Success: {} ({:.1f}/hr) | ' +
+        'Fails: {} ({:.1f}/hr) | ' +
+        'Empties: {} ({:.1f}/hr) | ' +
+        'Skips {} ({:.1f}/hr) | ' +
+        'Captchas: {} ({:.1f}/hr)|${:.1f}/hr|${:.1f}/mo | ' +
```
sebastienvercammen: Although it's good that you've shortened them, currencies should be displayed with a precision of two decimals.

static/js/status.js
```diff
@@ -371,8 +373,9 @@ function getStats(i, worker) {
 
 function addTotalStats(result) {
     var statmsg, title
-
     active = 0
+    busy = 0
+    idle = 0
     success = 0
```
sebastienvercammen: We should move away from non-descriptive/incomplete variable names.

Instead of `busy` and `idle`, consider
```
accounts_busy_count = 0
accounts_idle_count = 0
```
Feel free to be creative with the old variable names, they should be changed to be more descriptive as well.

pogom/search.py
Levonestral: Yeah, I was thinking that too, but based it off of the top header one only having 1 decimal point. I'll change it back.

static/js/status.js
Levonestral: Sounds good to me, was following the original "style" of the naming, but can make them more descriptive for sure.

Levonestral:
Changes completed. As per the commit message, I moved mine back to local method and added "scans" as a prefix to some of the others. I also decided to maintain the camel case format for the naming conventions to match the rest of the file.

tomballgithub:
You need to rebase.  You just pulled in a ton of unrelated commits

sebastienvercammen:
Agreed with @tomballgithub, you need to rebase (use interactive rebase), the PR just blew up.

Levonestral:
Sorry, I was rebasing my develop fork and must have broken this by accident. It should be all fixed now.

PR #2414, opened 2017-12-27
Weather

javisabalete:
## Description

Display weather image next to types on map. This PR is based on [sLoPPydrive's](https://github.com/sLoPPydrive/RocketMap) fork. Images from ~~[ZeChrales's](https://github.com/ZeChrales/PogoAssets) PogoAssets~~ [AdamWhitcroft's](http://adamwhitcroft.com/climacons/) website and extreme condition icon from https://commons.wikimedia.org/wiki/File:Ei-exclamation.svg

## Motivation and Context

since a few days ago, pokemons are affected by weather condition. That's very interesting to show on map.

## How Has This Been Tested?

local map

## Screenshots (if appropriate):
captura de pantalla 2017-12-27 a las 22 41 41
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2482, opened 2018-02-07
Just sent weather boosted condition in webhook

Sttug:
Pokealarm and others are enable to manage weather boosted condition, this just send weather info for any pokemon.

## Description

Get weather boosted condition in pokemon encountered and send it in webhook information.

## Motivation and Context

I dont use map only see pokemon info in pokealarm.

## How Has This Been Tested?

my own server 

https://screenshotscdn.firefoxusercontent.com/images/616b2635-e9ed-49ac-8b8f-b276d3cf434a.png
https://screenshotscdn.firefoxusercontent.com/images/00ae2679-89f9-4fe6-8c5d-9e3ec1854fde.png

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [  ] My change requires a change to the documentation.


michikrug:
In PokeAlarm this field is called 'boosted_weather'. So maybe we should stick to that.

billyjbryant:
#2414 also sends boosted weather webhooks but uses a different field name to do so.  I would also 2nd the suggestion to set the field to `boosted_weather` to match what PA already listens for.  Makes adoption and compatibility between systems easier to support.

sebastienvercammen:
Duplicate. Should be done in #2414.

@billyjbryant As a PA contributor, in the future PA should PM us about any requests like this one. We won't always be able to adopt the same structure/naming as other projects, so this will help to avoid conflicts. There are now 3 different webhook server projects that I know of, each with conflicting implementations.

Issue #2470, opened 2018-01-31
Rarity Exclusion first displays pokemon, then removes them


wfcook:
## Expected Behavior
When the "Exclude Rarity" option is used, e.g. to exclude pokemon that are "Uncommon and below", those pokemon should never be drawn.

## Current Behavior
RocketMap first displays all pokemon and then removes the excluded rarities en masse. This makes panning the map awkward and slow.

In contrast, when individual pokemon are excluded Via the "Hide Pokemon" feature, they are never drawn.

## Possible Solution
Move the check for rarity exclusion to the same location as the check for individual exclusion?

## Context
The "exclude by rarity" feature is potentially much more useful now with the new dynamic rarity, but currently runs slowly and awkwardly.

## Your Environment
* Version used: latest
* Python 2.7
* Windows 10

sebastienvercammen:
@wfcook 

`updateMap`: https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1914
* Loads data, then
* [calls `processPokemons`](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1916)
* [which calls `processPokemonChunked`](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1606)
* [which calls `processPokemon`](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1628)
* [which finds that the Pokémon is excluded by rarity](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1679) and so [it returns no markers for the Pokémon](https://github.com/RocketMap/RocketMap/blob/develop/static/js/map.js#L1675).

Maybe you can include a screen recording of the behavior issue you're having?

wfcook:
Will work on it

wfcook:
https://www.youtube.com/watch?v=IplFGdpOOQ8

Video quality is terrible, but you'll see what I mean

wfcook:
@sebastienvercammen see video above

BrownSlaughter:
I can confirm it does draw the pokemon then removes them afterwards, when I load the map it draws everything then removes common

trunksbomb:
The problem is that `excludedPokemonByRarity` is never sent to the server during the request for new Pokemon. So what happens is the server returns the entire list of available Pokemon (minus the ones you have explicitly excluded) to the client, then the client parses all the received Pokemon (puts them on the map) then filters them out, which causes the blinking.

[...]

PR #2489, opened 2018-02-10
Fix Excluded By Rarity to work with new Dynamic Rarity changes

trunksbomb:
excludedPokemon is now used as a global list of excluded Pokemon, made up of any number of individual exclusion lists (rarity and hidden, for now)
In accordance with the above two points, refactored all code that controls excluding Pokemon, including adding new functions to help add/remove from list
Fixes Issue #2470

## Description
Exclude By Rarity was introduced prior to the Dynamic Rarities change, which changed the way the excluded Pokemon list was sent to the server. This change broke Exclude By Rarity as described in Issue #2470 because the excluded Pokemon by rarity were not sent to the server and so the server sent back the entire list of nearby Pokemon.
Additionally, this PR sets the stage for additional exclusion lists to be easily implemented (such as by generation, or by weather boosted, etc).

## Motivation and Context

Fixes issue #2470 
New functions abstract excluding/including away from the main logic of decided what to exclude/include.
Moving forward, it should be easier to integrate new exclusion lists.

## How Has This Been Tested?

Tested on my live city map. Can switch between all rarity tiers reliably (although reducing the rarity filter doesn't repopulate quickly, moving the map will cause an update to reinclude Pokemon) and exclude/include individual Pokemon.
Map still loads all visible Pokemon on initial load, which can be a lot if clustering is disabled, because the first fetch request goes out before the rarity filter is processed.

## Types of changes

## Checklist:


sebastienvercammen:
Author is unresponsive, so I rebased and pushed a fix for Travis. Note: this doesn't serve as my review or approval.

PR #2498, opened 2018-02-18
[WIP/RFC] Gym Scan Scheduler

fosJoddie:
Implements a gym scan scheduler
## Description
Currently, the "best" way to mass scan gyms for raid alerts is to use hex scan in tandem with a carefully crafter geofence around the gyms involved. 

This is very tedious work and might even skip a gym or two if you are experimenting with tiny geofences. 

**WARNING** 
Since gyms can be scanned at any speed, this scheduler will most likely not be able to spin pokestops and make fresh accounts level 2.  This means you have to prepare accounts by using the levelup tool, or use already shadowbanned 2nd hand accounts.

This scheduler asumes that you have discovered all gyms in your area by using a traditional hex scan or speed scheduler first. Any gyms discovered in the process of scanning, won't be clustered until next restart of the instance.

TODOS:
-Currently the scheduler clusters first and geofences after. This might break stuff in some very weird edge cases. Logic has to be turned the other way
- Documentation for this new scheduler will be written if it's feasible to have such a scheduler in RM.

***Introducing*** 
The gym scan scheduler
- Finds any gym already in your database
- Clusters the gyms to optimize search area
- Supports geofencing to narrow down area
- uses scan-delay with # of workers to adjust how often all gyms are scanned.

## Motivation and Context
With the current short lifespan of accounts, this scheduler allows for a minimum of accounts to scan for 
raids.

## How Has This Been Tested?
Running for a week with little to no explosions

## Types of changes

## Checklist:


The-Real-Ketchum-Dev:
I like this option.. Will this have an option in the config sort of like spawn point scan or speed scan?

friscoMad:
Seems ok, my 3 comments:

- Spawnpoint cluster maybe can extend this cluster system, to avoid having mostly duplicated code (I know it is not the same as spawnpoint need to check appear/disappear times).
- Geofence before clustering, otherwise it will scan gyms outside the fence, it can happen anyway but in a lesser extent if done in this order.
- Maybe it is the right time to have a single flag with different values for the scheduler, you can not set several at the same time and some schedulers are defined by strange flags (skip if empty) so it will be a lot clearer if we had -scheduler: speed or -scheduler: hex

fosJoddie:
Yeah I looked at reusing existing cluster code. Can probably just reuse the same code from Spawnpoint Scan. 

Gf first is coming. 

I did not want to start redoing the logic of how schedulers are selected. That would make this pr a breaking change. I’m not against doing that tho, naturally.

SkOODaT:
thier was a fort only scanning mode PR created allready.... ive been using it to scan stops and gyms for months and its worked fine lol

https://github.com/RocketMap/RocketMap/pull/1957

i removed the need for dumping but to scan just stops and gyms this works great.... i use it to lure

fosJoddie:
@SkOODaT that PR was closed a long time ago, and used a somewhat different approach. Not sure what lures has to do with scanning gyms. The intent for this PR is to discover raids, and little more.

SkOODaT:
@fosJoddie the other PR was able to do forts in general so gyms for raids or gym data and pokestops for lured pokemon/luring stops for those who have it would be nice if this PR did ALL forts like the other one but just a suggestion

michikrug:
Also running #1957 for a really long time now. It works okay. But I do see a lot of space for improvement. Therefore, I am looking forward to see how this approach develops.

fosJoddie:
@SkOODaT  luring stops will never be a part of RM, it's directly interrupting the game economy in a negative way. 

Sorry for tagging you wrongly @The-Real-Ketchum-Dev  - my bad

The-Real-Ketchum-Dev:
I never assumed anything about lure i was asking about gym scanning

The-Real-Ketchum-Dev:
Sorry only seen the email and cut it off.. All good bud

SkOODaT:
@fosJoddie i never said anything about adding luring to this lmao but fort scanning all forts is beneficial in other ways, such as gathering stop information.....

fosJoddie:
@friscoMad  I started reworking the scheduler flags to use the single `-scheduler`. My only mystery guest is `skip-empty` which I feel like is a bad name for "hex scan minus the parts with no results". 

Code now reuses the clustering from spawnpoint - I'm not sure if it broke anything with spawnpoint scheduler.

Geofencing now happens before clustering

friscoMad:
Not tested but the code seems to be good, thanks for taking the effort to refactor clustering and looking at `-scheduler` I agree that it should be in a another PR. 
Skip-empty could be renamed to hex-noEmpty or something like that, right now that flag it is the worst one as no one identifies that flag with a scheduler and it takes precedence over other scheduler flags.

fosJoddie:
I felt like the extra work to make another pr for `-scheduler` was not worth it. Can split this into to PR if the scope is now too big.

PR can now scan for pokestops as well as gyms.
PR now has new argument: `-scheduler` that inputs what scheduler you want to use

TODO: 
I have a feeling `-bh` is not going to be friends with this scheduler

pogom/utils.py
```diff
@@ -299,19 +299,14 @@ def get_args():
...
                         action='store_true', default=False)
-    parser.add_argument('-ss', '--spawnpoint-scanning',
-                        help=('Use spawnpoint scanning (instead of hex ' +
-                              'grid). Scans in a circle based on step_limit ' +
-                              'when on DB.'),
-                        action='store_true', default=False)
+    parser.add_argument('-sch', '--scheduler',
+                        help=('Set scan scheduler to use. One of: '
+                              'hex, spawnpoint, speed, gym'),
+                        type=str, default='hex')
```
friscoMad: Add choices to force validation as -wh-types

friscoMad: --skip-empty is no longer needed

pogom/utils.py
```diff
@@ -780,14 +775,25 @@ def get_args():
...
-        else:
+        elif args.scheduler == 'fort':
+            args.scheduler = "FortSearch"
+            if args.beehive:
+                print(sys.argv[0] + "Beehive is not compatible with "
+                                    "Fort Scheduler!")
+                sys.exit(1)
+        elif args.scheduler == 'hex':
             args.scheduler = 'HexSearch'
+        else:
```
friscoMad: Not needed if choices is added to the add argument.
And maybe the valid values should be the full name of the scheduler, that saves us for this translation code and makes configuration more inline with the docs.

pogom/utils.py
friscoMad: Probably we should also change default to SpeedScan

fosJoddie:
Changed argument parser to use real names of scanners.

SpeedScan is now the default - gonna break configs that run without any scheduler flags.

I won't remove skip-empty, since that would be removing a full scheduler that I guess some people might be using. it is named **HexSearchSpawnpoint** now.

The names of the schedulers could be shortened a whole lot tho, since they are only referenced on Sceduler creation and a few odd places in code. Not sure what best course of action would be

friscoMad:
Umm just saying that we should remove the `--skip-empty` flag now that we have `-scheduler HexSearchSpawnpoint`

fosJoddie:
Ohh. Then we are in complete agreement. 

The flag is removed. 👌

darkelement1987:
Is this ready to be merged?

fosJoddie:
@darkelement1987 it's ready to be tested - I just rebased and fixed conflicts

billyjbryant:
@fosJoddie, needs to be rebased again.  Once it is I can test it.

fosJoddie:
@billyjbryant  thanks for the heads up, it has been rebased back into former glory

PR #2502, opened 2018-03-01
Replace weather images with emojis.

fosJoddie:
## Description
Replace pixely weather icons by pretty and scalable emojis for displaying weather boosted conditions on pokemon.

## Motivation and Context
Requested feature

## How Has This Been Tested?
Tested on local instance

image

## Types of changes

## Checklist:


static/sass/layout/_pokemon.scss
```diff
@@ -37,6 +37,11 @@
     font-size: 11px;
   }
 
+  &.weather-boosted {
```
pogo-excalibur: I think I'd prefer to use something like 'weather-boost' or 'weather-condition'.

But we don't have a consistent short-form name for this yet, so 🤷‍♂️ .

PR #2506, opened 2018-03-02
Updating Webhook Field for Boosted Weather

billyjbryant:
## Description
Re-present the webhook field for `weather_boosted_condition` as `boosted_weather` 

## Motivation and Context
PokeAlarm was updated in 3.6 to accept the boosted weather condition via webhooks using the `boosted_weather` field name.  This was to conform with multiple scanners who were already presenting this information in this way.  This PR simply conforms RM to also send this data in this manner, while also allowing it to be sent as `webhook_boosted_condition` to match the official field name. 

## How Has This Been Tested?
This was tested locally on a dev instance. 

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2072,7 +2072,8 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
                         'seconds_until_despawn': seconds_until_despawn,
                         'spawn_start': start_end[0],
                         'spawn_end': start_end[1],
-                        'player_level': encounter_level
+                        'player_level': encounter_level,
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
I'd agree with @sebastienvercammen here - this seems like something that should be handled by webhooks, rather than expecting RocketMap to fit in with everyone else's way of doing things.

In general (unless there is a **signifiant** cost involved) I would not expect RocketMap to change to support projects that want to work with RocketMap alternatives.

billyjbryant:
This PR is being closed as it goes against RM's guidelines for code quality.

PR #2508, opened 2018-03-02
Add DB execution context for gym_park.

friscoMad:
## Description
Just adds a context to the query to release the connection to the pool

## Motivation and Context
This query is done inside the worker threads and it was not using the context so it was not releasing the connection to the pool, that could cause 'too many conections opened' again.

## How Has This Been Tested?

Not tested I have no map anymore nor keys to test it.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -638,10 +638,11 @@ class Gym(LatLongModel):
 
     @staticmethod
     def get_gyms_park(id):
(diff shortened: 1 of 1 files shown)


friscoMad:
triggering the error is not easy but executing `SHOW FULL PROCESSLIST;` during normal operation with and without the PR should show easily if the number of connections are reduced.

PR #2509, opened 2018-03-02
Only spin stops with levelup script

friscoMad:
## Description
Filter forts to only spin stops

## Motivation and Context
Levelup as pointed by @fosJoddie was spinning both stops and gyms but it was using the stops method to spin gyms (calling a get_fort_info) that is not correct and never intended, we should filter to only spin stops.

## How Has This Been Tested?

Not tested I don't have map or key anymore.

## Types of changes

## Checklist:

### DIFF
--- a/tools/levelup.py
+++ b/tools/levelup.py
@@ -45,7 +45,8 @@ def get_location_forts(api, account, location):
     cells = response['responses']['GET_MAP_OBJECTS'].map_cells
     forts = []
     for i, cell in enumerate(cells):
-        forts += cell.forts
+        if fort.type == 1:
+            forts += cell.forts
     if not forts:
         return (ErrorType.no_stops, None)
     return (None, forts)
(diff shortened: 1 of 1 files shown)


tools/levelup.py
```diff
@@ -45,7 +45,8 @@ def get_location_forts(api, account, location):
     cells = response['responses']['GET_MAP_OBJECTS'].map_cells
     forts = []
     for i, cell in enumerate(cells):
-        forts += cell.forts
+        if fort.type == 1:
```
sebastienvercammen: * Variable `fort` doesn't exist.
* Add a comment to clarify what `1` means.

tools/levelup.py
```diff
@@ -45,7 +45,10 @@ def get_location_forts(api, account, location):
     cells = response['responses']['GET_MAP_OBJECTS'].map_cells
     forts = []
     for i, cell in enumerate(cells):
-        forts += cell.forts
+        for fort in cell.forts:
+            # Only use stops.
+            if fort.type == 1:
+                forts += cell.forts
```
sebastienvercammen: You're iterating over each `fort` in the list `cell.forts`, and if any of the forts is a Pokéstop, you add the entire list.

I think you meant to do
```
forts = []
...
for fort in cell.forts:
    if fort.type == 1:
        forts.append(fort)
```

PR #2512, opened 2018-03-03
Fix TTH count for one-hour spawns.

### LINKED ISSUE #2507: TTH Search Never Finishes
In January, PR #2416 changed the constraint for recording spawn timers in the RM DB. Since then, the TTH calculations on the Status page were not accurate. If you look, there may be a few TTHs that are yet to be found. This is caused by a spawn that has 'latest_seen' = '3600' and 'earliest_unseen' = '0' instead of being equal to each other.

The calculations done in schedulers.py would count that as a missing TTH because it is looking for spawns that have 'latest_seen' = 'earliest_unseen' but it shows in the DB as an hour long spawn ('kind' = 'ssss'). The log file notes them as 0 minute spawns "[   search-overseer][    schedulers][    INFO] Spawnpoints with a 0m range to find TTH: xx". 

Also noticed that the the spawnpointdetectiondata table does not record a tth_secs vaule for these spawnpoints. I looked through most of mine and saw that they stopped checking for a tth after 5 scans. I have reset the missed_count before noticing this so I am not sure if that is related. There is a mismatch between the last scan time in spawnpoints and spawnpointdetectiondata. So it probably stopped checking for a TTH once it marked 'latest_seen' = '3600' and 'earliest_unseen' = '0'.

Pokemon do show in these spots on the map every hour (I haven't verified if they are there in person yet). The spawnpoint does not note "Spawn times not yet determined. Current guess...". I once saw two pokemon on the same spawnpoint meaning these are still missing TTH.

image

image

image

image

image

I would like to hear if other people noticed this behavior also.

## Expected Behavior
The status page should show 100% TTH found once the TTH search is completed. The spawnpoint data for 'latest_seen' and 'earliest_unseen' should be equal once a TTH is found.

## Current Behavior
The status page should show < 100% TTH found because of the mismatching 'latest_seen' and 'earliest_unseen' variables. RM no long checks for a TTH on these spawns once 'latest_seen' = '3600' and 'earliest_unseen' = '0'.

## Possible Solution
Not sure of the best way to correct this. The issue could be that the tth_secs vaule is not being recorded before the spawn is marked with 'kind'='ssss'. Or that the 'kind' should be 'hhhs' since the log notes this as a spawn with less than 1 minute.

1. Rescan an area that used to suffer from the u_constraint error.
2. Wait to finish the initial and TTH scans (TTH may show a few missing and never complete).
3. Query the spawnpoints within that area for 'latest_seen' = '3600' and 'earliest_unseen' = '0'.
4. Query the spawnpointdetectdata table for those spawn IDs to see the last time to see if it stopped checking for a TTH.

## Context
Spawns still show in those areas but I am not sure if their timers are accurate. I rarely have people report that the timers are off but there was one case that a spawn was on my map with 20 minutes left and it had already despawned. I did not check into this since it was only one time but it could have been on one of these spots.

My current scanners do not schedule TTH searches anymore so these spots are just stuck in their current state.

This affects 46 out 58,000 spawnpoints (0.1%) in my database to give you an idea of how rare these outcomes are.

## Your Environment
RM pull to PR #2505 (latest)
MariaDB-10.2.11
Schema version 28
HiediSQL 9.4.0.5215
Ubuntu 17.04

sebastienvercammen:
## Description
* Fixed comparison to detect TTH found, fixing the TTH count.
* Added some extra newlines for readability.
* Edited one log method that used `format()` instead of the logger.

## Motivation and Context
Fixes #2507.

## How Has This Been Tested?
Untested.

## Types of changes

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -1285,11 +1285,11 @@ class SpawnPoint(LatLongModel):
 
         return list(spawnpoints.values())
 
-    # Confirm if tth has been found.
+    # Confirm if TTH has been found.
     @staticmethod
     def tth_found(sp):
-        # Fully indentified if no '?' in links and
-        # latest_seen == earliest_unseen.
+        # Fully identified if no '?' in links and
+        # latest_seen % 3600 == earliest_unseen % 3600.
         # Warning: python uses modulo as the least residue, not as
         # remainder, so we don't apply it to the result.
         latest_seen = (sp['latest_seen'] % 3600)
--- a/pogom/schedulers.py
+++ b/pogom/schedulers.py
@@ -758,6 +758,7 @@ class SpeedScan(HexSearch):
                     len(filter(lambda e: e['kind'] == 'spawn', Scanned_list))
                 spawns_missed = len(
                     filter(lambda e: e['kind'] == 'spawn', Missed_list))
+
                 band_percent = self.band_status()
                 kinds = {}
                 tth_ranges = {}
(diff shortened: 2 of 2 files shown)


Kneckter:
Tested this on my local database that had 1 hour spawns from the noted issue. The code changes worked to accurately display the TTH Found percentage on the Status page.

PR #2513, opened 2018-03-03
Update gender on encounter.

### LINKED ISSUE #2480: Scanned genders are inaccurate
Long story short, there is a discrepancy between the gender returned by GetMapObjects, and the actual gender you get when starting the encounter.

When using IV scanning accounts, the proper gender should easily be retrieved and the record updated - however, it is still using the original gender from the GMO response.

For non-IV scanned species there's no way to ensure accuracy, but when our local Discord server posts scanned results from the RM database, it's still using the original genders as returned by GMO, and not the true ones.

More information not pertinent to RM can be found at my initial write-up [here](https://www.reddit.com/r/TheSilphRoad/comments/7vl7ve/observation_pokémon_change_genders_when/)

sebastienvercammen:
## Description
Update Pokémon's gender if we encounter it.

## Motivation and Context
Fixes #2480.

## How Has This Been Tested?
Untested.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/models.py
+++ b/pogom/models.py
@@ -2055,7 +2055,8 @@ def parse_map(args, map_dict, scan_coords, scan_location, db_update_queue,
                     'height': pokemon_info.height_m,
                     'weight': pokemon_info.weight_kg,
                     'cp': pokemon_info.cp,
-                    'cp_multiplier': pokemon_info.cp_multiplier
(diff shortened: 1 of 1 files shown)


tomballgithub:
Approved via code review only.  I have been running a similar change for weeks with no more reports of incorrect gender.  I reviewed models.py and your one line change is correct.

PR #2517, opened 2018-03-04
Re-add #2513. Confirmed working.

sebastienvercammen:
Too silly to waste more words on this mess...

#2516 
#2513 

Details happened on Discord.

One user had modified RM files to use some shady, half-broken paid L30 scanning service, causing this update to break his setup.

The one user posts in our help channel because he thinks we broke things. Two others jump on the occasion to tell us they have the same problem.

We react quickly and revert the original commit.

2 hours of debugging later, it turns out neither of the three users had the same problem. None of them had any problems besides a serious case of [pebkac](https://en.wikipedia.org/wiki/User_error).

Let's act like this never happened. 👍


tomballgithub:
Approved again

PR #2519, opened 2018-03-05
Improvements to level-up script.

sebastienvercammen:
## Description
* Yay whitespace.
* Removed unused variables/unnecessary enumerate.
* Moved some code around to have some structure.
* Changed script to only spin a single pokéstop, not all of them. This should have been the case from the first version as we had agreed on, but somehow the previous version got through.
* Added a test to see if any pokéstops are in spinnable range. If none are in range, the script will exit and display a clear message.
* If no pokéstops are in spinnable range, the script will output the location of one of the Pokéstops in view range.

## Motivation and Context
Readability, consistency, exit script if no spins are in spin range.

## How Has This Been Tested?
Untested.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/tools/levelup.py
+++ b/tools/levelup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
+import os
 import sys
 import logging
 import time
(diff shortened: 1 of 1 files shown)


tools/levelup.py
```diff
@@ -111,37 +136,38 @@ def level_up_account(args, location, accounts, errors):
...
-startup_db(None, args.clear_db)
 args.player_locale = PlayerLocale.get_locale(args.location)
+
 if not args.player_locale:
     args.player_locale = gmaps_reverse_geolocate(
         args.gmaps_key,
         args.locale,
         str(position[0]) + ', ' + str(position[1]))
 
+startup_db(None, args.clear_db)
```
friscoMad: DB is needed for PlayerLocale.get_locale not sure how it does not complaint.

PR #2534, opened 2018-03-14
Improve --dump censor.

sebastienvercammen:
## Description
* Added some fields to censor.
* Return `<empty>` for censored fields that are empty.

## Motivation and Context
* Directories should always be censored.
* Helps to diagnose issues w/ empty fields (e.g. an empty `--status-page-password` = 404 on `/status`).

## How Has This Been Tested?
Untested.

## Types of changes
- [x] Enhancement

## Checklist:

### DIFF
--- a/pogom/utils.py
+++ b/pogom/utils.py
@@ -1245,9 +1245,10 @@ def check_output_catch(command):
         return result.strip()
 
 
-# Automatically censor all necessary fields. Lists will return
-# their length, all other items will return 'censored_tag'.
-def _censor_args_namespace(args, censored_tag):
+# Automatically censor all necessary fields. Lists will return their
+# length, all other items will return 'empty_tag' if they're empty
+# or 'censored_tag' if not.
+def _censor_args_namespace(args, censored_tag, empty_tag):
     fields_to_censor = [
         'accounts',
         'accounts_L30',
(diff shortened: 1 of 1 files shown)


pogo-excalibur:
Is there any easy way to check if a value is custom (i.e. not set to it's default value) and highlight it?

sebastienvercammen:
@pogo-excalibur Maybe by returning the parser itself and using its internal methods to determine the defaults. There's no actual "highlighting" in a text file, but we could add a custom tag.

But I would suggest that's something for a separate PR if anyone wants to pick that up.

billyjbryant:
Should `shared_config` also be censored since it outputs the directory in the path? ``` 'shared_config': '/home/pogo/data/RocketMap/config/shared.ini',
```

File: pogom/utils.py [hunk 1]
```diff
@@ -1289,7 +1290,9 @@ def _censor_args_namespace(args, censored_tag):
         'status_name',
         'status_page_password',
         'hash_key',
-        'trusted_proxies'
+        'trusted_proxies',
+        'data_dir',
+        'locales_dir'
```
billyjbryant: I would consider adding `shared_config`, everything else looks good :)

sebastienvercammen: Good catch. 👍

PR #2540, opened 2018-03-17
Efficiently exclude hidden Pokemon

animanmaster:
## Description

This PR optimizes the exclusion of Pokemon to speed up `raw_data` calls and improve the overall memory usage of the webserver.

## Motivation and Context

I run a map that covers a wide area with only the most common Pokemon excluded by default. Even with that default exclusion list, calls to `raw_data` would take 15-20 seconds, and I'd frequently go over the memory limitations of my cheap server. :)

## How Has This Been Tested?

Tested on my own map. Prior to this change, memory usage would go up to about 650-700MB with 2 gunicorn workers and `raw_data` calls were averaging about 15-20s. With this change, the same requests took around 5-8s and memory usage over the past week has dropped to about 350 - 400MB with 2 gunicorn workers. YMMV, of course, but letting MySQL use its pokemon_id index to filter before passing the data over to Python should provide a decent boost to speed and memory in most instances.

## Types of changes

## Checklist:


pogom/app.py
```diff
@@ -302,23 +302,29 @@ def raw_data(self):
         d['oNeLat'] = neLat
         d['oNeLng'] = neLng
 
+        eids = set([])
```
animanmaster: A micro-optimization, but using a `set` will make things like `int(x) not in eids` a bit more efficient.

pogom/app.py
```diff
@@ -329,17 +335,12 @@ def raw_data(self):
...
                                 oSwLat=oSwLat,
                                 oSwLng=oSwLng,
                                 oNeLat=oNeLat,
                                 oNeLng=oNeLng)))
 
-            if request.args.get('eids'):
-                # Exclude id's of pokemon that are hidden.
-                eids = [int(x) for x in request.args.get('eids').split(',')]
-                d['pokemons'] = [
-                    x for x in d['pokemons'] if x['pokemon_id'] not in eids]
```
animanmaster: Previously, we would be pulling all the active spawns in the current bounding box into memory and then iterating over them to exclude the Pokemon we want hidden. With thousands of spawns, this will take a toll, even if you expect to see fewer spawns on the map by having a large exclusion list (sometimes even more so in that case).

pogom/app.py
```diff
@@ -302,23 +302,29 @@ def raw_data(self):
...
+        if request.args.get('eids'):
+            # Exclude ids of Pokemon that are hidden.
+            for x in request.args.get('eids').split(','):
+                eids.add(int(x))
+
         if (request.args.get('pokemon', 'true') == 'true' and
                 not args.no_pokemon):
             if request.args.get('ids'):
-                ids = [int(x) for x in request.args.get('ids').split(',')]
+                ids = [int(x) for x in request.args.get('ids').split(',') if int(x) not in eids]
```
pogo-excalibur: With `eids` now being a set, is the `if int(x) not in eids` part necessary?

pogo-excalibur: Nevermind. `eids` is a set, `ids` isn't.

pogo-excalibur: As you've wrapped the use of `ids` in a `list()` could this perhaps be declared as a set to eliminate the `if int(x) not in eids` part?

pogom/models.py
```diff
@@ -136,9 +136,11 @@ class Meta:
 
     @staticmethod
     def get_active(swLat, swLng, neLat, neLng, timestamp=0, oSwLat=None,
-                   oSwLng=None, oNeLat=None, oNeLng=None):
+                   oSwLng=None, oNeLat=None, oNeLng=None, exclude=None):
         now_date = datetime.utcnow()
         query = Pokemon.select()
+        if exclude:
+          query = query.where(Pokemon.pokemon_id.not_in(list(exclude))
         if not (swLat and swLng and neLat and neLng):
```
pogo-excalibur: I'd suggest adding line breaks around these two lines:
```

        if exclude:
          query = query.where(Pokemon.pokemon_id.not_in(list(exclude))

```

pogom/app.py
animanmaster: Hmm sorry, not sure I follow. This is iterating over the `request.args.get('ids')` to make each value an `int` and make sure it only contains ids that are not in `eids`.

pogom/models.py
animanmaster: Sure, will do. 👍

pogom/app.py
pogo-excalibur: E.g.

```
ids = set([])
map(lambda id: ids.add(id), request.args.get('ids').split(','))
```

animanmaster: I still need to int-ify the string ids and need to exclude any ids in `eids`. For example, with a request like this:

```
/raw_data?ids=1,2,3,4,5,6,7,8,9&eids=1,2,4,5,7,8
```

ids should end up as `[3,6,9]`, hence the need for `if int(x) not in eids`.

The list comprehension isn't really removing duplicates, which your `set` change would accomplish; it's iterating over the ids, converting them to integers, and filtering out any that are in the exclusion list.

pogo-excalibur: Clearly I need to have my afternoon coffee :(
I was mising the `eids` on the end (and assumed it was `ids`).
👍

pogom/app.py
```diff
@@ -302,23 +302,32 @@ def raw_data(self):
         d['oNeLat'] = neLat
         d['oNeLng'] = neLng
 
+        eids = set([])
+        if request.args.get('eids'):
+            # Exclude ids of Pokemon that are hidden.
+            for x in request.args.get('eids').split(','):
+                eids.add(int(x))
```
sebastienvercammen: Since exclusion IDs are only for Pokémon, this code can be moved to below
```
         if (request.args.get('pokemon', 'true') == 'true' and
                 not args.no_pokemon):
```

animanmaster: Ooh good call. 👍 Changed in a12b8a9.

animanmaster:
Thanks @sebastienvercammen :) Updated with your suggestion. 👍

pogom/app.py
```diff
@@ -304,21 +304,31 @@ def raw_data(self):
 
         if (request.args.get('pokemon', 'true') == 'true' and
                 not args.no_pokemon):
+
+            # Exclude ids of Pokemon that are hidden.
+            eids = set([])
+            if request.args.get('eids'):
+                for x in request.args.get('eids').split(','):
+                    eids.add(int(x))
```
sebastienvercammen: Sorry for not realizing this in my previous review. You can simplify this code with set comprehension to:
```
eids = []
request_eids = request.args.get('eids')
if request_eids:
    eids = { int(x) for x in request_eids.split(',') }
```
*(Note I added a line to clarify `request_eids`, which also reduces the number of lookups that need to happen to retrieve its value.)*

animanmaster: No worries, I didn't think of it either, but it's definitely much more succinct this way. :) Updated in e96f9a7.

jagauthier:
Functionally, works exactly as expected.  I hid, and unhid pokemon. I zoomed in and out.  I don't have a very large map so I can't comment on performance improvements, but from reading the source, and the commentary it definitely seems like a nice process.

PR #2547, opened 2018-03-21
Fixed race condition

farstarss:
Fixed a race condition when loading map for the first time.

## Description

Although custom.js is listed above map.js in map.html it's not being executed prior to map.js since the settings inside custom.js are set on page load - which is fine as they depend on Store which is defined in map.js. 
The issue is that the google maps script is also going to be executed on page load (=> `defer` attribute) which results in a race condition between custom.js and google maps script calling `initMap()` and sometimes results in custom settings being ignored and default settings being used.

To reproduce this just disable clustering in custom.js and load the map in a private tab: You'll see that clustering sometimes happens and sometimes it doesn't.

## Motivation and Context

Race conditions should always be addressed and fixed.

## How Has This Been Tested?

Locally.

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -153,7 +153,7 @@ function createServiceWorkerReceiver() {
     })
 }
 
-function initMap() { // eslint-disable-line no-unused-vars
+function _initMap() { // eslint-disable-line
     map = new google.maps.Map(document.getElementById('map'), {
         center: {
             lat: Number(getParameterByName('lat')) || centerLat,
@@ -308,6 +308,12 @@ function initMap() { // eslint-disable-line no-unused-vars
     if (Push._agents.chrome.isSupported()) {
         createServiceWorkerReceiver()
     }
+}
+
+// wrapper function for real initMap (_initMap) to
+// avoid race conditions with custom.js
+function initMap() { // eslint-disable-line no-unused-vars
+    setTimeout(_initMap, 100)
 }
 
 function updateLocationMarker(style) {


static/js/map.js
```diff
@@ -310,6 +310,12 @@ function initMap() { // eslint-disable-line no-unused-vars
     }
 }
 
+// wrapper function for real initMap (_initMap) to
+// avoid race conditions with custom.js
+function initMap() { // eslint-disable-line no-unused-vars
+    setTimeout(_initMap, 100)
```
sebastienvercammen: Setting a 100ms timeout is a guess for execution order (an arbitrary delay). Instead, you have several options that guarantee execution order without delay:

- Replace jQuery's onLoad from custom.js with a self-executing function.
- Define a `customJsLoaded` (not the best naming) in `custom.js` and check for its existence in `map.js`. If it exists, call it before doing any initialization so it can properly set all defaults.

templates/map.html
```diff
@@ -519,7 +519,7 @@ <h3>Location Icon Marker</h3>
     </script>
     <script src="{{ url_for('static', filename='dist/js/app.min.js').lstrip('/') }}"></script>
     <script src="{{ url_for('static', filename='dist/js/map.common.min.js').lstrip('/') }}"></script>
-    <!-- Load custom JS before map.js so it can change Store defaults. -->
+    <!-- Custom JS is loaded but not actually executed prior to map.js as all its code is executed in jQuerys onload function. -->
```
sebastienvercammen: Read my other comment, and put this message back as it was.

sebastienvercammen:
@farstarss Are you still interested in updating this PR?

PR #2545, opened 2018-03-20
Adds buttons to the menu to export / import the settings

Ditto-IV100:
## Description
Adds buttons to the menu to export / import the settings.

## Motivation and Context

Because the settings are stored in "Local Storage", they can be lost. Not anymore!

## How Has This Been Tested?

I use it on my own server.

## Screenshots (if appropriate):
export-import

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2552, opened 2018-03-26
Adds a new menu item "Favorites Locations" in the navigation.

Ditto-IV100:
## Description

This will insert a new menu item "Favorites Locations" in the navigation with three functions:

- with the first function, users can search for places like neighborhoods, streets, etc. (This code was already in RM. I only changed it a bit.)

- the second function allows the user to save the current view as "quick button". Each view can be assigned an own name. This buttons can be deleted again at any time by the user.

- and the third feature allows the site admins to leave predefined locations as "quick button". These can be stored as deletable or non-deletable buttons. The predefined gps coordinates are stored in the _map.common.js_ 
A few examples are already inserted:
(result see Screenshots)

```
'favoritesLocations': {
    default: [{
        'Amsterdam': {lat: 52.3702157, lng: 4.895167899999933, zoom: 13, deletable: false},
        'Central Park (NY)': {lat: 40.7828647, lng: -73.96535510000001, zoom: 16, deletable: false},
        'London': {lat: 51.5073509, lng: -0.12775829999998223, zoom: 14, deletable: true},
        'Berlin': {lat: 52.52000659999999, lng: 13.404953999999975, zoom: 14}
    }],
    type: StoreTypes.JSON
}
```

## Motivation and Context

These functions are very helpful to save hotspots like nests, or if the site admin scans only some places (or a large area), he can deposit these places as a "quick button". The places created by the user are stored in the localstorage and will not be publicly available. Therefore, my other PR  (#2545)  is very useful to restore the settings in case of loss.

## How Has This Been Tested?

I use it on my map. I just changed it a bit for publishing and to make it more user-friendly.

favorites-locations

## Types of changes

## Checklist:

### DIFF
--- a/static/js/map.common.js
+++ b/static/js/map.common.js
@@ -1065,6 +1065,15 @@ var StoreOptions = {
     'isSearchMarkerMovable': {
         default: false,
         type: StoreTypes.Boolean
+    },
+    'favoritesLocations': {
+        default: [{
+            'Amsterdam': {lat: 52.3702157, lng: 4.895167899999933, zoom: 13, deletable: false},
+            'Central Park (NY)': {lat: 40.7828647, lng: -73.96535510000001, zoom: 16, deletable: false},
+            'London': {lat: 51.5073509, lng: -0.12775829999998223, zoom: 14, deletable: true},
+            'Berlin': {lat: 52.52000659999999, lng: 13.404953999999975, zoom: 14}
+        }],
+        type: StoreTypes.JSON
     }
 }
 
--- a/static/js/map.js
+++ b/static/js/map.js
@@ -509,6 +509,127 @@ function initSidebar() {
     }
 
     $('#pokemon-icon-size').val(Store.get('iconSizeModifier'))
+
+    initFavoLocation()
+
+    $('#add-favorit-location-button').on('click', function () {
+        var loc = map.getCenter()
+        var zoom = map.getZoom()
+        var name = $('#add-favorit-location-input').val()
+        addFarvoritLocationToSettings(name, loc.lat(), loc.lng(), zoom)
+        $('#add-favorit-location-input').val('')
+    })
+
+    var searchLocationInput = document.getElementById('search-favorit-location-input')
+    if (searchLocationInput) {
+        var googleSearchBox = new google.maps.places.Autocomplete(searchLocationInput)
+
+        googleSearchBox.addListener('place_changed', function () {
+            var place = googleSearchBox.getPlace()
+
+            if (!place.geometry) return
+
+            var loc = place.geometry.location
+            centerMap(loc.lat(), loc.lng(), 16)
+        })
+    }
+}
+
+/**
+ * Favorit-location
+ * first run load locations from settings / localStorage
+*/
+function initFavoLocation() {
+    var favoritLocationSettings = Store.get('favoritesLocations')[0]
+    Object.keys(favoritLocationSettings).forEach(function (location) {
+        addFavoritLocationToList(location,
+                                 favoritLocationSettings[location]['lat'],
+                                 favoritLocationSettings[location]['lng'],
+                                 favoritLocationSettings[location]['zoom'],
+                                 favoritLocationSettings[location]['deletable']
+                                )
+    })
+}
+
+/**
+ * Favorit-location
+ * add buttons to the navigation
+ *
+ * @param String name
+ * @param long lat – degree of latitude
+ * @param long lng – degree of longitude
+ * @param int zoom – zoomLevel
+ * @param boolean withDeleteButton (optional – default = true)
+*/
+function addFavoritLocationToList(name, lat, lng, zoom, withDeleteButton = true) {
+    var li = document.createElement('li')
+    li.setAttribute('name', name)
+
+    var favoritLocationButton = document.createElement('button')
+    if (!withDeleteButton) { favoritLocationButton.setAttribute('class', 'noDeleteButton') }
+    favoritLocationButton.innerHTML = name
+    favoritLocationButton.addEventListener('click', function () {
+        centerMap(lat, lng, parseInt(zoom))
+    })
+    li.appendChild(favoritLocationButton)
+
+    if (withDeleteButton) {
+        var favoritLocationDeleteButton = document.createElement('button')
+        favoritLocationDeleteButton.setAttribute('class', 'delete-favorit-location-button')
+        favoritLocationDeleteButton.innerHTML = '<span class="fa fa-times"></span>'
+        favoritLocationDeleteButton.addEventListener('click', function () {
+            confirm('Are you sure to delete "' + name + '"?') ? deleteFavoritLocation(li) : false
+        })
+        li.appendChild(favoritLocationDeleteButton)
+    }
+
+    var favoritLocationListElement = document.getElementById('favorit-location-ul')
+    favoritLocationListElement.appendChild(li)
+}
+
+/**
+ * Favorit-location
+ * delete location from localStorage and navigation
+ *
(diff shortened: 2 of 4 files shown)


static/js/map.common.js
```diff
@@ -1065,6 +1065,15 @@ var StoreOptions = {
     'isSearchMarkerMovable': {
         default: false,
         type: StoreTypes.Boolean
+    },
+    'favoritesLocations': {
+        default: [{
+            'Amsterdam': {lat: 52.3702157, lng: 4.895167899999933, zoom: 13, deletable: false},
```
pogo-excalibur: Just leave the default list blank.

static/js/map.js
```diff
@@ -509,6 +509,127 @@ function initSidebar() {
...
+    })
+
+    var searchLocationInput = document.getElementById('search-favorit-location-input')
+    if (searchLocationInput) {
+        var googleSearchBox = new google.maps.places.Autocomplete(searchLocationInput)
+
+        googleSearchBox.addListener('place_changed', function () {
+            var place = googleSearchBox.getPlace()
+
+            if (!place.geometry) return
```
pogo-excalibur: Because we're not using semicolons, please use:
```
if (!place.geometry) {
    return
}
```

static/js/map.js
```diff
@@ -509,6 +509,127 @@ function initSidebar() {
...
+            centerMap(loc.lat(), loc.lng(), 16)
+        })
+    }
+}
+
+/**
+ * Favorit-location
+ * first run load locations from settings / localStorage
+*/
+function initFavoLocation() {
```
pogo-excalibur: `initFavoLocation` -> `initFavoriteLocations`

static/js/map.common.js
```diff
@@ -1065,6 +1065,15 @@ var StoreOptions = {
     'isSearchMarkerMovable': {
         default: false,
         type: StoreTypes.Boolean
+    },
+    'favoritesLocations': {
```
pogo-excalibur: `favoritesLocations` -> `favoriteLocations`

static/js/map.common.js
Ditto-IV100: no problem, but then we need a documentation :)

templates/map.html
```diff
@@ -427,6 +427,46 @@ <h3>Use Pokémon cries</h3>
               </div>
           </div>
             {% endif %}
+
+          <h3><i class="fa fa-binoculars fa-fw"></i>Favorites Locations</h3>
```
pogo-excalibur: This should probably be under 'Location & Search Settings'.

static/js/map.js
Ditto-IV100: This is a code snippet from RM itself. But I can change that.

static/js/map.js
```diff
@@ -509,6 +509,127 @@ function initSidebar() {
...
+        var loc = map.getCenter()
+        var zoom = map.getZoom()
+        var name = $('#add-favorit-location-input').val()
+        addFarvoritLocationToSettings(name, loc.lat(), loc.lng(), zoom)
+        $('#add-favorit-location-input').val('')
+    })
+
+    var searchLocationInput = document.getElementById('search-favorit-location-input')
+    if (searchLocationInput) {
+        var googleSearchBox = new google.maps.places.Autocomplete(searchLocationInput)
```
pogo-excalibur: Is it possible to specify a locality here, using the location specified in the config?

Setting the locality helps reduce errors caused by multiple places with the same name.

pogo-excalibur:
You've misspelt 'favorite' throughout. Perhaps a language issue?

templates/map.html
Ditto-IV100: For the overview I have separated it, but if it is desired I can do that. It would fit in there too

pogo-excalibur: Is it a seperate category alongside 'Marker Settings', 'Location & Search Settings' etc?

static/js/map.js
Ditto-IV100: Well this part is from RM and uses the google API. I do not know if you can do something like that. But basically this is a google search and the locality is displayed.

static/js/map.common.js
pogo-excalibur: Yes - add an example to `custom.example.js`

Ditto-IV100:
Possible.. my native language is not english..

darkelement1987:
Also 'Favorite Locations' can be 'Favorite locations' and 'Search location' here (lowercase)

image

templates/map.html
Ditto-IV100: yes, I would say

Ditto-IV100:
@pogo-excalibur 
Apply changes

static/js/map.js
pogo-excalibur: So from memory Google allows you to pick a location to find results near.

e.g. If you were in London, Ohio (in the US) setting the locality to 'United States' would give you the result that you were looking for: London, Ohio, US, rather than the more searched-for: London, UK.

Ditto-IV100: Now it would look like this.
favorites-locations-ssearch

It's like "_Change search location_" which already exists in RM.

static/js/map.common.js
Ditto-IV100: done :)

darkelement1987:
I'd suggest changing the 'floppydisk' icon to a 'plus'-symbol. And maybe a few px's space between buttons/textfields? Just for the eye :p

Ditto-IV100:
@darkelement1987 why not a floppydisk? 

~~I found a bug, if no location is is preset. I will fix it.~~ _- done_

captbunzo:
Great feature btw. Similar to something I've done with custom.js on my map, but better! 👍

static/js/custom.js.example
```diff
@@ -11,6 +11,10 @@ $(function () {
     const upscalePokemon = false // Enable upscaling of certain Pokemon (upscaledPokemon and notify list). Default: false.
     const upscaledPokemon = [] // Add Pokémon IDs separated by commas (e.g. [1, 2, 3]) to upscale icons. Default: [].
 
+    // Add predefined buttons for favorit location
```
pogo-excalibur: `favorit` -> `favorite`

And because this is an example file you don't need to use a comment to specify the format, you can just use code:
```
const favoriteLocations = [
    {
        lat: 40.7828647,
        lng: -73.9653551,
        zoom: 16,
        deletable: false
    }
]
```

static/js/custom.js.example
```diff
@@ -83,6 +87,10 @@ $(function () {
     Store.set('showLocationMarker', showLocationMarker)
     Store.set('isLocationMarkerMovable', isLocationMarkerMovable)
 
+    if (Store.get('favoriteLocations')[0] === undefined) {
```
pogo-excalibur: Is the intention here to check if there are any favoriteLocations set?

If so use `if (Store.get('favoriteLocations').length === 0) {`

static/js/map.js
```diff
@@ -509,6 +509,134 @@ function initSidebar() {
     }
 
     $('#pokemon-icon-size').val(Store.get('iconSizeModifier'))
+
+    $('#add-favorit-location-button').on('click', function () {
```
pogo-excalibur: `favorit` -> `favorite`

(This needs changing throughout)

Ditto-IV100: you still found something.. shame on me :-/

static/js/custom.js.example
Ditto-IV100: this is also a possibility – _done_

static/js/map.js
pogo-excalibur: OK. Leave this for now then.

